PR #8, opened 2024-11-05
Fix publishing workflow

jprendes:
This PR:
* removes the version from `hyperlight-testing`. Having would force us to publish the crate to crates.io.
* remove the publishing of preview releases
* adds the ability to run the workflow as a dry-run
* publishes to crates.io instead of internal registry
* adds fields to Cargo.toml needed for publishing to crates.io

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,10 @@ exclude = [
 version = "0.9.0"
 edition = "2021"
 rust-version = "1.79.0"
+license = "Apache-2.0"
+homepage = "https://github.com/hyperlight-dev/hyperlight"
+repository = "https://github.com/hyperlight-dev/hyperlight"
+readme = "README.md"
 
 [workspace.dependencies]
 mshv-bindings = { version = "=0.2.1" }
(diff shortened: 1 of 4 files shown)


.github/workflows/CreateRelease.yml
```diff
@@ -85,6 +85,8 @@ jobs:
   cargo-publish:
     needs: [publish]
     uses: ./.github/workflows/CargoPublish.yml
+    with:
```
simongdavies, line 77: I think we should also update this workflow to remove all the dotnet stuff, make the version number calculation the same as it is in the CargoPublish.yml and get rid of the SHOULD_PUBLISH var

jprendes: With this changes CargoPublish.yml doesn't calculate any version, it uses whatever comes from the branch name.
Do you mind if we do the changes CreateRelease.yml in a different PR?

simongdavies: OK I will do a PR for that both need to be done so we can do a release

jprendes: I'm happy to do the PR for CreateReleaee.

PR #14, opened 2024-11-05
Rename references to dev branch to main

ludfjig:
These are the necessary changes for once we rename the `dev` branch to `main`.

### DIFF
--- a/.github/workflows/CreateRelease.yml
+++ b/.github/workflows/CreateRelease.yml
@@ -5,7 +5,7 @@ name: Create a Release
 on:
   workflow_dispatch:
   push:
-    branches: [dev]
+    branches: [main]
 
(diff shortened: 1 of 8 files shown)


.github/workflows/CreateRelease.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         shell: bash
 
 
-      - uses: hyperlight-dev/ci-setup-workflow@v1.0.0
+      - uses: hyperlight-main/ci-setup-workflow@v1.0.0
```
marosset: this is the org name and needs to stay hyperlight-dev :P

ludfjig: good catch!! fixed

PR #26, opened 2024-11-06
[performance] Move `set_up_hypervisor_partition` to inside the hv handler thread

danbugs:
As per KVM docs, vCPU ioctls should be issued from the same thread that was used to create the vCPU. In accordance to that, this PR moves `set_up_hypervisor_partition` to inside the hv handler thread, which manages vCPU interactions like init and dispatching guest funciton calls.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
+++ b/src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
@@ -16,6 +16,7 @@ limitations under the License.
 
 #[cfg(target_os = "windows")]
 use core::ffi::c_void;
+use std::ops::DerefMut;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use std::thread;
@@ -228,13 +229,9 @@ impl HypervisorHandler {
     #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
     pub(crate) fn start_hypervisor_handler(
         &mut self,
-        mut sandbox_memory_manager: SandboxMemoryManager<GuestSharedMemory>,
+        sandbox_memory_manager: SandboxMemoryManager<GuestSharedMemory>,
     ) -> Result<()> {
         let configuration = self.configuration.clone();
-        let mut hv = set_up_hypervisor_partition(
-            &mut sandbox_memory_manager,
-            configuration.outb_handler.clone(),
-        )?;
         #[cfg(target_os = "windows")]
         let in_process = sandbox_memory_manager.is_in_process();
 
@@ -267,18 +264,8 @@ impl HypervisorHandler {
         #[cfg(target_os = "linux")]
         self.execution_variables.run_cancelled.store(false);
 
-        #[cfg(target_os = "windows")]
-        if !in_process {
-            self.execution_variables
-                .set_partition_handle(hv.get_partition_handle())?;
-        }
-
         let to_handler_rx = self.communication_channels.to_handler_rx.clone();
-        #[cfg(target_os = "windows")]
-        let execution_variables = self.execution_variables.clone();
-        #[cfg(target_os = "linux")]
         let mut execution_variables = self.execution_variables.clone();
-        // ^^^ this needs to be mut on linux to set_thread_id
         let from_handler_tx = self.communication_channels.from_handler_tx.clone();
         let hv_handler_clone = self.clone();
 
@@ -295,9 +282,24 @@ impl HypervisorHandler {
             thread::Builder::new()
                 .name("Hypervisor Handler".to_string())
                 .spawn(move || -> Result<()> {
+                    let mut hv: Option<Box<dyn Hypervisor>> = None;
                     for action in to_handler_rx {
                         match action {
                             HypervisorHandlerAction::Initialise => {
+                                {
+                                    hv = Some(set_up_hypervisor_partition(
+                                        execution_variables.shm.try_lock().unwrap().deref_mut().as_mut().unwrap(),
+                                        configuration.outb_handler.clone(),
+                                    )?);
+                                }
+                                let hv = hv.as_mut().unwrap();
+
+                                #[cfg(target_os = "windows")]
+                                if !in_process {
+                                    execution_variables
+                                        .set_partition_handle(hv.get_partition_handle())?;
+                                }
+
                                 #[cfg(target_os = "linux")]
                                 {
                                     // We cannot use the Killable trait, so we get the `pthread_t` via a libc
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -82,7 +82,15 @@ impl KVMDriver {
         };
         let kvm = Kvm::new()?;
 
+        #[cfg(debug_assertions)]
+        let _now = std::time::Instant::now();
         let vm_fd = kvm.create_vm_with_type(0)?;
+        #[cfg(debug_assertions)]
+        println!(
```
ludfjig: let's not forget to remove these

danbugs: Done ✅

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -215,9 +216,23 @@ impl HypervisorHandler {
...
+    pub(crate) fn set_hypervisor(&mut self, hv: Box<dyn Hypervisor>) {
+        self.hypervisor = Some(Arc::new(Mutex::new(hv)));
+    }
+
+    /// Gets the hypervisor for the Hypervisor Handler.
+    pub(crate) fn get_hypervisor(&self) -> Result<MutexGuard<Box<dyn Hypervisor>>> {
+        self.hypervisor
+            .as_ref()
+            .ok_or_else(|| new_error!("Failed to get hypervisor: {}:{}:", file!(), line!()))
+            .map(|hv| hv.lock().unwrap())
```
ludfjig: Do we want an `.unwrap()` here?

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -64,6 +64,7 @@ pub(crate) struct HypervisorHandler {
     communication_channels: HvHandlerCommChannels,
     configuration: HvHandlerConfig,
     execution_variables: HvHandlerExecVars,
+    hypervisor: Option<Arc<Mutex<Box<dyn Hypervisor>>>>,
```
ludfjig: Just a conceptual question: Why do we need arc mutex for this hypervisor handler? Isn't everything handled from this thread?

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -228,13 +243,10 @@ impl HypervisorHandler {
     #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
     pub(crate) fn start_hypervisor_handler(
         &mut self,
-        mut sandbox_memory_manager: SandboxMemoryManager<GuestSharedMemory>,
+        sandbox_memory_manager: SandboxMemoryManager<GuestSharedMemory>,
     ) -> Result<()> {
+        let mut sandbox_memory_manager_clone = sandbox_memory_manager.clone();
```
ludfjig: nit: Is there any way to get rid of this clone?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -81,7 +81,6 @@ impl KVMDriver {
             log_then_return!("KVM is not present");
         };
         let kvm = Kvm::new()?;
-
```
ludfjig: is this intentioal?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -122,7 +122,7 @@ unsafe impl Send for ExclusiveSharedMemory {}
 /// unit that likely can't be discovered by the compiler) that _rust_
 /// users do not perform racy accesses to the guest communication
 /// buffers that are also accessed by HostSharedMemory.
-#[derive(Debug)]
```
ludfjig: nit: Same as above: is this clone necessary? (I haven't checked personally)

ludfjig (commented):
Nice job, I will be intersting to see if this increases performance :D. Just some minor questions

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
syntactically: I don't think you need to add this to the hypervisor handler struct, since it is only ever used from inside the hv handler thread; just storing the hypervisor in a local created inside the thread should be fine.

syntactically: It also then doesn't need to be Arc/Mutex; it can just be `Option<Box<dyn Hypervisor>>`.

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
syntactically: I don't believe that you need this clone, because the handler thread already has access to the memory manager through `execution_variables.shm`.

src/hyperlight_host/src/mem/shared_mem.rs
syntactically: I believe the GuestSharedMemory was purposefully not Clone because it was intended to reflect the unique ownership at any time of the guest side access to the shared memory. This may actually be somewhat important for soundness, actually; I can't recall.

syntactically (changes requested):
I think this can be simplified a little bit, and avoid some extra synchronisation and potential issues with aliasing the shared memory.

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
danbugs: Removed ✅

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
danbugs: Fixed! Thanks Lucy for the help ✅

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
danbugs: Oh, I missed that. Great catch. Fixed ✅

src/hyperlight_host/src/hypervisor/kvm.rs
danbugs: Reverted change ✅

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Removed ✅ . Thanks Lucy for the help again!

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -328,6 +330,7 @@ impl HypervisorHandler {
                                     .shared_mem
                                     .lock
                                     .try_read();
+
```
syntactically, line 333: intentional?

syntactically (approved):
LGTM

ludfjig (approved):

dblnz (approved):

PR #27, opened 2024-11-07
[chore] Modifying logo image in README to use an absolute path

danbugs:
Currently, our logo doesn't show on crates.io. This is because we are using a relative path to the image. This PR changes the path of the image to be an absolute path. For more info, see: https://github.com/rust-lang/crates.io/issues/982.


devigned (approved):
lgtm

ludfjig (approved):

PR #32, opened 2024-11-07
[github-actions] Made it so tests also run on intel

danbugs:
Note: I only modified the dep_rust matrix which will run the tests to assess differences between amd and intel. In the future, we might want to consider revisiting our matrix setup.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -17,17 +17,39 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        build: [windows-2022-release, linux-kvm-release, linux-hyperv-release]
+        build: [
+          windows-2022-release-amd,
+          linux-kvm-release-amd,
+          linux-hyperv-release-amd,
+          windows-2022-release-intel,
+          linux-kvm-release-intel,
+          linux-hyperv-release-intel,
+        ]
         include:
-          - build: windows-2022-release
+          - build: windows-2022-release-amd
             os: [self-hosted, Windows, X64, "1ES.Pool=hld-win2022-amd"]
             hypervisor: none
-          - build: linux-kvm-release
+            arch: amd
+          - build: linux-kvm-release-amd
             os: [self-hosted, Linux, X64, "1ES.Pool=hld-kvm-amd"]
             hypervisor: kvm
-          - build: linux-hyperv-release
+            arch: amd
+          - build: linux-hyperv-release-amd
             os: [self-hosted, Linux, X64, "1ES.Pool=hld-mshv-amd"]
             hypervisor: hyperv
+            arch: amd
+          - build: windows-2022-release-intel
+            os: [self-hosted, Windows, X64, "1ES.Pool=hld-win2022-intel"]
+            hypervisor: none
+            arch: intel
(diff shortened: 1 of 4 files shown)


ludfjig:
Could we add intel to benchmarking too? And include them in github release artifacts?

simongdavies:
@danbugs when this merges you will need to update the branch protection rules required checks as there are now new checks and some of the existing ones don't exist

README.md
```diff
@@ -1,6 +1,6 @@
-<div align="center">
+<div style="display: grid; place-items: center;">
```
ludfjig: Did you intend to include these changes in this PR? I think it would be better to make a separate PR for this

.github/workflows/Benchmarks.yml
```diff
@@ -17,17 +17,39 @@ jobs:
...
+          linux-hyperv-release-amd,
+          windows-2022-release-intel,
+          linux-kvm-release-intel,
+          linux-hyperv-release-intel,
+        ]
         include:
-          - build: windows-2022-release
+          - build: windows-2022-release-amd
             os: [self-hosted, Windows, X64, "1ES.Pool=hld-win2022-amd"]
             hypervisor: none
```
ludfjig: Optional: since we are refactoring some ci here, could we make the hypervisor different from `none` on windows? I'm thinking kvm/mshv/hyperv, where hyperv is for windows

.github/workflows/CreateRelease.yml
```diff
@@ -155,29 +155,50 @@ jobs:
...
-      - name: Download benchmarks (Linux kvm)
+      - name: Download benchmarks (Linux kvm amd)
         uses: actions/download-artifact@v4
         with:
-          name: benchmarks_Linux_kvm
-          path: benchmarks_Linux_kvm
+          name: benchmarks_Linux_kvm_amd
+          path: benchmarks_Linux_kvm_amd
+
+      - name: Download benchmarks (Windows intel)
```
ludfjig: Since there are a lot of this download-artifact action, maybe we could combine them into 1 using something like this? https://github.com/[actions/download-artifact](https://github.com/actions/download-artifact?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory)?tab=readme-ov-file#download-multiple-filtered-artifacts-to-the-same-directory

ludfjig (commented):
lgtm with some minor questions

README.md
danbugs: They are pretty minor README changes, so I figured it'd be alright. But ok, I'll move.

danbugs: Done: https://github.com/hyperlight-dev/hyperlight/pull/36

.github/workflows/Benchmarks.yml
danbugs: Changed the Windows name to "hyperv". Don't see a reason to change the Linux name, feel like that could lead to confusion.

.github/workflows/CreateRelease.yml
danbugs: Done ✅

simongdavies (approved):

ludfjig (approved):
LGTM

PR #35, opened 2024-11-07
[chore] Add devcontainer config

dblnz:
This adds devcontainer config to be used with codespaces and Visual Studio

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ These micro VMs operate without a kernel or operating system, keeping overhead l
 
 By default, Hyperlight restricts guest access to a minimal API. The only _host function_ available by default allows the guest to print messages, which are displayed on the host console or redirected to stdout, as configured. Hosts can choose to expose additional host functions, expanding the guest’s capabilities as needed.
 
-Below is an example demonstrating the use of the Hyperlight host library in Rust to execute a simple guest application and am example of a simple guest application using the Hyperlight guest library in also written in Rust.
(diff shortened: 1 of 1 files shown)


danbugs:
Is this possible now? I tried doing this about 2 years ago and, iirc, it wasn't possible to access `/dev/kvm`.

dblnz:
> Is this possible now? I tried doing this about 2 years ago and, iirc, it wasn't possible to access `/dev/kvm`.

I ran some tests locally and it seems to be working.

danbugs:
> > Is this possible now? I tried doing this about 2 years ago and, iirc, it wasn't possible to access `/dev/kvm`.
> 
> I ran some tests locally and it seems to be working.

Awesome! Might just not work if people try to use it from the browser? We can add a note on that after testing, though.

dblnz:
> > > Is this possible now? I tried doing this about 2 years ago and, iirc, it wasn't possible to access `/dev/kvm`.
> 
> > 
> 
> > I ran some tests locally and it seems to be working.
> 
> 
> 
> Awesome! Might just not work if people try to use it from the browser? We can add a note on that after testing, though.

I ran a couple of times today in the browser, but the issue was it built the container image too slowly and I believe it got terminated due to timeout.
Now I am looking into how to make the container smaller so that it is built faster

devigned:
@dblnz have you considered building a base image and then using that rather than base Debian?

dblnz:
> @dblnz have you considered building a base image and then using that rather than base Debian?

Yes, and that is the current plan, after I test locally and see everything works, then I'll push the image so it can be used with devcontainers

PR #36, opened 2024-11-07
[documentation] small README changes

danbugs:
- `align` is a deprecated attribute: https://stackoverflow.com/questions/14551305/align-attribute-deprecated
- changed logo image to be direct image link so it works outside GH, and renamed the logo name removing "tentative".
- small typos/nits.


README.md
```diff
@@ -212,9 +212,9 @@ If all worked as expected, you should the following message in your console:
 Hello, World! I am executing inside of a VM :)
 ```
 
-If you get the error `Error: NoHypervisorFound` and KVM or mshv is set up then this may be a permissions issue. In bash you can use `ls -l /dev/kvm` or  `ls -l /dev/mshv` to check which groups that owns the device and then `groups` to make sure your user is a member of that group. 
+If you get the error `Error: NoHypervisorFound` and KVM or mshv is set up then this may be a permissions issue. In bash, you can use `ls -l /dev/kvm` or  `ls -l /dev/mshv` to check which groups that owns the device and then `groups` to make sure your user is a member of that group.
```
simongdavies:
```suggestion
If you get the error `Error: NoHypervisorFound` and KVM or mshv is set up then this may be a permissions issue. In bash, you can use `ls -l /dev/kvm` or  `ls -l /dev/mshv` to check which group owns that device and then `groups` to make sure your user is a member of that group. 
```

simongdavies (approved):

devigned (approved):
lgtm

PR #44, opened 2024-11-08
Ignore markdown and txt files in CI

devigned:
This PR will ignore paths for markdown as well as txt files. The intent of this is to make it quicker to accept docs changes.

PR #48, opened 2024-11-08
Fix docs only CI job processing

devigned:
In #44, I attempted to ignore docs only CI job processing, but branch protection required jobs were still required to run. The only way to satisfy the branch protection required jobs is to have them run, but return with success. This PR does that by using conditional execution for the job rather than filtering on the PR content in the `pull_request` event handler.

### DIFF
--- a/.github/workflows/ValidatePullRequest.yml
+++ b/.github/workflows/ValidatePullRequest.yml
@@ -5,9 +5,6 @@ name: Validate Pull Request
 on:
   pull_request:
     branches: [main, "release/**"]
-    paths-ignore:
-      - '**.md'
-      - '**.txt'
   merge_group:
 
 permissions:
@@ -15,20 +12,52 @@ permissions:
   contents: read
 
 jobs:
+  docs-pr:
+    runs-on: ubuntu-latest
+    outputs:
+      docs-only: ${{ steps.docs-only.outputs.result }}
+    steps:
+      - uses: dorny/paths-filter@v3
+        id: changes
+        with:
+          filters: |
+            docs:
+              - '**/*.md'
+              - '**/*.txt'
+            all:
+              - '**/*'
+      - uses: actions/github-script@v7
+        id: docs-only
+        with:
+          script: |
+            let docs_file_count = ${{steps.changes.outputs.docs_count}};
+            let all_file_count = ${{steps.changes.outputs.all_count}};
+            return all_file_count === docs_file_count;
+          result-encoding: string
 
   rust:
+    needs:
+      - docs-pr
     uses: ./.github/workflows/dep_rust.yml
     secrets: inherit
+    with: 
+      docs_only: ${{needs.docs-pr.outputs.docs-only}}
   fuzzing:
+    needs:
+      - docs-pr
     uses: ./.github/workflows/dep_fuzzing.yml
     with:
       max_total_time: 300 # 5 minutes in seconds
+      docs_only: ${{needs.docs-pr.outputs.docs-only}}
     secrets: inherit
 
   #####
   # start build-on-windows
   #####
   build-on-windows:
+    needs:
+      - docs-pr
+    if: ${{needs.docs-pr.outputs.docs-only != 'true'}}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
@@ -72,6 +101,9 @@ jobs:
   # start build-on-linux
   #####
   build-on-linux:
+    needs:
+      - docs-pr
+    if: ${{needs.docs-pr.outputs.docs-only != 'true'}}
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: true
--- a/.github/workflows/dep_fuzzing.yml
+++ b/.github/workflows/dep_fuzzing.yml
@@ -7,6 +7,11 @@ on:
         description: Maximum total time for the fuzz run in seconds
         required: true
         type: number
+      docs_only:
+        description: Skip fuzzing if docs only
+        required: false
+        type: string
+        default: "false"
 
 permissions:
   id-token: write
@@ -14,6 +19,7 @@ permissions:
 
 jobs:
   fuzz:
+    if: ${{ inputs.docs_only == 'false' }}
     runs-on: [ self-hosted, Linux, X64, "1ES.Pool=hld-kvm-amd" ]
     steps:
       - name: Checkout code
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -5,6 +5,12 @@ name: Rust Tests and Lints
 # See README.md in this directory for more information about workflow_call
 on:
   workflow_call:
+    inputs:
+      docs_only:
+        description: Skip building if docs only
+        required: false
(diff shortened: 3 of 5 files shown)


.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,21 +5,47 @@ name: Validate Pull Request
...
+              - '**/*.txt'
+            all:
+              - '**/*'
+      - uses: actions/github-script@v7
+        id: not-just-docs
+        with:
+          script: |
+            let docs_file_count = "${{steps.changes.outputs.docs_count}}";
+            let all_file_count = "${{steps.changes.outputs.all_count}}";
+            let result = (all_file_count - docs_file_count) > 0;
```
jprendes:
```suggestion
            console.log(all_file_count > docs_file_count);
```

jprendes (approved):
LGTM

.github/workflows/ValidatePullRequest.yml
jprendes: Should also remove the line below, but the phone wouldn't let me do multiline selection

.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,30 +5,60 @@ name: Validate Pull Request
...
+            docs:
+              - '**/*.md'
+              - '**/*.txt'
+            all:
+              - '**/*'
+      - uses: actions/github-script@v7
+        id: not-just-docs
+        with:
+          script: |
+            let docs_file_count = "${{steps.changes.outputs.docs_count}}";
```
jprendes:
```suggestion
            let docs_file_count = ${{steps.changes.outputs.docs_count}};
```

.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,30 +5,60 @@ name: Validate Pull Request
...
+            let all_file_count = "${{steps.changes.outputs.all_count}}";
```
jprendes:
```suggestion
            let all_file_count = ${{steps.changes.outputs.all_count}};
```

jprendes (dismissed):

.github/workflows/ValidatePullRequest.yml
jprendes: With the `"..."` here, it's doing string comparison instead of numerical comparison

.github/workflows/dep_rust.yml
```diff
@@ -5,6 +5,12 @@ name: Rust Tests and Lints
 # See README.md in this directory for more information about workflow_call
 on:
   workflow_call:
+    inputs:
+      docs_only:
+        description: If docs only, don't fuzz, just complete successfully
```
jprendes:
```suggestion
        description: skip building if docs only
```

.github/workflows/dep_fuzzing.yml
```diff
@@ -7,13 +7,19 @@ on:
         description: Maximum total time for the fuzz run in seconds
         required: true
         type: number
+      docs_only:
+        description: If docs only, don't fuzz, just complete successfully
```
jprendes:
```suggestion
        description: skip fuzzing if docs only
```

devigned:
FYI, this make docs updates take ~15s to pass CI.

.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,30 +5,59 @@ name: Validate Pull Request
...
+        id: changes
+        with:
+          filters: |
+            docs:
+              - '**/*.md'
+              - '**/*.txt'
+            all:
+              - '**/*'
+      - uses: actions/github-script@v7
+        id: not-just-docs
```
jprendes: **suggestion:** rename `not-just-docs` to `docs-only` and invert the logic, to avoid the double negation.

.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,30 +5,59 @@ name: Validate Pull Request
...
+        with:
+          script: |
+            let docs_file_count = ${{steps.changes.outputs.docs_count}};
+            let all_file_count = ${{steps.changes.outputs.all_count}};
+            console.log(all_file_count > docs_file_count)
```
ludfjig: I think this maybe should be `return all_file_count > docs_file_count` instead of `console.log()`?

jprendes (dismissed):
LGTM

ludfjig (dismissed):
lgtm with a question

.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,30 +5,59 @@ name: Validate Pull Request
...
+              - '**/*.txt'
+            all:
+              - '**/*'
+      - uses: actions/github-script@v7
+        id: docs-only
+        with:
+          script: |
+            let docs_file_count = ${{steps.changes.outputs.docs_count}};
+            let all_file_count = ${{steps.changes.outputs.all_count}};
+            console.log(all_file_count > docs_file_count)
```
jprendes: With that name, this logic also needs to change
```suggestion
            console.log(all_file_count <= docs_file_count)
```

and unfortunately so do all the "true" and "false" conditions

devigned: I think I'm just going to back to `console.log((all_file_count - docs_file_count) == 0)`.

jprendes: In that case `console.log(all_file_count == docs_file_count)`

jprendes: or triple equal for type equality as well `console.log(all_file_count === docs_file_count)`

devigned: I feel real dumb right now LOL

.github/workflows/ValidatePullRequest.yml
```diff
@@ -5,30 +5,59 @@ name: Validate Pull Request
...
+            return all_file_count === docs_file_count;
+          result-encoding: string
 
   rust:
+    needs:
+      - docs-pr
     uses: ./.github/workflows/dep_rust.yml
     secrets: inherit
+    with: 
+      docs_only: ${{needs.docs-pr.outputs.docs-only != 'true'}}
```
jprendes: All these `'true'` need to change to `'false'`, and this PR should be running all the workflows as it's not a doc only change. Currently it's skipping all workflows.

jprendes (changes requested):

.github/workflows/ValidatePullRequest.yml
jprendes: Also, I don't know if this should be a string comparison or a boolean :-\

jprendes: string comparison is the answer

devigned: I think I had it reversed to test it and then started chasing my tail.

jprendes (approved):
LGTM.
Thanks @devigned !

devigned:
> LGTM. Thanks @devigned !

Thank you! Really appreciate your reviews.

danbugs (approved):

PR #54, opened 2024-11-12
Add devcontainer config

dblnz:
The following dev container configuration can be found in the .devcontainer directory.
It contains:
- devcontainer.json a configuration file for the Dev container that puts everything toghether
- Dockerfile config for a base container image that is pre-built by using the .github/workflows/CreateDevcontainerImage.yml action that is automatically triggered when a change is detected for Dockerfile or the workflow configuration.
- setup.sh script that changes the ownership for the /dev/kvm device inside the container so it can be accessible by the user


.github/workflows/CreateDevcontainerImage.yml
```diff
@@ -0,0 +1,73 @@
...
+# There is a single job in this workflow. It's configured to run on the latest available version of Ubuntu.
+jobs:
+  build-and-push-image:
+    runs-on: ubuntu-latest
+    # Sets the permissions granted to the `GITHUB_TOKEN` for the actions in this job.
+    permissions:
+      contents: read
+      packages: write
+      attestations: write
+      id-token: write
```
simongdavies: do you still need id-token write?

simongdavies (dismissed):
One small issue, this is great, thanks @dblnz

.github/workflows/CreateDevcontainerImage.yml
dblnz: I don't think so, I'll remove.
Thanks

simongdavies (dismissed):

simongdavies (approved):

ludfjig (approved):

PR #61, opened 2024-11-15
Fix otlp tracing

dblnz:
This PR fixes the otlp_tracing example compile issues with latest opentelemetry crate.

It required a slightly different approach to set up the tracer. The [examples ](https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/examples/opentelemetry-otlp.rs) from tracing-opentelemetry crate were useful.

One other side of the problem was that `jaeger` didn't receive the traces from the example anymore and it turns out it is because of the log filtering. So, using a different `RUST_LOG` should suffice.

The new test added sets up a TcpListener that binds to the same address the tracer subscriber does so it is able to receive the traces so it can validate whether the example runs correctly

### DIFF
--- a/docs/hyperlight-metrics-logs-and-traces.md
+++ b/docs/hyperlight-metrics-logs-and-traces.md
@@ -82,13 +82,13 @@ In the [examples/otlp_tracing](../src/hyperlight_host/examples/otlp_tracing) dir
 #### Linux
 
 ```bash
-RUST_LOG='none,hyperlight-host=info,tracing=info' cargo run --example otlp_tracing
+RUST_LOG='none,hyperlight_host=info,tracing=info' cargo run --example otlp_tracing
 ```
 
 #### Windows
 
 ```powershell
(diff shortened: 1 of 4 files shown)


simongdavies (approved):

docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -82,21 +82,21 @@ In the [examples/otlp_tracing](../src/hyperlight_host/examples/otlp_tracing) dir
 #### Linux
 
 ```bash
-RUST_LOG='none,hyperlight-host=info,tracing=info' cargo run --example otlp_tracing
+RUST_LOG='none,hyperlight_host=info,tracing=info' cargo run --example otlp_tracing
```
ludfjig, line 85: nice catch

ludfjig (approved):

ludfjig (approved):

marosset (approved):

PR #73, opened 2024-11-20
Avoid eagerly doing unnecessary string formatting

ludfjig:
Switch from `.ok_or` to `.ok_or_else` do only perform the string formatting in the failing case. 
Benchmarks show that time to create an uninitialized sandbox goes from 400µs to about 330µs (-17%) on my local machine

Also adds an `amd` tag that was forgotten in PR #74


dblnz (approved):
Good catch

simongdavies (approved):

danbugs (approved):

dblnz (approved):

PR #74, opened 2024-11-20
Fix 'download benchmark' workflow

ludfjig:
Currently, PRs are not able to compare performance to main branch, because downloading benchmarks from latest prerelease fails because the name of the artifact is incorrect. See https://github.com/hyperlight-dev/hyperlight/actions/runs/11925609379/job/33237994407#step:16:13 for an example error. 

Here is an example of it now working https://github.com/hyperlight-dev/hyperlight/actions/runs/11928115786/job/33244575940?pr=74#step:16:12. Note that some will still fail, since we haven't published benchmarks with the new naming conventions yet (It will be done as part of prerelease when this PR is merged). 

This pr also renames the matrix.hypervisor such that hyperv is used for windows, and mshv for linux

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -17,14 +17,6 @@ jobs:
     strategy:
       fail-fast: true
       matrix:
-        build: [
-          windows-2022-release-amd,
-          linux-kvm-release-amd,
-          linux-hyperv-release-amd,
-          windows-2022-release-intel,
-          linux-kvm-release-intel,
-          linux-hyperv-release-intel,
-        ]
         include:
           - build: windows-2022-release-amd
             os: [self-hosted, Windows, X64, "1ES.Pool=hld-win2022-amd"]
@@ -34,9 +26,9 @@ jobs:
             os: [self-hosted, Linux, X64, "1ES.Pool=hld-kvm-amd"]
             hypervisor: kvm
             arch: amd
-          - build: linux-hyperv-release-amd
+          - build: linux-mshv-release-amd
             os: [self-hosted, Linux, X64, "1ES.Pool=hld-mshv-amd"]
-            hypervisor: hyperv
+            hypervisor: mshv
             arch: amd
           - build: windows-2022-release-intel
             os: [self-hosted, Windows, X64, "1ES.Pool=hld-win2022-intel"]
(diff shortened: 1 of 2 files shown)


simongdavies (approved):

.github/workflows/Benchmarks.yml
```diff
@@ -34,9 +26,9 @@ jobs:
             os: [self-hosted, Linux, X64, "1ES.Pool=hld-kvm-amd"]
             hypervisor: kvm
             arch: amd
-          - build: linux-hyperv-release-amd
+          - build: linux-mshv-release-amd
```
danbugs, line 29: Do we want to name them different things? Afaik, people call it Linux Hyper-V in official documentation (https://learn.microsoft.com/en-us/windows-server/virtualization/hyper-v/best-practices-for-running-linux-on-hyper-v)?

.github/workflows/Benchmarks.yml
```diff
@@ -17,14 +17,6 @@ jobs:
...
       fail-fast: true
       matrix:
-        build: [
-          windows-2022-release-amd,
-          linux-kvm-release-amd,
-          linux-hyperv-release-amd,
-          windows-2022-release-intel,
-          linux-kvm-release-intel,
-          linux-hyperv-release-intel,
-        ]
```
danbugs, line 27: Do you not have to specify the build matrix?

.github/workflows/dep_rust.yml
```diff
@@ -155,15 +151,15 @@ jobs:
...
+        if: runner.os == 'Linux' && matrix.hypervisor == 'mshv'
         run: sudo dnf install gh -y
 
       - name: Install github-cli (Linux ubuntu)
         if: runner.os == 'Linux' && matrix.hypervisor == 'kvm'
         run: sudo apt install gh -y
 
       - name: Download benchmarks from "latest"
-        run: just bench-download ${{ runner.os }} ${{ matrix.hypervisor }} dev-latest # compare to prerelease
+        run: just bench-download ${{ runner.os }} ${{ matrix.hypervisor }} ${{ matrix.arch}} dev-latest # compare to prerelease
```
danbugs, line 162: Good catch!

.github/workflows/Benchmarks.yml
ludfjig: Nope it seems this was redundant

PR #77, opened 2024-11-20
[github_actions] update PRLabelChecker workflow to pull new labels

danbugs:


ludfjig (approved):

.github/workflows/PRLabelChecker.yml
```diff
@@ -14,16 +14,22 @@ jobs:
...
       - name: Check for specific labels
         run: |
-          PR_NUMBER=$(echo ${{ github.event.pull_request.number }})
-          LABELS_JSON=$(gh pr view $PR_NUMBER --json labels -q '.labels.[] | .name')
-          REQUIRED_LABELS=("chore" "ignore" "breaking-change" "enhancement" "feature" "dependencies" "bug" "security" "performance" "refactor" "testing" "documentation" "github-actions")
-          for REQUIRED_LABEL in "${REQUIRED_LABELS[@]}"; do
-            if echo "$LABELS_JSON" | grep -q "$REQUIRED_LABEL"; then
-              echo "One of the required labels is present"
+          # Fetch repository labels from the GitHub API
+          REPO_LABELS=$(gh api repos/${{ github.repository }}/labels --jq '.[].name')
```
simongdavies, line 18: Is it worth checking if there are any labels we don't want to be used, I seem to recall there were some that did not apply to PRs (e.g. needs-review)

danbugs: Considering we don't have established guidelines around labels, I think this should be okay as is. Who knows, maybe we'll add "needs review" to PRs in the future that we might want to discuss in the upcoming community meeting or smt like that.

simongdavies: Hmmm OK , not sure I agree but no matter

simongdavies (approved):

PR #81, opened 2024-11-21
Fix cargo feature 'dump_on_crash'

### LINKED ISSUE #70: hyperlight-host feature `dump_on_crash` doesn't compile
`cargo check -p hyperlight-host --features dump_on_crash` will show the problem.

We should probably add compile checks in CI to prevent this from happening again

ludfjig:
Fixes #70. Renames the feature to just "crashdump". Moves it into VirtualCPU::run, so each driver doesn't need to call it. Behavior should be the same: dumps registers,memory_regions and raw memory to file for certain hyperlightexits: mmio, accessviolation, unknown, and errors.

Also changes uses of `crate::debug!` macro to explicilty use the `crate` module prefix to distinguish from the `log` crate.

### DIFF
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -118,9 +118,7 @@ function_call_metrics = []
 executable_heap = []
 # This feature enables printing of debug information to stdout in debug builds
 print_debug = []
-# This feature enables dunping of the VMs details to a file when an unexpected or error exit occurs in a VM in debug mode
-# the name of the file is output to stdout and logged.
-dump_on_crash = ["dep:tempfile"]
+crashdump = ["dep:tempfile"] # Dumps the VM state to a file on unexpected errors or crashes. The path of the file will be printed on stdout and logged. This feature can only be used in debug builds.
 kvm = ["dep:kvm-bindings", "dep:kvm-ioctls"]
 mshv = ["dep:mshv-bindings", "dep:mshv-ioctls"]
 inprocess = []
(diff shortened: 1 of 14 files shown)


src/hyperlight_host/Cargo.toml
```diff
@@ -112,15 +112,15 @@ cfg_aliases = "0.2.1"
 built = { version = "0.7.0", features = ["chrono", "git2"] }
 
 [features]
-default = ["kvm", "mshv", "seccomp"]
+default = ["kvm", "mshv", "seccomp", "crashdump"]
```
ludfjig: remember to undo this

src/hyperlight_host/src/hypervisor/dump_on_crash.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Dump registers + memory regions + raw memory to a tempfile
+#[cfg(feature = "dump_on_crash")]
+pub(crate) fn dump_on_crash_to_tempfile(hv: &dyn Hypervisor) -> Result<()> {
+    let mem_regions = hv.get_memory_regions();
+    let mem_size = mem_regions
+        .iter()
+        .map(|region| region.host_region.len())
+        .sum();
+    let mem_start_addr = mem_regions[0].host_region.start as *const u8;
```
syntactically: It looks like there is an invariant here that the host regions all be contiguous, which is always the case now, but might not be in the future. Can we add a note on the definition of `get_memory_regions()` that this is an invariant imposed by this code, and/or change this code to not impose that invariant?

syntactically (dismissed):
LGTM except for that one documentation nit.

src/hyperlight_host/src/hypervisor/dump_on_crash.rs
ludfjig: fixed!

syntactically (dismissed):
LGTM

danbugs (dismissed):
LGTM!

danbugs (approved):

PR #85, opened 2024-11-26
[HIP] Hyperfile

danbugs:
This PR adds a HIP for Hyperfile—a TOML configuration file for sandbox configuration and run options.

The goal is to enable host applications built with Hyperlight to be configurable in a non-programmatic way, allowing users to modify microVM settings without directly altering the host code.

You can view the rendered file [here](https://github.com/danbugs/hyperlight/tree/config-proposal/proposals/0001-hip-hyperfile).

### DIFF
--- a/proposals/0001-hip-hyperfile/README.md
+++ b/proposals/0001-hip-hyperfile/README.md
@@ -1,0 +1,160 @@
+# HIP 0001 - Hyperfile
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+    - [User Stories (Optional)](#user-stories-optional)
+        - [Story 1](#story-1)
+        - [Story 2](#story-2)
+    - [Notes/Constraints/Caveats (Optional)](#notesconstraintscaveats-optional)
+    - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+    - [Test Plan](#test-plan)
+        - [Unit tests](#unit-tests)
+        - [Integration tests](#integration-tests)
+        - [e2e tests](#e2e-tests)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This proposal introduces the concept of a `hyperfile.toml` configuration file for specifying
+sandbox settings in Hyperlight. This feature will allow users to define configurations such
+as `stack_size_override`, `max_execution_time`, and execution modes (`RunInHypervisor` or
+`RunInProcess`) in a structured and human-readable format. The goal is to enable host 
+applications built with Hyperlight to be configurable in a non-programmatic way, 
+allowing users to modify microVM settings without directly altering the host code.
+
+## Motivation
+
+### Goals
+
+- Enable the creation and use of a `hyperfile.toml` configuration file for defining sandbox 
+settings.
+- Support all current configuration options (e.g., `stack_size_override`, `max_execution_time`) 
+in the new format.
+- Maintain backwards compatibility with direct code-based sandbox configuration.
+
+### Non-Goals
+
+- This proposal does not aim to change how sandbox configurations are applied programmatically.
+- It does not address dynamic configuration updates while a sandbox is running.
+
+## Proposal
+
+The `hyperfile.toml` will serve as a configuration file for sandbox initialization. Users will 
+specify parameters in the TOML format. The Hyperlight API will expose methods to parse the 
+file and use it to initialize sandboxes.
+
+### User Stories
+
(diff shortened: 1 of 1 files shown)


ludfjig:
Let's make sure we put every kv in the toml under a table such as [memory] or [runtimeoptinos] to make it forward compatible in case we need to add more fields. I also think we should avoid the name "override" in the fields. Btw, this is probably a good time as well to introduce a builder to UninitializedSandbox, and maybe even change the UninitializedSandbox::new api to be more ergonomic.

dblnz:
There are two aspects that would need clarifying:

1. How do we treat the case of an older or newer Hyperfile than Hyperlight is being used?
Example: I am using a Hyperfile that I took from a colleague but it matches a different version of Hyperlight than the one I have. The might not have some fields defined or has additional ones that are no longer valid.
Should we define some kind of versioning?

2. Is this meant to accommodate multiple configurations or is it used for setting defaults?
Example: An user has multiple types of sandboxes, how are their configurations chosen?
Do we add something like a special section for each type of configuration so that the user can mention which one to use, or do we use the Hyperfile for default values?

proposals/0001-hip-hyperfile/README.md
```diff
@@ -0,0 +1,160 @@
...
+#### Increased Complexity for New Users
+
+Introducing a configuration file may add initial complexity for users unfamiliar with TOML. 
+
+- **Mitigation**: Provide a well-documented template and examples in the Hyperlight 
+documentation.
+
+## Design Details
+
+### Rough Sketch of a `hyperfile.toml`
```
marosset, line 91: Have you considered being able to specify multiple different named sandbox configurations in the same file that can be referenced by name when creating the sandboxes?

I can see a use case for wanting to specify different configurations.

This would work similar to how runtime classes work in containerd.

marosset: If we do this I imagine one of the entries would be a default one which would be used if no named sandbox configuration is used.

devigned (commented):
+1 to the proposal. Nice work with the structure and content. 

I'm curious to see where @marosset's comment leads. I haven't fully thought through it, but I think I'm in favor a default and additional named configurations.

proposals/0001-hip-hyperfile/README.md
ludfjig: What about just doing this in separate files? `config1.toml` and `config2.toml`. It would simplify parsing alot, especially if we use the same struct for the config value itself, AND serde toml deserialization. And then you could do something like Config::FromFile(...) on the one you want to use.

proposals/0001-hip-hyperfile/README.md
```diff
@@ -0,0 +1,160 @@
...
+As a systems engineer, I want a standardized configuration file to share with teammates for
+debugging or reproducing issues without requiring them to modify source code.
+
+#### Story 3: Supporting Off-the-Shelf Hosts
+
+As a host provider offering an off-the-shelf host leveraging Hyperlight, I want to use a 
+`hyperfile.toml` to define configurable sandbox settings. This allows end users to adjust 
+microVM parameters without modifying the host's code, ensuring the integrity of the host 
+application while providing flexibility to users.
+
```
simongdavies, line 72: Would you see this as the host's responsibility to specify/resolve where these files are at runtime? We could possibly determine the default for a Sandbox via a well known environment variable , however if the host wanted to provide different types of Sandbox with different configurations for each type it would need to specify how this is done , I don't think that is something that can be done in Hyperlight itself, regardless of if this is done in single files or with named configurations in a single file.

simongdavies:
We should consider how the config file can be used by solutions built on top of hyperlight, e.g. hyperlight-wasm has its own builder and slightly modified configuration, solutions such as this may want to hide/add/restrict configuration items.

syntactically:
Thinking about a conversation I just had about snapshotting: a big part of the motivation for this HIP is that we have a _lot_ of tunable knobs for memory layout right now, but this is perhaps a historical artifact, and adds a lot of tunables that we have to be careful to "match up" in a sandbox save/restore scenario. An alternative that we have put some thought towards is moving to a situation where the host just provides an undifferentiated blob of memory (up to some maximum limit for the sandbox), and the guest itself handles assigning memory to different purposes, which would remove the vast majority of these knobs while also being a lot more flexible and removing some memory overhead due to the fact that the allocation of guest physical memory between these purposes isn't allowed to change through the lifetime of the sandbox. Similarly, we've talked about removing our sandbox execution wall-clock time limits and replacing them with a more general set of primitives that allow hyperlight hosts to implement their own, more sophisticated, accounting and cancelation policies. I think it's worth thinking some more about whether we want to move in these direction, and, if so, just how many sandbox settings we can remove entirely---perhaps ideally you would specify just a binary/list of snapshot layers and a run-mode, and that would be it, in which case having an external file format might be overkill.

syntactically:
Most of these tunables are gone, and hopefully more will follow, so we probably don't want to do this anymore.

PR #92, opened 2024-12-02
Fix broken realloc, rename and mark hlmalloc/hlfree/hlcalloc/hlrealloc unsafe

ludfjig:
First commit fixes the actual bug, the second commit is rename+docs+cleanup


syntactically (approved):
LGTM. Good catches on switching to marking things unsafe and using the checked arithmetic ops.

marosset (approved):

PR #93, opened 2024-12-03
Issue template for reporting bugs

marosset:
Adding an issue template for reporting bugs


.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,71 @@
...
+        ```console
+        On Linux:
+        $ cat /etc/os-release
+        # paste output here
+
+        $ cat /uname -a
+        # paste output here
+
+        KVM version:
+        $ /usr/bin/qemu-system-x86_64 --version
```
ludfjig: Is this really the kvm version? Unless I am mistaken, kvm version is tied to kernel version. Idea: Maybe we can create some bash/python kind of script and check it in, that users can run to get all information we need.

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,71 @@
...
+      description: The version of the OS you are using.
+      value: |
+        <details>
+
+        ```console
+        On Linux:
+        $ cat /etc/os-release
+        # paste output here
+
+        $ cat /uname -a
```
ludfjig: I think just `uname -a` is better?

.github/ISSUE_TEMPLATE/bug.yaml
```diff
@@ -0,0 +1,71 @@
+name: Bug Report
```
ludfjig, line 1: Should we add a "how to reproduce" section or something like that? Perhaps a link to a gist/repo or something along those lines.

ludfjig (dismissed):
lgtm with some minor questons

.github/ISSUE_TEMPLATE/bug.yaml
marosset: yes, copy paste issue :-/

.github/ISSUE_TEMPLATE/bug.yaml
marosset: I can update to `kvm --version` but that gives the exact same output

ludfjig: I think this is the qemu version, not kvm

marosset: updated

ludfjig: sorry what I meant is that both are actually the qemu version

marosset: Ah,
Should I update the description to say qemu version? or do you know a better way to get versions we want?

syntactically: There isn't a separate kvm version from the kernel version, so uname -a will capture it.

simongdavies (approved):

ludfjig (approved):

PR #97, opened 2024-12-04
Add many missing musl files

ludfjig:
It's currently hard to compile certain code as a hyperlight guest, because big parts of libc is missing. This PR aims to add more parts of libc to hyperlight guest library


syntactically:
I wonder if we can switch to using git-subtree more properly here, especially if we don't care about removing so many files.

ludfjig:
Closing in favor of #104

PR #103, opened 2024-12-10
Add fast trace collection & use it for instrumenting guest memory operations

syntactically:
This adds a framework for quickly tracing events of interest in a guest, including getting stack traces out of the guest based on dwarf unwinding information. This capability is used to allow collecting stack traces of memory allocations/frees, providing various statistics on what causes the maximum memory consumption for a sandbox.

### DIFF
--- a/src/hyperlight_guest/src/lib.rs
+++ b/src/hyperlight_guest/src/lib.rs
@@ -83,8 +83,66 @@ fn panic(info: &core::panic::PanicInfo) -> ! {
 }
 
 // Globals
+#[cfg(feature = "mem_profile")]
+struct ProfiledLockedHeap<const ORDER: usize>(LockedHeap<ORDER>);
+#[cfg(feature = "mem_profile")]
+unsafe impl<const ORDER: usize> alloc::alloc::GlobalAlloc for ProfiledLockedHeap<ORDER> {
+    unsafe fn alloc(&self, layout: core::alloc::Layout) -> *mut u8 {
+        let addr = self.0.alloc(layout);
+        unsafe {
+            core::arch::asm!("out dx, al",
+                in("dx") OutBAction::TraceMemoryAlloc as u16,
+                in("rax") layout.size() as u64,
+                in("rcx") addr as u64);
+        }
+        addr
+    }
+    unsafe fn dealloc(&self, ptr: *mut u8, layout: core::alloc::Layout) {
+        unsafe {
+            core::arch::asm!("out dx, al",
+                in("dx") OutBAction::TraceMemoryFree as u16,
+                in("rax") layout.size() as u64,
+                in("rcx") ptr as u64);
+        }
+        self.0.dealloc(ptr, layout)
+    }
+    unsafe fn alloc_zeroed(&self, layout: core::alloc::Layout) -> *mut u8 {
+        let addr = self.0.alloc_zeroed(layout);
+        unsafe {
+            core::arch::asm!("out dx, al",
+                in("dx") OutBAction::TraceMemoryAlloc as u16,
+                in("rax") layout.size() as u64,
+                in("rcx") addr as u64);
+        }
+        addr
+    }
+    unsafe fn realloc(
+        &self,
+        ptr: *mut u8,
+        layout: core::alloc::Layout,
+        new_size: usize,
+    ) -> *mut u8 {
+        let new_ptr = self.0.realloc(ptr, layout, new_size);
+        unsafe {
+            core::arch::asm!("out dx, al",
+                in("dx") OutBAction::TraceMemoryFree as u16,
+                in("rax") layout.size() as u64,
+                in("rcx") ptr);
+            core::arch::asm!("out dx, al",
+                in("dx") OutBAction::TraceMemoryAlloc as u16,
+                in("rax") new_size as u64,
+                in("rcx") new_ptr);
+        }
+        new_ptr
+    }
+}
+#[cfg(not(feature = "mem_profile"))]
 #[global_allocator]
 pub(crate) static HEAP_ALLOCATOR: LockedHeap<32> = LockedHeap::<32>::empty();
+#[cfg(feature = "mem_profile")]
+#[global_allocator]
+pub(crate) static HEAP_ALLOCATOR: ProfiledLockedHeap<32> =
+    ProfiledLockedHeap(LockedHeap::<32>::empty());
 
 ///cbindgen:ignore
 #[no_mangle]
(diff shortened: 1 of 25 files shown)


ludfjig (approved):
This looks great to me. Stack unwinding will definitely be very helpful. Btw, it's not totally clear to me how to use the two added features, nor the new trace_dump crate. Would it be possible to add a simple example or minimal docs somewhere? Lastly, should we try to compile some feature matrix of this in CI to make sure we don't break it in the future? Great job 👍

src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -146,6 +148,35 @@ pub fn is_hypervisor_present() -> bool {
...
 }
 
+#[cfg(feature = "trace_guest")]
+#[derive(Clone)]
+/// The information that trace collection requires in order to write
+/// an accurate trace.
+pub(crate) struct TraceInfo {
+    /// The epoch against which trace events are timed; at least as
+    /// early as the creation of the sandbox being traced.
+    #[allow(dead_code)]
```
simongdavies, line 162: Curious why these need allow(dead_code)?

src/hyperlight_host/src/mem/elf.rs
```diff
@@ -23,13 +26,85 @@ use goblin::elf64::program_header::PT_LOAD;
...
+    va_size: u64,
+    base_svma: u64,
+    shdrs: Vec<ResolvedSectionHeader>,
+}
+
+#[cfg(feature = "unwind_guest")]
+impl super::exe::UnwindInfo for UnwindInfo {
+    fn as_module(&self) -> framehop::Module<Vec<u8>> {
+        framehop::Module::new(
+            // TODO: plumb through a name from from_file if this
```
simongdavies, line 59: Is this todo still relevant? Do we need a follow up to make the file details available?

src/hyperlight_host/src/mem/elf.rs
```diff
@@ -23,13 +26,85 @@ use goblin::elf64::program_header::PT_LOAD;
...
+    fn base_svma(&self) -> u64 {
+        self.base_svma
+    }
+    fn section_svma_range(&mut self, name: &[u8]) -> Option<std::ops::Range<u64>> {
+        let shdr = self.resolved_section_header(name)?;
+        Some(shdr.addr..shdr.addr + shdr.size)
+    }
+    fn section_data(&mut self, name: &[u8]) -> Option<Vec<u8>> {
+        if name == b".eh_frame" && self.resolved_section_header(b".debug_frame").is_some() {
+            /* Rustc does not always emit enough information for stack
```
simongdavies, line 92: Does this imply that we need debug builds of the guest to make this work?

src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -160,20 +164,53 @@ pub(crate) struct TraceInfo {
     /// The file to which the trace is being written
     #[allow(dead_code)]
     pub file: Arc<Mutex<std::fs::File>>,
```
simongdavies, line 166: should this be behind the unwind_guest feature flag? Also as before not clear why we need allow(dead_code) here?

src/hyperlight_guest/src/lib.rs
```diff
@@ -83,8 +83,66 @@ fn panic(info: &core::panic::PanicInfo) -> ! {
 }
 
 // Globals
+#[cfg(feature = "mem_profile")]
+struct ProfiledLockedHeap<const ORDER: usize>(LockedHeap<ORDER>);
+#[cfg(feature = "mem_profile")]
+unsafe impl<const ORDER: usize> alloc::alloc::GlobalAlloc for ProfiledLockedHeap<ORDER> {
+    unsafe fn alloc(&self, layout: core::alloc::Layout) -> *mut u8 {
+        let addr = self.0.alloc(layout);
```
simongdavies, line 91: Heap provides some statistics (alloc, user, total) about allocations , it might be useful to emit those (maybe total at least so we can see how big the heap got?)

simongdavies (commented):
This is really great @syntactically , I added a few review comments.

I agree with @ludfjig it would be nice to have some docs that explain how to use these features and how they interact (it isn't immediately obvious to me at least if the `trace_guest` feature is useful on its own).

I also think it would be useful to perhaps have a README for the trace_dump tool that specifies how to use it , last I think we should maybe include the trace_dump binary in the GH release?

src/hyperlight_host/src/sandbox/mod.rs
syntactically: At least right now, if you build with trace_guest and on other features, there is no code that emits trace events compiled in so the trace sink structure is unused.

src/hyperlight_host/src/mem/elf.rs
syntactically: I couldn't find much this was actually being used for, so I don't think it actually matters for now; at the same time just coming up with a random name like this is a bit unpleasant, so I think it justifies a comment but not doing more work unless we find a use for having a better name here.

src/hyperlight_host/src/mem/elf.rs
syntactically: I think so, at least for now. Because we have panic="abort" in the guest, release builds don't need to support unwinding the stack at all, since panic is the only native stack unwinding mechanism in rust. There is a rust codegen flag `-C force-unwind-tables` that can probably force it on dev builds, but I don't know an easy way to convince Cargo to apply that flag just for feature = unwind_guest builds.

src/hyperlight_host/src/sandbox/mod.rs
syntactically: I think it makes sense to have the basic trace sink infrastructure, which could be used by other trace sources in the future (e.g. trace every page table change, trace every write to a register, etc), factored out from them the specific unwinding source. The dead_code is there for the same reason, which is that building with trace_guest to get the common trace sink infrastructure, but no actual trace sources, makes this dead.

src/hyperlight_guest/src/lib.rs
syntactically: The max size of the user heap is easy to calculate from the trace file. The overhead from the buddy system allocator is probably worth getting those kind of statistics for at least occasionally, but I also don't know if it's worth defining a trace packet format for them when we're planning on replacing the allocator in the near future?

simongdavies:
@syntactically can we get this updated and merged please?

simongdavies:
@dblnz can this be closed now your trace work is merged?

syntactically:
This was rolled into #695

PR #104, opened 2024-12-10
Musl subtree

ludfjig:
Adds musl as a git subtree. Adds more of libc, as required by for example a quickjs guest.

The 6 commits does the following

1. Removes current musl
2. Adds musl v1.2.5 (latest) as a git subtree (squashed)
3. Merge commit for git subtree
4. Removes musl files we don't need
5. Patches musl for Hyperlight
6. Updates docs/build.rs/justfiles

[having some problems gpg signing the squash commit...]


marosset (approved):
LGTM (but I mostly just reviewed https://github.com/hyperlight-dev/hyperlight/pull/104/commits/6fa239ed7299b1619b89a616c0098e8f1541f86b)

src/hyperlight_guest/third_party/musl/arch/x86_64/bits/float.h
```diff
@@ -4,6 +4,22 @@
 #define FLT_EVAL_METHOD 0
 #endif
 
+#ifdef _MSC_VER
```
simongdavies, line 7: Do we need this one now since we are not use MSVC any longer?

simongdavies (approved):
One small comment

src/hyperlight_guest/third_party/musl/arch/x86_64/bits/float.h
ludfjig: I believe we need it, because without it clang-cl gives a bunch of warnings about floating point magnitudes

PR #106, opened 2024-12-11
Remove custom alloca

ludfjig:
Removes custom alloca in favor of clang builtin alloca. Also removes the cargo alloca feature.
The test guest StackAllocate fn is now only available in the c guest.

Requires #104


src/hyperlight_host/tests/integration_test.rs
```diff
@@ -266,76 +268,30 @@ fn guest_malloc_abort() {
...
-    let bytes = 10_000; // some low number that can be allocated on stack
-
-    sbox.call_guest_function_by_name(
-        "StackAllocate",
-        ReturnType::Int,
-        Some(vec![ParameterValue::Int(bytes)]),
-    )
-    .unwrap();
-}
-
```
simongdavies, line 283: Since we still hve this in C guest should we not have tests for the C guest case?

ludfjig: I kept the test for the c guest, test `dynamic_stack_allocate_c_guest` tests a passing small value, and makes sure large values give an error of overflow

syntactically (approved):

PR #108, opened 2024-12-12
Removes requirement of specific environment variables during testing

ludfjig:
These environment `KVM_SHOULD_BE_PRESENT` `HYPERV_SHOULD_BE_PRESENT` variables just add unnecessary complexity, especially for newcomers who just cloned the repo.  They're not very useful anyway

### DIFF
--- a/src/hyperlight_host/src/sandbox/mod.rs
+++ b/src/hyperlight_host/src/sandbox/mod.rs
@@ -158,12 +158,6 @@ mod tests {
     use crossbeam_queue::ArrayQueue;
     use hyperlight_testing::simple_guest_as_string;
 
-    #[cfg(target_os = "linux")]
-    use super::is_hypervisor_present;
-    #[cfg(mshv)]
-    use crate::hypervisor::hyperv_linux::test_cfg::TEST_CONFIG as HYPERV_TEST_CONFIG;
-    #[cfg(kvm)]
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -177,29 +171,19 @@ mod tests {
     #[test]
     // TODO: add support for testing on WHP
     #[cfg(target_os = "linux")]
-    fn test_is_hypervisor_present() {
+    fn is_hypervisor_present() {
         // TODO: Handle requiring a stable API
```
simongdavies: This comment can go since mshv got rid of the stable api flag

simongdavies (changes requested):
One small nit

src/hyperlight_host/src/sandbox/mod.rs
ludfjig: fixed

simongdavies (approved):

simongdavies (approved):

simongdavies:
@ludfjig once this merges we should update the common workflow setup to get rid of the variables there too

dblnz (dismissed):

ludfjig:
> @ludfjig once this merges we should update the common workflow setup to get rid of the variables there too

https://github.com/hyperlight-dev/ci-setup-workflow/pull/4

syntactically (approved):

marosset (approved):

PR #111, opened 2024-12-13
Add support for Hyperlight KVM guest debugging using gdb 

dblnz:
The current implementation supports:
- add/remove HW breakpoints (max 4)
- add/remove SW breakpoints
- read/write registers
- read/write addresses
- step/continue
- get code offset from target

The overall architecture is designed to work like a `Request - Response` protocol from the gdb thread to the hypervisor handler over a `DebugCommChannel`.
- All the functionality is implemented on the hypervisor side so it has access to the shared memory and VcpuFd
- The gdb thread uses `gdbstub` to handle the communication with the gdb client
- when the gdb client requests the supported features mentioned above, a request is sent forward to the hypervisor handler for the running sandbox to run

### DIFF
--- a/docs/how-to-debug-a-hyperlight-guest.md
+++ b/docs/how-to-debug-a-hyperlight-guest.md
@@ -1,0 +1,193 @@
+# How to debug a Hyperlight **KVM** guest using gdb
+
+Hyperlight supports gdb debugging of a **KVM** guest running inside a Hyperlight sandbox.
+When Hyperlight is compiled with the `gdb` feature enabled, a Hyperlight KVM sandbox can be configured
+to start listening for a gdb connection.
+
+## Supported features
+
+The Hyperlight `gdb` feature enables **KVM** guest debugging:
+   - an entry point breakpoint is automatically set for the guest to stop
+   - add and remove HW breakpoints (maximum 4 set breakpoints at a time)
+   - add and remove SW breakpoints
+   - read and write registers
+   - read and write addresses
+   - step/continue
+   - get code offset from target
+
+## Expected behavior
+
+Below is a list describing some cases of expected behavior from a gdb debug 
+session of a guest binary running inside a KVM Hyperlight sandbox.
+
+- when the `gdb` feature is enabled and a SandboxConfiguration is provided a
+  debug port, the created sandbox will wait for a gdb client to connect on the
+  configured port
+- when the gdb client attaches, the guest vCPU is expected to be stopped at the
+  entry point
+- if a gdb client disconnects unexpectedly, the debug session will be closed and
+  the guest will continue executing disregarding any prior breakpoints
+- if multiple sandbox instances are created, each instance will have its own
+  gdb thread listening on the configured port
+- if two sandbox instances are created with the same debug port, the second
+  instance logs an error and the gdb thread will not be created, but the sandbox
+  will continue to run without gdb debugging
+
+## Example
+
+### Sandbox configuration
+
+The `guest-debugging` example in Hyperlight demonstrates how to configure a Hyperlight
+sandbox to listen for a gdb client on a specific port.
+
+### CLI Gdb configuration
+
+One can use a gdb config file to provide the symbols and desired configuration.
+
+The below contents of the `.gdbinit` file can be used to provide a basic configuration
+to gdb startup.
+
+```gdb
+# Path to symbols
+file path/to/symbols.elf
+# The port on which Hyperlight listens for a connection
+target remote :8080
+set disassembly-flavor intel
+set disassemble-next-line on
+enable pretty-printer
+layout src
+```
+One can find more information about the `.gdbinit` file at [gdbinit(5)](https://www.man7.org/linux/man-pages/man5/gdbinit.5.html).
+
+### End to end example
+
+Using the example mentioned at [Sandbox configuration](#sandbox-configuration) 
+one can run the below commands to debug the guest binary:
+
+```bash
+# Terminal 1
+$ cargo run --example guest-debugging --features gdb
+```
+
+```bash
+# Terminal 2
+$ cat .gdbinit
+file src/tests/rust_guests/bin/debug/simpleguest
+target remote :8080
+set disassembly-flavor intel
+set disassemble-next-line on
+enable pretty-printer
+layout src
+
+$ gdb
+```
+
+### Using VSCode to debug a Hyperlight guest
+
+To replicate the above behavior using VSCode follow the below steps:
+- install the `gdb` package on the host machine
+- install the `C/C++` extension in VSCode to add debugging capabilities
+- create a `.vscode/launch.json` file in the project directory with the below content:
+    ```json
+    {
+        "version": "0.2.0",
+        "configurations": [
+            {
+                "name": "GDB",
+                "type": "cppdbg",
+                "request": "launch",
+                "program": "${workspaceFolder}/src/tests/rust_guests/bin/debug/simpleguest",
+                "args": [],
+                "stopAtEntry": true,
+                "hardwareBreakpoints": {"require": false, "limit": 4},
+                "cwd": "${workspaceFolder}",
+                "environment": [],
+                "externalConsole": false,
+                "MIMode": "gdb",
+                "miDebuggerPath": "/usr/bin/gdb",
+                "miDebuggerServerAddress": "localhost:8080",
+                "setupCommands": [
+                    {
+                        "description": "Enable pretty-printing for gdb",
+                        "text": "-enable-pretty-printing",
+                        "ignoreFailures": true
+                    },
+                    {
+                        "description": "Set Disassembly Flavor to Intel",
+                        "text": "-gdb-set disassembly-flavor intel",
+                        "ignoreFailures": true
+                    }
+                ]
+            }
+        ]
+    }
+    ```
+- in `Run and Debug` tab, select the `GDB` configuration and click on the `Run`
+  button to start the debugging session.
+  The gdb client will connect to the Hyperlight sandbox and the guest vCPU will
+  stop at the entry point.
+
+
+## How it works
+
+The gdb feature is designed to work like a Request - Response protocol between
+a thread that accepts commands from a gdb client and the hypervisor handler over
+a communication channel.
+
+All the functionality is implemented on the hypervisor side so it has access to
+the shared memory and the vCPU.
+
+The gdb thread uses the `gdbstub` crate to handle the communication with the gdb client.
+When the gdb client requests one of the supported features mentioned above, a request
+is sent over the communication channel to the hypervisor handler for the sandbox
+to resolve.
+
+Below is a sequence diagram that shows the interaction between the entities
+involved in the gdb debugging of a Hyperlight guest running inside a KVM sandbox.
+
+```
+                               ┌───────────────────────────────────────────────────────────────────────────────────────────────┐
+                               │                                       Hyperlight Sandbox                                      │
+     USER                      │                                                                                               │
+┌────────────┐                 │  ┌──────────────┐                      ┌───────────────────────────┐              ┌────────┐  │
+│ gdb client │                 │  │  gdb thread  │                      │ hypervisor handler thread │              │  vCPU  │  │
+└────────────┘                 │  └──────────────┘                      └───────────────────────────┘              └────────┘  │
+      |                        │          |               create_gdb_thread           |                                 |      │
+      |                        │          |◄─────────────────────────────────────────┌─┐         vcpu stopped          ┌─┐     │
+      |    attach              │         ┌─┐                                         │ │◄──────────────────────────────┴─┘     │
+     ┌─┐───────────────────────┼────────►│ │                                         │ │     entrypoint breakpoint      |      │
+     │ │   attach response     │         │ │                                         │ │                                |      │
+     │ │◄──────────────────────┼─────────│ │                                         │ │                                |      │
+     │ │                       │         │ │                                         │ │                                |      │
+     │ │  add_breakpoint       │         │ │                                         │ │                                |      │
+     │ │───────────────────────┼────────►│ │          add_breakpoint                 │ │                                |      │
+     │ │                       │         │ │────────────────────────────────────────►│ │  add_breakpoint                |      │
+     │ │                       │         │ │                                         │ │────┐                           |      │
+     │ │                       │         │ │                                         │ │    │                           |      │
+     │ │                       │         │ │                                         │ │◄───┘                           |      │
+     │ │                       │         │ │          add_breakpoint response        │ │                                |      │
+     │ │  add_breakpoint response        │ │◄────────────────────────────────────────│ │                                |      │
+     │ │◄──────────────────────┬─────────│ │                                         │ │                                |      │
+     │ │   continue            │         │ │                                         │ │                                |      │
+     │ │───────────────────────┼────────►│ │            continue                     │ │                                |      │
+     │ │                       │         │ │────────────────────────────────────────►│ │         resume vcpu            |      │
+     │ │                       │         │ │                                         │ │──────────────────────────────►┌─┐     │
+     │ │                       │         │ │                                         │ │                               │ │     │
+     │ │                       │         │ │                                         │ │                               │ │     │
+     │ │                       │         │ │                                         │ │                               │ │     │
+     │ │                       │         │ │                                         │ │                               │ │     │
+     │ │                       │         │ │                                         │ │         vcpu stopped          │ │     │
+     │ │                       │         │ │        notify vcpu stop reason          │ │◄──────────────────────────────┴─┘     │
+     │ │   notify vcpu stop reason       │ │◄────────────────────────────────────────│ │                                |      │
+     │ │◄──────────────────────┬─────────│ │                                         │ │                                |      │
+     │ │    continue until end │         │ │                                         │ │                                |      │
+     │ │───────────────────────┼────────►│ │            continue                     │ │         resume vcpu            |      │
+     │ │                       │         │ │────────────────────────────────────────►│ │──────────────────────────────►┌─┐     │
+     │ │                       │         │ │                                         │ │                               │ │     │
+     │ │                       │         │ │        comm channel disconnected        │ │         vcpu halted           │ │     │
+     │ │   target finished exec│         │ │◄────────────────────────────────────────┤ │◄──────────────────────────────┴─┘     │
+     │ │◄──────────────────────┼─────────┴─┘          target finished exec           └─┘                                |      │
+     │ │                       │          |                                           |                                 |      │
+     └─┘                       │          |                                           |                                 |      │
+      |                        └───────────────────────────────────────────────────────────────────────────────────────────────┘
+```
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -71,6 +71,8 @@ sha256 = "1.4.0"
 windows-version = "0.1"
 
 [target.'cfg(unix)'.dependencies]
+gdbstub = { version = "0.7.3", optional = true }
+gdbstub_arch = { version = "0.3.1", optional = true }
 seccompiler = { version = "0.4.0", optional = true }
 kvm-bindings = { version = "0.11", features = ["fam-wrappers"], optional = true }
 kvm-ioctls = { version = "0.20", optional = true }
@@ -128,8 +130,8 @@ kvm = ["dep:kvm-bindings", "dep:kvm-ioctls"]
 mshv2 = ["dep:mshv-bindings2", "dep:mshv-ioctls2"]
 mshv3 = ["dep:mshv-bindings3", "dep:mshv-ioctls3"]
 inprocess = []
-# This enables compilation of gdb stub for easy debug in the guest
-gdb = []
+# This enables easy debug in the guest
+gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 
 [[bench]]
 name = "benchmarks"
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -128,6 +128,8 @@ kvm = ["dep:kvm-bindings", "dep:kvm-ioctls"]
 mshv2 = ["dep:mshv-bindings2", "dep:mshv-ioctls2"]
 mshv3 = ["dep:mshv-bindings3", "dep:mshv-ioctls3"]
 inprocess = []
+# This enables compilation of gdb stub for easy debug in the guest
+gdb = []
 
 [[bench]]
 name = "benchmarks"
--- a/src/hyperlight_host/build.rs
+++ b/src/hyperlight_host/build.rs
@@ -89,6 +89,7 @@ fn main() -> Result<()> {
     // Essentially the kvm and mshv features are ignored on windows as long as you use #[cfg(kvm)] and not #[cfg(feature = "kvm")].
     // You should never use #[cfg(feature = "kvm")] or #[cfg(feature = "mshv")] in the codebase.
     cfg_aliases::cfg_aliases! {
+        gdb: { all(feature = "gdb", debug_assertions, feature = "kvm", target_os = "linux") },
         kvm: { all(feature = "kvm", target_os = "linux") },
         mshv: { all(any(feature = "mshv2", feature = "mshv3"), target_os = "linux") },
         // inprocess feature is aliased with debug_assertions to make it only available in debug-builds.
--- a/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
@@ -101,6 +101,9 @@ pub fn event_loop_thread(
         Ok(disconnect_reason) => match disconnect_reason {
             DisconnectReason::Disconnect => { 
                 log::info!("Gdb client disconnected");
+                if let Err(e) = target.disable_debug() {
+                    log::error!("Cannot disable debugging: {:?}", e);
+                }
             }
             DisconnectReason::TargetExited(_) => {
                 log::info!("Guest finalized execution and disconnected");
--- a/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
@@ -1,0 +1,117 @@
+/*
+Copyright 2024 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use gdbstub::conn::ConnectionExt;
+use gdbstub::stub::run_blocking::{self, WaitForStopReasonError};
+use gdbstub::stub::{BaseStopReason, DisconnectReason, GdbStub, SingleThreadStopReason};
+
+use super::x86_64_target::HyperlightSandboxTarget;
+use super::{DebugResponse, VcpuStopReason};
+
+pub struct GdbBlockingEventLoop;
+
+impl run_blocking::BlockingEventLoop for GdbBlockingEventLoop {
+    type Connection = Box<dyn ConnectionExt<Error = std::io::Error>>;
+    type StopReason = SingleThreadStopReason<u64>;
+    type Target = HyperlightSandboxTarget;
+
+    fn wait_for_stop_reason(
+        target: &mut Self::Target,
+        conn: &mut Self::Connection,
+    ) -> Result<
+        run_blocking::Event<Self::StopReason>,
+        run_blocking::WaitForStopReasonError<
+            <Self::Target as gdbstub::target::Target>::Error,
+            <Self::Connection as gdbstub::conn::Connection>::Error,
+        >,
+    > {
+        loop {
+            match target.try_recv() {
+                Ok(DebugResponse::VcpuStopped(stop_reason)) => {
+                    log::debug!("VcpuStopped with reason {:?}", stop_reason);
+
+                    // Resume execution if unknown reason for stop
+                    let stop_response = match stop_reason {
+                        VcpuStopReason::DoneStep => BaseStopReason::DoneStep,
+                        VcpuStopReason::SwBp => BaseStopReason::SwBreak(()),
+                        VcpuStopReason::HwBp => BaseStopReason::HwBreak(()),
+                        VcpuStopReason::Unknown => {
+                            target
+                                .resume_vcpu()
+                                .map_err(WaitForStopReasonError::Target)?;
+
+                            continue;
+                        }
+                    };
+
+                    return Ok(run_blocking::Event::TargetStopped(stop_response));
+                }
+                Ok(msg) => {
+                    log::error!("Unexpected message received {:?}", msg);
+                }
+                Err(crossbeam_channel::TryRecvError::Empty) => (),
+                Err(crossbeam_channel::TryRecvError::Disconnected) => {
+                    return Ok(run_blocking::Event::TargetStopped(BaseStopReason::Exited(
+                        0,
+                    )));
+                }
+            }
+
+            // Check if there is any data to read from the connection
+            // If there is, return the data as an incoming data event
+            // Otherwise, continue waiting for a stop reason from the target
+            if conn.peek().map(|b| b.is_some()).unwrap_or(false) {
+                let byte = conn
+                    .read()
+                    .map_err(run_blocking::WaitForStopReasonError::Connection)?;
+
+                return Ok(run_blocking::Event::IncomingData(byte));
+            }
+        }
+    }
+
+    /// Handle an interrupt from the GDB client.
+    /// This function is called when the GDB client sends an interrupt signal.
+    /// Passing `None` defers sending a stop reason to later (e.g. when the target stops).
+    fn on_interrupt(
+        _target: &mut Self::Target,
+    ) -> Result<Option<Self::StopReason>, <Self::Target as gdbstub::target::Target>::Error> {
+        Ok(None)
+    }
+}
+
+pub fn event_loop_thread(
+    debugger: GdbStub<HyperlightSandboxTarget, Box<dyn ConnectionExt<Error = std::io::Error>>>,
+    target: &mut HyperlightSandboxTarget,
+) {
+    match debugger.run_blocking::<GdbBlockingEventLoop>(target) {
+        Ok(disconnect_reason) => match disconnect_reason {
+            DisconnectReason::Disconnect => { 
+                log::info!("Gdb client disconnected");
+            }
+            DisconnectReason::TargetExited(_) => {
+                log::info!("Guest finalized execution and disconnected");
+            }
+            DisconnectReason::TargetTerminated(sig) => {
+                log::info!("Gdb target terminated with signal {}", sig)
+            }
+            DisconnectReason::Kill => log::info!("Gdb sent a kill command"),
+        },
+        Err(e) => {
+            log::error!("fatal error encountered: {e:?}");
+        }
+    }
+}
--- a/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
@@ -99,7 +99,7 @@ pub fn event_loop_thread(
 ) {
     match debugger.run_blocking::<GdbBlockingEventLoop>(target) {
         Ok(disconnect_reason) => match disconnect_reason {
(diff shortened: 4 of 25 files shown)


docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,26 @@
+# How to debug a Hyperlight guest
+
+Currently Hyperlight support gdb debugging of a guest running inside a Hyperlight sandbox.
```
devigned:
```suggestion
Hyperlight supports gdb debugging of a guest running inside a Hyperlight sandbox.
```

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,26 @@
...
+When the option is enabled Hyperlight starts listening on port `8081` for a gdb connection.
```
devigned:
```suggestion
When Hyperlight is compiled with the `gdb` feature enabled, Hyperlight starts listening on port `8081` for a gdb connection.
```

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,26 @@
...
+## Example
+```bash
+cargo run --example hello-world --features gdb
+```
+The execution will wait for gdb to attach.
+
+One can use a simple gdb config to provide the symbols and desired configuration:
+
+For the above example, when running from the repository root directory, the below contents
+of `.gdbinit` file can be used.
```
devigned: Please either explain what the file does specifically, or provide a link to .gdbinit file documentation.

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+            _ => Self::UnexpectedError,
+        }
+    }
+}
+
+impl From<DebugMessage> for GdbTargetError {
+    fn from(value: DebugMessage) -> Self {
+        match value {
+            DebugMessage::VcpuStoppedEv => GdbTargetError::UnexpectedMessageError,
+            _ => GdbTargetError::UnexpectedMessageError,
```
devigned: Why are you matching on value only to return the same error?

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -0,0 +1,184 @@
...
+
+/// Creates a thread that handles gdb protocol
+pub fn create_gdb_thread<T: GdbDebug + Send + 'static>(
+    mut target: T,
+) -> Result<(), <T as Target>::Error>
+where
+    <T as Target>::Error:
+        std::fmt::Debug + Send + From<io::Error> + From<DebugMessage> + From<TryRecvError>,
+{
+    // TODO: Address multiple sandboxes scenario
```
devigned: Rather than have a code comment, please create a new issue in the repo to track the feature.

src/hyperlight_host/src/hypervisor/gdb/target.rs
```diff
@@ -0,0 +1,536 @@
...
+    /// This method provides a way to set a breakpoint at the entrypoint
+    /// it does not keep this breakpoint set after the vcpu already stopped at the address
+    pub fn set_entrypoint_bp(&mut self) -> Result<bool, GdbTargetError> {
+        let mut entrypoint_debug = KvmDebug::new();
+        entrypoint_debug.set_breakpoints(&self.vcpu_fd.lock().unwrap(), &[self.entrypoint], false)
+    }
+
+    /// Translates the guest address to physical address
+    fn translate_gva(&self, gva: u64) -> Result<u64, GdbTargetError> {
+        // TODO: Properly handle errors
```
devigned: What would you like to do based on this TODO?

A related nit, it would be nice to have some preservation of the data in the underlying error, so as the error bubbles up, we can determine the root cause.

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -88,7 +88,8 @@ impl HypervisorHandler {
 #[derive(Clone)]
 struct HvHandlerExecVars {
     join_handle: Arc<Mutex<Option<JoinHandle<Result<()>>>>>,
-    shm: Arc<Mutex<Option<SandboxMemoryManager<GuestSharedMemory>>>>,
+    #[allow(clippy::type_complexity)] // TODO: Change this type
+    shm: Arc<Mutex<Option<Arc<Mutex<SandboxMemoryManager<GuestSharedMemory>>>>>>,
```
devigned: Do you intend on changing this prior to PR merging?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -112,9 +125,30 @@ impl KVMDriver {
...
+        mgr: Arc<Mutex<SandboxMemoryManager<GuestSharedMemory>>>,
+        vcpu_fd: Arc<Mutex<VcpuFd>>,
+        entrypoint: u64,
+    ) -> Result<GdbConnection> {
+        let (gdb_conn, hyp_conn) = GdbConnection::new_pair();
+
+        let mut target = HyperlightKvmSandboxTarget::new(mgr, vcpu_fd, entrypoint, hyp_conn);
+        let _ = target
+            .set_entrypoint_bp()
+            .map_err(|_| new_error!("Cannot set entrypoint breakpoint"))?;
```
devigned: Is there useful information in the underlying error?

devigned (changes requested):
Super excited to see this landing. Nice work! My feedback mostly consists of nits and questions.

src/hyperlight_host/src/hypervisor/gdb/mod.rs
dblnz: This was something I missed, I'll change it, thanks for pointing it out

src/hyperlight_host/src/hypervisor/gdb/mod.rs
dblnz: I'll do that after I gather all the feedback as this might be addressed before merging

src/hyperlight_host/src/hypervisor/gdb/target.rs
dblnz: This was a left over comment from the development phase, I added some error handling in the meantime. 
I would like to maybe get an idea as to how to redo the error handling to be easier to track where errors originated

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
dblnz: I would like to change it.
Cloning  `SandboxMemoryManager` directly might be a better idea, but the `GuestSharedMemory` type does not implement Clone, I need to check what implementing that would mean.

src/hyperlight_host/src/hypervisor/kvm.rs
dblnz: It is a KVM generated error, I need to think about how to propagate the errors

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
danbugs: Does `KVMDriver` need a clone of SandboxMemoryManager or does a mutable borrow maybe suffice? `GuestSharedMemory` was purposefully not `Clone` because it was intended to reflect unique ownership at any time of the guest side access to the shared memory.

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -545,6 +545,16 @@ impl HypervisorHandler {
...
+        #[cfg(gdb)]
+        match self.communication_channels.from_handler_rx.recv() {
+            Ok(msg) => match msg {
+                HandlerMsg::Error(e) => Err(e),
+                HandlerMsg::FinishedHypervisorHandlerAction => Ok(()),
+            },
+            Err(_) => Err(HyperlightError::HypervisorHandlerMessageReceiveTimedout()),
+        }
+
+        #[cfg(not(gdb))]
```
danbugs: Disabling the timeout might be helpful for scenarios other than debugging too. Might want to put this under a different feature flag or even as a runtime option. Though, that's not too related to this PR, so maybe leaving it like this is OK for now 👍

src/hyperlight_host/Cargo.toml
```diff
@@ -71,6 +71,8 @@ sha256 = "1.4.0"
 windows-version = "0.1"
 
 [target.'cfg(unix)'.dependencies]
+gdbstub = "0.7.3"
```
ludfjig: Is it possible to make these optional dependencies that are only included when the `gdb` feature is turned on?

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -0,0 +1,193 @@
...
+}
+
+impl From<GdbTargetError> for TargetError<GdbTargetError> {
+    fn from(value: GdbTargetError) -> TargetError<GdbTargetError> {
+        TargetError::Io(std::io::Error::other(value))
+    }
+}
+
+#[derive(Debug, Default)]
+/// Struct that contains the x86_64 core registers
```
ludfjig, line 69: nit: I personally prefer the comment above the #[derive], but up to you :D

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -136,6 +139,9 @@ impl SandboxConfiguration {
     pub const MIN_KERNEL_STACK_SIZE: usize = 0x1000;
     /// The default value for kernel stack size
     pub const DEFAULT_KERNEL_STACK_SIZE: usize = Self::MIN_KERNEL_STACK_SIZE;
+    #[cfg(gdb)]
+    /// The minimum value for debug port
+    pub const MIN_GUEST_DEBUG_PORT: u16 = 9000;
```
ludfjig: Is there a reason the port must be >=9000? Otherwise the name could be DEFAULT_GUEST_DEBUG_PORT

src/hyperlight_host/src/hypervisor/gdb/x86_64_target.rs
```diff
@@ -0,0 +1,342 @@
...
+                regs.regs[9] = read_regs.r9;
+                regs.regs[10] = read_regs.r10;
+                regs.regs[11] = read_regs.r11;
+                regs.regs[12] = read_regs.r12;
+                regs.regs[13] = read_regs.r13;
+                regs.regs[14] = read_regs.r14;
+                regs.regs[15] = read_regs.r15;
+                regs.rip = read_regs.rip;
+                regs.eflags = u32::try_from(read_regs.rflags)
+                    .expect("Couldn't convert rflags from u64 to u32");
```
ludfjig: Perhaps we should return an error instead of panic here

ludfjig (dismissed):
This is awesome!! Looked briefly though everything and I will give a more in-depth review tomorrow

src/hyperlight_host/Cargo.toml
dblnz: good point, I'll do that

src/hyperlight_host/src/sandbox/config.rs
dblnz: I wanted to have this constraint to not interfere with any other services that might be running on the machine, but I thing the MIN should be around 1024.

src/hyperlight_host/src/hypervisor/gdb/x86_64_target.rs
dblnz: I've missed this, I'll change it

src/hyperlight_host/src/hypervisor/gdb/mod.rs
dblnz: You are right, it looks weird this way. I changed it

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -346,6 +352,12 @@ impl SandboxConfiguration {
         );
     }
 
+    #[instrument(skip_all, parent = Span::current(), level= "Trace")]
+    /// Sets the configuration for the guest debug
+    pub fn set_guest_debug_port(&mut self, port: u16) {
+        self.guest_debug_port = Some(max(port, Self::MIN_GUEST_DEBUG_PORT));
```
ludfjig: I called this function but forgot to enable the gdb feature. The code ran fine, but didn't wait for gdb to connect, and I got confused. In my opinion, maybe we should feature-gate this function, or return error somewhere (maybe not in this func) if a user is trying to use a sandbox with this setting set, without using the gdb feature. Maybe others have better ideas. Maybe some tests for this could be good to

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,111 @@
...
+set disassembly-flavor intel
+set disassemble-next-line on
+enable pretty-printer
+layout src
+```
+One can find more information about the `.gdbinit` file at [gdbinit(5)](https://www.man7.org/linux/man-pages/man5/gdbinit.5.html).
+
+### End to end example
+
+Using the [Sandbox configuration](#sandbox-configuration) above to configure the [hello-world](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_host/examples/hello-world/main.rs) example
```
ludfjig: I think it would be useful if we created a brand new example for this instead of using hello-world. I read through this doc but missed the part where it said "Using the sandbox configuration above", so I was confused why it wasn't working. What do you think?

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,111 @@
...
+
+```bash
+# Terminal 1
+$ cargo run --example hello-world --features gdb
+```
+
+```bash
+# Terminal 2
+$ cat .gdbinit
+file file src/tests/rust_guests/bin/debug/simpleguest
```
ludfjig: file file typo?

src/hyperlight_host/Cargo.toml
```diff
@@ -128,6 +130,8 @@ kvm = ["dep:kvm-bindings", "dep:kvm-ioctls"]
 mshv2 = ["dep:mshv-bindings2", "dep:mshv-ioctls2"]
 mshv3 = ["dep:mshv-bindings3", "dep:mshv-ioctls3"]
 inprocess = []
+# This enables compilation of gdb stub for easy debug in the guest
```
ludfjig: I think you can drop the first part of this comment, since we probably don't care too much about what compilation it enables, but rather that it allows for easy gdb debugging of a guest.

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -55,6 +65,527 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
+                    "The breakpoints array is larger than expected: {} >= {}",
+                    addrs.len(),
+                    Self::MAX_NO_OF_HW_BP
+                ));
+            }
+
+            self.dbg_cfg.arch.debugreg = [0; 8];
+            for (k, addr) in addrs.iter().enumerate() {
+                self.dbg_cfg.arch.debugreg[k] = *addr;
+                self.dbg_cfg.arch.debugreg[7] |= 1 << (k * 2);
```
ludfjig, line 139: Would you mind adding a comment about what these bit operations does/means?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -101,15 +638,36 @@ impl KVMDriver {
         let mut vcpu_fd = vm_fd.create_vcpu(0)?;
         Self::setup_initial_sregs(&mut vcpu_fd, pml4_addr)?;
 
+        #[cfg(gdb)]
+        let (debug, gdb_conn) = if let Some(DebugInfo { port }) = debug_info {
+            (
+                Some(debug::KvmDebug::new()),
+                Some(create_gdb_thread(*port).map_err(|_| new_error!("Cannot create GDB thread"))?),
```
ludfjig: To me it doesn't seem right to create the gdb thread from the KVM driver. In the future, we might support debugging on mshv, in which case we'd probably want to create this after the particular driver has been created. Do you think it's possible to move this somewhere else?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -55,6 +65,527 @@ pub(crate) fn is_hypervisor_present() -> bool {
     }
 }
 
+#[cfg(gdb)]
+mod debug {
```
ludfjig, line 67: Is there any chance we could move this to it's own file? I understand kvm_guest_debug struct is kvm-specific, but otherwise the rest seems pretty generic to me, and to me it makes sense moving the gdb specific stuff out of the kvm.rs driver code. Hopefully it's possible :P

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -55,6 +65,527 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
+        }
+
+        /// Gdb expects the target to be stopped when connected.
+        /// This method provides a way to set a breakpoint at the entrypoint
+        /// it does not keep this breakpoint set after the vcpu already stopped at the address
+        pub fn set_entrypoint_bp(&self) -> Result<()> {
+            if self.debug.is_some() {
+                log::debug!("Setting entrypoint bp {:X}", self.entrypoint);
+                let mut entrypoint_debug = KvmDebug::new();
+                entrypoint_debug.hw_breakpoints.push(self.entrypoint);
```
ludfjig: Could we make an method on KvmDebug for adding a breakpoint that does error checking (>4 breapoints), instead of manually modifying the vec? It seems fragile to do the error checking during the `set_breakpoints`, and to expose the inner vec like this.

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -55,6 +65,527 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
+            } else {
+                Ok(false)
+            }
+        }
+
+        pub fn remove_hw_breakpoint(&mut self, addr: u64) -> Result<bool> {
+            let addr = self.translate_gva(addr)?;
+
+            if let Some(debug) = self.debug.as_mut() {
+                if debug.hw_breakpoints.contains(&addr) {
```
ludfjig: Seems like the `.contains` is unnecessary given that you use `.position` later anyway. You might be able to simplify this whole block using `hw_breakpoints.retain(|a| a != addrv)` or something along those lines

src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
```diff
@@ -0,0 +1,118 @@
...
+                }
+                Err(crossbeam_channel::TryRecvError::Empty) => (),
+                Err(crossbeam_channel::TryRecvError::Disconnected) => {
+                    return Ok(run_blocking::Event::TargetStopped(BaseStopReason::Exited(
+                        0,
+                    )));
+                }
+            }
+
+            if conn.peek().map(|b| b.is_some()).unwrap_or(false) {
```
ludfjig, line 86: I'm not sure I really understand what this is doing. Maybe a comment here could help

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    RemoveSwBreakpoint(u64),
+    Step,
+    WriteAddr(u64, Vec<u8>),
+    WriteRegisters(X86_64Regs),
+}
+
+/// Enumerates the possible responses that a hypervisor can provide to a debugger
+#[derive(Debug)]
+pub enum DebugResponse {
+    AddHwBreakpoint(bool),
```
ludfjig, line 122: Do we need all of these responses, or would it be sufficient to combine some of them into something like Success()? I looked briefly and I didn't see `AddHwBreakpoint` and `AddSwBreakpoint` being used for anything for example. ( can see DebugResponse::Step | DebugResponse::Continue | DebugResponse::DisableDebug are being used though)

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,111 @@
...
+target remote :9050
+set disassembly-flavor intel
+set disassemble-next-line on
+enable pretty-printer
+layout src
+
+$ gdb
```
ludfjig, line 82: This markdown is great. One thing this gdb feature introduces is some complexity regarding threading. Would you be able to write down (or maybe create a diagram) for how the threads interact with each other, for example the handler thread and so on? It's a bit confusing to me which thread waits for what thread, etc :D. A simple list about what happens in what order when you create a sandbox with gdb enabled would be sufficient

ludfjig (approved):
This is awesome. I tried it out and it works great. Left some more comments, feel free to dismiss if you disagree with any of them. I realize some of them might be hard to implement as well. GREAT WORK!!

src/hyperlight_host/src/sandbox/config.rs
dblnz: This was the previous intention, however, I disabled it with the last commit.
The reasoning behind was that the rust-analyzer didn't know about the feature being enabled and it showed errors in IDE as it doesn't know the methods.
I am not sure which is the best approach, you might be right with being confusing when it doesn't do anything.

docs/how-to-debug-a-hyperlight-guest.md
dblnz: Good point, I can add another example

src/hyperlight_host/src/hypervisor/kvm.rs
dblnz: It doesn't need to be there anymore, since the big change.
However, the specific hypervisor would have to own the connection to `gdb` for it to be able to receive/send messages.

src/hyperlight_host/src/hypervisor/kvm.rs
dblnz: Well, part of it is possible, the KvmDebug type specifically, however the rest of the implementation is for the `KVMDriver` so it cannot be moved out of the file the type is defined, right?

src/hyperlight_host/src/hypervisor/kvm.rs
dblnz: good point, I'll change it 👍

src/hyperlight_host/src/hypervisor/kvm.rs
dblnz: thanks for the suggestion

src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
dblnz: I'll add a comment

src/hyperlight_host/src/hypervisor/gdb/mod.rs
dblnz: The need stems from the return value the `add_sw_breakpoint` and the others expect.
An example signature is:
```
    fn add_sw_breakpoint(
        &mut self,
        addr: <Self::Arch as Arch>::Usize,
        _kind: <Self::Arch as Arch>::BreakpointKind,
    ) -> TargetResult<bool, Self> {
```

and it expects true/false on the success path signaling whether the breakpoint is active, true being it was just added, false being it was already active and the new call didn't actually do anything.

I could put all of them together to make a generic response that returns `Result<bool, Error>` but I am not sure it would make it look a lot better getting rid of only 3 variants. Let me know what you think

docs/how-to-debug-a-hyperlight-guest.md
dblnz: this is a good idea, I thought about adding a diagram but I was afraid it would get desynchronized with the code when there would be a change as the documentation usually doesn't get updated unless it is close to the code being changed.

I'll work on a diagram to help with the understanding of the inner workings of the feature.

src/hyperlight_host/src/sandbox/config.rs
ludfjig: You should be able to enable the feature for rust-analyzer in .vscode/settings.json

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,111 @@
...
+
+Hyperlight supports gdb debugging of a guest running inside a Hyperlight sandbox.
+When Hyperlight is compiled with the `gdb` feature enabled, a Hyperlight sandbox can be configured
+to start listening for a gdb connection.
+
+## Supported features
+
+The Hyperlight `gdb` feature enables:
+
+1. KVM guest debugging:
```
simongdavies: Should we call out more clearly that this only works on KVM for now?

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,111 @@
...
+  configured port
+- when the gdb client attaches, the guest vCPU is expected to be stopped at the
+  entrypoint
+- if a gdb client disconnects unexpectedly, the debug session will be closed and
+  the guest will continue executing disregarding any prior breakpoints
+
+## How it works
+
+The gdb feature is designed to work like a Request - Response protocol between
+a thread that accepts commands from a gdb cliend and the hypervisor handler over
```
simongdavies:
```suggestion
a thread that accepts commands from a gdb client and the hypervisor handler over
```

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,111 @@
...
+$ cat .gdbinit
+file file src/tests/rust_guests/bin/debug/simpleguest
+target remote :9050
+set disassembly-flavor intel
+set disassemble-next-line on
+enable pretty-printer
+layout src
+
+$ gdb
+```
```
simongdavies, line 83: I think it would be useful to have a walkthrough of how to set up debugging using vscode with a few screenshots etc. ?

src/hyperlight_host/build.rs
```diff
@@ -89,6 +89,7 @@ fn main() -> Result<()> {
     // Essentially the kvm and mshv features are ignored on windows as long as you use #[cfg(kvm)] and not #[cfg(feature = "kvm")].
     // You should never use #[cfg(feature = "kvm")] or #[cfg(feature = "mshv")] in the codebase.
     cfg_aliases::cfg_aliases! {
+        gdb: { all(feature = "gdb", debug_assertions, target_os = "linux") },
```
simongdavies: Since right now we only support debugging with kvm should we be explicit about that here? (and maybe fail with a compilation error if kvm is not set and debug is)

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub r11: u64,
+    pub r12: u64,
+    pub r13: u64,
+    pub r14: u64,
+    pub r15: u64,
+    pub rip: u64,
+    pub rflags: u64,
+}
+
+/// Defines the possible reasons for which a vCPU ce be stopped when debugging
```
simongdavies:
```suggestion
/// Defines the possible reasons for which a vCPU can be stopped when debugging
```

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -55,6 +65,527 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
+
+    use super::KVMDriver;
+    use crate::hypervisor::gdb::{DebugMsg, DebugResponse, VcpuStopReason, X86_64Regs};
+    use crate::hypervisor::handlers::DbgMemAccessHandlerCaller;
+    use crate::mem::layout::SandboxMemoryLayout;
+    use crate::{new_error, HyperlightError, Result};
+
+    /// Software Breakpoint size in memory
+    pub const SW_BP_SIZE: usize = 1;
+    /// Software Breakpoinnt opcode
```
simongdavies:
```suggestion
    /// Software Breakpoint opcode
```

src/hyperlight_host/src/sandbox/mem_access.rs
```diff
@@ -44,3 +46,56 @@ pub(crate) fn mem_access_handler_wrapper(
     let mem_access_hdl = MemAccessHandler::from(mem_access_func);
     Arc::new(Mutex::new(mem_access_hdl))
 }
+
```
simongdavies, line 49: is there a reason why these functions are not gated by the gdb feature?

simongdavies (changes requested):
a few minor issues, but overall this is great. I don't know why the spelling issues are not getting picked up by the spell check in GH actions

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -597,11 +597,15 @@ impl HypervisorHandler {
     /// and still have to receive after sorting that out without sending
     /// an extra message.
     pub(crate) fn try_receive_handler_msg(&self) -> Result<()> {
-        match self
+        #[cfg(gdb)]
```
syntactically, line 621: Maybe include a comment here mentioning why this is? (I presume it's because the thread may pause for a long time waiting for gdb commands?)

src/hyperlight_host/src/sandbox/config.rs
syntactically: I agree with @ludfjig that an arbitrary minimum of 9000 here is not great. You probably don't need to do anything special to prevent ports below 1024, since Linux won't allow binding those without CAP_NET_BIND_SERVICE (by default granted only to root processes) anyway.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -36,6 +36,13 @@ use crate::sandbox_state::sandbox::EvolvableSandbox;
 use crate::sandbox_state::transition::Noop;
 use crate::{log_build_details, log_then_return, new_error, MultiUseSandbox, Result};
 
+#[cfg(gdb)]
+/// Used for passing debug configuration to a sandbox
+pub struct DebugInfo {
```
syntactically: Why do we have this struct? If the goal is to make it easy to pass more information that's needed for the debug configuration in the future, should the `SandboxConfiguration` directly contain one of these, instead of a raw port?

docs/how-to-debug-a-hyperlight-guest.md
dblnz: It might be better to advertise it as only for KVM for now

docs/how-to-debug-a-hyperlight-guest.md
dblnz: Weird I haven't spotted (I enabled a couple of local tools for spelling) it and neither did the typo checker.

src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
```diff
@@ -0,0 +1,99 @@
...
+                let byte = conn
+                    .read()
+                    .map_err(run_blocking::WaitForStopReasonError::Connection)?;
+
+                return Ok(run_blocking::Event::IncomingData(byte));
+            }
+        }
+    }
+
+    fn on_interrupt(
```
syntactically, line 99: I'm slightly confused here---it looks to me like this is reporting to the gdbstub state machine that the thread has stopped, but isn't actually stopping it? We could accomplish actually stopping it by sending a signal to the handler thread to kick the vcpu, or report an error if we don't want to support it?

src/hyperlight_host/src/hypervisor/gdb/x86_64_target.rs
syntactically: I think that the extra bits in `rflags` vs `eflags` are currently reserved, but going by the general x86 conventions, I would expect `eflags` to always refer to the bottom half of `rflags`, so maybe instead of doing a checked conversion here it's semantically cleaner to just truncate?

src/hyperlight_host/src/sandbox/mem_access.rs
```diff
@@ -44,3 +46,56 @@ pub(crate) fn mem_access_handler_wrapper(
...
+#[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
+pub(super) fn handle_dbg_get_code_addr_impl(
+    wrapper: &mut MemMgrWrapper<HostSharedMemory>,
+) -> Result<usize> {
+    Ok(wrapper.unwrap_mgr().layout.get_guest_code_address())
+}
+
+#[instrument(skip_all, parent = Span::current(), level= "Trace")]
+pub(crate) fn dbg_mem_access_handler_wrapper(
+    mut wrapper: MemMgrWrapper<HostSharedMemory>,
```
syntactically: Instead of cloning the hshm wraper 3 times to allow it to be closed over by 3 independent closures, why not define a struct that contains a single hshm wrapper and implements the handler trait directly, or perhaps do away with the trait entirely and pass the wrapper directly to the kvm/gdb thread?

src/hyperlight_host/src/hypervisor/gdb/mod.rs
syntactically: I think the extra semantic information of the extra variants is probably useful

src/hyperlight_host/src/hypervisor/gdb/x86_64_target.rs
```diff
@@ -83,6 +83,11 @@ impl Target for HyperlightSandboxTarget {
     type Arch = GdbTargetArch;
     type Error = GdbTargetError;
 
+    #[inline(always)]
+    fn guard_rail_implicit_sw_breakpoints(&self) -> bool {
+        true
```
syntactically: The gdbstub docs seem to imply that this should only be true when we don't implement our own support for sw breakpoints, since this lets gdb do it by itself?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -380,6 +391,69 @@ mod debug {
...
+                    .ok_or_else(|| new_error!("Debug is not enabled"))?;
+                let addr = self.translate_gva(addr)?;
+                if debug.sw_breakpoints.contains_key(&addr) {
+                    return Ok(true);
+                }
+
+                addr
+            };
+
+            let mut save_data = [0; SW_BP_SIZE];
```
syntactically, line 390: Do we need to make sure that when resuming after a software breakpoint we put the saved instruction back, single step, and then replace it with brk again? Or does the gdb frontend guarantee it will call remove/add_breakpoint in the correct sequence for that, in which case we should probably have a note about that?

syntactically (commented):
I really like how this looks overall! I've left a few small things in the code comments. I also have a couple of nits about the commit organization:

Commit message for 579b3249: s/tread/thread/

Commits b175f416, af0a52a5, 6cc61849, and 98ac60ea: please convert
each one to a series of fixup! commits to the original commits
introducing code, so that we can autosquash before merging, or just
rebase the changes into the original commits now if nobody else
objects to that.

src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
syntactically: The new version seems less likely to mess up the state machine, but is it better UX to eagerly report the fact that we don't stop this by sending an error back?

src/hyperlight_host/src/hypervisor/handlers.rs
```diff
@@ -102,3 +102,25 @@ impl MemAccessHandlerCaller for MemAccessHandler {
...
+    /// Function that gets called for a request to get guest code offset.
+    fn get_code_offset(&mut self) -> Result<usize>;
+}
+
+/// A convenient type representing an implementer of `DbgMemAccessHandlerCaller`
+///
+/// Note: This needs to be wrapped in a Mutex to be able to grab a mutable
+/// reference to the underlying data
+#[cfg(gdb)]
+pub type DbgMemAccessHandlerWrapper = Arc<Mutex<dyn DbgMemAccessHandlerCaller>>;
```
syntactically, line 126: Honestly, given that there's only really one sensible implementation of this, I somewhat question the utility of making it into a trait like this at all... though that also applies to the other exact handlers, tbh.

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -0,0 +1,193 @@
+# How to debug a Hyperlight **KVM** guest using gdb
+
+Hyperlight supports gdb debugging of a **KVM** guest running inside a Hyperlight sandbox.
```
ludfjig, line 3: this is great document! And nice diagram :)

ludfjig (approved):
LGTM!

simongdavies (dismissed):

devigned (dismissed):
lgtm! Great work, @dblnz!

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -61,14 +61,190 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
+            }
+        }
+
+        /// This method sets the kvm debugreg fields to enable breakpoints at
+        /// specific addresses
+        ///
+        /// The first 4 debug registers are used to set the addresses
+        /// The 4th and 5th debug registers are obsolete and not used
+        /// The 7th debug register is used to enable the breakpoints
+        /// For more information see: https://en.wikipedia.org/wiki/X86_debug_register
```
syntactically: Instead of a (mutable) wikipedia page, can we cite the architecture manual here?

syntactically (approved):

ludfjig (approved):

PR #125, opened 2024-12-26
Remove SingleUseSandbox

ludfjig:
Removes the singleusesandbox, as it serves no purpose. This is a first step towards refactoring the evolve, transition-func, etc., abstractions


simongdavies (approved):

simongdavies:
We should consider renaming MultiUseSandbox after this, although I guess that will be a breaking change (I know this one is but AFAIK no one is using `SingleUseSandbox`)

danbugs (approved):
LGTM. If you haven't already, might be worth searching for all text references of `SingleUseSandbox` to make sure we are not missing any mentions in documentation and so on.

PR #135, opened 2025-01-16
Use `CreateFileMapping\MapViewOfFile` and `UnmapViewOfFile\CloseHandle` instead of `VitualAllocEx` and `VirtualFreeEx` on Windows

simongdavies:
This PR replaces the use of `VirtualAllocEx` and `VirtualFreeEx` with `CreateFileMapping`, `MapViewOfFile`, `UnmapViewOfFile` and `CloseHandle` on Windows. 

This enables the mapping of memory into the surrogate processes without zero copying.

There are 9 commits as part of this PR each with its own description/explanation.


marosset (approved):
Looks good to me

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -82,6 +92,8 @@ macro_rules! generate_writer {
 pub struct HostMapping {
     ptr: *mut u8,
     size: usize,
+    #[cfg(target_os = "windows")]
+    handle: HandleWrapper,
```
ludfjig: question: is it possible to put a raw HANDLE here and remove our wrapper?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -385,16 +408,65 @@ impl ExclusiveSharedMemory {
...
+            dwmaximumsizelow = (total_size & 0xFFFFFFFF) as u32;
+        }
+
+        // Allocate the memory use CreateFileMapping instead of VirtualAlloc
+        // This allows us to map the memory into the surrogate process using MapViewOfFile2
+        let handle = unsafe {
+            CreateFileMappingA(
+                INVALID_HANDLE_VALUE,
+                None,
+                PAGE_EXECUTE_READWRITE,
```
ludfjig: Does this need to be execute even when not using in-process mode?

ludfjig (approved):
LGTM!

src/hyperlight_host/src/mem/shared_mem.rs
simongdavies: Yes, fixed

src/hyperlight_host/src/mem/shared_mem.rs
simongdavies: Good catch, fixed

ludfjig (dismissed):
LGTM. Regarding the HANDLE insteade of HandleWrapper, it just seemed unnecessary to have send/sync on handle when it wasn't needed, but if it's indeed in needed, or that made things more complicated feel free to change it back

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -394,16 +415,65 @@ impl ExclusiveSharedMemory {
...
         }
 
-        // TODO protect the guard pages
+        // Set the first and last pages to be guard pages
+
+        let mut unused_out_old_prot_flags = PAGE_PROTECTION_FLAGS(0);
+
+        let first_guard_page_start = addr.Value;
+        if let Err(e) = unsafe {
+            VirtualProtect(
```
syntactically, line 458: IIUC, these `VirtualProtect` calls need to stay in sync with the `VirtualProtectEx` calls in the surrogate manager so that we have consistent permissions between our own mapping and the one that the VM uses?  Should we add a comment here (and there) mentioning that they should be kept in sync, and/or see if we can factor things so that they are closer to each other?

syntactically (approved):

simongdavies:
> LGTM. Regarding the HANDLE insteade of HandleWrapper, it just seemed unnecessary to have send/sync on handle when it wasn't needed, but if it's indeed in needed, or that made things more complicated feel free to change it back

It is needed in some places but not in the one where you pointed it out, since I have to make another small change I will check and see if there are other places where it can be removed.

syntactically (approved):

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -145,35 +144,56 @@ impl SurrogateProcessManager {
...
+        let surrogate_process_handle: HANDLE = self.process_receiver.recv()?.into();
+        let mapping_file_handle: HANDLE = mmap_file_handle.into();
 
-        // allocate memory
+        // Allocate the memory by creating a view over the memory mapped file
+
+        // Use MapViewOfFile2 to map memoy into the surrogate process, the MapViewOfFile2 API is implemented in as an inline function in a windows header file
+        // (see https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile2#remarks) so we use the same API it uses in the header file here instead of
+        // MapViewOfFile2 which does not exist in the rust crate (see https://github.com/microsoft/windows-rs/issues/2595)
+        const NUMA_NO_PREFERRED_NODE: u32 = 0xFFFFFFFF;
```
ludfjig: this const exists. windows::Win32::System::SystemServices::NUMA_NO_PREFERRED_NODE, but requires crate feature Win32_System_SystemServices

ludfjig (dismissed):

ludfjig (approved):

syntactically (approved):

PR #154, opened 2025-01-21
Add a check for dependabot PR updating guest or common lib dependencies

simongdavies:
We have a check in the PR workflow to ensure that Cargo.lock files are up to date as a part of any PR. This works fine for most scenarios but when dependabot updates depdendencies in either `hyperlight_common` or `hyperlight_guest` this may make the Cargo.lock file out of date for the `simpleguest` and `callbackguest` test guest binaries. 

See #148 for an example of this.

This PR updates the PR workflow to update those files if there are dependabot changes to the dependencies in either of those crates.

### DIFF
--- a/.github/workflows/ValidatePullRequest.yml
+++ b/.github/workflows/ValidatePullRequest.yml
@@ -40,6 +40,21 @@ jobs:
             let all_file_count = ${{steps.changes.outputs.all_count}};
             return all_file_count === docs_file_count;
           result-encoding: string
+      # Check to see if this a dependabot PR and if it updates either the common or guest Cargo.toml files
+      # if it does we need to try and generate a new Cargo.lock file as the PR validation checks that these are up to date
+      - name: Update Cargo.lock for dependabot changes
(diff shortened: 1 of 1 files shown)


dblnz (approved):

ludfjig:
I suspect it we add the directory of simpleguest to dependabot.yml, it would update those cargo.toml as well, and we should be fine. It might be an easier solution that this. I'm thinking replace 

```yml
  - package-ecosystem: "cargo"
    directory: "/"
    schedule:
      interval: "daily"
``` 

with 
```yml
  - package-ecosystem: "cargo"
    directory: ["/", "src/tests/rust_guests/simpleguest"]
    schedule:
      interval: "daily"
``` 
https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--

simongdavies:
> It might be worth a try anyway?

Yes , definitely, I will close this one and open a new one with your suggestion, if it doesnt work we can aways re-open this

ludfjig (approved):
LGTM! I'm sad the other way didn't work :/

Issue #47, opened 2024-11-08
`kvm_create` performance regression for Linux 6.x


andreiltd:
We observed the regression in KVM startup times when switching from Kernel `5.x` to `6.x`. This regression is captured by running `create_sandbox` benchmark tests:

- Linux 5.15
```
sandboxes/create_sandbox
                        time:   [1.8367 ms 1.8681 ms 1.9014 ms]
```

- LInux 6.9.3
```
 sandboxes/create_sandbox
                        time:   [7.8108 ms 8.7506 ms 9.7426 ms]
```

A similar issue has been seen in Firecracker: [Linux 6.1 boot time regression](https://github.com/firecracker-microvm/firecracker/blob/c75c86103d464f61851134eb3211b35323be4ffc/docs/prod-host-setup.md#linux-61-boot-time-regressions)

We should try proposed mitigation strategies from the link above and see if it improves the performance.

danbugs:
- Benchmarks from KVM on Ubuntu Server 20.04 (using Linux 5.15):

![image](https://github.com/user-attachments/assets/a55afb38-70c0-4fdf-85a3-b85792c4153b)

- Benchmarks from KVM on Ubuntu Server 24.04 (using Linux 6.8):

![image](https://github.com/user-attachments/assets/1854e19b-5f83-48e7-9449-87bb14b09ad5)

danbugs:
After running the first mitigation suggested [here](https://github.com/firecracker-microvm/firecracker/blob/c75c86103d464f61851134eb3211b35323be4ffc/docs/prod-host-setup.md#linux-61-boot-time-regressions) (i.e., `sudo mount -o remount,favordynmods /sys/fs/cgroup`), we see 18% improvement on Ubuntu 24.04:

![image](https://github.com/user-attachments/assets/9137bb2d-992b-4915-a969-36275508b1ea)

LTBT2530:
Thanks you

simongdavies:
Added a script in #155 to apply the mitigation. Also added a PR to the dev-images repo that adds this mitigation to the GH runners

PR #155, opened 2025-01-21
Adds a script for KVM perf issue mitigation on 6.x kernel

simongdavies:
Adds a script to mitigate for the issue #47


dblnz (approved):

simongdavies:
> Nice! Does it makes sense to run the script in the CI, at least for the benchmarks?

@andreiltd I did do this (https://github.com/hyperlight-dev/ci-setup-workflow/pull/3) but i closed it and added it to the scripts that build the base CI Image instead (https://github.com/hyperlight-dev/dev-images/pull/9), once that PR is merged and a new image is built and used (which happens every Sunday) then it will be applied permanently

andreiltd (approved):

PR #178, opened 2025-01-28
Add clarification on why we must take GuestSharedMemory::lock

syntactically:
Previously, the code comments were not sufficiently clear as to why apparently-useless locks during the execution of the guest are necessary for the invariants of the GuestSharedMemory/HostSharedMemory split.  This led to confusion as to whether it would be a good idea to remove them, so this commit clarifies that there is a good reason for them to be there.


ludfjig (approved):
LGTM. I feel like there should exist a way to have the type system help with this...

danbugs (approved):

PR #214, opened 2025-01-30
Updates rand crate to 0.9

simongdavies:
See #170 

Updates the rand crate to 0.9 and fixes breaking changes

### DIFF
--- a/src/hyperlight_host/src/mem/layout.rs
+++ b/src/hyperlight_host/src/mem/layout.rs
@@ -13,14 +13,12 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 */
-
 use std::fmt::Debug;
 use std::mem::{offset_of, size_of};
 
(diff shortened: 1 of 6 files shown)


src/hyperlight_host/src/mem/layout.rs
```diff
@@ -1100,7 +1099,7 @@ impl SandboxMemoryLayout {
 
         // Set up the security cookie seed
         let mut security_cookie_seed = [0u8; 8];
-        OsRng.fill_bytes(&mut security_cookie_seed);
+        OsRng.try_fill_bytes(&mut security_cookie_seed)?;
```
ludfjig: What do you think about `rand::rng().fill_bytes` here instead? In that way, we won't need rand_core. Or `rand::fill(...)` which doesn't require an explicit call to `rng()`. This is if we don't care about the difference between OsRng and ThreadRng, of course

ludfjig (dismissed):

syntactically (approved):

ludfjig (approved):

PR #230, opened 2025-01-31
Only use general purpose registers in c guest

ludfjig:
There is a mismatch in the ffi ABI between c and rust. Rust library doesn't use SSE, so won't use XMM registers for passing floats as arguments, but clang will. This is a temporary fix until I figure out how to properly set this


danbugs:
Are you still planning on merging this, @ludfjig ?

ludfjig:
> Are you still planning on merging this, @ludfjig ?

Clang's docs state that this is an aarch64 only flag so probably not in this state at least. Unsure why this works at all tbh :D

syntactically:
Subsumed by the custom target work.

Issue #138, opened 2025-01-17
Failed test cases not causing CI jobs to fail


marosset:
A test failure was observed that should have caused the CI jobs to fail but didn't

https://github.com/hyperlight-dev/hyperlight/actions/runs/12827492104/job/35770312541?pr=135#step:12:412

Capturing the relevant test output here in case the github actions logs get deleted

I looked briefly and didn't see anything obviously wrong with the just targets so this might require a deeper look

```log
cargo test  --profile=dev  --lib
   Compiling hyperlight-common v0.1.0 (D:\a\_work\hyperlight\hyperlight\src\hyperlight_common)
   Compiling hyperlight-testing v0.0.0 (D:\a\_work\hyperlight\hyperlight\src\hyperlight_testing)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.75s
     Running unittests src\lib.rs (target\debug\deps\hyperlight_common-c286475b9fdc32cb.exe)

running 1 test
test flatbuffer_wrappers::function_call::tests::read_from_flatbuffer ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running unittests src\lib.rs (target\debug\deps\hyperlight_host-952c13f888be941c.exe)

running 78 tests
test func::call_ctx::tests::singleusesandbox_multi_call ... ok
test func::call_ctx::tests::ensure_multiusesandbox_multi_calls_dont_reset_state ... ok
test func::call_ctx::tests::ensure_multiusesandbox_single_calls_do_reset_state ... ok
test func::call_ctx::tests::singleusesandbox_single_call ... ok
test func::guest_dispatch::tests::test_call_guest_function_by_name_hv ... ok
test func::call_ctx::tests::test_multi_call_multi_thread ... ok
test func::guest_dispatch::tests::test_execute_in_host ... ok
test hypervisor::hyperv_windows::tests::test_init ... ok
test hypervisor::hypervisor_handler::tests::create_1000_sandboxes ... ignored
test func::guest_dispatch::tests::test_terminate_vcpu_spinning_cpu ... ok
test hypervisor::hypervisor_handler::tests::create_10_sandboxes ... ok
test hypervisor::hypervisor_handler::tests::hello_world ... ok
test func::guest_dispatch::tests::test_terminate_vcpu_and_then_call_guest_function_on_the_same_host_thread ... ok
[...]

PR #234, opened 2025-02-03
Fix issues with using `CreateMapViewOfFile` with `inprocess` feature

simongdavies:
#135 introduced a change to the way memory is allocated and shared with surrogate process on Windows, this change causes failures when the `inprocess` feature is enabled , tests failed but did not trigger CI failures because of #138 .

This PR fixes that issue.

### DIFF
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -27,12 +27,18 @@ use tracing::{instrument, Span};
 use windows::core::PCSTR;
 #[cfg(target_os = "windows")]
 use windows::Win32::Foundation::{CloseHandle, HANDLE, INVALID_HANDLE_VALUE};
+#[cfg(all(target_os = "windows", inprocess))]
+use windows::Win32::System::Memory::FILE_MAP_EXECUTE;
+#[cfg(all(target_os = "windows", not(inprocess)))]
+use windows::Win32::System::Memory::PAGE_READWRITE;
 #[cfg(target_os = "windows")]
 use windows::Win32::System::Memory::{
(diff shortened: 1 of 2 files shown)


ludfjig:
Still some tests failing, for example
https://github.com/hyperlight-dev/hyperlight/actions/runs/13123886665/job/36616025945#step:12:409
https://github.com/hyperlight-dev/hyperlight/actions/runs/13123886665/job/36616025945#step:12:571

simongdavies:
> Still some tests failing, for example https://github.com/hyperlight-dev/hyperlight/actions/runs/13123886665/job/36616025945#step:12:409 https://github.com/hyperlight-dev/hyperlight/actions/runs/13123886665/job/36616025945#step:12:571

Yes, I think they are new failures, hopefully fixed now

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -437,10 +438,24 @@ impl ExclusiveSharedMemory {
         };
 
         #[cfg(inprocess)]
-        let addr =
-            unsafe { MapViewOfFile(handle, FILE_MAP_ALL_ACCESS | FILE_MAP_EXECUTE, 0, 0, 0) };
+        let handle = unsafe {
+            CreateFileMappingA(
```
ludfjig, line 435: nit: Seems like this CreateFileMappingA and the below MapViewOfFile are the same inprocess and when running out of process, except for a flag. If you want, you could just change the flag depending on the #cfg instead of the entire funcitno call.

ludfjig (dismissed):

src/hyperlight_host/src/mem/shared_mem.rs
simongdavies: Fixed and added a missing error check

ludfjig (approved):
nice

marosset (approved):

PR #237, opened 2025-02-04
Run clippy on guests in CI

### LINKED ISSUE #216: Should the type of  `function_pointer` in `GuestFunctionDefinition` be changed from i64
The field  `function_pointer` in struct `GuestFunctionDefinition` is defined as a i64.

This requires that a function pointer passed to `register_function` is cast to i64

When running clippy on a guest that casts from a func directly to i64 (e.g. `myfunc as i64`) results in the clippy error "casting function pointer `myfunc` to `i64`" and the suggested fix "try: `myfunc as usize`" which does not work.

The options are to disable the clippy check (`[allow(clippy::fn_to_numeric_cast)]`) or to double cast the pointer (e.g. `myfunc as *const() as i64` or `myfunc as usize as i64`).

Should the definition of the field function pointer be changed to either `usize` or `fn(&FunctionCall) -> Result<Vec<u8>>`

### LINKED ISSUE #217: Run clippy on the test guests
Clippy is not currently run against the test guests in CI.

ludfjig:
closes #217 by running cargo clippy on the guests in CI. Also fixes existing clippy violations that have snuck in. Note that the change from i64->usize is a breaking change

also closes #216

### DIFF
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -60,7 +60,7 @@ jobs:
       - name: clippy
         run: |
           just clippy ${{ matrix.config }}
-          just clippy-guests
+          just clippy-guests ${{ matrix.config }}
 
       # Does not check for updated Cargo.lock files for test rust guests as this causes an issue with this checkwhen deoendabot updates dependencies in common crates
       - name: Ensure up-to-date Cargo.lock
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -58,7 +58,9 @@ jobs:
         run: just fmt-check
 
       - name: clippy
-        run: just clippy ${{ matrix.config }}
+        run: |
+          just clippy ${{ matrix.config }}
+          just clippy-guests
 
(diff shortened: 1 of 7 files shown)


.github/workflows/dep_rust.yml
```diff
@@ -48,7 +48,9 @@ jobs:
         run: just fmt-check
 
       - name: clippy
-        run: just clippy ${{ matrix.config }}
+        run: |
+          just clippy ${{ matrix.config }}
+          just clippy-guests
```
marosset:
```suggestion
          just clippy-guests ${{ matrix.config }}
```
?

ludfjig: nice catch

simongdavies:
Related to the change from i64 -> usize #216

simongdavies (approved):

marosset (approved):
LGTM

devigned (approved):
lgtm!

marosset (approved):

dblnz:
lgtm

dblnz (commented):
lgtm

dblnz (approved):

marosset (approved):

devigned (approved):
lgtm

PR #249, opened 2025-02-08
[HIP] Simpler creation of sandboxes

ludfjig:
Currently creating sandboxes is more complicated than it should be. Here is a small proposal for how to make it simpler and more ergonomic


syntactically:
This is out of date with the things we might want a builder for now.

PR #250, opened 2025-02-08
[interrupt-handlers] adding exception/interrupt handler support for hyperlight guest

### LINKED ISSUE #228: Implement exception/interrupt handlers in hyperlight-guest
Currently, the guest library does not have interrupt/exception handlers. This is unfortunate for several reasons: it prevents us from using memory fault handlers to do things like dynamically grow guest stacks, does not satisfy wasmtime's requirements for floating point and memory permission exceptions to be returned to it, etc.

danbugs:
// TODO: description

closes #228

### DIFF
--- a/src/hyperlight_guest/src/entrypoint.rs
+++ b/src/hyperlight_guest/src/entrypoint.rs
@@ -30,6 +30,8 @@ use crate::{
     __security_cookie, HEAP_ALLOCATOR, MIN_STACK_ADDRESS, OS_PAGE_SIZE, OUTB_PTR,
     OUTB_PTR_WITH_CONTEXT, P_PEB, RUNNING_MODE,
 };
+use crate::idt::init_idt;
+use crate::idtr::load_idt;
 
 #[inline(never)]
 pub fn halt() {
@@ -85,6 +87,10 @@ pub extern "win64" fn entrypoint(peb_address: u64, seed: u64, ops: u64, max_log_
             P_PEB = Some(peb_address as *mut HyperlightPEB);
             let peb_ptr = P_PEB.unwrap();
             __security_cookie = peb_address ^ seed;
+
+            // Set up IDT
+            init_idt();
+            load_idt();
 
             let srand_seed = ((peb_address << 8 ^ seed >> 4) >> 32) as u32;
 
--- a/src/hyperlight_guest/src/interrupt_entry.rs
+++ b/src/hyperlight_guest/src/interrupt_entry.rs
@@ -1,0 +1,229 @@
+/*
+Copyright 2024 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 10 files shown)


src/hyperlight_guest/src/interrupt_entry.rs
```diff
@@ -0,0 +1,202 @@
...
+            generate_excp!(5),
+            generate_excp!(6),
+            generate_excp!(7),
+            generate_excp!(8),
+            generate_excp!(9),
+            generate_excp!(10),
+            generate_excp!(11),
+            generate_excp!(12),
+            generate_excp!(13),
+            generate_excp!(14, pagefault),
```
ludfjig, line 142: I believe more exception types than pagefault take additional parameters

src/hyperlight_guest/src/entrypoint.rs
```diff
@@ -86,6 +89,14 @@ pub extern "win64" fn entrypoint(peb_address: u64, seed: u64, ops: u64, max_log_
             let peb_ptr = P_PEB.unwrap();
             __security_cookie = peb_address ^ seed;
 
+            // Set up GDT/IDT
+            load_gdt();
+            init_idt();
```
ludfjig: question: does it make sense to distinguish `init_idt` from `load_idt`?

ludfjig (dismissed):
lgtm

src/hyperlight_guest/src/interrupt_entry.rs
danbugs: Fixed in [bb7c680](https://github.com/hyperlight-dev/hyperlight/pull/250/commits/bb7c680dcdddc86e622be28a56bcaf9cf925bfc7) ✅

src/hyperlight_guest/src/entrypoint.rs
danbugs: Changed in [bb7c680](https://github.com/hyperlight-dev/hyperlight/pull/250/commits/bb7c680dcdddc86e622be28a56bcaf9cf925bfc7) ✅

src/hyperlight_guest/src/interrupt_entry.rs
```diff
@@ -0,0 +1,202 @@
...
+        concat!(
+            // Common exception handler
+            ".global _do_excp_common\n",
+            "_do_excp_common:\n",
+            // In SysV ABI, the first argument is passed in rdi
+            // rdi is the stack pointer.
+            "    mov rdi, rsp\n",
+            "    call {hl_exception_handler}\n",
+            context_restore!(),
+            "    mov r14, 495\n",
```
syntactically: Why are we clobbering r14 here?

syntactically (changes requested):
This looks to be in pretty good shape overall. A general comment: for structure, bitfield, etc. layouts, please cite the Intel or AMD manuals (either is fine, but we should pick one and stick with it, unless the explanation in the other is far better), instead of the osdev wiki. The manuals, unlike the wiki, are authoritative and unlikely to change :)

On a procedural note: can we squash the "fixes" commits into the original commits (or at least turn them into fixup!s immediately after the original commit, that we have checked will squash nicely without conflicts, and then run rebase --autosquash before merging?)

src/hyperlight_guest/src/interrupt_entry.rs
danbugs: Oops, was using that for debugging. Good catch!

danbugs: Fixed ✅

syntactically (dismissed):

ludfjig (approved):

simongdavies (approved):

PR #272, opened 2025-02-18
Static variable `__security_cookie` should have UPPER_SNAKE_CASE name. `__SECURITY_COOKIE

myadav27:


simongdavies:
@myadav27 please can you signoff the commits to satisfy the DCO check?

simongdavies:
@myadav27 do you have time to update this?

myadav27:
I will be setting this aside for now. The linker expects `__security_cookie `in lowercase, which would require extensive and unnecessary changes across multiple linker files. These changes were also causing existing test cases to break. Given the impact, I’m dropping this effort for the time being.

myadav27:
I will be setting this aside for now. The linker expects `__security_cookie `in lowercase, which would require extensive and unnecessary changes across multiple linker files. These changes were also causing existing test cases to break. Given the impact, I’m dropping this effort for the time being.

Issue #286, opened 2025-02-25
Devcontainer does not have access to `/dev/kvm` after resuming a devcontainer in codespaces


dblnz:
### What happened?

After resuming a previously created devcontainer in github Codespaces, the user does not have access to /dev/kvm

### What did you expect to happen?

It is expected that the user has access to /dev/kvm

### Steps to reproduce the behavior

1. Create new codespace in Github
2. Check access to /dev/kvm - has access
3. Stop the codespace
4. Start the codespace
5. The user does not have access to /dev/kvm anymore

### Hyperlight Version

0.1.0

### OS version

<details>

```console
On Linux:
$ cat /etc/os-release
# paste output here

$ uname -a
# paste output here

On Windows:
C:\> cmd /c ver
# paste output here

````
</details>

### Additional Information

By running a `chown/chmod` command to change permissions for the running user, the issue is solved.
This needs to be done every time the container is closed.

myadav27:
PR #287

ludfjig:
Closing as this was fixed in #326

PR #287, opened 2025-02-25
fix access issue in devconatiner #286

myadav27:
Devcontainer does not have access to /dev/kvm after resuming a devcontainer in codespaces
#286

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -17,10 +17,7 @@
       "--device=/dev/kvm"
   ],
 
-  // Use 'postCreateCommand' to run commands after the container is created
-  "postCreateCommand": "bash .devcontainer/setup.sh",
-
-  // Runs each time the container starts/restarts and set kvm permissions
+  // Runs each time the container starts/restarts 
   "postStartCommand": "bash .devcontainer/setup.sh",
   
   "customizations": {
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -20,6 +20,9 @@
   // Use 'postCreateCommand' to run commands after the container is created
   "postCreateCommand": "bash .devcontainer/setup.sh",
 
+  // Runs each time the container starts/restarts
+  // Only run these commands if /dev/kvm exists
+  "postStartCommand": "if [ -e /dev/kvm ]; then sudo chown root:kvm /dev/kvm && sudo chmod 660 /dev/kvm && sudo usermod -aG kvm vscode; else echo 'KVM device not found, skipping...'; fi",
   "customizations": {
     "vscode": {
       "extensions": [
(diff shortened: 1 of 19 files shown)


.devcontainer/devcontainer.json
```diff
@@ -20,6 +20,9 @@
   // Use 'postCreateCommand' to run commands after the container is created
   "postCreateCommand": "bash .devcontainer/setup.sh",
```
dblnz: I think this would suffice to deal with the issue of `/dev/kvm` access.
The issues is that now the command only works after the container is created, but if it is instantiated the `postCreateCommand` is not called again, so the below suggestion should work.

```suggestion
  "postStartCommand": "bash .devcontainer/setup.sh",
```

.devcontainer/devcontainer.json
```diff
@@ -20,6 +20,9 @@
...
 
+  // Runs each time the container starts/restarts
+  // Only run these commands if /dev/kvm exists
+  "postStartCommand": "if [ -e /dev/kvm ]; then sudo chown root:kvm /dev/kvm && sudo chmod 660 /dev/kvm && sudo usermod -aG kvm vscode; else echo 'KVM device not found, skipping...'; fi",
```
simongdavies: @myadav27 Thanks for the PR, could you update it to use the `setup.sh` script instead of the embedded command?

simongdavies (changes requested):

.devcontainer/devcontainer.json
myadav27: Good Idea ,will change it and update.

myadav27: Updated, please refer latest commit

.devcontainer/setup.sh
```diff
@@ -3,3 +3,11 @@
 # Change device ownership
 sudo chown -R $REMOTE_USER:$REMOTE_GROUP $DEVICE
 
+# Check for the KVM device and adjust permissions.
```
dblnz: I don't think you need this at all.
Only using the file as it was should address the issue.
I remember I tried that last week. Can you also check this is the case for you?

.devcontainer/devcontainer.json
dblnz: I think we can remove the `postCreateCommand` as the `postStartCommand` is called each time we start the container, we need not call the script after creation anymore.

.devcontainer/setup.sh
myadav27: yes, tested worked after restarts. Will remove

dblnz:
To pass the DCO check, please sign your commits.

myadav27:
> To pass the DCO check, please sign your commits.

@dblnz  could you please refer #326 with correct code and sign off. I will abandon sign off, since I wan committing with msft account and was causing the issue. I will abandon/reject this pr

PR #291, opened 2025-02-26
devcontainer: install gdb and vscode ext for debugging

dblnz:
This installs `gdb`  and `C/C++ Extension pack` for a better experience debugging in a devcontainer either locally or in Github Codespaces.
Not having these installed means the user needs to do it every time the container gets started


ludfjig (approved):

devigned (approved):
lgtm

PR #301, opened 2025-02-28
Add fuzzing target to fuzz return type and parameters

ludfjig:
- Adds 2 new fuzzing target for fuzzing parameters to guest funcs and host funcs respectively.
- Renames existing fuzzing target to `host_print`.
- Make fuzzing more efficient by re-using sandbox
- Fixes our cron fuzz target which is supposed to fuzz for a long time every sunday, but was only fuzzing for 5 minutes
- Moves fuzzing up to root directory
---
- Second commit fixes a bug that `host_call` fuzz target found, where RSP kept decreasing after each dispatch, and eventually caused an error. RSP should be reset every dispatch. The new logic resets RSP **before** each dispatch to the original RSP (which points to beginning of stack). I applied the new logic to all 3 drivers, even though the bug only manifested on KVM.

In order to merge this PR, we need to remove `fuzz/fuzzing` status check from branch protection, as this PR changes the name of that job to `fuzz/fuzzing (guest_call)` and `fuzz/fuzzing (host_print)`


dblnz (approved):
LGTM!
Good work on improving the verifications we do on the changes.

simongdavies:
> In order to merge this PR, we need to remove `fuzz/fuzzing` status check from branch protection, as this PR changes the name of that job to `fuzz/fuzzing (guest_call)` and `fuzz/fuzzing (host_print)`

I removed the fuzz check temporarily, once this merges let me know and I will add the 2 new ones

simongdavies (dismissed):

marosset (approved):
LGTM

danbugs (approved):

PR #324, opened 2025-03-04
Update fuzzing test to check for simpleguest in same directory as executable

marosset:
Some fuzzing scenarios we want to report require building a fuzzing binary with `cargo fuzz build` and submitting that binary to a fuzzing harness. In those instances we also want to submit a guestbinary and this PR makes sure that the guestbinary can be found in the same directory as the fuzzing binary.

### DIFF
--- a/src/hyperlight_testing/src/lib.rs
+++ b/src/hyperlight_testing/src/lib.rs
@@ -16,6 +16,7 @@ limitations under the License.
 
 // This crate contains testing utilities which need to be shared across multiple
 // crates in this project.
+use std::env;
 use std::path::PathBuf;
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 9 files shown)


src/hyperlight_testing/src/lib.rs
```diff
@@ -129,3 +130,28 @@ pub fn c_callback_guest_as_string() -> Result<String> {
...
+/// Get a fully qualified path to a simple guest binary preferring a binary
+/// in the same directory as the parent executable. This will be used in
+/// fuzzing scenarios where pre-built binaries will be build and submitted to
+/// a fuzzing framework.
+pub fn simple_guest_for_fuzzing_as_string() -> Result<String> {
+    let exe_dir = env::current_exe()
+        .ok()
+        .and_then(|path| path.parent().map(|p| p.to_path_buf()));
+
+    if exe_dir.is_some() {
```
ludfjig: is this `.is_some()` redundant given the next line?

ludfjig (dismissed):

src/hyperlight_testing/src/lib.rs
marosset: yes, i think is is :P

marosset: fixed

ludfjig (approved):

ludfjig (dismissed):

ludfjig (approved):

danbugs (approved):

PR #325, opened 2025-03-05
Review/remove and prevent unintentional use of `expect/assert*/panic/unwrap` in hyperlight_host and hyperlght_common

### LINKED ISSUE #320: Review use of `expect/panic/assert/unwrap` across the hyperlight codebase
Since hyperlight is a library the use of `expect/panic/assert/unwrap` especially in release builds is problematic since it use implies that we are happy to terminate the process, in most cases these usages should be turned into errors returned to the host program, the only cases where we should `expect/panic/assert/unwrap` are where we detect out of bounds memory access attempts or actions that would corrupt data.

simongdavies:
This PR:

Adds a clippy lints to check for use of expect, panic and unwrap in hyperlight-host for non test/non dev code in hyperlight_host and hyperlight_common
Removes all uses of expect in non test/dev code in hyperlight-host.
Removes all uses of panic in non test/dev code in hyperlight-host.
Removes all uses of unwrap in non test/dev code in hyperlight-host.

Adds a clippy.toml file to hyperlight_host/src and hyperlight_common/src to prevent the use of assert* using disallowed_macros lint
Disables the disallowed_macros lint for test and dev code

Fixes #320

### DIFF
--- a/src/hyperlight_common/src/clippy.toml
+++ b/src/hyperlight_common/src/clippy.toml
@@ -1,0 +1,7 @@
+disallowed-macros = [
+    { path = "std::assert", reason = "no asserts in release builds" },
+    { path = "std::assert_eq", reason = "no asserts in release builds" },
+    { path = "std::assert_ne", reason = "no asserts in release builds" },
+    { path = "std::assert_true", reason = "no asserts in release builds" },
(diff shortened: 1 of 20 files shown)


ludfjig (approved):
LGTM

Issue #220, opened 2025-01-30
Hyperlight Debugging (MSHV)


devigned:
Enable native debugging when running on Linux with `/dev/mshv`.

PR #327, opened 2025-03-05
Gdb support for mshv guests

dblnz:
This PR address the #220 issue.
Additionally, the first 7 commits refactor the existing guest debugging code and moves code from `kvm.rs` to a new file that is to contain the hypervisor specific debugging functionality.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -117,9 +117,9 @@ test-rust-feature-compilation-fail target=default-target:
     {{ if os() == "linux" { "! cargo check -p hyperlight-host --no-default-features 2> /dev/null"} else { "" } }}
 
 # Test rust gdb debugging
-test-rust-gdb-debugging target=default-target: (build-rust target)
-    {{ set-trace-env-vars }} cargo test --profile={{ if target == "debug" { "dev" } else { target } }} --example guest-debugging --features gdb
-    {{ set-trace-env-vars }} cargo test --profile={{ if target == "debug" { "dev" } else { target } }} --features gdb -- test_gdb
+test-rust-gdb-debugging target=default-target features="": (build-rust target)
+    {{ set-trace-env-vars }} cargo test --profile={{ if target == "debug" { "dev" } else { target } }} --example guest-debugging {{ if features =="" {'--features gdb'} else { "--features gdb," + features } }}
+    {{ set-trace-env-vars }} cargo test --profile={{ if target == "debug" { "dev" } else { target } }} {{ if features =="" {'--features gdb'} else { "--features gdb," + features } }} -- test_gdb
 
 test target=default-target: (test-rust target)
 
--- a/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
@@ -23,7 +23,7 @@ use libc::{pthread_kill, SIGRTMIN};
 use super::x86_64_target::HyperlightSandboxTarget;
 use super::{DebugResponse, GdbTargetError, VcpuStopReason};
 
-pub struct GdbBlockingEventLoop;
+struct GdbBlockingEventLoop;
 
 impl run_blocking::BlockingEventLoop for GdbBlockingEventLoop {
     type Connection = Box<dyn ConnectionExt<Error = std::io::Error>>;
@@ -115,7 +115,7 @@ impl run_blocking::BlockingEventLoop for GdbBlockingEventLoop {
     }
 }
 
-pub fn event_loop_thread(
+pub(crate) fn event_loop_thread(
     debugger: GdbStub<HyperlightSandboxTarget, Box<dyn ConnectionExt<Error = std::io::Error>>>,
     target: &mut HyperlightSandboxTarget,
 ) {
--- a/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
@@ -16,8 +16,9 @@ limitations under the License.
 
 use gdbstub::common::Signal;
 use gdbstub::conn::ConnectionExt;
-use gdbstub::stub::run_blocking::{self, WaitForStopReasonError};
-use gdbstub::stub::{BaseStopReason, DisconnectReason, GdbStub, SingleThreadStopReason};
+use gdbstub::stub::{
+    run_blocking, BaseStopReason, DisconnectReason, GdbStub, SingleThreadStopReason,
+};
 use libc::{pthread_kill, SIGRTMIN};
 
 use super::x86_64_target::HyperlightSandboxTarget;
@@ -57,13 +58,10 @@ impl run_blocking::BlockingEventLoop for GdbBlockingEventLoop {
                             signal: Signal(SIGRTMIN() as u8),
                         },
                         VcpuStopReason::Unknown => {
-                            log::warn!("Unknown stop reason - resuming execution");
+                            log::warn!("Unknown stop reason received");
 
-                            target
-                                .resume_vcpu()
-                                .map_err(WaitForStopReasonError::Target)?;
-
-                            continue;
+                            // Marking as a SwBreak so the gdb inspect where/why it stopped
+                            BaseStopReason::SwBreak(())
                         }
                     };
 
--- a/src/hyperlight_host/src/hypervisor/gdb/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/mod.rs
@@ -50,6 +50,17 @@ const SW_BP: [u8; SW_BP_SIZE] = [SW_BP_OP];
 /// Maximum number of supported hardware breakpoints
 const MAX_NO_OF_HW_BP: usize = 4;
 
+/// Check page 19-4 Vol. 3B of Intel 64 and IA-32
+/// Architectures Software Developer's Manual
+/// Bit position of BS flag in DR6 debug register
+const DR6_BS_FLAG_POS: usize = 14;
+/// Bit mask of BS flag in DR6 debug register
+const DR6_BS_FLAG_MASK: u64 = 1 << DR6_BS_FLAG_POS;
+/// Bit position of HW breakpoints status in DR6 debug register
+const DR6_HW_BP_FLAGS_POS: usize = 0;
+/// Bit mask of HW breakpoints status in DR6 debug register
+const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
+
 #[derive(Debug, Error)]
 pub(crate) enum GdbTargetError {
     #[error("Error encountered while binding to address and port")]
--- a/src/hyperlight_host/src/hypervisor/gdb/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/mod.rs
@@ -21,6 +21,7 @@ mod x86_64_target;
 
 use std::io::{self, ErrorKind};
 use std::net::TcpListener;
+use std::sync::{Arc, Mutex};
 use std::thread;
 
 use crossbeam_channel::{Receiver, Sender, TryRecvError};
@@ -28,23 +29,26 @@ use event_loop::event_loop_thread;
 use gdbstub::conn::ConnectionExt;
 use gdbstub::stub::GdbStub;
 use gdbstub::target::TargetError;
+use hyperlight_common::mem::PAGE_SIZE;
 #[cfg(kvm)]
 pub(crate) use kvm_debug::KvmDebug;
 use thiserror::Error;
 use x86_64_target::HyperlightSandboxTarget;
 
+use crate::hypervisor::handlers::DbgMemAccessHandlerCaller;
+use crate::mem::layout::SandboxMemoryLayout;
 use crate::new_error;
 
 /// Software Breakpoint size in memory
-pub(crate) const SW_BP_SIZE: usize = 1;
+const SW_BP_SIZE: usize = 1;
 /// Software Breakpoint opcode - INT3
 /// Check page 7-28 Vol. 3A of Intel 64 and IA-32
 /// Architectures Software Developer's Manual
 const SW_BP_OP: u8 = 0xCC;
 /// Software Breakpoint written to memory
-pub(crate) const SW_BP: [u8; SW_BP_SIZE] = [SW_BP_OP];
+const SW_BP: [u8; SW_BP_SIZE] = [SW_BP_OP];
 /// Maximum number of supported hardware breakpoints
-pub(crate) const MAX_NO_OF_HW_BP: usize = 4;
+const MAX_NO_OF_HW_BP: usize = 4;
 
 #[derive(Debug, Error)]
 pub(crate) enum GdbTargetError {
@@ -159,10 +163,16 @@ pub(crate) trait GuestDebug {
 
     /// Returns true whether the provided address is a hardware breakpoint
     fn is_hw_breakpoint(&self, addr: &u64) -> bool;
+    /// Returns true whether the provided address is a software breakpoint
+    fn is_sw_breakpoint(&self, addr: &u64) -> bool;
     /// Stores the address of the hw breakpoint
     fn save_hw_breakpoint(&mut self, addr: &u64) -> bool;
+    /// Stores the data that the sw breakpoint op code replaces
+    fn save_sw_breakpoint_data(&mut self, addr: u64, data: [u8; 1]);
     /// Deletes the address of the hw breakpoint from storage
     fn delete_hw_breakpoint(&mut self, addr: &u64);
+    /// Retrieves the saved data that the sw breakpoint op code replaces
+    fn delete_sw_breakpoint_data(&mut self, addr: &u64) -> Option<[u8; 1]>;
 
     /// Read registers
     fn read_regs(&self, vcpu_fd: &Self::Vcpu, regs: &mut X86_64Regs) -> crate::Result<()>;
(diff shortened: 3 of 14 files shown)


ludfjig:
This looks great! I noticed the `how-to-debug-a-hyperlgiht-guest.md` only mentions KVM guests, so we might want to update that doc. Additionally we have a `debugging-hyperlgiht.md` doc that has no mention of gdb debugging at all. Perhaps we could add a link from that doc to `how-to-debug-a-hyperlight-guest.md`

dblnz:
> This looks great! I noticed the `how-to-debug-a-hyperlgiht-guest.md` only mentions KVM guests, so we might want to update that doc. Additionally we have a `debugging-hyperlgiht.md` doc that has no mention of gdb debugging at all. Perhaps we could add a link from that doc to `how-to-debug-a-hyperlight-guest.md`

Thanks, I forgot about the documentation.
I'll add another commit that addresses the documentation

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -26,6 +27,7 @@ use event_loop::event_loop_thread;
 use gdbstub::conn::ConnectionExt;
 use gdbstub::stub::GdbStub;
 use gdbstub::target::TargetError;
+pub use hyp_debug::{kvm, mshv, GuestMemoryDebug, GuestVcpuDebug};
```
ludfjig: Does this need to be pub?

ludfjig (approved):
LGTM! It would be good if @syntactically could review as well

src/hyperlight_host/src/hypervisor/gdb/mod.rs
dblnz: Yes, it is used in each hypervisor driver.

ludfjig: Could you use `pub(crate) use hyp_debug::{kvm, mshv, GuestMemoryDebug, GuestVcpuDebug};` instead?

ludfjig (dismissed):

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,758 @@
...
+
+            Ok(true)
+        } else {
+            Ok(false)
+        }
+    }
+}
+
+/// Trait used for defining memory debugging functionality
+pub trait GuestMemoryDebug: GuestVcpuDebug {
```
danbugs: Why split this into a separate trait? Also–is naming appropriate? If so, might be worth further elaborating in the comment why memory debugging = software breakpoints and vCPU debugging = hardware breakpoints.

src/hyperlight_host/src/hypervisor/gdb/mod.rs
danbugs: +1 on pub(crate)

src/hyperlight_host/src/hypervisor/gdb/x86_64_target.rs
```diff
@@ -89,6 +84,11 @@ impl HyperlightSandboxTarget {
         }
     }
 
+    /// Non-Blocking check for a response over the communication channel
+    pub fn try_recv(&self) -> Result<DebugResponse, TryRecvError> {
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,758 @@
...
+                .set_guest_debug(&self.dbg_cfg)
+                .map_err(|e| new_error!("Could not set guest debug: {:?}", e))?;
+
+            self.single_step = step;
+
+            Ok(())
+        }
+
+        /// Get the reason the vCPU has stopped
+        pub fn get_stop_reason(
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,758 @@
...
+        pub hw_breakpoints: Vec<u64>,
+        /// Saves the bytes modified to enable SW breakpoints
+        pub sw_breakpoints: HashMap<u64, [u8; SW_BP_SIZE]>,
+
+        /// Sent to KVM for enabling guest debug
+        pub dbg_cfg: kvm_guest_debug,
+    }
+
+    impl KvmDebug {
+        pub fn new() -> Self {
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,758 @@
...
+        hw_breakpoints: Vec<u64>,
+        /// Saves the bytes modified to enable SW breakpoints
+        sw_breakpoints: HashMap<u64, [u8; SW_BP_SIZE]>,
+
+        /// Debug registers
+        pub dbg_cfg: DebugRegisters,
+    }
+
+    impl MshvDebug {
+        pub fn new() -> Self {
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,758 @@
...
+
+            vcpu_fd
+                .set_regs(&regs)
+                .map_err(|e| new_error!("Could not set registers: {:?}", e))?;
+
+            Ok(())
+        }
+
+        /// Returns the vCPU stop reason
+        pub fn get_stop_reason(
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -55,8 +63,152 @@ use crate::hypervisor::hypervisor_handler::HypervisorHandler;
...
+
+            debug.set_single_step(&self.vcpu_fd, false)?;
+
+            self.debug = Some(debug);
+
+            Ok(())
+        }
+
+        /// Get the reason the vCPU has stopped
+        pub fn get_stop_reason(&mut self) -> Result<VcpuStopReason> {
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -55,8 +63,152 @@ use crate::hypervisor::hypervisor_handler::HypervisorHandler;
...
+            let debug = self
+                .debug
+                .as_mut()
+                .ok_or_else(|| new_error!("Debug is not enabled"))?;
+
+            debug.get_stop_reason(&self.vcpu_fd, self.entrypoint)
+        }
+
+        pub fn process_dbg_request(
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -55,8 +63,152 @@ use crate::hypervisor::hypervisor_handler::HypervisorHandler;
...
+
+                        Ok(DebugResponse::WriteRegisters)
+                    }
+                }
+            } else {
+                Err(new_error!("Debugging is not enabled"))
+            }
+        }
+
+        pub fn recv_dbg_msg(&mut self) -> Result<DebugMsg> {
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -55,8 +63,152 @@ use crate::hypervisor::hypervisor_handler::HypervisorHandler;
...
+            gdb_conn.recv().map_err(|e| {
+                new_error!(
+                    "Got an error while waiting to receive a
+                    message: {:?}",
+                    e
+                )
+            })
+        }
+
+        pub fn send_dbg_msg(&mut self, cmd: DebugResponse) -> Result<()> {
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -65,477 +67,110 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
-
-            if ip == self.entrypoint {
-                return Ok(VcpuStopReason::HwBp);
-            }
-
-            Ok(VcpuStopReason::Unknown)
+            debug.get_stop_reason(&self.vcpu_fd, debug_exit, self.entrypoint)
         }
 
         pub fn process_dbg_request(
```
danbugs: pub(crate), maybe?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    /// Adds hardware breakpoint
+    fn add_hw_breakpoint(&mut self, vcpu_fd: &Self::Vcpu, addr: u64) -> Result<bool> {
+        let addr = self.translate_gva(vcpu_fd, addr)?;
+
+        if self.is_hw_breakpoint(&addr) {
+            Ok(true)
+        } else {
+            let res = self.save_hw_breakpoint(&addr);
+            if res {
```
simongdavies:
```suggestion
            if self.save_hw_breakpoint(&addr) {
```

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,481 @@
...
+    fn read_regs(&self, vcpu_fd: &Self::Vcpu, regs: &mut X86_64Regs) -> Result<()>;
+    /// Enables or disables stepping and sets the vCPU debug configuration
+    fn set_single_step(&mut self, vcpu_fd: &Self::Vcpu, enable: bool) -> Result<()>;
+    /// Translates the guest address to physical address
+    fn translate_gva(&self, vcpu_fd: &Self::Vcpu, gva: u64) -> Result<u64>;
+    /// Write registers
+    fn write_regs(&self, vcpu_fd: &Self::Vcpu, regs: &X86_64Regs) -> Result<()>;
+
+    /// Adds hardware breakpoint
+    fn add_hw_breakpoint(&mut self, vcpu_fd: &Self::Vcpu, addr: u64) -> Result<bool> {
```
simongdavies: Question: Does this fn and the remove one below need to return Result<bool> or can it be Result<()>? it seems like the error case could be mapped to `DebugResponse::AddHwBreakpoint(false)?  , if this is the case the logic in the functions can be simplified and we should also perhaps log when the error case occurs?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,481 @@
...
+    fn is_sw_breakpoint(&self, addr: &u64) -> bool;
+    /// Stores the data that the sw breakpoint op code replaces
+    fn save_sw_breakpoint_data(&mut self, addr: u64, data: [u8; 1]);
+    /// Retrieves the saved data that the sw breakpoint op code replaces
+    fn delete_sw_breakpoint_data(&mut self, addr: &u64) -> Option<[u8; 1]>;
+
+    /// Overwrites the guest memory with the SW Breakpoint op code that instructs
+    /// the vCPU to stop when is executed and stores the overwritten data to be
+    /// able to restore it
+    fn add_sw_breakpoint(
```
simongdavies: Same as above does the add/remove need to be `Result<bool>` or can it be `Result<()>`

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -0,0 +1,481 @@
...
+
+            data = &data[write_len..];
+            gva += write_len as u64;
+        }
+
+        Ok(())
+    }
+}
+
+#[cfg(kvm)]
```
simongdavies: Maybe this should be in a separate file?

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
```diff
@@ -479,3 +479,15 @@ pub mod kvm {
         }
     }
 }
+
```
simongdavies: Can this be moved into an mshv specific file?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -188,6 +199,31 @@ impl HypervLinuxDriver {
...
 
+            // The bellow intercepts make the vCPU exit with the Exception Intercept exit code
+            // Install intercept for #DB exception
+            vm_fd
+                .install_intercept(mshv_install_intercept {
+                    access_type_mask: HV_INTERCEPT_ACCESS_MASK_EXECUTE,
+                    intercept_type: hv_intercept_type_HV_INTERCEPT_TYPE_EXCEPTION,
+                    // Exception handler #DB
+                    intercept_parameter: hv_intercept_parameters {
+                        exception_vector: 0x1,
```
simongdavies, line 358: Shall we make these magic numbers constants or at least comment what they are and where they come from?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -188,6 +199,31 @@ impl HypervLinuxDriver {
...
+                .map_err(|e| new_error!("Cannot install debug exception intercept: {}", e))?;
+
+            // Install intercept for #BP exception
+            vm_fd
+                .install_intercept(mshv_install_intercept {
+                    access_type_mask: HV_INTERCEPT_ACCESS_MASK_EXECUTE,
+                    intercept_type: hv_intercept_type_HV_INTERCEPT_TYPE_EXCEPTION,
+                    // Exception handler #BP
+                    intercept_parameter: hv_intercept_parameters {
+                        exception_vector: 0x3,
```
simongdavies, line 370: As above

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -438,6 +476,13 @@ impl Hypervisor for HypervLinuxDriver {
                         None => HyperlightExit::Mmio(gpa),
                     }
                 }
+                #[cfg(gdb)]
```
simongdavies, line 629: Might be worth commenting here that the only case where we are expecting an intercept exit is the case we are debugging as we are only enabling exception exits for debug exceptions?

Justfile
```diff
@@ -111,9 +111,9 @@ test-rust-feature-compilation-fail target=default-target:
     {{ if os() == "linux" { "! cargo check -p hyperlight-host --no-default-features 2> /dev/null"} else { "" } }}
 
 # Test rust gdb debugging
-test-rust-gdb-debugging target=default-target: (build-rust target)
-    {{ set-trace-env-vars }} cargo test --profile={{ if target == "debug" { "dev" } else { target } }} --example guest-debugging --features gdb
-    {{ set-trace-env-vars }} cargo test --profile={{ if target == "debug" { "dev" } else { target } }} --features gdb -- test_gdb
+test-rust-gdb-debugging target=default-target features="": (build-rust target)
```
simongdavies, line 120: nit shouldnt the gdb feature be always set in this recipe rather than passed. Otherwise we have a recipe called `test-rust-gdb-debugging` that only actually does that if we pass a "gdb" feature argument?

simongdavies (approved):
Some minor questions/suggestions but otherwise this LGTM

src/hyperlight_host/src/hypervisor/gdb/mod.rs
dblnz: Oh, I see what you meant, I was thinking about something else.
Good catch!

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
dblnz: I think you might be right, good catch

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
dblnz: Yes, I'll split them in two separate files.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
dblnz: I did mention above the handler for `#DB`. This is the way it is mentioned in the Intel Spec.
I'll add the specific page here also.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
dblnz: yes, I'll add a comment

Justfile
dblnz: Good catch

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
dblnz: This is more of a logical split, rather than functional, meant to allow only vCPU manipulation without memory access.
Indeed, I don't think we'll ever have a use case where we don't allow memory manipulation of the guest.
So, I'll merge them together and rename the trait

dblnz: Refactored and merged them together.

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
dblnz: Changed the method signature

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
dblnz: Moved to kvm specific different file

src/hyperlight_host/src/hypervisor/gdb/hyp_debug.rs
dblnz: Moved to mshv specific different file

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
dblnz: Added comment

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -124,6 +340,43 @@ impl HypervLinuxDriver {
...
+        let (debug, gdb_conn) = if let Some(gdb_conn) = gdb_conn {
+            let mut debug = MshvDebug::new();
+            debug.add_hw_breakpoint(&vcpu_fd, entrypoint_ptr.absolute()?)?;
+
+            // The bellow intercepts make the vCPU exit with the Exception Intercept exit code
+            // Check Table 6-1. Exceptions and Interrupts at Page 6-13 Vol. 1
+            // of Intel 64 and IA-32 Architectures Software Developer's Manual
+            // Install intercept for #DB (1) exception
+            vm_fd
+                .install_intercept(mshv_install_intercept {
```
danbugs, line 353: Curious about how this works w/ the guest's exception handler.

danbugs (approved):
LGTM–great work, Doru!

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
dblnz: That would be interesting to see

src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
```diff
@@ -57,13 +58,10 @@ impl run_blocking::BlockingEventLoop for GdbBlockingEventLoop {
...
-                            log::warn!("Unknown stop reason - resuming execution");
+                            log::warn!("Unknown stop reason received");
 
-                            target
-                                .resume_vcpu()
-                                .map_err(WaitForStopReasonError::Target)?;
-
-                            continue;
+                            // Marking as a SwBreak so the gdb inspect where/why it stopped
+                            BaseStopReason::SwBreak(())
```
syntactically, line 64: Did you test whether GDB gets upset when it receives a breakpoint where pc isn't matching one of the ones it knows it's set? It's probably fine, but worth double-checking.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
syntactically: Generally, the HV needs to figure out whether the exception corresponds to an HV breakpoint or a guest kernel breakpoint and then reinject the exception if the latter. Similarly, you need to intercept when the guest writes to the hardware breakpoint/single-step/etc registers, and emulate the functionality. I don't imagine that we will be in a situation where we care about the kernel inside the hypervisor being able to debug its own userspace, so I don't think the extra complexity is worth it.

src/hyperlight_host/src/hypervisor/gdb/mshv_debug.rs
```diff
@@ -14,11 +14,255 @@ See the License for the specific language governing permissions and
...
+
+        vcpu_fd
+            .set_regs(&regs)
+            .map_err(|e| new_error!("Could not set registers: {:?}", e))?;
+
+        Ok(())
+    }
+
+    /// Returns the vCPU stop reason
+    pub(crate) fn get_stop_reason(
```
syntactically, line 133: This code looks almost identical to the equivalent in the KVM driver, and seems to me to mostly encode architectural things that are independent from how you talk to the HV. It might be interesting to see if that could be factored out, assuming a factored-out interface for getting DR6/PC/etc? That might be worth looking into in the future, at least.

src/hyperlight_host/src/hypervisor/gdb/mshv_debug.rs
```diff
@@ -163,6 +170,14 @@ impl MshvDebug {
             return Ok(VcpuStopReason::HwBp);
         }
 
+        // mshv does not provide a way to specify which exception triggered the
```
syntactically, line 173: As far as I'm aware, the `hv_message` for the exception intercept does contain the exception vector that was intercepted. See [`hv_message::to_exception_info`](https://docs.rs/mshv-bindings/latest/mshv_bindings/struct.hv_message.html#method.to_exception_info) and [`hv_x64_exception_intercept_message`](https://docs.rs/mshv-bindings/latest/mshv_bindings/struct.hv_x64_exception_intercept_message.html)

syntactically (approved):
Overall LGTM. Great work, especially given how little documentation there is for the HV interfaces, @dblnz !

src/hyperlight_host/src/hypervisor/gdb/mshv_debug.rs
syntactically: It may or may not be worth using here, but it's probably a good idea to update the comment?

dblnz: Will be addressed in #380

src/hyperlight_host/src/hypervisor/gdb/mshv_debug.rs
dblnz: Will be addressed in #380

src/hyperlight_host/src/hypervisor/gdb/event_loop.rs
dblnz: Will be addressed in #380

PR #359, opened 2025-03-14
Remove unused `page_size` guest parameter

ludfjig:
The page_size is currently passed to the guest's entrypoint, but is unused.


ludfjig:
This will be addressed in #297 so I'm closing this

PR #361, opened 2025-03-15
Rework metrics to use metrics crate

ludfjig:
This simplifies metrics alot to use the `metrics` crate. The `metrics` crate functions similar to the `log` crate in that libraries just call `counter!`, `gauge!` or `histogram!` to record metrics, and the user of the library can decide where to send the metrics by installing an appropriate recorder/exporter.


Justfile
```diff
@@ -95,6 +95,7 @@ test-rust target=default-target features="": (test-rust-int "rust" target featur
     cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} --test integration_test log_message -- --ignored
     cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} sandbox::uninitialized::tests::test_log_trace -p hyperlight-host --lib -- --ignored
     cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} hypervisor::hypervisor_handler::tests::create_1000_sandboxes -p hyperlight-host --lib -- --ignored
+    cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} metrics::tests::test_description_called_once -p hyperlight-host --lib -- --ignored
```
simongdavies: shouldn't this have the metrics feature enabled?

src/hyperlight_host/src/metrics/metrics_macro.rs
```diff
@@ -0,0 +1,68 @@
+/// Crate-internal trait for defining metrics.
```
simongdavies: this needs a copy right header.

src/hyperlight_host/src/metrics/metrics_macro.rs
```diff
@@ -0,0 +1,68 @@
...
+                        unit: $unit:expr $(,)?
+                    }
+                ),* $(,)?
+            }
+        )*
+    ) => {
+        $(
+            #[derive(Debug, Clone)]
+            #[allow(dead_code)]
+            pub(crate) enum $metric_type $( < $( $gen ),+ > )? {
```
simongdavies: does this get used if the feature is not enabled? i.e. shouldn't the feature be on the enum not in the body, in fact its not clear right now why this macro is even invoked if the feature is not enabled?

src/hyperlight_host/src/metrics/mod.rs
```diff
@@ -1,355 +1,367 @@
-/*
-Copyright 2024 The Hyperlight Authors.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
+use std::sync::Once;
```
simongdavies: Where did the copyright header go?

src/hyperlight_host/src/metrics/mod.rs
```diff
@@ -1,355 +1,367 @@
...
-pub trait HyperlightMetricOps {
-    /// Converts the enum into a HyperlightMetric
-    fn get_metric(&self) -> Result<&HyperlightMetric>;
-}
-
-/// A trait that should be implemented by all hyperlight metric definitions to convert the metric into a HyperlightMetric
-pub trait GetHyperlightMetric<T> {
-    /// Converts the metric into a HyperlightMetric
-    fn metric(&self) -> Result<&T>;
-}
```
simongdavies: If this is private to the crate how would something like hyperlight_wasm use it?

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -435,38 +432,22 @@ impl HypervisorHandler {
...
                                     .lock
-                                    .try_read();
+                                        .try_read();
 
-                                let res = {
-                                    #[cfg(feature = "function_call_metrics")]
-                                    {
-                                        let start = std::time::Instant::now();
-                                        let result = hv.dispatch_call_from_host(
+                                let res = HistogramMetric::time_and_emit_guest_call(
```
simongdavies: I am not sure I like this way of emitting metrics, it requires a clone of the function name even if we are not emitting metrics, so its not really a no-op

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -181,20 +182,7 @@ fn call_host_func_impl(
...
-                &crate::sandbox::metrics::SandboxMetric::HostFunctionCallsDurationMicroseconds,
-                &[name],
-                start.elapsed().as_micros() as f64
-            );
-            result
-        }
-
-        #[cfg(not(feature = "function_call_metrics"))]
-        func.call(args)
+        HistogramMetric::time_and_emit_host_call(name.to_string(), || func.call(args.clone()))
```
simongdavies: Again I am not sure its worth doing this this way, we need to call .to_string and args.clone each time we make a call which is overhead when not recording metrics

simongdavies (changes requested):
A couple of general questions:

Did we get rid of Gauges? There was a use of them previously and if I am not mistaken somewhere we have an issue that asks for a running total of memory in use, in addition the intent was that these could be used by things built on top of hyperlight such as hyperlight-wasm (we should maintain consistency in the way metrics are collected and emitted between hyperlight and crates built on top)

I may have commented on this in the code review but I dont see how a crate such as hyperlight-wasm would be able to use the metrics structs and macros?

src/hyperlight_host/src/metrics/metrics_macro.rs
ludfjig: fixed

src/hyperlight_host/src/metrics/mod.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
ludfjig: Fixed

Justfile
ludfjig: I made it so metrics exist regardless of features. But 2 types of metrics are emitted only with the feature enabled. This test emits metrics "no-matter-what", so it's fine to not enable metrics feature here.

src/hyperlight_host/src/metrics/metrics_macro.rs
ludfjig: Oh yeah this is old code. I used to allow specifying #[feature] on a specific enum variant. But I removed that bit when calling the macro to make all variants always exit regardless of features.

src/hyperlight_host/src/sandbox/host_funcs.rs
ludfjig: Whoops fixed

ludfjig:
I totally changed the direction of this PR. We now simply use the macros directly where we want to emit metrics. I dont think we need any further abstractions @simongdavies

src/hyperlight_host/src/metrics/mod.rs
ludfjig: removed all these apis from PR

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -435,38 +431,22 @@ impl HypervisorHandler {
...
                                     .lock
-                                    .try_read();
+                                        .try_read();
 
-                                let res = {
-                                    #[cfg(feature = "function_call_metrics")]
-                                    {
-                                        let start = std::time::Instant::now();
-                                        let result = hv.dispatch_call_from_host(
+                                let res = crate::metrics::time_and_emit_guest_call(
```
danbugs: Seems a little confusing to me that, when we call the `time_and_emit_guest_call`, it might not do either (i.e., time and emit a guest call) if the `function_call_metrics` feature is not enabled. I think the previous approach having the explicit `#[cfg(feature = "function_call_metrics")]` here was a little more intuitive. Regardless, this is non-blocking.

src/hyperlight_host/src/func/guest_err.rs
```diff
@@ -31,7 +29,8 @@ pub(crate) fn check_for_guest_error(mgr: &MemMgrWrapper<HostSharedMemory>) -> Re
         ErrorCode::NoError => Ok(()),
         ErrorCode::OutbError => match mgr.as_ref().get_host_error()? {
             Some(host_err) => {
-                increment_guest_error_count(&guest_err);
+                metrics::counter!(METRIC_GUEST_ERROR, METRIC_GUEST_ERROR_LABEL_CODE => (guest_err.code as u64).to_string()).increment(1);
```
danbugs, line 32: In here, we are technically getting a host error. Is it worth considering adding an extra "host error" metric?

danbugs (dismissed):
LGTM

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
ludfjig: Fixed by renaming them to `maybe_*` instead.

src/hyperlight_host/src/func/guest_err.rs
ludfjig: I realized we are never using host errors currently. So no :D. We need to clean up our error handling. Maybe this is part of your guest memory changes? If not I can create an issue for it

danbugs: Yeah, I'm doing part of that in #297 . We can reconvene then 👍

danbugs (approved):

PR #348, opened 2025-03-11
Switch to mostly manual release notes

ludfjig:
### Summary

This PR introduces a `CHANGELOG.md` file to the Hyperlight repository to systematically track and document changes. The goal is for this file to be updated with each PR that contributes to the release notes. However, in practice, it will be manually edited by the individual responsible for creating a Hyperlight release.

### Motivation

The need for manual release notes arose from the limitations of our automatic release notes system. The current automatic notes were generated from PR titles, which often don't adequately reflect the changes in a manner that's easily consumable by users.

### Changes

- A `CHANGELOG.md` file has been added to document key changes.
- The GitHub release notes will now include both the content from the `CHANGELOG.md` and a full list of PRs with their titles at the end of the release notes.

This approach draws inspiration from the release notes format used by Runwasi, as seen in [this example](https://github.com/containerd/runwasi/releases).

After this merges, I will replace the old release notes in 0.2.0 with the new ones

PR #370, opened 2025-03-19
Switch release job to use linux runner

ludfjig:
This change allows extract-changelog.sh to run, which is used for creating release notes, which was introduced in #348.
The release job is currently failing, see https://github.com/hyperlight-dev/hyperlight/actions/runs/13932381161/job/38993460531

gh-cli is already included in the ubuntu image so removing that step

### DIFF
--- a/.github/workflows/CreateRelease.yml
+++ b/.github/workflows/CreateRelease.yml
@@ -78,7 +78,7 @@ jobs:
   publish:
     # see https://github.com/orgs/community/discussions/26286#discussioncomment-3251208 for why we need to check the ref
     if: ${{ contains(github.ref, 'refs/heads/release/') }} ||  ${{ github.ref=='refs/heads/main' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/CreateRelease.yml
```diff
@@ -78,7 +78,7 @@ jobs:
   publish:
     # see https://github.com/orgs/community/discussions/26286#discussioncomment-3251208 for why we need to check the ref
     if: ${{ contains(github.ref, 'refs/heads/release/') }} ||  ${{ github.ref=='refs/heads/main' }}
-    runs-on: windows-2022
+    runs-on: ubuntu-latest
```
marosset: Can you use our KVM runner?
we pre-install the some cargo components

ludfjig: fixed

marosset (approved):

PR #371, opened 2025-03-19
Fix benchmark artifact names in release job

ludfjig:
Fixes the tarfiles of our benchmarking artifacts to match the ones that we create. When switching to linux runners in #370, this became a hard-errror: https://github.com/hyperlight-dev/hyperlight/actions/runs/13954211641/job/39062218141


marosset:
how exactly did this break?

marosset (approved):

PR #376, opened 2025-03-24
Add component bindgen macros

syntactically:
This adds a macro that allows generating a host-side interface to a Hyperlight sandbox based on a WebAssembly Component Type (roughly equivalent to a WIT file).


src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -622,13 +622,20 @@ impl HypervisorHandler {
...
         // In this case, we will wait indefinitely for a message from the handler thread.
         // Note: This applies to all the running sandboxes, not just the one being debugged.
-        #[cfg(gdb)]
-        let response = self.communication_channels.from_handler_rx.recv();
         #[cfg(not(gdb))]
-        let response = self
-            .communication_channels
-            .from_handler_rx
-            .recv_timeout(self.execution_variables.get_timeout()?);
+        let timeout = self.execution_variables.get_timeout()?;
```
simongdavies: nit: `cfg_if!` would probably be better here

src/hyperlight_host/src/func/host_functions.rs
```diff
@@ -36,15 +85,15 @@ macro_rules! host_function {
                 /// Register the host function with the given name in the sandbox.
                 fn register(
                     &self,
-                    sandbox: &mut UninitializedSandbox,
+                    sandbox: &mut dyn Registerable,
```
simongdavies: nit: should we change the name of the variable `sandbox` too?

src/hyperlight_common/src/resource.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+unsafe impl<T: Send> Send for ResourceEntry<T> {}
+
+pub struct LentResourceGuard<'a> {
+    flag: Arc<RwLock<bool>>,
+    already_revoked: bool,
+    _phantom: core::marker::PhantomData<&'a mut ()>,
+}
+impl<'a> LentResourceGuard<'a> {
+    pub fn revoke_nonblocking(&mut self) -> bool {
```
simongdavies, line 73: Can this be called multiple times? If so is it worth checking `already_revoked` is true at the top?

simongdavies (commented):
I left a couple of minor nits.

This definitely needs a couple of other folks to review but generally I think that this needs two things:

More comments (especially at the method/function level in th hyperlight_comppnent_util crate)

Unit tests

src/hyperlight_component_macro/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+extern crate proc_macro;
+
+use hyperlight_component_util::*;
+
+#[proc_macro]
+pub fn host_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
+    let path: Option<syn::LitStr> = syn::parse_macro_input!(input as Option<syn::LitStr>);
+    let path = path
+        .map(|x| x.value().into())
+        .unwrap_or_else(|| std::env::var_os("HYPERLIGHT_WASM_WORLD").unwrap());
```
simongdavies: Could we find a different name for the env var , IMO its a bit strange/confusing that this is called `HYPERLIGHT_WASM_WORLD`

syntactically: How about a simpler `HYPERLIGHT_WORLD` then?

src/hyperlight_common/src/resource.rs
syntactically: I think it's semantically a bit nonsensical to call it repeatedly, but I don't think it really matters. The reason why this is here is for a caller that wants to avoid blocking behaviour on `Drop` if possible, which is why that flag is set when the nonblocking mutex acquire succeeds.

src/hyperlight_host/src/func/host_functions.rs
syntactically: I think sandbox is a more descriptive name, since it is _some_ kind of sandbox (whether a hyperlight-core one or a hyperlight-wasm one)

src/hyperlight_component_macro/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+description = """
+Procedural macros to generate Hyperlight host and guest bindings from component types
+"""
+
+[lib]
+name = "hyperlight_component_macro"
+proc-macro = true
+
+[dependencies]
+wasmparser = { version = "0.224.0" }
```
Mossaka, line 19: I thought hyperlight is independent of the wasm guest so perhaps we should move this crate and hyperlight-component-util to hyperlight-wasm repo?

src/hyperlight_component_macro/src/lib.rs
Mossaka: TinyGo uses `--wit-package` and `--wit-world` flags to specify a specific WIT World from a WIT package has one package may have multiple worlds (think of `wasi:cli/imports` and `wasi:cli/command` worlds). 

I would suggest for `WIT_WOLRD` and `WIT_PACKAGE` env vars.

src/hyperlight_component_macro/Cargo.toml
carlos2martinize: If you look through lang it has to abide with apple, no third party has to follow a special viewing node for tri direction control

syntactically: > I thought hyperlight is independent of the wasm guest so perhaps we should move this crate and hyperlight-component-util to hyperlight-wasm repo?

We had a general consensus that we wanted to use WIT to express the interface between host and guest even for hyperlight-core, which brings us a lot of benefits as to analyzability, performance, etc.

src/hyperlight_component_macro/src/lib.rs
jsturtevant: would it make sense to model this like other wit bindgen tools and provide a parameter to macro with the path?

```
generate!({
    world: "my-world",
    path: "../path/to/wit",
```

https://docs.rs/wit-bindgen/0.42.1/wit_bindgen/macro.generate.html

syntactically:
This now includes a guest bindgen macro as well, because that turned out to be the easiest way to test the marshalling code without having to put all the tests in hyperlight-wasm.

src/hyperlight_common/src/resource.rs
```diff
@@ -0,0 +1,170 @@
...
+/// bunch of them going at once is fine.  Safe Rust in the host can't
+/// use any earlier borrows (potentially invalidating these) until
+/// borrow passed into [`ResourceEntry::lend`] has expired.  Because
+/// that borrow outlives the [`LentResourceGuard`], it will not expire
+/// until that destructor is called. That destructor ensures that (a)
+/// there are no outstanding [`BorrowedResourceGuard`]s alive (since
+/// they would be holding the read side of the [`RwLock`] if they
+/// were), and that (b) the shared flag has been set to false, so
+/// [`ResourceEntry::borrow`] will never create another borrow
+pub enum ResourceEntry<T> {
```
danbugs, line 60: Might be good to point out that these structures are used by things like the `host_bindgen!` macro?

src/hyperlight_component_util/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "hyperlight-component-util"
```
danbugs, line 2: Not sure if you do this later, but we might want to add publishing this crate to crates.io to our CI.

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,94 @@
...
+        None => match ctx.types.get(ce.index as usize) {
+            Some(t) => t,
+            t => panic!("bad component type 2 {:?}", t),
+        },
+        _ => panic!("non-component ascribed type"),
+    }
+}
+
+/// Find the last exported type in a component, since in wasm-encoded
+/// WIT this is typically the main world to use.  This is a very
```
danbugs, line 57: What happens currently if this is not the case? Do we have tests for this?

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,94 @@
...
+/// for a specific named world, instead of simply grabbing the last
+/// export.
+pub fn read_component_single_exported_type<'a>(
+    items: impl Iterator<Item = wasmparser::Result<Payload<'a>>>,
+) -> Component<'a> {
+    let mut ctx = Ctx::new(None, false);
+    let mut last_idx = None;
+    for x in items {
+        match x {
+            Ok(Version { .. }) => (),
```
danbugs: Is this line needed considering we have a do-nothing catch-all at the end? Did you plan to check for a specific version?

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,94 @@
...
+                        Ok(ce) => {
+                            if ce.kind == ComponentExternalKind::Type {
+                                last_idx = Some(ctx.types.len());
+                                ctx.types.push(raw_type_export_type(&ctx, &ce).clone());
+                            }
+                        }
+                    }
+                }
+            }
+            _ => {}
```
danbugs, line 149: Should we perhaps log that we are ignoring part of the Wasm binary?

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,94 @@
...
+                    match t {
+                        Ok(ComponentType::Component(ct)) => {
+                            let ct_ = ctx.elab_component(&ct);
+                            ctx.types.push(Defined::Component(ct_.unwrap()));
+                        }
+                        _ => panic!("non-component type"),
+                    }
+                }
+            }
+            Ok(ComponentExportSection(es)) => {
```
danbugs, line 109: Not being familiar w/ wasmparser payload types, I think adding a comment indicating that smt like `ComponentExportSection` equates to `import`s/`export`s of a world and the world itself would be helpful for the maintainability of the codebase (same idea for `ComponentTypeSection` too!). 

For reference of other reviewers, the `ComponentExportSection`s for our `witguest/interface.wasm` are the following:

```
ComponentExportSection:
  Export: name='ComponentExportName(
    "roundtrip",
)', kind="type", index=0
ComponentExportSection:
  Export: name='ComponentExportName(
    "host-resource",
)', kind="type", index=2
ComponentExportSection:
  Export: name='ComponentExportName(
    "test-host-resource",
)', kind="type", index=4
ComponentExportSection:
  Export: name='ComponentExportName(
    "test",
)', kind="type", index=6
```

This is for:
```
world test {
  import roundtrip;
  import host-resource;
  export roundtrip;
  export test-host-resource;
}
```

You can see that the last export is the `test` `world` itself.

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,94 @@
...
+/// From [`wasmparser::ComponentExport`], elaborate a deftype_e as per
+/// the specification.
+fn raw_type_export_type<'p, 'a, 'c>(
+    ctx: &'c Ctx<'p, 'a>,
+    ce: &'c wasmparser::ComponentExport<'a>,
+) -> &'c Defined<'a> {
+    match ce.ty {
+        Some(ComponentTypeRef::Component(n)) => match ctx.types.get(n as usize) {
+            Some(t) => t,
+            t => panic!("bad component type 1 {:?}", t),
```
danbugs: What are bad component types 1 and 2? 🙂

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,94 @@
...
+                }
+            }
+            _ => {}
+        }
+    }
+    match last_idx {
+        None => panic!("no exported type"),
+        Some(n) => match ctx.types.into_iter().nth(n) {
+            Some(Defined::Component(c)) => c,
+            _ => panic!("final export is not component"),
```
danbugs, line 156: Is this true? If we got here and `last_idx` is `Some`, we did parse a `ComponentExportSection`.

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+        &'c mut self,
+        ed: &ComponentTypeRef,
+    ) -> Result<(Vec<BoundedTyvar<'a>>, ExternDesc<'a>), Error<'a>> {
+        match ed {
+            ComponentTypeRef::Module(i) => match &self.core.types[*i as usize] {
+                CoreDefined::Module(mt) => Ok((vec![], ExternDesc::CoreModule(mt.clone()))),
+                _ => {
+                    panic!("internal invariant violation: bad sort for ComponentTypeRef to Module")
+                }
+            },
```
danbugs, line 479: If we got here, this means we elaborated to `ComponentTypeDeclaration::Export` so now we are parsing things inside that exported interface. What does it mean to be parsing a `Module`? Can a component interface contain a module? Would be curious to see what that looks like w/ WIT. I tried to get here parsing our `interface.wasm` and I think this code path is untested. It might be worth creating some more Wasm components and test this stuff outside its macro usage. There's probably more untested code paths that I think we should catch. Perhaps we can leverage something like Wasm smith for coverage?

src/hyperlight_component_util/src/wf.rs
```diff
@@ -0,0 +1,379 @@
...
+        Err(e(x))
+    } else {
+        Ok(())
+    }
+}
+
+/// # Well-formedness
+///
+/// Most of this is a very direct translation of the specification
+/// (Well-formedness subsections of section 3.4 Type Elaboration).
```
danbugs, line 156: I'm assuming this is the _component_ spec? Would be good to be able to link to the spec version we are implementing...

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+                imports.push(import);
+            }
+            if let Some(export) = export {
+                exports.push(export);
+            }
+        }
+        ctx.finish_component(&imports, &exports)
+    }
+
+    fn elab_core_module_decl<'c>(
```
danbugs, line 162: `elab_core_module_decl` is currently untested by our `witguest`. If we haven't tested this logic, should we instead just not support elabbing this?

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+    }
+
+    fn elab_core_type_rec<'c>(&'c mut self, rg: &wasmparser::RecGroup) -> CoreDefined<'a> {
+        match &rg.types().nth(0).unwrap().composite_type.inner {
+            CompositeInnerType::Func(ft) => CoreDefined::Func(ft.clone()),
+            _ => panic!("GC core types are not presently supported"),
+        }
+    }
+
+    fn elab_core_type<'c>(&'c mut self, ct: &wasmparser::CoreType<'a>) -> CoreDefined<'a> {
```
danbugs, line 230: Not being exhaustive here, but this is also untested.

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+                .map(|(n, vt)| {
+                    Ok(Param {
+                        name: Name { name: n },
+                        ty: self.elab_value(vt)?,
+                    })
+                })
+                .collect::<Result<Vec<_>, Error<'a>>>()?,
+            result: match &ft.results {
+                ComponentFuncResult::Unnamed(vt) => etypes::Result::Unnamed(self.elab_value(vt)?),
+                ComponentFuncResult::Named(rs) => etypes::Result::Named(
```
danbugs, line 451: What makes a function Unamed vs. Named? From our parsing of `interface.wasm`, most functions are `unnamed`—is that correct?

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+            ComponentType::Instance(ids) => Ok(Defined::Instance(self.elab_instance(ids)?)),
+            ComponentType::Resource { dtor, .. } => {
+                let rid = ResourceId {
+                    id: self.rtypes.len() as u32,
+                };
+                self.rtypes.push(Resource { _dtor: *dtor });
+                Ok(Defined::Handleable(Handleable::Resource(rid)))
+            }
+        }
+    }
```
danbugs, line 711: Should we instead just elab the stuff in here + externs and remove untested elabs if we want to avoid spending a lot of time on test coverage rn?

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
+/*
```
danbugs, line 1: This elaboration logic is not specific to Hyperlight in any way, right? Is there no "wasmelaborate" crate or alike that we could use to lessen the burden of maintaining this or are we creating our own definitions to reduce the elaboration scope?

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+                CoreDefined::Module(mt) => Ok((vec![], ExternDesc::CoreModule(mt.clone()))),
+                _ => {
+                    panic!("internal invariant violation: bad sort for ComponentTypeRef to Module")
+                }
+            },
+            ComponentTypeRef::Func(i) => match &self.types[*i as usize] {
+                Defined::Func(ft) => Ok((vec![], ExternDesc::Func(ft.clone()))),
+                _ => panic!("internal invariant violation: bad sort for ComponentTypeRef to Func"),
+            },
+            ComponentTypeRef::Value(_) => panic!("First-class values are not yet supported"),
```
danbugs, line 484: Should this be an issue or is this not supported in the spec? Same for async and alike.

src/hyperlight_component_util/src/guest.rs
```diff
@@ -0,0 +1,366 @@
...
+            let tn = kebab_to_type(wn.name);
+            quote! {
+                type #tn = Self;
+                #[allow(refining_impl_trait)]
+                fn #getter<'a>(&'a mut self) -> &'a mut Self {
+                    self
+                }
+            }
+        }
+        ExternDesc::Component(_) => {
```
danbugs, line 129: We can elaborate nested components but not create rust bindings. Should we remove the support for elaborating nested components?

src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -0,0 +1,684 @@
...
+    fn elab_instance<'c>(
+        &'c mut self,
+        decls: &[InstanceTypeDeclaration<'a>],
+    ) -> Result<QualifiedInstance<'a>, Error<'a>> {
+        let mut ctx = Ctx::new(Some(self), false);
+        let mut exports = Vec::new();
+        for decl in decls {
+            let export = ctx.elab_instance_decl(decl)?;
+            if let Some(export) = export {
+                exports.push(export);
```
danbugs, line 556: Is it fine to end up w/ multiple repeated exports? For example, while testing, I got multiple `roundtrip-u16` pushes.

src/hyperlight_component_util/src/tv.rs
```diff
@@ -0,0 +1,119 @@
...
+};
+use crate::substitute::{self, Substitution, Unvoidable};
+
+/// The most information we possibly have about a type variable
+pub enum ResolvedTyvar<'a> {
+    /// Invariant: the head of this [`Defined`] is not `[Defined::Handleable]([HHandleable::Var](...))`
+    Definite(Defined<'a>),
+    /// It's just some bound var... so there is no way to look it up.
+    #[allow(unused)]
+    Bound(u32),
```
danbugs, line 28: Do we want this unused?

src/hyperlight_component_util/src/etypes.rs
```diff
@@ -0,0 +1,506 @@
+/*
```
danbugs, line 1: This is another section that is not really specific to Hyperlight. Is there no definition of these in any crate that we can use out-of-the-box?

src/hyperlight_component_util/src/util.rs
```diff
@@ -0,0 +1,70 @@
...
+    let manifest_dir = std::path::Path::new(&manifest_dir);
+    let path = manifest_dir.join(path);
+
+    let bytes = std::fs::read(path).unwrap();
+    let i = wasmparser::Parser::new(0).parse_all(&bytes);
+    let ct = crate::component::read_component_single_exported_type(i);
+
+    // because of the two-level encapsulation scheme, we need to look
+    // for the single export of the component type that we just read
+    #[allow(unused_parens)]
```
danbugs: Needed?

src/hyperlight_component_util/src/util.rs
```diff
@@ -0,0 +1,70 @@
...
+        || !ct.imports.is_empty()
+        || !ct.instance.evars.is_empty()
+        || ct.instance.unqualified.exports.len() != 1
+    {
+        panic!("malformed component type container for wit type");
+    };
+    let export = &ct.instance.unqualified.exports[0];
+    use etypes::ExternDesc;
+    let ExternDesc::Component(ct) = &export.desc else {
+        panic!("component type container does not contain component type");
```
danbugs: Is this possible?

src/hyperlight_component_util/src/util.rs
```diff
@@ -0,0 +1,70 @@
...
+    };
+    dbg_println!("hcm: considering component type {:?}", ct);
```
danbugs: Can we do logging here instead?

src/hyperlight_component_util/src/guest.rs
```diff
@@ -0,0 +1,366 @@
+/*
+Copyright 2024 The Hyperlight Authors.
```
danbugs: All the copyright headers should be changed to 2025!

src/tests/rust_guests/witguest/src/main.rs
```diff
@@ -0,0 +1,168 @@
...
+extern crate alloc;
+extern crate hyperlight_guest;
+
+mod bindings;
+use bindings::*;
+
+struct Guest {}
+
+impl test::wit::Roundtrip for Guest {
+    fn roundtrip_bool(&mut self, x: bool) -> bool {
```
danbugs, line 29: I see that the witguest has deps on hyperlight-guest and hyperlight-guest-bin. Should we try adding functions that use those dependencies to call host functions directly and do stuff like that? What currently happens if they try? Is this something we want to support considering it circumvents the bindings?

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -0,0 +1,295 @@
...
+                assert_eq!(self_.x, "strab");
+                assert_eq!(self_.last, 'b');
+            }
+            _ => panic!(),
+        };
+        self_.n_calls += 1;
+        self_.x.push(c);
+        self_.last = c;
+    }
+}
```
danbugs: Add comments on the behaviour you are verifying here?

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -0,0 +1,295 @@
...
+
+impl test::wit::HostResource for Host {
+    fn roundtrip_own(&mut self, owned: Arc<Mutex<TestResource>>) -> Arc<Mutex<TestResource>> {
+        owned
+    }
+
+    fn return_own(&mut self, _: Arc<Mutex<TestResource>>) {}
```
danbugs: Is this supposed to do nothing?

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -0,0 +1,295 @@
...
+
+struct TestResource {
+    n_calls: u32,
+    x: String,
+    last: char,
+}
+
+use std::sync::atomic::AtomicBool;
+use std::sync::atomic::Ordering::Relaxed;
+static HAS_BEEN_DROPPED: AtomicBool = AtomicBool::new(false);
```
danbugs, line 134: If we test this w/ parallelism, this breaks—yeah?

ludfjig (commented):
I really appreciate your commitment to documenting all of this so very thoroughly. This is some fantastic work! I'm also wondering what exact component model spec you are referencing throughout the code.

I'm not going to pretend to understand it all but looks reasonable on an initial read-through. Btw, I'm curious what the reason is for useing wasm-parser instead of wit-parser, and not read the wit directly?

src/hyperlight_component_util/src/component.rs
syntactically: We did indeed get a `ComponentExportSection` that exports a type, but in this case, the type that is being exported is not a component type (e.g. it might be a resource type, a value type like a struct definition, an instance type, etc).

src/hyperlight_component_util/src/elaborate.rs
syntactically: An instance/component can indeed export(/import) a core module, but just like higher-order components that import or export entire components (this, like those, is a feature intended primarily for "linker script"-like components), it's not yet super clear whether that makes sense to surface in Hyperlight.  Currently, if we actually try to emit macro code for these it will error out in rtypes, just like it would for nested component imports/exports.

The code is here partly to keep this more in sync with the spec document and OCaml interpreter, which makes it a bit more maintainable because it's easy to see the correspondences, and partly because I would, when we have an official test suite set up upstream in the component model spec repo, like to try to run this on the official `.wast`s, which I think is the sensible path to having the elaborator tested in more isolation.  It's not super relevant for what we are using this for right now, though, so I don't think it makes sense for us to block merging this and using the portions of it that we do need/test on waiting for the good collection of wasts to exist.

src/hyperlight_component_util/src/wf.rs
syntactically: The formal spec is a bit in flux with the move from 0.2 to 0.3 right now, so this is referring to section numbers that I think are correct for the very old spec draft that's floating around. I'm going to be getting a more up-to-date draft (although perhaps not with all of 0.3) out on the upstream component model repo in the next few weeks, and will try to find a way to host a built copy of it somewhere; once that's done, I was planning on coming back here to add a link.

src/hyperlight_component_util/src/elaborate.rs
syntactically: See response above

src/hyperlight_component_util/src/elaborate.rs
syntactically: See response above

src/hyperlight_component_util/src/elaborate.rs
syntactically: This is a little bit of a leftover from a version of the upstream specification where functions could return an anonymous tuple of named results, in order to make them symmetric with parameters. The spec upstream has replaced that with only a single optional return value, but this still has the IR using the old options because my IR spec in other places is coupled with this for 0.2 and only has it changed in 0.3. So, right now, anything that returns something will return an unnamed result, and anything that does not return anything returns an empty named tuple.

src/hyperlight_component_util/src/elaborate.rs
syntactically: See comment above

src/hyperlight_component_util/src/elaborate.rs
syntactically: Correct, it is not specific to hyperlight. I would like to factor it out into a more reusable crate at some point, but it's not a high priority right now.

In the medium term, actually, what I would really like is to package the OCaml reference interpreter that we have in the upstream specification as a component, and then use component tooling to run that component in the proc macro at compile time to do elaboration, and get rid of all this extra Rust duplication of something that already exists. Unfortunately, the wasm-targeting toolchain is not there in OCaml yet, so it ended up being a lot faster to just semi-mechanically translate OCaml to rust :(.

src/hyperlight_component_util/src/elaborate.rs
syntactically: The upstream spec is in evolution on these topics (related to 0.3.x releases). We don't support async or first-class component values right now because they were not in 0.2, but wasmparser is also used by some of the projects that are working out the details of 0.3 before it is released.

src/hyperlight_component_util/src/guest.rs
syntactically: See comment above

src/hyperlight_component_util/src/tv.rs
syntactically: Yes. I think currently the downstream code gives up and panics when this case is encountered, but it is the semantically sensible thing to return, and there might be cases where it is useful later. This is all stuff that should morally be library interfaces if we separate out a more reusable crate.

src/hyperlight_component_util/src/etypes.rs
syntactically: See comment above

src/hyperlight_component_util/src/util.rs
syntactically: As with the stuff in component.rs, the component grammar certainly lets you write files that would get here, but for the specific thing that we care about, which is components of a specific form with one exported component type, with wit-style package names everywhere allowing codegen to be nice, you shouldn't, so it would imply something was wrong with the input file. Changed the error message to more clearly state that the input is malformed.

src/tests/rust_guests/witguest/src/main.rs
syntactically: I think that should not be supported, in large part because I would like the bindgen'd wit interfaces to be a way for users to be sure that as we overhaul the whole set of actual communication primitives over the next few months, they won't feel that. I don't see any reasonable way to prohibit it right now, though, unfortunately---the guest crate has to depend on these "implementation detail" crates because the code generated by the macros needs to be able to access names in them, and an appropriate level of hygiene to make that not an issue is not available to us.

src/hyperlight_component_util/src/component.rs
syntactically: It was just for clarity (to make it clear that that section is expected/handled), but it's not a bad idea to add a version check---added.

src/hyperlight_common/src/resource.rs
syntactically: Added a mention to the comment that this is here primarily for the use of the hyperlight_component_macro users.

src/hyperlight_component_util/Cargo.toml
syntactically: Added

src/hyperlight_component_util/src/component.rs
syntactically: Added a clarifying comment

src/hyperlight_component_util/src/component.rs
syntactically: Added a more descriptive comment explaining why

src/hyperlight_component_util/src/component.rs
syntactically: Replaced with more descriptive errors

src/hyperlight_component_util/src/component.rs
syntactically: Added a bit about the structure of WIT-encoded components that covers this to the header of the function

src/hyperlight_component_util/src/elaborate.rs
syntactically: Good question. Most of this code assumes that the component type in question has already passed full validation, but this is really low-hanging fruit to guard against by adding an appropriate wf_component() call, so I've added that.

src/hyperlight_component_util/src/util.rs
syntactically: Fixed

src/hyperlight_component_util/src/util.rs
syntactically: Good catch; this may be obsolete.

src/hyperlight_component_util/src/guest.rs
syntactically: Fixed

src/hyperlight_host/tests/wit_test.rs
syntactically: Added a comment

src/hyperlight_host/tests/wit_test.rs
syntactically: Added a comment

src/hyperlight_component_macro/src/lib.rs
syntactically: Changed to WIT_WORLD (although, component types don't necessarily have to come from the WIT language).

syntactically:
> Btw, I'm curious what the reason is for useing wasm-parser instead of wit-parser, and not read the wit directly?

WIT is in even more a state of flux than the rest of the specification, and is a bit of a pain to work with, being arbitrarily limited in a bunch of ways. We are trying to extend it to be a bit more in line with the actual core language of component types, but that's not done. The component type language is a bit more general and a lot easier to parse/elaborate/deal with and better specified at the moment, so I just used it directly.  From a devux perspective, we could probably pretty easily add a dependency on wit-parser to be able to desugar wit files provided by the user without asking them to do it first, but I think that's better left for a follow-up.

src/hyperlight_host/src/sandbox/callable.rs
```diff
@@ -0,0 +1,28 @@
+/*
+Copyright 2024 The Hyperlight Authors.
```
danbugs: Just missed this one 2024 copyright label. If you update this one, might also want to update the ones in `src/hyperlight_host/src/func/utils.rs` and `src/hyperlight_host/src/hypervisor/crashdump.rs`. I think we missed them in a previous PR.

src/hyperlight_component_util/src/component.rs
```diff
@@ -0,0 +1,159 @@
...
+                        | Ok(ComponentAlias::Outer {
+                            kind: ComponentOuterAliasKind::Type,
+                            ..
+                        }) => {
+                            panic!("Component outer type aliases are not supported")
+                        }
+                        // Anything else doesn't affect the index
+                        // space that we are interested in, so we can
+                        // safely ignore
+                        _ => {}
```
danbugs, line 138: Before, you used dbgprintln! to print some generally useful debugging info. Would it be useful to log what we are ignoring here?

ludfjig (dismissed):

src/hyperlight_component_util/src/util.rs
```diff
@@ -0,0 +1,69 @@
...
+    let export = &ct.instance.unqualified.exports[0];
+    use etypes::ExternDesc;
+    let ExternDesc::Component(ct) = &export.desc else {
+        panic!("malformed component type container: does not contain component type");
+    };
+    log::debug!("hcm: considering component type {:?}", ct);
+    cb(export.kebab_name.to_string(), ct)
+}
+
+/// Deal with `$HYPERLIGHT_COMPONENT_MACRO_DEBUG`: if it is present,
```
danbugs, line 53: Maybe document somewhere that we can set this env var to the path of a debug file to view the result of the macro expression? I think this is the only place it is mentioned. Adding a `...-development-notes.md` document about the component work to `docs/` in general could also be helpful too.

.github/workflows/CargoPublish.yml
```diff
@@ -71,6 +71,18 @@ jobs:
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_PUBLISH_TOKEN }}
 
+      - name: Publish hyperlight-component-util
```
danbugs, line 74: Might want to get these names reserved on crates.io 👍

src/hyperlight_component_util/src/elaborate.rs
danbugs: Ok. That sounds reasonable. Could you maybe just add an issue documenting the code paths that are currently untested to track that effort?

danbugs (commented):
Mostly LGTM. Just some minor comments.

dblnz (dismissed):
There are a lot of changes to review but overall it looks good to me.
Awesome work, @syntactically .

simongdavies:
> > Btw, I'm curious what the reason is for useing wasm-parser instead of wit-parser, and not read the wit directly?
> 
> WIT is in even more a state of flux than the rest of the specification, and is a bit of a pain to work with, being arbitrarily limited in a bunch of ways. We are trying to extend it to be a bit more in line with the actual core language of component types, but that's not done. The component type language is a bit more general and a lot easier to parse/elaborate/deal with and better specified at the moment, so I just used it directly. From a devux perspective, we could probably pretty easily add a dependency on wit-parser to be able to desugar wit files provided by the user without asking them to do it first, but I think that's better left for a follow-up.

1. Can we get an issue for that?
2. Is there a point in the future where we will be able to move to using a wit-parser instead?

src/hyperlight_common/src/resource.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+impl<'a> Drop for LentResourceGuard<'a> {
+    fn drop(&mut self) {
+        if !self.already_revoked {
+            #[allow(unused_mut)] // it isn't actually unused
+            let mut guard = self.flag.write();
+            #[cfg(feature = "std")]
+            // If a mutex that is just protecting us from our own
+            // mistakes is poisoned, something is so seriously
+            // wrong that dying is a sensible response.
```
simongdavies, line 95: Are you sure you want to kill our host for this?

src/hyperlight_common/src/resource.rs
```diff
@@ -0,0 +1,172 @@
...
+            ResourceEntry::Owned(t) => Some(BorrowedResourceGuard {
+                _flag: None,
+                reference: t,
+            }),
+            ResourceEntry::Borrowed(flag, t) => {
+                let guard = flag.read();
+                // If a mutex that is just protecting us from our own
+                // mistakes is poisoned, something is so seriously
+                // wrong that dying is a sensible response.
+                #[allow(clippy::unwrap_used)]
```
simongdavies, line 144: Same here, killing the host should be a last resort I believe, I am just making sure that we really want to do that?

simongdavies (changes requested):
The abstractions are great.

One request, can we put the bindgen functionality behind a feature flag please?

src/hyperlight_common/src/resource.rs
syntactically: I don't feel confident right now that soundness would be preserved if the mutex was poisoned & we can't revoke the borrow properly?

syntactically:
> 1. Can we get an issue for that?

For adding support for doing the wit desugaring in one go? Sure: #575.

> 2. Is there a point in the future where we will be able to move to using a wit-parser instead?

I don't see any reason we would want to, as long as the desugaring is implemented reasonably correctly somewhere else. As I mentioned before, I would love to move more of the generic component-y code out of this crate and into something reusable (or replace it entirely with a component built from the reference implementation), so I don't see it making sense to add more of that kind of generic thing here.

src/hyperlight_component_util/src/component.rs
syntactically: All the uses of `dbg_println!` were transformed 1-1 to `log::debug!`. I don't think this is useful to log for the same reasons as it isn't for the unknown component definitions.

src/hyperlight_component_util/src/util.rs
syntactically: Added to the crate-level documentation for `hyperlight_component_macro`, which I think is a sensible place for it to be.

src/hyperlight_common/src/resource.rs
syntactically: I will leave this as is for now, but make a follow-up issue for us to investigate whether there is a better solution: #578.

src/hyperlight_host/src/sandbox/callable.rs
danbugs: Saw you updated the copyright label in `callable.rs`. We can leave the other ones for another PR, this is non-blocking, so I'm resolving.

src/hyperlight_component_util/src/elaborate.rs
danbugs: Issue #576 created. Resolving this conversation.

.github/workflows/CargoPublish.yml
danbugs: Done!

https://crates.io/crates/hyperlight-component-macro
https://crates.io/crates/hyperlight-component-util

src/hyperlight_common/src/resource.rs
danbugs: Same as above.

PR #384, opened 2025-03-28
[bugfix] Fix case where memory was reset, but host still tried to read result value from memory

ludfjig:
`terminate_hypervisor_handler_execution_and_reinitialise` is always considered an 'error', except if `HypervisorHandlerExecutionCancelAttemptOnFinishedExecution` was returned.


ludfjig:
Misread the code - this is PR does nothing

PR #410, opened 2025-04-09
Fix issues in guest logging

simongdavies:
Allow explicit setting of max guest log level, , if this is not set check `RUST_LOG` environment variable.

The guest log level impacts the amount of logs produced by the guest, its presently set by examining the value of max_log_level in the host.

This may not be appropriate since logging in the guest is expensive and it may cause logs to be emitted from the guest but ignored in the host. This change introduces a function on an `UninitializedSandbox` to set the level explicitly, if this is not set the env var `RUST_LOG` is parsed to determine the log level so that logs  are only produced if they are going to be consumed.  If `RUST_LOG` is not set then the max guest log level is set to error.

Updates inconsistent crate name `hyperlight-guest` in guest logs to `hyperlight_guest`

### DIFF
--- a/docs/hyperlight-metrics-logs-and-traces.md
+++ b/docs/hyperlight-metrics-logs-and-traces.md
@@ -66,13 +66,13 @@ In the [examples/tracing](../src/hyperlight_host/examples/tracing) directory, th
 #### Linux
 
 ```bash
-RUST_LOG='none,hyperlight-host=info,tracing=info' cargo run --example tracing
+RUST_LOG='none,hyperlight_host=info,tracing=info' cargo run --example tracing
 ```
 
 #### Windows
 
 ```powershell
-$env:RUST_LOG='none,hyperlight-host=info,tracing=info'; cargo run --example tracing
+$env:RUST_LOG='none,hyperlight_host=info,tracing=info'; cargo run --example tracing
 ```
 
 ### Using OTLP exporter and Jaeger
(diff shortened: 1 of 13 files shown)


docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -66,13 +66,13 @@ In the [examples/tracing](../src/hyperlight_host/examples/tracing) directory, th
 #### Linux
 
 ```bash
-RUST_LOG='none,hyperlight-host=info,tracing=info' cargo run --example tracing
+RUST_LOG='none,hyperlight_host=info,tracing=info' cargo run --example tracing
```
dblnz, line 69: I also remember making this change somewhere when the tracing logs were not visible

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
```diff
@@ -99,13 +101,15 @@ pub(super) fn evolve_impl_multi_use(u_sbox: UninitializedSandbox) -> Result<Mult
 }
 
 #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
+#[allow(clippy::too_many_arguments)]
```
dblnz, line 104: nit: I think that we'll need to refactor the functions that take too many arguments at some point.

dblnz (approved):

PR #417, opened 2025-04-14
Support ELF core dump creation on guest crash

### LINKED ISSUE #310: Improve experience when an unexpected VM exit or crash occurs
At the moment, when an unexpected VM exit or crash occurs in a debug build, the VM configuration and memory are dumped to a file.
The contents of the file are pretty difficult to navigate, we need to create some tooling that can parse the file and display formatted output about its contents to aid in diagnosing failures.

This is related to #222 which partially addresses this issue.

dblnz:
## Purpose
- Add support for core dump file generation when a guest crashes #310 

## Description
This depends on [PR](https://github.com/microsoft/elfcore/pull/34).
This closes #310 

- The `crashdump` feature toggles the logic of the creation of core dumps for Hyperlight Sandboxes.
Additionally, core dumps can be configured at the Sandbox level. By default, all the Sandboxes have core dump generation enabled. The user can set the `guest_core_dump = false` in the `SandboxConfiguration` to disable core dumps for specific Sandboxes.
- By default, Hyperlight places the core dumps in the temporary directory (platform specific).
To change this, use the `HYPERLIGHT_CORE_DUMP_DIR` to specify a directory. If the directory does not exist, it will be created.

### DIFF
--- a/docs/how-to-debug-a-hyperlight-guest.md
+++ b/docs/how-to-debug-a-hyperlight-guest.md
@@ -211,6 +211,13 @@ To make Hyperlight dump the state of the vCPU (general purpose registers, regist
 This will result in a dump file being created in the temporary directory.
 The name and location of the dump file will be printed to the console and logged as an error message.
 
+**NOTE**: By enabling the `crashdump` feature, you instruct Hyperlight to create core dump files for all sandboxes when an unhandled crash occurs.
+To selectively disable this feature for a specific sandbox, you can set the `guest_core_dump` field to `false` in the `SandboxConfiguration`.
+```rust
+    let mut cfg = SandboxConfiguration::default();
+    cfg.set_guest_core_dump(false); // Disable core dump for this sandbox
+```
+
 ### Inspecting the core dump
 
 After the core dump has been created, to inspect the state of the guest, load the core dump file using `gdb` or `lldb`.
--- a/docs/how-to-debug-a-hyperlight-guest.md
+++ b/docs/how-to-debug-a-hyperlight-guest.md
@@ -207,10 +207,14 @@ involved in the gdb debugging of a Hyperlight guest running inside a **KVM** or 
 When a guest crashes because of an unknown VmExit or unhandled exception, the vCPU state is dumped to an `ELF` core dump file.
 This can be used to inspect the state of the guest at the time of the crash.
 
-To make Hyperlight dump the state of the vCPU (general purpose registers, registers) to an `ELF` core dump file, set the feature `crashdump` and run a debug build.
-This will result in a dump file being created in the temporary directory.
+To make Hyperlight dump the state of the vCPU (general purpose registers, registers) to an `ELF` core dump file, enable the `crashdump`
+feature and run.
+The feature enables the creation of core dump files for both debug and release builds of Hyperlight hosts.
+By default, Hyperlight places the core dumps in the temporary directory (platform specific).
+To change this, use the `HYPERLIGHT_CORE_DUMP_DIR` environment variable to specify a directory.
 The name and location of the dump file will be printed to the console and logged as an error message.
 
+**NOTE**: If the directory provided by `HYPERLIGHT_CORE_DUMP_DIR` does not exist, Hyperlight places the file in the temporary directory.
 **NOTE**: By enabling the `crashdump` feature, you instruct Hyperlight to create core dump files for all sandboxes when an unhandled crash occurs.
 To selectively disable this feature for a specific sandbox, you can set the `guest_core_dump` field to `false` in the `SandboxConfiguration`.
 ```rust
--- a/docs/how-to-debug-a-hyperlight-guest.md
+++ b/docs/how-to-debug-a-hyperlight-guest.md
@@ -204,15 +204,17 @@ involved in the gdb debugging of a Hyperlight guest running inside a **KVM** or 
 
 ## Dumping the guest state to an ELF core dump when an unhandled crash occurs
 
-When a guest crashes, the vCPU state is dumped to an `ELF` core dump file. This can be used to inspect the state of the guest at the time of the crash.
+When a guest crashes because of an unknown VmExit or unhandled exception, the vCPU state is dumped to an `ELF` core dump file.
+This can be used to inspect the state of the guest at the time of the crash.
 
-To dump the state of the vCPU (general purpose registers, registers) to an `ELF` core dump file set the feature `crashdump` and run a debug build. This will result in a dump file being created in the temporary directory.
+To make Hyperlight dump the state of the vCPU (general purpose registers, registers) to an `ELF` core dump file, set the feature `crashdump` and run a debug build.
+This will result in a dump file being created in the temporary directory.
 The name and location of the dump file will be printed to the console and logged as an error message.
 
 ### Inspecting the core dump
 
 After the core dump has been created, to inspect the state of the guest, load the core dump file using `gdb` or `lldb`.
-A `gdb` version later than `15.0` and `lldb` version later than `17` have been used to test this feature.
+**NOTE: This feature has been tested with version `15.0` of `gdb` and version `17` of `lldb`, earlier versions may not work, it is recommended to use these versions or later.**
 
 To do this in vscode, the following configuration can be used to add debug configurations:
 
@@ -268,7 +270,7 @@ To do this in vscode, the following configuration can be used to add debug confi
     ]
 }
 ```
-NOTE: The `CodeLldb` debug session does not stop after launching. To see the code, stack frames and registers you need to
+**NOTE: The `CodeLldb` debug session does not stop after launching. To see the code, stack frames and registers you need to
 press the `pause` button. This is a known issue with the `CodeLldb` extension [#1245](https://github.com/vadimcn/codelldb/issues/1245).
-The `cppdbg` extension works as expected and stops at the entry point of the program.
+The `cppdbg` extension works as expected and stops at the entry point of the program.**
 
--- a/docs/how-to-debug-a-hyperlight-guest.md
+++ b/docs/how-to-debug-a-hyperlight-guest.md
@@ -274,3 +274,149 @@ To do this in vscode, the following configuration can be used to add debug confi
 press the `pause` button. This is a known issue with the `CodeLldb` extension [#1245](https://github.com/vadimcn/codelldb/issues/1245).
 The `cppdbg` extension works as expected and stops at the entry point of the program.**
 
+## Compiling guests with debug information for release builds
+
+This section explains how to compile a guest with debugging information but still have optimized code, and how to separate the debug information from the binary.
+
+### Creating a release build with debug information
+
+To create a release build with debug information, you can add a custom profile to your `Cargo.toml` file:
+
+```toml
+[profile.release-with-debug]
+inherits = "release"
+debug = true
+```
+
+This creates a new profile called `release-with-debug` that inherits all settings from the release profile but adds debug information.
+
+### Splitting debug information from the binary
+
+To reduce the binary size while still having debug information available, you can split the debug information into a separate file.
+This is useful for production environments where you want smaller binaries but still want to be able to debug crashes.
+
+Here's a step-by-step guide:
+
+1. Build your guest with the release-with-debug profile:
+   ```bash
+   cargo build --profile release-with-debug
+   ```
+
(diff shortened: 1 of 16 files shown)


src/hyperlight_host/Cargo.toml
```diff
@@ -46,6 +46,7 @@ tempfile = { version = "3.19", optional = true }
 serde_yaml = "0.9"
 anyhow = "1.0"
 metrics = "0.24.2"
+elfcore = { git = "https://github.com/dblnz/elfcore.git", branch = "split-linux-impl-from-elfcore" }
```
dblnz: This needs to be change after the `elfcore` PR is merged

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -1,44 +1,274 @@
...
+            .collect();
 
-    // write the raw memory dump for each memory region
-    for region in hv.get_memory_regions() {
-        if region.host_region.start == 0 || region.host_region.is_empty() {
-            continue;
+        // The filename and command line are set to null-terminated strings
+        let filename = ctx
+            .filename
+            .clone()
```
ludfjig:
```suggestion
            .as_ref()
```

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -1,44 +1,274 @@
...
+            .map_or_else(|| "\0".to_string(), |s| format!("{}\0", s));
+
+        let cmd = ctx
+            .binary
+            .clone()
```
ludfjig:
```suggestion
            .as_ref()
```

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -1,44 +1,274 @@
...
+                },
+                mapped_file_name: None,
+            })
+            .collect();
 
-    // write the raw memory dump for each memory region
-    for region in hv.get_memory_regions() {
-        if region.host_region.start == 0 || region.host_region.is_empty() {
-            continue;
+        // The filename and command line are set to null-terminated strings
```
ludfjig: What do you think about setting them to something explicit instead of just null? Maybe `c"<unknown>"` or something like it?

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -1,44 +1,274 @@
...
+            threads: vec![thread],
+            aux_vector: auxv,
+        }
     }
-    temp_file.flush()?;
+}
 
-    // persist the tempfile to disk
-    let persist_path = temp_file.path().with_extension("dmp");
+impl ProcessInfoSource for GuestView {
```
ludfjig, line 179: Are all these values below fine to be hardcoded?

ludfjig (dismissed):
Looks great. Left some nit comments.

By the way (unrelated to your PR), we should simplify each driver to include only the core VM functionality, (like run, set_reg, get_regs, etc.). Then, we can rely on a dyn Hypervisor to avoid duplicating functionality across different implementations.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -142,15 +150,25 @@ impl UninitializedSandbox {
                 let path = Path::new(&binary_path)
                     .canonicalize()
                     .map_err(|e| new_error!("GuestBinary not found: '{}': {}", binary_path, e))?;
-                GuestBinary::FilePath(
-                    path.into_os_string()
-                        .into_string()
-                        .map_err(|e| new_error!("Error converting OsString to String: {:?}", e))?,
-                )
+                let path = path
```
ludfjig: Can we combine the `into_*` calls with the previous line?

src/hyperlight_host/src/hypervisor/crashdump.rs
dblnz: That is better indeed

src/hyperlight_host/src/hypervisor/crashdump.rs
dblnz: Well, at this point, as we're running a bare metal guest we don't have the Linux kernel to manage processes inside. So we have no way of filling them in.
The need to have them stems from how linux uses the `ELF` file format to represent the general purpose registers. The [`NT_PRSTATUS` note](https://elixir.bootlin.com/linux/v6.12.1/source/include/linux/elfcore.h#L48) is used to describe the registers, but beside the registers there are these process specific information.
If you have ideas about what values to use to fill them in, I am open to feedback 😄

dblnz: Well, for the PID I am not sure, but for the page size, that is used as the alignment for the ELF segments, so I don't think it is an issue. I can add some constants to make it look better

docs/debugging-hyperlight.md
```diff
@@ -37,11 +37,74 @@ Alternatively, this can be enabled when running a test from the command line:
 cargo test --package hyperlight-host --test integration_test --features print_debug -- static_stack_allocate --exact --show-output
 ```
 
-## Dumping the memory configuration, virtual processor register state and memory contents on a crash or unexpected VM Exit
-
-To dump the details of the memory configuration, the virtual processors register state and the contents of the VM memory set the feature `crashdump` and run a debug build. This will result in a dump file being created in the temporary directory. The name and location of the dump file will be printed to the console and logged as an error message.
+## Dumping the guest state to an ELF core dump
+
+To dump the state of the vCPU (general purpose registers, registers) to an `ELF` core dump file set the feature `crashdump` and run a debug build. This will result in a dump file being created in the temporary directory.
```
ludfjig: btw what happens if you run in release mode?

ludfjig (approved):

docs/debugging-hyperlight.md
dblnz: For now, the `crashdump` feature is gated by the debug build.
There are two scenarios here:
1. Run release mode with `crashdump` which will only make the guest crash and the guest return an error that the guest panicked (no core dump)
2. After I remove the `debug` only option for `crashdump`, the `core dump` file is correctly generated in the release mode, but the symbols are scarce (it just shows the function names). I haven't checked, but for release builds the debug symbols might be stripped

simongdavies: I think that  we should add support for release builds, and maybe document the changes needed to the release profile when building the guest to support full symbols (or perhaps even add a profile that does that as an example) The other option (and maybe preferable) would be to set the option to create full debug info but to use split-debug-info on our release builds, we would still need to test this and document it

simongdavies: Unless I am mistaken this only happens if there is an unexpected VM exit or unhandled exception in the guest not just by turning on the feature?

docs/debugging-hyperlight.md
```diff
@@ -37,11 +37,74 @@ Alternatively, this can be enabled when running a test from the command line:
...
+The name and location of the dump file will be printed to the console and logged as an error message.
+
+### Inspecting the core dump
+
+After the core dump has been created, to inspect the state of the guest, load the core dump file using `gdb` or `lldb`.
+A `gdb` version later than `15.0` and `lldb` version later than `17` have been used to test this feature.
```
simongdavies:
```suggestion
**NOTE: This feature has been tested with version `15.0` of `gdb` and version `17` of `lldb` , earlier versions may not work, it is recommended to use these versions or later.**
```

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -1,44 +1,280 @@
...
+            .collect();
 
-    // write the raw memory dump for each memory region
-    for region in hv.get_memory_regions() {
-        if region.host_region.start == 0 || region.host_region.is_empty() {
-            continue;
+        let filename = ctx
+            .filename
+            .as_ref()
+            .map_or_else(|| "<unknown>".to_string(), |s| s.to_string());
```
simongdavies:
```suggestion
            .map_or("<unknown>".to_string(), |s| s.to_string());
```

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -1,44 +1,280 @@
...
+
+        let cmd = ctx
+            .binary
+            .as_ref()
+            .map_or_else(|| "<unknown>".to_string(), |s| s.to_string());
```
simongdavies:
```suggestion
            .map_or("<unknown>".to_string(), |s| s.to_string());
```

simongdavies (dismissed):
A couple of minor comments/suggestions, the main thing is that I think we should enable this for release builds and show how to get debug info in a release build as above

docs/debugging-hyperlight.md
dblnz: Yes, I'll specify it

dblnz: I made the changes

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
```diff
@@ -0,0 +1,1082 @@
+/*
```
ludfjig: This file should probably not be included as it's been deleted.

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -201,3 +201,222 @@ involved in the gdb debugging of a Hyperlight guest running inside a **KVM** or
      └─┘                       │          |                                           |                                 |      │
       |                        └───────────────────────────────────────────────────────────────────────────────────────────────┘
 ```
+
+## Dumping the guest state to an ELF core dump when an unhandled crash occurs
+
+When a guest crashes because of an unknown VmExit or unhandled exception, the vCPU state is dumped to an `ELF` core dump file.
+This can be used to inspect the state of the guest at the time of the crash.
+
+To make Hyperlight dump the state of the vCPU (general purpose registers, registers) to an `ELF` core dump file, set the feature `crashdump` and run a debug build.
```
ludfjig: Am I correct that is requires both the cargo feature, and also the runtime configuration to be set to true? If yes, then maybe update this

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -236,6 +262,8 @@ impl Default for SandboxConfiguration {
             Self::INTERRUPT_VCPU_SIGRTMIN_OFFSET,
             #[cfg(gdb)]
             None,
+            #[cfg(crashdump)]
+            true,
```
ludfjig, line 266: Do we want `true` or `false` to be the default when the feature is enabled? I'm leaning towards `false` but unsure

ludfjig (approved):
Nice, looks great!

src/hyperlight_host/src/hypervisor/hypervisor_handler.rs
dblnz: Wow, yeah, rebase artifact.
I'll remove it.

src/hyperlight_host/src/sandbox/config.rs
dblnz: We want it to be true.
This feature only impacts a guest at runtime on the crash path (after the guest did something bad), so it doesn't do anything on the happy path.
However, if the user wants to avoid having guests create crash dumps, the host application can dynamically change the configuration for new sandboxes to `false`.

docs/how-to-debug-a-hyperlight-guest.md
dblnz: No, by default the crashdump generation is enabled when the `--features crashdump` is provided.

dblnz: I added a note that explains how to disable the core dump creation for specific sandboxes.

docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -201,3 +201,229 @@ involved in the gdb debugging of a Hyperlight guest running inside a **KVM** or
...
```
simongdavies: Is it true that it needs to be a debug build? I thought that with the changes to enable runtime configuration that we would make this work with release builds as well?

dblnz: No, I removed that constraint. I'll change the docs, thanks for noticing this.

simongdavies (approved):

PR #419, opened 2025-04-16
Clean up justfile and testing

ludfjig:
First commit:
- Renames the just recipe for testing to `test` (instead of `test-rust`) and for building to `build` (instead of `build-rust`)
- Reorders the justfile after category: building, testing, linting, rust-examples, artifacts, benchmarks, fuzzing, flatbuffrers
- Modifies several crates' cargo.toml to set `doctest=false`, so that `cargo test` doesn't show noise (running 0 tests)
- Removes the need to set environment variables for certain rust examples/tests

Second commit:
- fixes a bug where the gdb test failed if the terminal size was too small (due to pagination).

### DIFF
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -85,7 +85,7 @@ jobs:
         run: just build-and-move-c-guests
 
       - name: Build
-        run: just build-rust ${{ matrix.config }}
+        run: just build ${{ matrix.config }}
 
       - name: Verify MSRV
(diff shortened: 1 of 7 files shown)


danbugs (approved):
Do we have to change any documentation that maybe used `build-rust` or alike? I saw the README uses `just build` already. Was it wrong or a different build?

ludfjig:
> Do we have to change any documentation that maybe used `build-rust` or alike? I saw the README uses `just build` already. Was it wrong or a different build?

We had a `build` recipe, but it was just forwarding to build-rust

PR #421, opened 2025-04-16
fix: Add -fPIC option for building C guests.

prydt:
I ran into an issue running `just guests` on Fedora 41, and adding the -fPIC option in the C flags fixes this.

Please let me know if I need to make any changes to this PR. Thank you.


ludfjig (approved):
Thanks for this!

PR #422, opened 2025-04-17
Fix confusing log message when checking present hypervisor

### LINKED ISSUE #379: Incorrect log level for failing to create driver when multiple cargo features are enabled.
When multiple hypervisors are enabled via features (example both mshv and kvm), and the host system has kvm, an log message of level ERROR will be reported because `HypervLinuxDriver::new(` will be called, which fails, but later `KVMDriver::new` will be called, which succeeds. 

We should not log an error when this happens.

ludfjig:
When checking if a specific hypervisor is present using for example `kvm::is_hypervisor_present`, and the hypervisor is not present, a log message is printed with the word 'Error' in it, which was confusing since it's not actually an error.

The second commit does a slight cleanup by switching to Mshv::new() to avoid having to manually close the fd with raw libc calls.

Closes #379

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
@@ -278,8 +278,8 @@ pub(crate) fn is_hypervisor_present() -> bool {
             } // must explicitly close fd to avoid a leak
             true
         }
-        Err(e) => {
-            log::info!("Error creating MSHV object: {:?}", e);
+        Err(_) => {
+            log::info!("MSHV is not available on this system");
             false
         }
     }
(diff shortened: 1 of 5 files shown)


src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -271,15 +271,10 @@ mod debug {
...
-        Ok(fd) => {
-            unsafe {
-                libc::close(fd);
-            } // must explicitly close fd to avoid a leak
-            true
-        }
-        Err(e) => {
-            log::info!("Error creating MSHV object: {:?}", e);
+    match Mshv::new() {
+        Ok(_) => true,
```
danbugs, line 275: Why don't you need to close the fd anymore? I added this a while back because we were getting an "out of resources" error due to leaking this fd in long running intensive tests.

ludfjig: Because Mshv::new() internally takes ownership of the filedescriptor, and it will be dropped when Mshv is dropped. The reason we used to leak is because we used `Mshv::open_with_cloexec` with returns a raw owned fd, and we didn't use to drop it. 

 thought this way of doing it is more clean

danbugs (approved):

PR #435, opened 2025-04-25
Remove tracing from hot functions

ludfjig:
When using tracing for profiling purposes, certain functions are called over and over, adding overhead specifically when tracing is turned on, rendering the traces useless. This PR removes some tracing for hot functions.
- Adds an example of how to use tracy
- Renames all tracing-related examples


src/hyperlight_host/examples/tracing-tracy/README.md
```diff
@@ -0,0 +1,9 @@
+This is an example of using the tracing-tracy tracing-subscriber. When ran, it will generate traces that can be viewed in the tracy profiler.
+
+You can run it with:
+
+```console
+TRACY_NO_EXIT=1 RUST_LOG=trace cargo run --package hyperlight-host --example tracing-tracy --profile release-with-debug
+```
+
+and then the client should show up in the profiler GUI.
```
danbugs: Would be cool to see some images for this in the README :)

src/hyperlight_host/examples/tracing-tracy/main.rs
```diff
@@ -0,0 +1,60 @@
+/*
```
danbugs, line 1: Is it worth adding this example to CI?

danbugs (approved):
Long time coming. Great stuff. We should probably at some point establish guidelines to what we want to trace or not.

src/hyperlight_host/examples/tracing-tracy/README.md
ludfjig: fixed

src/hyperlight_host/examples/tracing-tracy/main.rs
ludfjig: Hmm I don't think so, we don't run all examples currently. Maybe something to consider in the future.

danbugs (approved):

PR #444, opened 2025-04-29
Optimize serialization of function calls

ludfjig:
Removes a vec allocation + copy by avoiding calling `to_vec` on the serialized flatbuffer slice. A downside is that the implementation now leaks an internal FlatBufferBuilder, which is unavoidable.  

`guest_call` benchmark shows a -6% execution speedup on my local machine.


marosset (dismissed):

src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
```diff
@@ -288,9 +223,40 @@ impl TryFrom<FunctionCall> for Vec<u8> {
             },
         );
         builder.finish_size_prefixed(function_call, None);
-        let res = builder.finished_data().to_vec();
```
ludfjig, line 291: This `to_vec` is the clone I was able to avoid, which is the most important part of this PR

simongdavies (dismissed):

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -617,18 +616,20 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
-        })?;
-
+    pub(crate) fn write_guest_function_call(&mut self, function_call: &FunctionCall) -> Result<()> {
+        match function_call.function_call_type() {
+            FunctionCallType::Host => {
+                log_then_return!(
+                    "Tried to serialize a host function call as a guest function call"
+                );
+            }
+            FunctionCallType::Guest => {}
```
dblnz, line 626: Nit: maybe move the logic below on the `Guest` match arm or use an `if let FunctionCallType::Host`.

src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
```diff
@@ -72,91 +71,29 @@ impl FunctionCall {
...
-    match guest_function_call_fb.function_call_type() {
-        FbFunctionCallType::guest => Ok(()),
-        other => {
-            bail!("Invalid function call type: {:?}", other);
-        }
-    }
-}
 
-#[cfg_attr(feature = "tracing", instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace"))]
-pub fn validate_host_function_call_buffer(function_call_buffer: &[u8]) -> Result<()> {
```
dblnz, line 90: Maybe I missed it, but where was this function called?
If it wasn't used at all, I am curious how the compiler hasn't warned us.

dblnz (approved):
LGTM. I left some small comments

ludfjig:
closed in favor of #778

PR #453, opened 2025-05-02
Removing HostFunctionDefinitions region

danbugs:
We only used the `HostFunctionDefinitions` region to validate that a host function exists before calling it. But, if the host function doesn't exist, we'll just exit anyway, so that utility is questionable. Also, with the removal of the region, because we no longer need to serialize the `host_function_*` structures, I updated our Flatbuffer schema.

> Note: Just like #451 , this PR is also part of the effort of breaking #297 into more digestible bits.


src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -133,22 +150,11 @@ fn register_host_function_helper(
             .get_host_funcs_mut()
             .insert(hfd.function_name.to_string(), func, None);
     }
-    self_
-        .get_host_func_details_mut()
-        .insert_host_function(hfd.clone());
+    self_.insert_host_function(hfd.clone());
     // Functions need to be sorted so that they are serialised in sorted order
-    // this is required in order for flatbuffers C implementation used in the Gues Library
+    // this is required in order for flatbuffers C implementation used in the Guest Library
```
ludfjig: I don't think is relevant anymore? Do we really need to sort?

docs/how-to-use-flatbuffers.md
```diff
@@ -1,5 +1,7 @@
 # How to use FlatBuffers
 
+> Note: the last generation of the flatbuffer code was with done with flatc version 25.2.10 (i.e., the last version as of May 1st, 2025).
```
ludfjig, line 3: nice

src/hyperlight_host/src/func/host_functions.rs
```diff
@@ -27,6 +28,32 @@ use crate::sandbox::{ExtraAllowedSyscall, UninitializedSandbox};
 use crate::HyperlightError::UnexpectedNoOfArguments;
 use crate::{log_then_return, new_error, Result};
 
+/// The definition of a function exposed from the host to the guest
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct HostFunctionDefinition {
```
ludfjig, line 33: What do we use this struct for?

ludfjig (approved):
lgtm with some questions

src/hyperlight_host/src/func/host_functions.rs
danbugs: When we register host functions, we save them as `HostFunctionDefinition`.

src/hyperlight_host/src/sandbox/host_funcs.rs
danbugs: Yeah, you're right. We don't even use flatcc anymore.

danbugs: Removed ✅

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -133,22 +140,7 @@ fn register_host_function_helper(
...
-        .get_host_func_details_mut()
-        .sort_host_functions_by_name();
-    let buffer: Vec<u8> = self_.get_host_func_details().try_into().map_err(|e| {
-        new_error!(
-            "Error serializing host function details to flatbuffer: {}",
-            e
-        )
-    })?;
-    mgr.write_buffer_host_function_details(&buffer)?;
+    self_.insert_host_function(hfd.clone());
```
ludfjig, line 143: nit: is there any way to avoid this clone?

ludfjig (approved):
LGTM! Feel free to ignore nit

src/hyperlight_host/src/sandbox/host_funcs.rs
danbugs: Don't see an obvious way to do it rn, but I think a lot of this code can be cleaned up when we add a `SandboxBuilder` and add registering host functions onto it, so I think it's okay to defer this optimization.

PR #456, opened 2025-05-02
Gdb debug improvements

### LINKED ISSUE #380: Improve the way vCPU stop reason is implemented to reduce code duplication
This is a follow up issue for the #327 PR.

1. Check the comments for more context.
The way `get_stop_reason` method is implemented introduces code duplication between kvm and mshv, this can be improved.

2. it is worth investigating whether `mshv` has a way of checking which exception was intercepted.
[Link 1](https://docs.rs/mshv-bindings/latest/mshv_bindings/struct.hv_message.html#method.to_exception_info) and [Link 2](https://docs.rs/mshv-bindings/latest/mshv_bindings/struct.hv_x64_exception_intercept_message.html) are relevant to this.

3. Check whether `gdb` does not complain when an unknown vCPU exit reason is reported as SwBp

dblnz:
## Purpose
I opened up this PR to address some leftover comments from a previous PR.
This closes #380 .

## Description
These changes do the following
- move the x86 vCPU specific constants to `gdb/arch.rs` to avoid duplicating them
- add a function that determines the vCPU stop reason based on a series of parameters (DR6 register, exception vector)
- modify kvm `get_stop_reason` method to use the hypervisor independent function described above 
- retrieve additional information from the mshv `hv_message` returned after a `run` call.
This information serves the `get_stop_reason` with information on the exception that was triggered.
- modify the mshv debugging code to use the hypervisor independent function which determines the stop reason of an x86 vCPU.

Note: I tested the scenario specified in [this comment](https://github.com/hyperlight-dev/hyperlight/pull/327#discussion_r2016765045) and it does not present any issue

### DIFF
--- a/src/hyperlight_host/src/hypervisor/gdb/arch.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/arch.rs
@@ -16,6 +16,17 @@ limitations under the License.
 
 //! This file contains architecture specific code for the x86_64
 
+use std::collections::HashMap;
+
+use super::VcpuStopReason;
+
+// Described in Table 6-1. Exceptions and Interrupts at Page 6-13 Vol. 1
+// of Intel 64 and IA-32 Architectures Software Developer's Manual
+/// Exception id for #DB
+const DB_EX_ID: u32 = 1;
+/// Exception id for #BP - triggered by the INT3 instruction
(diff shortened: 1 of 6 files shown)


ludfjig (approved):
Looks good to me

PR #457, opened 2025-05-02
Removed host error region

danbugs:
This region was unused. Ever since we ported our code from C# to Rust, if there's an error calling a host function, the host will just error out instead of returning to the guest, which is probably more sensible.

> Note: Just like #453 , this PR is also part of the effort of breaking #297 into more digestible bits.


ludfjig (approved):
LGTM. Can we remove our dependency on serde now? I don't know what we use serde for

ludfjig (approved):

PR #464, opened 2025-05-07
Unify HostFunctionXX traits into a single HostFunction

jprendes:
This PR removes the HostFunctionXX traits, and replaces them with a single HostFunction trait.

### DIFF
--- a/src/hyperlight_host/src/func/host_functions.rs
+++ b/src/hyperlight_host/src/func/host_functions.rs
@@ -20,7 +20,6 @@ use std::sync::{Arc, Mutex};
 use hyperlight_common::flatbuffer_wrappers::function_types::{
     ParameterType, ParameterValue, ReturnType,
 };
-use paste::paste;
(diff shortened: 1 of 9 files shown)


src/hyperlight_host/src/func/host_functions.rs
```diff
@@ -20,7 +20,6 @@ use std::sync::{Arc, Mutex};
 use hyperlight_common::flatbuffer_wrappers::function_types::{
     ParameterType, ParameterValue, ReturnType,
 };
-use paste::paste;
```
danbugs, line 23: Actually–can we remove this dependency now? That would allow us to close #341 !

jprendes: It's still being used in layout.rs, I'll remove it as a follow up if that's ok with you

dblnz (approved):
Great work @jprendes

PR #468, opened 2025-05-07
Improve the ergonomics of registering host functions

jprendes:
This PR adds the `register` and `register_with_extra_allowed_syscalls` methods to the `UninitializedSandbox` to register host methods.
e.g.,
```rust
uninitialized_sandbox.register("add", |a: i32, b: i32| {
    Ok(a + b)
})?;
```

This method accepts either:
* `FnMut(..) -> Result<R>`
* `Arc<Mutex<FnMut(..) -> Result<R>>>`
* `&Arc<Mutex<FnMut(..) -> Result<R>>>`

This is controlled by the IntoHostFunction trait.

This is not a breaking change. It's an additive change, and the previous registration mechanism continues to works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,16 +47,11 @@ fn main() -> hyperlight_host::Result<()> {
         None, // default host print function
     )?;
 
-    // Register a host function
-    fn sleep_5_secs() -> hyperlight_host::Result<()> {
+    // Registering a host function makes it available to be called by the guest
+    uninitialized_sandbox.register("Sleep5Secs", || {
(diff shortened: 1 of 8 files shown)


danbugs (commented):
LGTM. Wonder if we should just remove the previous registration mechanism...

jprendes:
> Wonder if we should just remove the previous registration mechanism...

The new mechanism is a wrapper around the old one. I would need to think how we can stop exposing the old mechanism without violating some "X is more private than Y" rule.

ludfjig (approved):

PR #469, opened 2025-05-08
Vm trait for reducing code duplication

### LINKED ISSUE #312: Unify register representations across all Hypervisor implementations

### LINKED ISSUE #465: Rethink driver API
With our current driver API for kvm/mshv/whp, certain functionality needs to be reimplemented 3 times over. We should probably change the existing Hypervisor trait to be "dumber" and only include things like `set_regs`, `get_regs`, `run`, and not include `dispatch` nor `initialize`, and then we can implement `dispatch` and `initialize` only once

### LINKED ISSUE #503: Avoid reopening /dev/kvm or /dev/mshv for every new sandbox
We should just reuse the same one to avoid unnecessary file operations

ludfjig:
This PR introduces a minimal `Vm` trait. It's implemented by `KvmVm`, `MshvVm` and `WhpVm`. The goal is to reduce code duplication. The existing `Hypervisor` trait is renamed to `HyperlightVm`, and is now only implemented by `HyperlightSandbox` (and Inprocess, but not for long since we're getting rid of inproces mode). In addition, the `Vm` trait also reduces a bunch of code duplication related to GDB debugging.

About half of the lines of this PR is introducing common register structs used for all Vms. They Are CommonRegisters, CommonStandardRegisters, and CommonFpu, with tests. The rest should be pretty straight forward, and is mostly just plumbing.

This PR will be squashed, so it is not worth it to review commit by commit.

Closes #465 
Closes #312 
Closes #503


dblnz:
Nice work, this is a very good idea.

src/hyperlight_host/src/hypervisor/inprocess.rs
dblnz, line 1: This should be removed, as we removed the inprocess mode, right?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -604,11 +604,8 @@ impl SandboxMemoryLayout {
         }
 
         // guard page
-        let stack_offset = builder.push_page_aligned(
-            PAGE_SIZE_USIZE,
-            MemoryRegionFlags::READ | MemoryRegionFlags::STACK_GUARD,
-            GuardPage,
-        );
+        let stack_offset =
+            builder.push_page_aligned(PAGE_SIZE_USIZE, MemoryRegionFlags::READ, GuardPage);
```
dblnz, line 608: Why remove the `STACK_GUARD` flag?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,750 @@
...
+                data = &data[write_len..];
+                gva += write_len as u64;
+            }
+
+            Ok(())
+        }
+    }
+}
+
+/// A Hypervisor driver for KVM on Linux
```
dblnz: Remove comment

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -64,509 +52,113 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
+                exception: debug_exit.exception,
+            }),
             Err(e) => match e.errno() {
                 // In case of the gdb feature, the timeout is not enabled, this
                 // exit is because of a signal sent from the gdb thread to the
                 // hypervisor thread to cancel execution
-                #[cfg(gdb)]
-                libc::EINTR => HyperlightExit::Debug(VcpuStopReason::Interrupt),
+                // #[cfg(gdb)]
+                // libc::EINTR => Ok(HyperlightExit::Debug(VcpuStopReason::Interrupt)),
```
dblnz: This allows a user to stop a running vcpu from the debugger using pause or Ctrl-C in CLI.
Without this, in `event_loop.rs` we cannot confirm to the debugger that the host stopped due to a signal.
I am curious how this compiled without removing `VcpuStopReason::Interrupt` from the `event_loop.rs`

dblnz (changes requested):
Awesome work @ludfjig! This really removes a lot of duplicated code.
I think this should have been done when the guest debugging feature was added. But I wanted to keep the debugging logic separate from the normal flow.

src/hyperlight_host/src/mem/layout.rs
ludfjig: It was not a real permissions flag like READ or WRITE or EXECUTE, it was only an indicator. But that makes no sense because we already have another indicator for whether a region is a guard page. I don't think flag serves any purpose, and only messes things up when coverting a memory region to a vm-specific region. For example it makes us unable to map a region with only READ (which means readonly) to for example KVM_MEM_READONLY, without special handling of STACK_GUARD, which I am not a big fan of.

I admit I should have made some separate PRs for many of my changes in this PR

src/hyperlight_host/src/hypervisor/inprocess.rs
ludfjig: yes good catch, was a bad merge from main. Will clean up

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -2,21 +2,21 @@ use std::io::Write;
 
 use tempfile::NamedTempFile;
 
-use super::Hypervisor;
+use super::hyperlight_vm::HyperlightSandbox;
+use crate::hypervisor::HyperlightVm;
 use crate::{new_error, Result};
 
 /// Dump registers + memory regions + raw memory to a tempfile
-#[cfg(crashdump)]
```
ludfjig, line 9: if you're wondering why this was removed, it's because this entire crashdump module is already gated on this

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: ah got it. I added it back

PR #474, opened 2025-05-08
Guest panic context refactor

danbugs:
This PR removes the guest panic context memory region and refactors the functionality to use a new OutBAction that prints to stderr. 

I recommend you review this commit-by-commit.

> Note: Just like #462, this PR is also part of the effort of breaking #297 into more digestible bits.


src/hyperlight_guest/src/host_function_call.rs
```diff
@@ -75,24 +70,40 @@ pub fn call_host_function(
...
+    outb(OutBAction::CallFunction as u16, &[0]);
 
     Ok(())
 }
 
-pub fn outb(port: u16, value: u8) {
+pub fn outb(port: u16, data: &[u8]) {
     unsafe {
         match RUNNING_MODE {
             RunMode::Hypervisor => {
```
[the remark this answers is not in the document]
danbugs: Added comment ✅

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -117,7 +99,7 @@ fn handle_outb_impl(
     mem_mgr: &mut MemMgrWrapper<HostSharedMemory>,
     host_funcs: Arc<Mutex<HostFuncsWrapper>>,
     port: u16,
-    byte: u64,
+    data: Vec<u8>,
```
ludfjig, line 102: I'm not sure I understand why this changed. Do I understand correctly that this vec will only ever have length 1?

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -136,23 +118,31 @@ fn handle_outb_impl(
...
+                        None => "See stderr for panic context".into(),
+                    };
+
+                    Err(HyperlightError::GuestAborted(byte as u8, message))
+                }
             }
         }
+        OutBAction::DebugPrint => {
+            let s = String::from_utf8_lossy(&data);
+            eprint!("{}", s);
```
simongdavies, line 143: should we not append a new line at the end of the data, it seems like this is going to potentially concatenate a lot of data to stderr? Given you are padding with 0x7f then we could replace that with a new line although you would have to send an extra 4 bytes in the case that the string length was divisible by 4.

Also I think it might be nice to allow the host to override or alternatively maybe we should just use error!() here , which would mean gathering the string into a buffer until you get the final 0x7f?

src/hyperlight_host/src/sandbox/outb.rs
simongdavies: It can be up to 4 bytes now

danbugs: I changed the outb instruction to now send 4 bytes instead of 1. This is useful for less exits w/ `debug_print` and to also be able to send more information w/ exceptions.

src/hyperlight_host/src/sandbox/outb.rs
danbugs: We send 4 bytes at a time. If we added a new line, we'd get print output that looks like this:
```
hell
o, w
orld
```

I chose 0x7f for padding because it is a non-printable character. A new-line is 0x0A, which is the same value as the `InvalidTSS` exception. I considered using a macro from the log crate, but it felt like a bit of overengineering for just `debug_print`. Do we want to allow people to customize an `OutBAction`?

ludfjig (approved):

marosset (approved):
LGTM

PR #475, opened 2025-05-08
Remove generics from SupportedParameterType and SupportedReturnType traits

jprendes:
As far as I can tell those generics are not needed
They also mean that for each generic value it's actually a different trait, making it harder to use them as generics themselves.


ludfjig (approved):
Nice, looks reasonable to me :D

PR #476, opened 2025-05-08
Improve ergonomics of SupportedParameterType and SupportedReturnType

jprendes:
This PR:
* reduces code duplication using macros
* replaces `get_hyperlight_type` with a `TYPE` associated const
* replaces `get_hyperlight_value(&self)` with `into_value(self)`, for a better name and avoiding a copy.
* replaces `get_inner(v: XxxValue)` with `from_value(v: XxxValue)`, for a better name.


ludfjig (approved):
clean

PR #479, opened 2025-05-09
Remove unused FunctionDefinition and redundant FunctionsMap

jprendes:
The `FunctionDefinition` struct is unused and can be simple removed.
The `FunctionsMap` struct is just a thin wrapper around a `HashMap` that re-exposes the same API, so it can also be removed.


jprendes:
@danbugs PTAL :-)

danbugs (approved):
Awesome work, @jprendes —I'm loving this cleanup 😁

PR #485, opened 2025-05-12
Remove support from hyperlight_host for PE formatted guests

### LINKED ISSUE #470: Remove support for PE guests from hyperlight
There is no real reason for us to support both ELF and PE binaries in hyperlight, building and supporting PE based guests just adds complexity to hyperlight and its build processes.

We should remove this support and all associated tests and build actions/jobs.

simongdavies:
Remove support for PE formatted guests from hyperlight_host.

Follow up PRs will remove in-process execution mode, fix up build and test and refactor where necessary.

Fixes #470


ludfjig (approved):
lgtm

PR #486, opened 2025-05-13
Changes to remove build steps for PE formatted guests.

simongdavies:
Now that we have removed support for PE formatted guests (#470), we no longer need to build/archive these guests.

This completes the work for issue #470


dblnz (approved):
LGTM

ludfjig:
@simongdavies there is still some PE info in the cargo/config.toml for each guest that we can probably remove

PR #480, opened 2025-05-09
Refactor HostFuncsWrapper

jprendes:
This PR is a continuation of #479. I'll keep this PR in draft until #479 is merged.
The new changes in this PR (excluding #479) are here:
https://github.com/jprendes/hyperlight/compare/host_function_2...jprendes:hyperlight:host_function_3

PR #487, opened 2025-05-13
Remove `host_print_writer` from the arguments to `UninitializedSandbox::new`

jprendes:
With #480 merged, registering a host function with a name that is already in use, just replaces the previous registration.
This means that we can pre-emptively register the default `HostPrint` method, and replace it afterwards.

This PR removes `host_print_writer` from the arguments to `UninitializedSandbox::new`.
The default host print is pre-emptively registered, and a new `HostPrint` can be registered like any other host function.
Alternatively, the method `register_print` can be used, which uses the correct name and (compile-time-)checks the function signature.

This partly addresses the following comment (i.e., it still doesn't implement a builder pattern, but does allow adding extra syscalls to the writer method)
https://github.com/hyperlight-dev/hyperlight/blob/a862d4ad73c72ab89e4504f410ef6541d87472b9/src/hyperlight_host/src/sandbox/uninitialized.rs#L195-L197

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,6 @@ fn main() -> hyperlight_host::Result<()> {
         hyperlight_host::GuestBinary::FilePath(hyperlight_testing::simple_guest_as_string().unwrap()),
         None, // default configuration
         None, // default run options
-        None, // default host print function
     )?;
 
     // Registering a host function makes it available to be called by the guest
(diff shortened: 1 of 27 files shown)


jprendes:
@danbugs @ludfjig PTAL :-)

ludfjig (dismissed):
LGTM. Could you update the example in the root README as well?

danbugs (approved):
LGTM

jprendes:
@danbugs @ludfjig PTAL :-)
Basically rebased and fixed doctest.

ludfjig (approved):

Issue #477, opened 2025-05-08
Improve the correctness of the OutB API


danbugs:
In #474 , we modified our `out` mechanism to transmit 4 bytes instead of 1. With this, we should rename our `OutB` APIs to reflect that we're no longer transmitting just a `B`.

Also, our OutB mechanism now splits a &[u8] in chunks of 4–when using `debug_print`, to ensure we don't split a UTF8 code point we could leverage something like `msg.bytes().align_to::<u32>()`.

PR #489, opened 2025-05-13
Bring back panic info to abort

danbugs:
In #474, I removed the guest panic context region. With that, aborting was split into two steps:
    (1) printing panic info w/ debug_print, and
    (2) exiting w/ abort and codes.
    
This PR addresses feedback I got in #474 to join the operations by buffering string output when aborting.
    
This PR also cleans up our manual chunking logic w/ outb and leverages out8, out16, and out32 to minimize VM exits. Abort detects it should finishing buffering by a terminator code (0xFF).
    
Edited tests to check for panic message in aborts as they were prior to #474 merged.

This PR can be reviewed commit by commit. I also made some other minor changes like:
- removing some legacy unused structs, and
-  adding a feature flag to logging some of our build metadata, which uses a git2 dependency that is not available on every target.

### DIFF
--- a/src/hyperlight_guest/src/entrypoint.rs
+++ b/src/hyperlight_guest/src/entrypoint.rs
@@ -25,7 +25,7 @@ use spin::Once;
 use crate::gdt::load_gdt;
 use crate::guest_function_call::dispatch_function;
 use crate::guest_logger::init_logger;
-use crate::host_function_call::{debug_print, outb};
+use crate::host_function_call::outb;
 use crate::idtr::load_idt;
 use crate::{
     __security_cookie, HEAP_ALLOCATOR, MIN_STACK_ADDRESS, OS_PAGE_SIZE, OUTB_PTR,
@@ -43,11 +43,12 @@ pub fn halt() {
 
 #[no_mangle]
 pub extern "C" fn abort() -> ! {
-    abort_with_code(&[0])
+    abort_with_code(&[0, 0xFF])
 }
 
 pub fn abort_with_code(code: &[u8]) -> ! {
     outb(OutBAction::Abort as u16, code);
+    outb(OutBAction::Abort as u16, &[0xFF]); // send abort terminator (if not included in code)
     unreachable!()
 }
 
@@ -56,13 +57,19 @@ pub fn abort_with_code(code: &[u8]) -> ! {
 /// # Safety
 /// This function is unsafe because it dereferences a raw pointer.
 pub unsafe fn abort_with_code_and_message(code: &[u8], message_ptr: *const c_char) -> ! {
-    debug_print(
-        CStr::from_ptr(message_ptr)
-            .to_str()
-            .expect("Invalid UTF-8 string"),
-    );
-
+    // Step 1: Send abort code (typically 1 byte, but `code` allows flexibility)
     outb(OutBAction::Abort as u16, code);
+
+    // Step 2: Convert the C string to bytes
+    let message_bytes = CStr::from_ptr(message_ptr).to_bytes(); // excludes null terminator
+
+    // Step 3: Send the message itself in chunks
+    outb(OutBAction::Abort as u16, message_bytes);
+
+    // Step 4: Send abort terminator to signal completion (e.g., 0xFF)
+    outb(OutBAction::Abort as u16, &[0xFF]);
+
+    // This function never returns
     unreachable!()
 }
 
--- a/src/hyperlight_guest/src/host_function_call.rs
+++ b/src/hyperlight_guest/src/host_function_call.rs
@@ -17,7 +17,7 @@ limitations under the License.
 use alloc::format;
 use alloc::string::ToString;
 use alloc::vec::Vec;
-use core::arch::global_asm;
+use core::arch;
 
 use hyperlight_common::flatbuffer_wrappers::function_call::{FunctionCall, FunctionCallType};
 use hyperlight_common::flatbuffer_wrappers::function_types::{
@@ -79,21 +79,16 @@ pub fn outb(port: u16, data: &[u8]) {
     unsafe {
         match RUNNING_MODE {
             RunMode::Hypervisor => {
-                for chunk in data.chunks(4) {
-                    // Process the data in chunks of 4 bytes. If a chunk has fewer than 4 bytes,
-                    // pad it with 0x7F to ensure it can be converted into a 4-byte array.
-                    // The choice of 0x7F as the padding value is arbitrary and does not carry
-                    // any special meaning; it simply ensures consistent chunk size.
-                    let val = match chunk {
-                        [a, b, c, d] => u32::from_le_bytes([*a, *b, *c, *d]),
-                        [a, b, c] => u32::from_le_bytes([*a, *b, *c, 0x7F]),
-                        [a, b] => u32::from_le_bytes([*a, *b, 0x7F, 0x7F]),
-                        [a] => u32::from_le_bytes([*a, 0x7F, 0x7F, 0x7F]),
-                        [] => break,
-                        _ => unreachable!(),
-                    };
-
-                    hloutd(val, port);
+                let mut i = 0;
+                while i < data.len() {
+                    let remaining = data.len() - i;
+                    let chunk_len = remaining.min(3);
+                    let mut chunk = [0u8; 4];
+                    chunk[0] = chunk_len as u8;
+                    chunk[1..1 + chunk_len].copy_from_slice(&data[i..i + chunk_len]);
+                    let val = u32::from_le_bytes(chunk);
+                    out32(port, val);
+                    i += chunk_len;
                 }
             }
             RunMode::InProcessLinux | RunMode::InProcessWindows => {
@@ -119,11 +114,25 @@ pub fn outb(port: u16, data: &[u8]) {
     }
 }
 
-extern "win64" {
-    fn hloutd(value: u32, port: u16);
+pub(crate) unsafe fn out32(port: u16, val: u32) {
+    arch::asm!("out dx, eax", in("dx") port, in("eax") val, options(preserves_flags, nomem, nostack));
 }
 
-pub fn print_output_as_guest_function(function_call: &FunctionCall) -> Result<Vec<u8>> {
+/// Prints a message using `OutBAction::DebugPrint`. It transmits bytes of a message
+/// through several VMExists and, with such, it is slower than
+/// `print_output_with_host_print`.
+///
+/// This function should be used in debug mode only. This function does not
+/// require memory to be setup to be used.
+pub fn debug_print(msg: &str) {
+    outb(OutBAction::DebugPrint as u16, msg.as_bytes());
+}
+
+/// Print a message using the host's print function.
+///
+/// This function requires memory to be setup to be used. In particular, the
+/// existence of the input and output memory regions.
+pub fn print_output_with_host_print(function_call: &FunctionCall) -> Result<Vec<u8>> {
     if let ParameterValue::String(message) = function_call.parameters.clone().unwrap()[0].clone() {
         call_host_function(
             "HostPrint",
@@ -135,20 +144,7 @@ pub fn print_output_as_guest_function(function_call: &FunctionCall) -> Result<Ve
     } else {
         Err(HyperlightGuestError::new(
             ErrorCode::GuestError,
-            "Wrong Parameters passed to print_output_as_guest_function".to_string(),
+            "Wrong Parameters passed to print_output_with_host_print".to_string(),
         ))
     }
 }
-
-pub fn debug_print(msg: &str) {
-    outb(OutBAction::DebugPrint as u16, msg.as_bytes());
-}
-
-global_asm!(
-    ".global hloutd
-     hloutd:
-        mov eax, ecx
-        mov dx, dx
-        out dx, eax
-        ret"
-);
--- a/src/hyperlight_guest/src/interrupt_handlers.rs
+++ b/src/hyperlight_guest/src/interrupt_handlers.rs
@@ -31,10 +31,9 @@ pub extern "sysv64" fn hl_exception_handler(
 ) {
     let exception = Exception::try_from(exception_number as u8).expect("Invalid exception number");
     let msg = format!(
-        "EXCEPTION: {:#?}\n\
-            Page Fault Address: {:#x}\n\
+        "Page Fault Address: {:#x}\n\
             Stack Pointer: {:#x}",
-        exception, page_fault_address, stack_pointer
+        page_fault_address, stack_pointer
     );
 
     unsafe {
--- a/src/hyperlight_guest/src/lib.rs
+++ b/src/hyperlight_guest/src/lib.rs
@@ -17,16 +17,13 @@ limitations under the License.
 #![no_std]
 // Deps
 use alloc::string::ToString;
-use core::hint::unreachable_unchecked;
 
 use buddy_system_allocator::LockedHeap;
 use guest_function_register::GuestFunctionRegister;
-use host_function_call::debug_print;
 use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
 use hyperlight_common::mem::{HyperlightPEB, RunMode};
-use hyperlight_common::outb::OutBAction;
 
-use crate::host_function_call::outb;
+use crate::entrypoint::abort_with_code_and_message;
 extern crate alloc;
 
 // Modules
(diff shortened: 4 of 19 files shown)


src/hyperlight_guest/src/entrypoint.rs
```diff
@@ -43,11 +43,12 @@ pub fn halt() {
 
 #[no_mangle]
 pub extern "C" fn abort() -> ! {
-    abort_with_code(&[0])
+    abort_with_code(&[0, 0xFF])
```
ludfjig, line 46: looks like the 0xFF is already added by `abort_with_code`? It would make sense to me to not require the caller to think about it

src/hyperlight_guest/src/entrypoint.rs
```diff
@@ -56,13 +57,19 @@ pub fn abort_with_code(code: &[u8]) -> ! {
...
-    );
-
+    // Step 1: Send abort code (typically 1 byte, but `code` allows flexibility)
     outb(OutBAction::Abort as u16, code);
+
+    // Step 2: Convert the C string to bytes
+    let message_bytes = CStr::from_ptr(message_ptr).to_bytes(); // excludes null terminator
+
+    // Step 3: Send the message itself in chunks
+    outb(OutBAction::Abort as u16, message_bytes);
```
ludfjig, line 67: what happens if bytes contains 0xFF? In addition, maybe we could have different OutbAction for the start, the data, and the end?

src/hyperlight_guest/src/entrypoint.rs
```diff
@@ -56,13 +57,19 @@ pub fn abort_with_code(code: &[u8]) -> ! {
 /// # Safety
 /// This function is unsafe because it dereferences a raw pointer.
 pub unsafe fn abort_with_code_and_message(code: &[u8], message_ptr: *const c_char) -> ! {
```
ludfjig, line 59: Do we ever use the fact that code is a slice? If not, it would make more sense to keep it an integer, because it's not entirely clear what it means to send both a code buffer and a message buffer

src/hyperlight_guest/src/lib.rs
```diff
@@ -73,9 +70,11 @@ pub(crate) static _fltused: i32 = 0;
 // to satisfy the clippy when cfg == test
 #[allow(dead_code)]
 fn panic(info: &core::panic::PanicInfo) -> ! {
-    debug_print(info.to_string().as_str());
-    outb(OutBAction::Abort as u16, &[ErrorCode::UnknownError as u8]);
-    unsafe { unreachable_unchecked() }
+    let msg = info.to_string();
+    let c_string = alloc::ffi::CString::new(msg)
```
ludfjig, line 74: Is there a reason we're using **C** string here over regular string?

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -93,6 +94,58 @@ pub(super) fn outb_log(mgr: &mut SandboxMemoryManager<HostSharedMemory>) -> Resu
...
+
+        for &b in data {
+            if b == ABORT_TERMINATOR {
+                let full = &buffer[..];
+
+                let guest_error_code = *full.first().unwrap_or(&0);
+                let guest_error = ErrorCode::from(guest_error_code as u64);
+
+                match guest_error {
+                    ErrorCode::StackOverflow => {
```
ludfjig: What do you think about making stackoverflow it's own abort?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -847,7 +837,19 @@ mod tests {
     // from their workstation to be successful without needed to know about test interdependencies
     // this test will be run explicitly as a part of the CI pipeline
     #[ignore]
+    #[cfg(feature = "build-metadata")]
```
simongdavies, line 800: Don't understand why this is behind the build metadata feature? its testing logging and tracing?

ludfjig (commented):
Looks reasonable but have some questions. Also I'm not sure thread_local is right for this. What happens when two vms are running on the same thread?

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: This is only used by tests that are gated behind the build-metadata feature, so, without gating this, we get errors about dead_code.

src/hyperlight_host/src/sandbox/outb.rs
danbugs: It's simpler that the other aborts, but I don't really see a reason to separate it. This sets precedence to custom handling of different error codes, which I think should be okay to do without creating a new OutBAction.

src/hyperlight_guest/src/entrypoint.rs
danbugs: That's right. We shouldn't require the caller to think about it. I can remove it from the `abort` fxn, but this is just a shortcut to not require 2 VM exits.

src/hyperlight_guest/src/entrypoint.rs
danbugs: 0xFF is outside the printable range, so a message should not contain that. I'm in favour of keeping it as one OutBAction to avoid the extra error handling we'd have to do to express the relationship between the separate actions in code...

src/hyperlight_guest/src/entrypoint.rs
danbugs: We do use it—e.g., our interrupt handler sends 2 codes: one error code + one exception code.

src/hyperlight_guest/src/lib.rs
danbugs: We used a C string before too. I think it's for compatibility w/ our C-API: `pub extern "C" fn hl_abort_with_code_and_message(err: i32, message: *const c_char)`.

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -93,6 +94,58 @@ pub(super) fn outb_log(mgr: &mut SandboxMemoryManager<HostSharedMemory>) -> Resu
     Ok(())
 }
 
+const ABORT_TERMINATOR: u8 = 0xFF;
+
+thread_local! {
+    static GUEST_ABORT_BUFFER: RefCell<Vec<u8>> = const { RefCell::new(Vec::new()) }
```
ludfjig: Can we limit the size of this? Could potentially be a DOS target otherwise

danbugs: Will add a max length and check for it.

danbugs:
> Looks reasonable but have some questions. Also I'm not sure thread_local is right for this. What happens when two vms are running on the same thread (not currenly possible but will be very soon)?

Removed thread_local buffer and moved buffer to `MemMgrWrapper` so that it is exclusive to a sandbox and not to a thread.

src/hyperlight_guest/src/entrypoint.rs
ludfjig: If the 0xFF is unconditionally added by `abort_with_code`, then wouldn't adding it here add an additional exit, not reduce?

ludfjig (changes requested):
LGTM, but old file are added back

src/hyperlight_guest/src/entrypoint.rs
```diff
@@ -56,13 +57,19 @@ pub fn abort_with_code(code: &[u8]) -> ! {
...
-        CStr::from_ptr(message_ptr)
-            .to_str()
-            .expect("Invalid UTF-8 string"),
-    );
-
+    // Step 1: Send abort code (typically 1 byte, but `code` allows flexibility)
     outb(OutBAction::Abort as u16, code);
+
+    // Step 2: Convert the C string to bytes
+    let message_bytes = CStr::from_ptr(message_ptr).to_bytes(); // excludes null terminator
```
syntactically, line 64: Why are we doing this? We can define the ABI for hostcalls here, since this isn't "real" port I/O and we know outb will lead to a synchronous exit.

I always assumed this would be something like
- set <some well-known register, e.g. rax> to address of {length, data} FFI string
- call outb with an abort action

And the host would take care of reading the guest pointer and then reading the bytes of the string from guest memory.

src/hyperlight_guest/src/host_function_call.rs
```diff
@@ -119,11 +114,25 @@ pub fn outb(port: u16, data: &[u8]) {
...
 
-pub fn print_output_as_guest_function(function_call: &FunctionCall) -> Result<Vec<u8>> {
+/// Prints a message using `OutBAction::DebugPrint`. It transmits bytes of a message
+/// through several VMExists and, with such, it is slower than
+/// `print_output_with_host_print`.
+///
+/// This function should be used in debug mode only. This function does not
+/// require memory to be setup to be used.
+pub fn debug_print(msg: &str) {
+    outb(OutBAction::DebugPrint as u16, msg.as_bytes());
```
syntactically, line 128: Same comment as on abort above - why are we relying on the semantics of the legacy x86 port I/O instruction this way, when we really are just using it as a fast way to cause a synchronous vmexit after which the host can inspect the guest state howsoever we choose?

src/hyperlight_guest/src/interrupt_handlers.rs
```diff
@@ -31,10 +31,9 @@ pub extern "sysv64" fn hl_exception_handler(
 ) {
     let exception = Exception::try_from(exception_number as u8).expect("Invalid exception number");
     let msg = format!(
-        "EXCEPTION: {:#?}\n\
```
syntactically, line 34: Why does this remove useful logging?

src/hyperlight_guest/src/lib.rs
syntactically: Hm, I think there is probably a decent argument for using proper strings with their length passed properly in all our core functions and only converting to C strings at the C FFI boundaries.

src/hyperlight_host/src/sandbox/mem_mgr.rs
```diff
@@ -29,27 +29,39 @@ pub type StackCookie = [u8; STACK_COOKIE_LEN];
 /// A container with methods for accessing `SandboxMemoryManager` and other
 /// related objects
 #[derive(Clone)]
-pub(crate) struct MemMgrWrapper<S>(SandboxMemoryManager<S>, StackCookie);
+pub(crate) struct MemMgrWrapper<S> {
+    mgr: SandboxMemoryManager<S>,
+    stack_cookie: StackCookie,
+    abort_buffer: Vec<u8>,
```
syntactically, line 35: Is there a reason to have this buffer be part of this wrapper, rather than just have the abort message be returned as an Error variant and hopefully be bubbled up to the host code that initially invoked a guest function?

syntactically (commented):
I didn't take a super detailed look at this yet, but I'm somewhat confused by the approach.

src/hyperlight_guest/src/entrypoint.rs
danbugs: Part of the requirement here is being able to panic before memory is set on the guest side, so I can't just put the string in memory. Otherwise, yeah—I would use our output stack.

src/hyperlight_guest/src/host_function_call.rs
danbugs: Like you said, currently, we use `out` to perform exits. `out` enables us to transmit a port and up to 32 bits of value. This is just another usage of the mechanism we already rely on. Here, we are transmitting a message through multiple VM exits with characters being values. It's a slow but sometimes useful mechanism to print something before memory is set. It was previously introduced in #474 as a way to print abort messages. This PR removes its usage there in favor of still passing panic info with the HyperlightError but I'm leaving it here because I think it can still be useful when debugging.

src/hyperlight_guest/src/interrupt_handlers.rs
danbugs: I pass the exception now w/ the abort. I just removed the double printing.

src/hyperlight_guest/src/entrypoint.rs
danbugs: Responded w/ a top-level msg.

src/hyperlight_guest/src/lib.rs
danbugs: Fair play. Because this way of passing a message is already in main, for now, I'll keep this as is and maybe target it in a follow-up 👍

src/hyperlight_host/src/sandbox/mem_mgr.rs
danbugs: The buffer holds the entire panic context info. In #474, I had changed this logic to essentially just use error variants but one of the requirements here was to not lose that info.

src/hyperlight_guest/src/entrypoint.rs
syntactically: > so I can't just put the string in memory

The string already looks to be in memory here... Surely you can just exit with `message_bytes.as_ptr()` stashed in a register where the host can get it.

src/hyperlight_guest/src/host_function_call.rs
syntactically: > Like you said, currently, we use out to perform exits. out enables us to transmit a port and up to 32 bits of value. 

`out` the way we use it is a way of forcing a guest exit to VMM code, nothing more. _If we were a real hardware device on an actual port I/O bus_, yes, it would transmit a word of value to us, but we are not---the host can inspect the state of the guest arbitrarily upon exit.

src/hyperlight_host/src/sandbox/mem_mgr.rs
syntactically: Hm, is this an artifact of the fact that the message is being split across multiple exits?

src/hyperlight_guest/src/entrypoint.rs
danbugs: W/ the way our `out` handler is rn, there's no easy way to of doing something like ` vcpu_fd.get_regs()` to get a register value from a guest. Doing that would incur a bigger change than I had planned for this PR. I have an issue tracking improving the correctness of our `out` usage (#477). Maybe I can elaborate this as a follow-up there? I can see that removing this reliance on `out` would make it easier to transition to some other doorbell mechanism in the future.

src/hyperlight_guest/src/host_function_call.rs
danbugs: Responded in the thread above.

src/hyperlight_guest/src/entrypoint.rs
danbugs: Also, I should add—before, we already used the value part of `outb` to get data: https://github.com/hyperlight-dev/hyperlight/blob/ae139b02117ff3451a3e0a6745061ce9b4db6c34/src/hyperlight_host/src/sandbox/outb.rs#L139. In here and in #474, we're just getting more bytes.

src/hyperlight_host/src/sandbox/mem_mgr.rs
danbugs: No. Before, we also put the entire panic context info into the guest panic context memory region.

src/hyperlight_guest/src/entrypoint.rs
syntactically: I remember adding a similar interface for the trace stuff I did a bit ago (https://github.com/hyperlight-dev/hyperlight/pull/103/commits/8a23d78b456d7640be0b9a0f4384646e84a10753#diff-ff9dc5a5e903b284efc74dbfad22d47789d9b83e293fd82d4ab6ee3d2fc34c84, https://github.com/hyperlight-dev/hyperlight/pull/103/commits/0500ce85baf951509c01cf7aead26582673e709a), and it was pretty simple IIRC---basically just pass in the hypervisor and make sure you can get regs from it (possibly the debugging changes have already done the latter?). As @ludfjig pointed out, having this semi-state-machine that requires a bunch of different out calls in order makes me rather nervous, so I would rather avoid introducing more code that does this pattern here.

ludfjig: I agree, I think it would be good to simplify this

danbugs: I agree on this being the way forward. I'll need the outb handler to be able to extract register values in about 2 PRs anyway when I move the memory setup to the guest. I just feel like this would* be a hassle to do without #312 . How about we merge this to bring context back to aborts and implement your request as a follow-up? Hoping to do it for v0.5.0 still.

simongdavies: I agree lets merge this add an issue for the improvements that Lucy suggested and get it done when you make the other changes, lets make sure if gets in v0.5.0

simongdavies (approved):
I agree this can be improved, but its on the unhappy path, lets add an issue for the improvements that Lucy suggested and get this done after you make the other changes, lets make sure if gets in v0.5.0 and mark the issue as such

src/hyperlight_guest/src/entrypoint.rs
danbugs: Added info to this issue: #477 .

PR #491, opened 2025-05-14
Refactor HostFunction

jprendes:
Some more refactor of `HostFunction`.
In this PR:
* the registration can't be made from the function anymore, and has to be done from the sandbox.
* the registered function can return either the plain type `T`, or a `Result<T>`.
* adds a new `ParameterTuple` trait that is used to define serializable input to a function.


jprendes:
@ludfjig @danbugs PTAL, thank! :-)

danbugs (approved):
LGTM

Issue #495, opened 2025-05-15
Remove `OutBHandlerWrapper` type


simongdavies:
Now that inprocess mode has been removed we can also remove `OutBHandlerWrapper` since there is only a single implementation of the outb handler function now.

simongdavies:
@copilot Look at PR #519 to see how to implement this

Issue #472, opened 2025-05-08
Remove In Process Mode from hyperlight


simongdavies:
In process execution is incomplete , buggy and now unnecessary since we don't need it for debugging any longer (at least on Linux and also on Windows when #241 is implemented) so we should remove it.

PR #496, opened 2025-05-15
Remove in-process mode from hyperlight-guest and remaining clean up in hyperlight-host

simongdavies:
Completes implementation of #472

One final simplification we can make is to get rid of `OutBHandlerWrapper` since we now only have one implementation of outb handler function now. 

Added #495 to track this


ludfjig (approved):
Nice :)

danbugs (approved):
LGTM

PR #499, opened 2025-05-16
Cleanup of win64 and alike ABI usage and MSVC legacy

danbugs:
See individual commit messages for more details.


src/hyperlight_guest/third_party/musl/include/setjmp.h
```diff
@@ -0,0 +1,49 @@
+#ifndef	_SETJMP_H
```
ludfjig, line 1: What was the reason @syntactically removed the musl setjmp before? Are we fine to bring this back?

ludfjig (approved):
looks good to me

dblnz (approved):
LGTM

src/hyperlight_guest/third_party/musl/include/setjmp.h
danbugs: We never used musl's setjmp—we used musl's `setjmp.h`, which had an undersized `jmp_buf` for the implementation that we did use 👍

PR #500, opened 2025-05-16
guest: make `call_host_function` generic to avoid two steps to retrieve return value 

dblnz:
## Description
As the title says, this changes the `call_host_function` function to return a generic type.

## Details
The reason behind this change is to avoid calling the host function and getting the result in a separate step.
For the C guests we still need to do it in two steps as it does not support generic types.
To avoid code duplication I added a `call_host_function_internal` that does what `call_host_function` used to. In addition I added `call_host_function<T>` which calls the `call_host_function_internal` and also retrieves the return value.

This approach allows us to use a generic function to call and retrieve the return value in Rust and we can use `call_host_function_internal` for C guests.


src/hyperlight_host/src/func/utils.rs
```diff
@@ -4,6 +4,8 @@
 ///
 /// Usage:
 /// ```rust
+/// use hyperlight_host::func::for_each_tuple;
```
ludfjig, line 7: was this change intended?

ludfjig (approved):
LGTM, nice improvement

src/hyperlight_host/src/func/utils.rs
dblnz: Yes, there were two failing docs tests that I fixed. I don't know how on CI they did not fail

PR #517, opened 2025-05-21
Refactor Vm trait

### LINKED ISSUE #312: Unify register representations across all Hypervisor implementations

### LINKED ISSUE #465: Rethink driver API

### LINKED ISSUE #503: Avoid reopening /dev/kvm or /dev/mshv for every new sandbox

ludfjig:
This PR introduces a minimal `Vm` trait. It's implemented by `KvmVm`, `MshvVm` and `WhpVm`. The goal is to reduce code duplication. The existing `Hypervisor` trait is renamed to `HyperlightVm`, and is now only implemented by `HyperlightSandbox` (and Inprocess, but not for long since we're getting rid of inproces mode). In addition, the `Vm` trait also reduces a bunch of code duplication related to GDB debugging.

About half of the lines of this PR is introducing common register structs used for all Vms. They Are CommonRegisters, CommonStandardRegisters, and CommonFpu, with tests. The rest should be pretty straight forward, and is mostly just plumbing.

Closes #465 
Closes #312 
Closes #503

There are some more cleanup to do (like cleaning up some windows surrogate process things, tests, no more need for a `dyn` HyperligtVM, etc.), but figured I'll do that in separate PRs as this one is quite big already.


src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,768 @@
...
+                    .write(offset, data)?;
+
+                data = &data[write_len..];
+                gva += write_len as u64;
+            }
+
+            Ok(())
+        }
+    }
+}
```
danbugs, line 312: Maybe @dblnz wants to have a look at this?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,768 @@
...
+                Ok(HyperlightExit::Halt()) => {
+                    break;
+                }
+                Ok(HyperlightExit::IoOut(port, data)) => {
+                    self.handle_io(port, data, outb_handle_fn.clone())?
+                }
+                Ok(HyperlightExit::MmioRead(addr)) => {
+                    #[cfg(crashdump)]
+                    crashdump::crashdump_to_tempfile(self)?;
+
```
danbugs, line 542: Currently, on Windows, we have some very useful `debug!`s for different `HyperlightExit`s that capture VCPU state. I think these were removed from `whp.rs`. Would be good to re-add them here, so we have consistent `debug!`s for all drivers.

src/hyperlight_host/src/hypervisor/whp.rs
```diff
@@ -0,0 +1,410 @@
...
+    }
+}
+
+unsafe fn try_load_whv_map_gpa_range2() -> Result<WHvMapGpaRange2Func> {
+    let library = unsafe {
+        LoadLibraryExA(
+            s!("winhvplatform.dll"),
+            None,
+            LOAD_LIBRARY_SEARCH_DEFAULT_DIRS,
+        )
```
danbugs, line 391: Don't think we need this anymore now that in-process is gone. No?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -604,11 +604,8 @@ impl SandboxMemoryLayout {
         }
 
         // guard page
-        let stack_offset = builder.push_page_aligned(
-            PAGE_SIZE_USIZE,
-            MemoryRegionFlags::READ | MemoryRegionFlags::STACK_GUARD,
-            GuardPage,
-        );
+        let stack_offset =
+            builder.push_page_aligned(PAGE_SIZE_USIZE, MemoryRegionFlags::READ, GuardPage);
```
danbugs, line 608: No more `STACK_GUARD`? Guess we don't need it w/ the `RegionType` being there...

src/hyperlight_host/src/hypervisor/whp.rs
```diff
@@ -0,0 +1,410 @@
...
+    partition: WHV_PARTITION_HANDLE,
+    // Lazily create the surrogate process when we need to map memory
+    surrogate_process: Option<SurrogateProcess>,
+    mmap_file_handle: HandleWrapper,
+}
+
+unsafe impl Send for WhpVm {}
+unsafe impl Sync for WhpVm {}
+
+#[repr(C, align(16))]
```
danbugs, line 91: Maybe add a comment here on why we do this :)

danbugs (dismissed):
LGTM. Awesome work!

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: This is copied from previous implementation

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Isn't the crashdump feature enough for this? I can add them back if you prefer though

src/hyperlight_host/src/hypervisor/whp.rs
ludfjig: No I believe this is different. Here we're checking whether WHvMapGpaRange2 exists.

src/hyperlight_host/src/mem/layout.rs
ludfjig: Yeah I commented earlier in #469

> It was not a real permissions flag like READ or WRITE or EXECUTE, it was only an indicator. But that makes no sense because we already have another indicator for whether a region is a guard page. I don't think flag serves any purpose, and only messes things up when coverting a memory region to a vm-specific region. For example it makes us unable to map a region with only READ (which means readonly) to for example KVM_MEM_READONLY, without special handling of STACK_GUARD, which I am not a big fan of.

src/hyperlight_host/src/hypervisor/whp.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
simongdavies: I tend to agree with Dan, debug output might be used when crashdump feature is off, however we should probably not do this in release builds since registers might contain sensitive data

src/hyperlight_host/src/hypervisor/whp.rs
simongdavies: Yes , this is a crappy way on getting a relevant error if running on Windows 10

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -54,58 +55,51 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
+        if dr6 & DR6_BS_FLAG_MASK != 0 {
+            return Ok(VcpuStopReason::DoneStep);
         }
 
         // If any of the B0-B3 flags in DR6 register is set, it means a
         // hardware breakpoint triggered the exit
         // Check page 19-4 Vol. 3B of Intel 64 and IA-32
         // Architectures Software Developer's Manual
-        if DR6_HW_BP_FLAGS_MASK & dr6 != 0 && hw_breakpoints.contains(&rip) {
+        if DR6_HW_BP_FLAGS_MASK & dr6 != 0 {
```
simongdavies, line 77: why did this logic change?

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -54,58 +55,51 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
+                return Ok(VcpuStopReason::EntryPointBp);
             }
-            return VcpuStopReason::HwBp;
+            return Ok(VcpuStopReason::HwBp);
         }
     }
 
-    if BP_EX_ID == exception && sw_breakpoints.contains_key(&rip) {
-        return VcpuStopReason::SwBp;
+    if BP_EX_ID == exception {
```
simongdavies, line 86: why did this if change?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
simongdavies: Given the scope of the changes I think we should have @dblnz  review this

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,768 @@
...
+                gva += write_len as u64;
+            }
+
+            Ok(())
+        }
+    }
+}
+
+#[derive(Debug)]
+pub(crate) struct HyperlightSandbox {
```
simongdavies, line 315: I am a bit confused as we now seem to have a new struct called HyperlightSandbox and we already have a Sandbox , think this needs a different name , and probably needs to be in its own module file

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,768 @@
...
+
+            #[cfg(gdb)]
+            gdb_conn,
+        };
+
+        Ok(ret)
+    }
+}
+
+impl HyperlightVm for HyperlightSandbox {
```
simongdavies, line 388: #Why are we calling this HyperlightVm? what other Vms are we going to have? Should it be called Hypervisor?

src/hyperlight_host/src/hypervisor/mshv.rs
```diff
@@ -0,0 +1,470 @@
+/*
```
simongdavies, line 1: I think renaming these files is wrong, the old names were more descriptive, but, if we want to rename them then lets do it as a separate PR , it makes it even harder to review

simongdavies (changes requested):
I had to give up reviewing this, it is too big, please split this up into smaller sequential commits and a set of smaller PRs to make it reviewable. For example the changes to optimise /dev/kvm and /dev/mshv could easily be a seperate PR

dblnz:
Do you plan on updating this after the `hypervisor_handler change`?
I have partially reviewed the previous draft of this #469 .
Let me know when to have a look on this

ludfjig:
> Do you plan on updating this after the `hypervisor_handler change`? I have partially reviewed the previous draft of this #469 . Let me know when to have a look on this

I won't merge yet, I'm very sorry for delaying your windows PR. Let's get your windows PR merged first if it's ready

syntactically:
We still want to do this, but this is getting stale, so we will make another attempt. Desire to do this is tracked by #465

PR #524, opened 2025-05-23
Make common and guest libs portable

danbugs:
This PR changes our pointer usage in the guest and common lib to be portable (i.e., using `u64s` instead of `usize` or pointer types that vary between architectures).

Plus, I did some general housekeeping like removing legacy fields/comments, refactored the PEB structs to avoid code duplication, and madeour PEB related struct items snake case.

Reviews can be made commit-by-commit, if you preferred.

### DIFF
--- a/src/hyperlight_common/src/mem.rs
+++ b/src/hyperlight_common/src/mem.rs
@@ -46,10 +46,6 @@ pub struct GuestStackData {
     pub minUserStackAddress: u64,
     /// This is the user stack pointer
     pub userStackAddress: u64,
-    /// This is the stack pointer for the kernel mode stack
-    pub kernelStackAddress: u64,
-    /// This is the initial stack pointer when init is called its used before the TSS is set up
-    pub bootStackAddress: u64,
 }
 
 #[repr(C)]
--- a/src/hyperlight_common/src/mem.rs
+++ b/src/hyperlight_common/src/mem.rs
@@ -20,24 +20,22 @@ pub const PAGE_SHIFT: u64 = 12;
 pub const PAGE_SIZE: u64 = 1 << 12;
 pub const PAGE_SIZE_USIZE: usize = 1 << 12;
 
-use core::ffi::{c_char, c_void};
-
 #[repr(C)]
 pub struct InputData {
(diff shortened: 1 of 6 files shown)


jsturtevant:
from one of the commits:

> If we use usize and the guest and host are compiled for different architectures, there will be a mismatch
in sizing between the guest and host, so we should use u64s to address that (tested w/ Nanvix already).

What happens if one side is 32 bit?  could a cast down to the smaller pointer lose information?

jsturtevant:
> Reviews can be made commit-by-commit, if you preferred.

This made it really easy to review. Thanks!

devigned (approved):
LGTM. 

I will reserve a bit of concern about the assumptions made in this PR about the host targeting u64 and casting into usize. We do not intend on targeting Hyperlight host to 32 bit. If we were to pursue that, this code would likely not work as intended.

danbugs:
> LGTM.
> 
> I will reserve a bit of concern about the assumptions made in this PR about the host targeting u64 and casting into usize. We do not intend on targeting Hyperlight host to 32 bit. If we were to pursue that, this code would likely not work as intended.

Just to elaborate on this...

Before this PR, the guest writes a `usize` stack ptr and, regardless, the host always reads a u64 stack ptr. This works fine when both guest and host are compiled to a 64-bit architecture. But, if the guest is compiled to a 32-bit architecture, the guest will update the stack pointer incorrectly and the host will end up reading something beyond the stack ptr—potentially, part of a serialized flatbuffer structure.

W/ this PR in, regardless of architecture, the guest now also always writes a 64-bit stack ptr matching the expectation of the host.

The problem @devigned is pointing out occurs when we try to run a host on a 32-bit architecture. Currently, the host does this when reading the stack ptr:

```
let stack_pointer_rel = self.read::<u64>(buffer_start_offset)? as usize;
```

... which, on a 32-bit architecture, this could cause a silent truncation and UB.

That said, as things are currently, it's unlikely we could ever get any meaningful truncation because the ptr would have to be bigger than the maximum memory size we support. In any case, this is a good reminder we definitely don't support 32-bit hosts :)

PR #533, opened 2025-05-27
Remove hypervisor_handler thread

ludfjig:


src/hyperlight_host/src/func/call_ctx.rs
```diff
@@ -72,7 +71,8 @@ impl MultiUseGuestCallContext {
         // !Send (and !Sync), we also don't need to worry about
         // synchronization
 
-        call_function_on_guest(&mut self.sbox, func_name, func_ret_type, args)
+        self.sbox
```
simongdavies: Why did this change to the `_no_reset` version?

Justfile
```diff
@@ -87,10 +87,7 @@ test-isolated target=default-target features="":
...
     cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host --test integration_test -- log_message --exact --ignored
-    @# metrics tests
-    cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host --lib -- metrics::tests::test_metrics_are_emitted --exact --ignored
-    cargo test {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F function_call_metrics," + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host --lib -- metrics::tests::test_metrics_are_emitted --exact --ignored
-    
+
```
simongdavies: Why did these tests get deleted? They don't see related to this change, I guess we potentially have different metrics now (although arguably we might want to have a kill metric and metrics about success of killing etc.) but even if we don't we still have some metrics?

src/hyperlight_host/src/func/guest_dispatch.rs
simongdavies, line 1: It seems that deleting this file has deleted a bunch of tests that we would want to keep (e.g. seccomp tests etc.) I couldn't see that these had been relocated anywhere?

src/hyperlight_host/examples/logging/main.rs
```diff
@@ -89,17 +89,29 @@ fn main() -> Result<()> {
     let no_op = Noop::<UninitializedSandbox, MultiUseSandbox>::default();
 
     let mut multiuse_sandbox = usandbox.evolve(no_op)?;
+    let interrupt_handle = multiuse_sandbox.interrupt_handle();
+
+    const NUM_CALLS: i32 = 5;
```
simongdavies, line 90: Question: Are we logging attempts and results of kills? I assume so given this code has changed , if we are not logging when we kill we should

src/hyperlight_host/examples/metrics/main.rs
```diff
@@ -102,10 +102,22 @@ fn do_hyperlight_stuff() {
     let no_op = Noop::<UninitializedSandbox, MultiUseSandbox>::default();
 
     let mut multiuse_sandbox = usandbox.evolve(no_op).expect("Failed to evolve sandbox");
+    let interrupt_handle = multiuse_sandbox.interrupt_handle();
```
simongdavies, line 99: Same as logging? are we creating metrics on how many times kill has been called/ succeeded etc.? If not we probably should

src/hyperlight_host/examples/tracing/main.rs
```diff
@@ -117,8 +117,18 @@ fn run_example() -> Result<()> {
     let no_op = Noop::<UninitializedSandbox, MultiUseSandbox>::default();
 
     let mut multiuse_sandbox = usandbox.evolve(no_op)?;
+    let interrupt_handle = multiuse_sandbox.interrupt_handle();
 
     // Call a function that gets cancelled by the host function 5 times to generate some log entries.
+    const NUM_CALLS: i32 = 5;
```
simongdavies, line 117: Same question as the logging example, if we log then we should be creating trace records as well

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -390,6 +393,11 @@ impl HypervLinuxDriver {
             mem_regions,
             entrypoint: entrypoint_ptr.absolute()?,
             orig_rsp: rsp_ptr,
+            interrupt_handle: Arc::new(LinuxInterruptHandle {
+                running: AtomicBool::new(false),
+                tid: AtomicU64::new(unsafe { libc::pthread_self() }),
```
simongdavies, line 401: What is the thread id here? what happens if the sandbox is created on thread 1 , then used on thread 2 and killed will we send the kill signal to the wrong thread? 

Also is it possible that sandbox a and b are created on thread 1 then guest func is called on sandbox a , then host tries to kill guest func call ,meanwhile func_call ends on sandbox a and thread 1 calls func on sandbox b , then we send kill signal and kill the wrong one?

If we dont have tests for these scenarios then we should (and more that I haven't thought of)

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -577,13 +581,38 @@ impl Hypervisor for HypervLinuxDriver {
...
+        // Then this is fine because the call to `VcpuFd::run()` is already finished,
+        // the signal handler itself is a no-op, and the signals will stop being sent
+        // once we've set the `running` to false.
+        self.interrupt_handle
+            .running
+            .store(false, Ordering::Relaxed);
+        // Note: if a `InterruptHandle::kill()` signal is delivered to this thread **here**
+        // - after we've set the running to true,
+        // Then the signal does not have any effect, because the signal handler is a no-op.
+        // This is fine since we are already done with the `VcpuFd::run()` call.
```
simongdavies: I don't think that this statement is true, for example the reason for the exit may be to call a host function in which case we want to stop the execution, I think that this scenario is handled in the current implementation

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -634,58 +666,8 @@ impl Hypervisor for KVMDriver {
     }
 }
 
-#[cfg(test)]
```
simongdavies, line 637: Why did these tests get deleted?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -331,84 +320,49 @@ impl VirtualCPU {
...
+    /// # Note
+    /// This function will block for the duration of the time it takes for the vcpu thread to be interrupted.
+    fn kill(&self) -> bool;
 
-    use hyperlight_testing::dummy_guest_as_string;
+    /// Returns true iff the corresponding sandbox has been dropped
+    fn dropped(&self) -> bool;
+}
 
-    #[cfg(gdb)]
```
simongdavies, line 342: Why did this get deleted? We still want to test initialize don't we? Its still being executed just not via the handler?

simongdavies (commented):
Left some comments/questions , not quite finished reviewing , I will finish it in the morning

src/hyperlight_host/src/func/call_ctx.rs
ludfjig: previous `call_function_on_guest` in removed file `guest_dispatch.rs` didn't reset memory either, I just renamed it to make that more clear. I moved the method to MultiUseSandbox because I'm not sure what purpose `guest_dispatch.rs` served now that SingleUsedSandbox is removed.

Justfile
ludfjig: We don't have different metrics now. The test itself changed to use a local metrics recorder, which makes it capable to only record the metrics emitted in the given test. The local metrics recorder is able to only capture metrics that were emitted while the local recorder is active, and that are emitted on the same thread. We couldn't use the local metrics recorder before, because the metrics were emitted from a separate thread (hypervisor_handler thread), and local metrics recorder only worked on the given thread. 

In short, we no longer require #[ignore] on the test, so it is ran as part of the other tests. Your comment made me realize we should still test it with the metrics feature flag so I added back that one

src/hyperlight_host/examples/logging/main.rs
ludfjig: The metrics are still emitted, unchanged from previous behavior https://github.com/ludfjig/hyperlight/blob/210e5069b88e60d5398ff72abbb8ed031a152464/src/hyperlight_host/src/hypervisor/mod.rs#L297-L302.

Added logging to kill attempts

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: In the current implementation, if the user-thread times out while waiting for hypervisor_handler thread, the user-thread returns control back. However, it doesn't actually stop the host function call from happening, after the host-thread has returned control. Let's add host-function cancellation in another PR since it was never properly supported.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: We'll never send the signal to the wrong thread, because `tid` is always set before `running`, and we only send signal when `running` is set.

However, it's possible that killing a sandbox on thread 1 will accidentally kill a new sandbox on the same thread. As far as I know, I don't know of any way to avoid this.

- Thread 1 creates sandbox, finishes its call to vcpu::run(), but not yet set `running` to false
- Thread 2 calls `kill()`, and enters the while loop because `running` is still set to `true`
- Thread 1 sets `running` to false, and on this thread we create a new sandbox, and call vcpu::run on the new sandbox
- Thread 2 sends the signal
- Thread 1's vcpu of the new sandbox gets interrupted, but the user intended to kill the old one :/

This occurs because the checking `running` and sending the signal is not an atomic operation, and I'm not sure how to avoid it. Basically, between the green and red line, anything could happen on another thread

image

We'll at most send 1 incorrect signal, because the next checking of `running` in the while-loop is guaranteed to be false

src/hyperlight_host/src/func/guest_dispatch.rs
ludfjig: Whoops you are correct. I added them back to initializedmultiuse.rs (I didn't keep the ones that made no sense)

src/hyperlight_host/examples/metrics/main.rs
ludfjig: We have same metrics as before, which is how many sandbox cancellations have occured (successfully)

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: Deleted in kvm/windows/mshv in favor of running the test from mod.rs, since they all do the same thing regardless

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Added back! Whopsie

src/hyperlight_host/examples/tracing/main.rs
ludfjig: Yep the metrics are still there, and I added logging for kill attempts

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Fixed by ignoring signals unless `cancel_requested` is true

danbugs: So, say, we create sandbox_A and sandbox_B on thread_1. 

Then, in thread_1, we create threads thread_2 and thread_3 to execute guest functions of sandbox_A and sandbox_B. If we interrupt sandbox_A, that would kill sandbox_B too. Correct?

If this is right and we're ok w/ it, I think we should at least add a note on the `kill` function highlighting this. Another idea is to keep track of how many sandboxes we have running per thread or smt like that.

Agree that it would be nice to add a test that does this.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -572,15 +582,57 @@ impl Hypervisor for HypervLinuxDriver {
...
-        let run_result = {
-            let hv_message: hv_message = Default::default();
-            &self.vcpu_fd.run(hv_message)
+        self.interrupt_handle
+            .tid
+            .store(unsafe { libc::pthread_self() as u64 }, Ordering::Relaxed);
+        // Note: if a `InterruptHandle::kill()` called while this thread is **here**
+        // Then this is fine since `cancel_requested` is set to true, so we will skip the `VcpuFd::run()` call
+        self.interrupt_handle.running.store(true, Ordering::Relaxed);
+        // Don't run the vcpu is `cancel_requested` is true
```
danbugs: nit: 'Don't run the vcpu **if**'

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -572,15 +582,57 @@ impl Hypervisor for HypervLinuxDriver {
...
-        let result = match run_result {
+        // Note: if a `InterruptHandle::kill()` called while this thread is **here**
+        // Then signals will be sent to this thread until `running` is set to false.
+        // This is fine since the signal handler is a no-op.
+        let cancel_requested = self
+            .interrupt_handle
+            .cancel_requested
+            .swap(false, Ordering::Relaxed);
+        // Note: if a `InterruptHandle::kill()` called while this thread is **here**
+        // Then `cancel_requested` will be set to true again, which will cancel the **next vcpu run**.
```
danbugs, line 636: Should we really allow cancelling guest function calls before they are made like this?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -508,7 +517,50 @@ impl Hypervisor for KVMDriver {
 
     #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
     fn run(&mut self) -> Result<HyperlightExit> {
-        let exit_reason = self.vcpu_fd.run();
+        self.interrupt_handle
```
danbugs, line 524: Is this logic the same as in hyperv_linux? Should we make this a function? Perhaps related to #517 ?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -329,6 +318,66 @@ impl VirtualCPU {
...
+    /// True when an "interruptor" has requested the VM to be cancelled. Set immediately when
+    /// `kill()` is called, and cleared when the vcpu is no longer running.
+    /// This is used to
+    /// 1. make sure stale signals do not interrupt the
+    ///     the wrong vcpu (a vcpu may only be interrupted iff `cancel_requested` is true),
+    /// 2. ensure that if a vm is killed while a host call is running,
+    ///     the vm will not re-enter the guest after the host call returns.
+    cancel_requested: AtomicBool,
+    /// Whether the corresponding vm is dropped
+    dropped: AtomicBool,
```
danbugs, line 370: Do we use `dropped` outside integration tests?

src/hyperlight_host/src/signal_handlers/mod.rs
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 #[cfg(feature = "seccomp")]
 pub mod sigsys_signal_handler;
 
+pub(crate) const INTERRUPT_VCPU_SIGRTMIN_OFFSET: i32 = 0;
```
danbugs: Maybe add a comment on what this is?

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -14,19 +14,236 @@ See the License for the specific language governing permissions and
...
+};
 use log::LevelFilter;
 
 pub mod common; // pub to disable dead_code warning
 use crate::common::{new_uninit, new_uninit_rust};
 
+// A host function cannot be interrupted, but we can at least make sure after requesting to interrupt a host call,
+// we don't re-enter the guest again once the host call is done
+#[test]
+fn interrupt_host_call() {
```
danbugs, line 40: Idk if you do this later in the PR, but aside from integration tests, I'd also add a new `hyperlight_host` example where you make a cancellation thread like this and so on.

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -14,19 +14,236 @@ See the License for the specific language governing permissions and
...
+
+    thread.join().unwrap();
+}
+
+/// Makes sure a running guest call can be interrupted by the host
+#[test]
+fn interrupt_in_progress_guest_call() {
+    let mut sbox1: MultiUseSandbox = new_uninit_rust().unwrap().evolve(Noop::default()).unwrap();
+    let barrier = Arc::new(Barrier::new(2));
+    let barrier2 = barrier.clone();
```
danbugs, line 88: Is it worth writing a test like this but without the barriers?

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -14,19 +14,236 @@ See the License for the specific language governing permissions and
...
+/// Makes sure interrupting a vm before the guest call has started also prevents the guest call from being executed
+#[test]
+fn interrupt_guest_call_in_advance() {
+    let mut sbox1: MultiUseSandbox = new_uninit_rust().unwrap().evolve(Noop::default()).unwrap();
+    let barrier = Arc::new(Barrier::new(2));
+    let barrier2 = barrier.clone();
+    let interrupt_handle = sbox1.interrupt_handle();
+    assert!(!interrupt_handle.dropped()); // not yet dropped
+
+    // kill vm before the guest call has started
```
danbugs, line 127: I asked this before, but I'm curious to hear about why we want this feature.

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -14,19 +14,236 @@ See the License for the specific language governing permissions and
...
+/// even when multiple sandboxes share the same thread.
+/// This test runs several interleaved iterations where `sbox2` is interrupted,
+/// and ensures that:
+/// - `sbox1` and `sbox3` are never affected by the interrupt.
+/// - `sbox2` either completes normally or fails with `ExecutionCanceledByHost`.
+///
+/// This test is not foolproof and may not catch
+/// all possible interleavings, but can hopefully increases confidence somewhat.
+#[test]
+fn interrupt_same_thread() {
```
danbugs, line 162: Would be cool to try:
(1) create sbox1 and sbox2 on thread_a
(2) run Spin guest function for sbox1 on thread_b + run a guest function that can conclude for sbox2 on thread_c.
(3) cancel execution for sbox1.

Will (3) kill sbox2 as well or does this work fine because we re-set the tid on `run`?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -329,6 +318,66 @@ impl VirtualCPU {
...
+        let signal_number = libc::SIGRTMIN() + INTERRUPT_VCPU_SIGRTMIN_OFFSET;
+        let mut sent_signal = false;
+
+        while self.running.load(Ordering::Relaxed) {
+            log::info!("Sending signal to kill vcpu thread...");
+            sent_signal = true;
+            unsafe {
+                libc::pthread_kill(self.tid.load(Ordering::Relaxed) as _, signal_number);
+            }
+            std::thread::sleep(std::time::Duration::from_micros(50));
```
danbugs: Before, this was 500 micros: https://github.com/hyperlight-dev/hyperlight/blob/a68ff5eac911cb6c89ecb1aade5f810018e57ab9/src/hyperlight_host/src/hypervisor/hypervisor_handler.rs#L766 . Should we make this configurable?

src/hyperlight_host/src/metrics/mod.rs
```diff
@@ -132,27 +136,17 @@ mod tests {
...
-                let histogram_value = &snapshot.get(&histogram_key).unwrap().2;
-                assert!(
-                    matches!(
-                        histogram_value,
-                        metrics_util::debugging::DebugValue::Histogram(histogram) if histogram.len() == 1
-                    ),
-                    "Histogram metric does not match expected value"
-                );
+                let expected_num_metrics = if cfg!(all(feature = "seccomp", target_os = "linux")) {
+                    3 // if seccomp enabled, the host call duration metric is emitted on a separate thread which this local recorder doesn't capture
```
danbugs, line 141: Why did this change? Wasn't this always the case?

src/hyperlight_host/src/metrics/mod.rs
```diff
@@ -195,9 +189,28 @@ mod tests {
                     ),
                     "Histogram metric does not match expected value"
                 );
+
+                if !cfg!(all(feature = "seccomp", target_os = "linux")) {
+                    // 4. Host call duration
```
danbugs, line 194: Why make this (4) instead of (1)?

src/hyperlight_host/src/metrics/mod.rs
```diff
@@ -195,9 +189,28 @@ mod tests {
...
+                            histogram_value,
+                            metrics_util::debugging::DebugValue::Histogram(histogram) if histogram.len() == 1
+                        ),
+                        "Histogram metric does not match expected value"
+                    );
+                }
             } else {
                 // Verify that the counter metrics are recorded correctly
-                assert_eq!(snapshot.len(), 1, "Expected two metrics in the snapshot");
+                assert_eq!(snapshot.len(), 1);
```
danbugs, line 213: Why did this change?

danbugs (commented):
Really awesome work, Ludvig. Just got some comments, nits, and questions.

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: Yeah, had I gotten that PR merged first then I'd only need to write it 1 time. But @simongdavies wanted only this feature first since it's more urgent

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: We don't use it, it's intended for users to know whether or not they can discard a given InterruptHandle. Otherwise they have no way of knowing.

src/hyperlight_host/tests/integration_test.rs
ludfjig: For a host call specifically? logging/metrics/tracing-examples all cancel execution.

src/hyperlight_host/src/metrics/mod.rs
ludfjig: It was an incorrect message so removed it while I was editing the test

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: The alternative is having `kill()` potentially doing nothing if you call it before the vcpu thread is scheduled, which I think is worse. Additionally, without it we cannot cancel host calls (we can't properly cancel them either, but we can prevent them from returning control into the guest after executing)

src/hyperlight_host/src/signal_handlers/mod.rs
ludfjig: If you don't mind I think I'll leave it like this, I think the name makes it pretty clear

src/hyperlight_host/tests/integration_test.rs
ludfjig: For this one the barriers guarantee the thread-interleaving I want. Without them, I'm not sure what this test would do? There is another test for calling `kill` before the guest function is invoked.

src/hyperlight_host/tests/integration_test.rs
ludfjig: I think we want it. The alternative is having kill() potentially doing nothing if you call it before the vcpu thread is scheduled, which I think is worse.

Additionally, without it we cannot cancel host calls (we can't properly cancel them either, but we can prevent them from returning control into the guest after executing)

src/hyperlight_host/src/metrics/mod.rs
ludfjig: No reason, I just did it like this

src/hyperlight_host/src/metrics/mod.rs
ludfjig: We can run this test together with other tests now since it's not ignore anymore

src/hyperlight_host/tests/integration_test.rs
ludfjig: Added a new test for moving sandbox to new thread

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Good idea. Updated in latest commit

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -310,11 +314,12 @@ impl HypervLinuxDriver {
     /// `apply_registers` method to do that, or more likely call
     /// `initialise` to do it for you.
     #[instrument(skip_all, parent = Span::current(), level = "Trace")]
-    pub(super) fn new(
+    pub(crate) fn new(
         mem_regions: Vec<MemoryRegion>,
         entrypoint_ptr: GuestPtr,
         rsp_ptr: GuestPtr,
         pml4_ptr: GuestPtr,
+        config: &SandboxConfiguration,
```
simongdavies, line 322: Maybe just pass the delay here rather than the config?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -577,15 +585,57 @@ impl Hypervisor for HypervLinuxDriver {
...
-        #[cfg(mshv2)]
-        let run_result = {
-            let hv_message: hv_message = Default::default();
-            &self.vcpu_fd.run(hv_message)
+        self.interrupt_handle
+            .tid
+            .store(unsafe { libc::pthread_self() as u64 }, Ordering::Relaxed);
+        // Note: if a `InterruptHandle::kill()` called while this thread is **here**
+        // Then this is fine since `cancel_requested` is set to true, so we will skip the `VcpuFd::run()` call
+        self.interrupt_handle.running.store(true, Ordering::Relaxed);
```
simongdavies: should this be on line 607 since you dont actually start running till then?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
simongdavies: I think this is correct, it also mirrors what we did in the first implementation

dblnz (approved):
Awesome work! 💯 
I tested it locally. 😺

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -496,28 +520,28 @@ impl Hypervisor for HypervWindowsDriver {
...
+}
 
-    use crate::hypervisor::handlers::{MemAccessHandler, OutBHandler};
-    use crate::hypervisor::tests::test_initialise;
-    use crate::Result;
+impl InterruptHandle for WindowsInterruptHandle {
+    fn kill(&self) -> bool {
+        self.cancel_requested.store(true, Ordering::Relaxed);
+        self.running.load(Ordering::Relaxed)
+            && unsafe { WHvCancelRunVirtualProcessor(self.partition_handle, 0, 0).is_ok() }
```
dblnz, line 541: This will come in handy when the windows debug support comes in.
Currently we cannot interrupt a Hyper-V vCPU from the GDB thread, but having this will allow a user to stop a vCPU with the debugger

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -662,7 +712,15 @@ impl Hypervisor for HypervLinuxDriver {
             },
             Err(e) => match e.errno() {
                 // we send a signal to the thread to cancel execution this results in EINTR being returned by KVM so we return Cancelled
-                libc::EINTR => HyperlightExit::Cancelled(),
+                libc::EINTR => {
+                    // If cancellation was not requested for this specific vm, the vcpu was interrupted because of stale signal
+                    // that was meant to be delivered to a previous/other vcpu on this same thread, so let's ignore it
+                    if !cancel_requested {
+                        HyperlightExit::Retry()
```
simongdavies: Do we have any tests to ensure that this behaves the way we expect?

simongdavies: ALso can we invert this if ?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -577,15 +585,57 @@ impl Hypervisor for HypervLinuxDriver {
...
+            .running
+            .store(false, Ordering::Relaxed);
+        // At this point, `running` is false so no more signals will be sent to this thread,
+        // but we may still receive async signals that were sent before this point.
+        // To prevent those signals from interrupting subsequent calls to `run()`,
+        // we make sure to check `cancel_requested` before cancelling (see `libc::EINTR` match-arm below).
+        let result = match exit_reason {
             Ok(m) => match m.header.message_type {
                 HALT_MESSAGE => {
                     crate::debug!("mshv - Halt Details : {:#?}", &self);
```
simongdavies, line 647: If a kill is received after the VM has finished running but before IO message has been evaluated we are still going to let a host function call proceed , this is probably a small windows but I wonder if we can improve this?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -565,7 +615,15 @@ impl Hypervisor for KVMDriver {
                 libc::EINTR => HyperlightExit::Debug(VcpuStopReason::Interrupt),
                 // we send a signal to the thread to cancel execution this results in EINTR being returned by KVM so we return Cancelled
                 #[cfg(not(gdb))]
-                libc::EINTR => HyperlightExit::Cancelled(),
+                libc::EINTR => {
+                    // If cancellation was not requested for this specific vm, the vcpu was interrupted because of stale signal
+                    // that was meant to be delivered to a previous/other vcpu on this same thread, so let's ignore it
+                    if !cancel_requested {
```
simongdavies: Can you invert this if

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -634,58 +696,8 @@ impl Hypervisor for KVMDriver {
...
-        fn write(&mut self, _offset: usize, _data: &[u8]) -> Result<()> {
-            Ok(())
-        }
-
-        fn get_code_offset(&mut self) -> Result<usize> {
-            Ok(0)
-        }
-    }
-
-    #[test]
```
simongdavies, line 665: Did these tests get reinstated somwhere?

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -56,31 +56,8 @@ pub struct SandboxConfiguration {
...
-    max_wait_for_cancellation: u8,
-    // The max_initialization_time represents the maximum time the host should wait for a guest to initialize
-    // If set to 0, the max_initialization_time will be set to the default value of 2000ms.
-    // The minimum value is 1ms.
-    //
-    // Note: this is a C-compatible struct, so even though this optional
-    // field should be represented as an `Option`, that type is not
-    // FFI-safe, so it cannot be.
-    max_initialization_time: u16,
+    /// Interrupt retry delay
```
simongdavies: might be worth expanding on this a little to make it cler to a csonsumer what this value impacts

src/hyperlight_host/src/signal_handlers/mod.rs
```diff
@@ -45,7 +47,10 @@ pub(crate) fn setup_signal_handlers() -> crate::Result<()> {
             original_hook(panic_info);
         }));
     }
-    vmm_sys_util::signal::register_signal_handler(libc::SIGRTMIN(), handle_hltimeout)?;
+    vmm_sys_util::signal::register_signal_handler(
+        libc::SIGRTMIN() + INTERRUPT_VCPU_SIGRTMIN_OFFSET,
```
simongdavies: Is INTERRUPT_VCPU_SIGRTMIN_OFFSET configurable? just wondering what purpose is if it is constant and zero?

simongdavies: given we dont do chaining for the reasons below maybe we should make this offset configurable

src/hyperlight_host/tests/integration_test.rs
simongdavies: I thiknk we want it , if the host starts killing things because its under CPU pressure then we want to be able to kill as  quickly as possible , a host might choose to kill based on wall time in which case I think this makes sense

simongdavies (changes requested):

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: The behavior is largely the same, but the way I did it prevents a window where the the vcpu will run but it shouldn't:
With your version, the following is possible
1. Vcpu thread: reads `cancel_requested == false`
2. Main thread calls `kill()`
3. Main thread Sets `cancel_requested == true` (won't get read)
4. Main threads sees that `running` is false, so won't send any signals
5. Vcpu thread: sets `running = true`, and then calls `vcpu::run()`

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Fixed

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: It's a small window, but it's possible. It would be equivalent to calling `kill` while the vm is in a host call, so at least the control won't be returned to the guest after the host call

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: Yes, in `hypervisor/mod.rs`. The behavior should be identical to before.

src/hyperlight_host/src/sandbox/config.rs
ludfjig: Added (at the expense of exposing some implementation details)

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Mind if I keep it like this? We might need more things in the future

src/hyperlight_host/tests/integration_test.rs
danbugs: No, I meant for cancellation in general. Though I guess we can point people to the logging/metrics/tracing examples for reference. Just a little counter-intuitive.

src/hyperlight_host/src/signal_handlers/mod.rs
ludfjig: done!

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: I needed it again for the configurable signal number

src/hyperlight_host/src/metrics/mod.rs
danbugs: For future reference, `local_recorder` is what now allows us to run this test together w/ others because it only captures metrics sent on that test thread (ignoring others). W/ this, and because w/ seccomp ON the host functions run on a separate thread, we expect less metrics when the feature is ON.

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -357,14 +355,16 @@ pub(super) struct LinuxInterruptHandle {
     dropped: AtomicBool,
     /// Retry delay between signals sent to the vcpu thread
     retry_delay: Duration,
+    /// The offset of the SIGRTMIN signal used to interrupt the vcpu thread
+    sig_rt_min_offset: usize,
```
danbugs: This should not be a `usize`. `register_signal_handler` expects an `i32`. Later on, you also cast this to a `c_int`, which can lead to overflows. Here's an example where I set this value to `usize::MAX-2` in a test, caused an overflow, and replaced our `SIGSYS` (signalno 31) handler w/ the vm kill one: 
image

Imo, we should have this be a `u8` and also make sure that the value of `libc::SIGRTMIN() + self.sig_rt_min_offset as libc::c_int` stays under `libc::SIGRTMAX()`.

Let's also add tests using different config values here.

simongdavies (dismissed):
LGTM , nice work @ludfjig

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -55,6 +56,13 @@ pub struct SandboxConfiguration {
     /// field should be represented as an `Option`, that type is not
     /// FFI-safe, so it cannot be.
     heap_size_override: u64,
+    /// Delay between interrupt retries. This duration specifies how long to wait
+    /// between attempts to send signals to the thread running the sandbox's VCPU.
+    /// Multiple retries may be necessary because signals only interrupt the VCPU
+    /// thread when the vcpu thread is in kernel space. There's a narrow window during which a
+    /// signal can be delivered to the thread, but the thread may not yet
+    /// have entered kernel space.
+    interrupt_retry_delay: Duration,
```
danbugs, line 67: I think we should add a test for using different interrupt retry delays.

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Whoops, should be fixed by u8. Also added tests that exercises different signal number and retry delay

src/hyperlight_host/src/sandbox/config.rs
ludfjig: done

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -629,9 +636,7 @@ impl Hypervisor for HypervLinuxDriver {
         // Then `cancel_requested` will be set to true again, which will cancel the **next vcpu run**.
         // Additionally signals will be sent to this thread until `running` is set to false.
```
danbugs, line 637: Might want to correct these comments w/ changing this to an `AtomicU64`.

danbugs (commented):
Looking good. Thanks for the fix on the `SIGRTMIN` offset stuff!

To finish off this PR, I think we just have to update some docs. Doing a quick search for "hypervisor handler" in the project shows references in `shared_mem.rs`, `guest.rs`, `how-to-debug-a-hyperlight-guest.md`, and the `copilot-instructions.md`.

Also, I think we should update the `signal-handlers-development-notes.md` doc to be appropriate to what we have rn.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: I think it makes sense to keep, while not technically correct, there's still a concept of `running` being true or false (even if it's a bit instead of a bool now)

PR #535, opened 2025-05-28
Remove some dev-dependecies cargo features to speed up compilation

ludfjig:
Decrease the amount of crates we have to compile.
`cargo test` goes from compiling 374 crates to 323


ludfjig:
> Did you test that the instructions at https://github.com/hyperlight-dev/hyperlight/blob/main/docs/hyperlight-metrics-logs-and-traces.md#using-otlp-exporter-and-jaeger still work?

Yep just tested, it's still working fine

dblnz (dismissed):
need to solve conflict

ludfjig:
> need to solve conflict

fixed

> Did you test that the instructions at https://github.com/hyperlight-dev/hyperlight/blob/main/docs/hyperlight-metrics-logs-and-traces.md#using-otlp-exporter-and-jaeger still work?

yes

docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -81,13 +81,13 @@ The sample will run and generate trace data until any key is pressed.
...
 ```console
- docker run -d --name jaeger -e COLLECTOR_OTLP_ENABLED=true -p 4317:4317 -p 16686:16686 jaegertracing/all-in-one:1.60
+ docker run -p 16686:16686 -p 4317:4317 -p 4318:4318 -e COLLECTOR_OTLP_ENABLED=true jaegertracing/all-in-one:latest
 ```
 
 NOTE: when running this on windows that this is a linux container, so you will need to ensure that docker is configured to run linux containers using WSL2. Alternatively, you can download the Jaeger binaries from [here](https://www.jaegertracing.io/download/). Extract the archive and run the `jaeger-all-in-one` executable as follows:
 
 ```powershell
-.\jaeger-all-in-one.exe  --collector.otlp.grpc.host-port=4317
+.\jaeger-all-in-one.exe
```
jsturtevant, line 90: any reason for moving from grpc endpoint to http?

ludfjig: nothing other than that it brings in less dependencies

jprendes (approved):
LGTM

marosset (approved):

PR #557, opened 2025-06-03
Update steps for making a patch release

ludfjig:
I understand this is not a pretty process, but updating it to what the current process looks like so we can improve it later

### DIFF
--- a/docs/how-to-make-releases.md
+++ b/docs/how-to-make-releases.md
@@ -59,7 +59,13 @@ When this job is done, a new [GitHub release](https://github.com/hyperlight-dev/
 - `hyperlight-host`
 
 ## Patching a release
+> Note: for this example we'll assume `v0.4.0` is already released, and you want to release a `v0.4.1` patch for it.
 
-If you need to update a previously released version of Hyperlight then you should open a Pull Request against the release branch you want to patch, for example if you wish to patch the release `v0.4.0` then you should open a PR against the `release/v0.4.0` branch.
+1. Make sure the patch/patches you want include are already merged to `main` branch.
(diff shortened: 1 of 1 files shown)


docs/how-to-make-releases.md
```diff
@@ -59,7 +59,13 @@ When this job is done, a new [GitHub release](https://github.com/hyperlight-dev/
...
 
 ## Patching a release
+> Note: for this example we'll assume `v0.4.0` is already released, and you want to release a `v0.4.1` patch for it.
 
-If you need to update a previously released version of Hyperlight then you should open a Pull Request against the release branch you want to patch, for example if you wish to patch the release `v0.4.0` then you should open a PR against the `release/v0.4.0` branch.
+1. Make sure the patch/patches you want include are already merged to `main` branch.
+2. Make sure `CHANGELOG.md` is updated with the changes you want to include in the patch release (see instructions above) and is merged to `main` branch.
+3. Make sure the `Cargo.toml` versions are updated to the new patch version (e.g. `v0.4.1`), and is merged to `main` branch.
+4. Manually create a new branch **from the `release/v0.4.0` branch** and name it `release/v0.4.1`. Important: do not create the branch from `main` branch.
+5. Cherry-pick the commits from `main` that you want to include in the patch. You must include the commit that updated the `CHANGELOG.md` and the commit that updated the `Cargo.toml` versions.
```
dblnz: Nit: This step might result in conflicts which need to be addressed.

danbugs (approved):

docs/how-to-make-releases.md
ludfjig: correct. Added a small sentence about it

danbugs (approved):

PR #559, opened 2025-06-03
Make `GuestBinary::Buffer` variant take slice instead of owned vec

ludfjig:
This avoids copies if making several GuestBinaries from a single buffer. I will follow up this PR with another one that makes `Uninitialized::new` take a GuestBinary by reference instead of value


qbradley:
Thanks!

dblnz (approved):

PR #560, opened 2025-06-03
Take guestbinary by ref instead of value

ludfjig:
Depends on #559 

Making `Uninitialized::new` take a `GuestBinary` by ref instead of value is more flexible, allowing reuse of the same guest binary across many different sandboxes. This PR allows passing either a ref or value, to be maximally flexible and not be a breaking change

### DIFF
--- a/src/hyperlight_host/src/sandbox/uninitialized.rs
+++ b/src/hyperlight_host/src/sandbox/uninitialized.rs
@@ -132,7 +132,10 @@ impl UninitializedSandbox {
         skip(guest_binary),
         parent = Span::current()
     )]
-    pub fn new(guest_binary: GuestBinary, cfg: Option<SandboxConfiguration>) -> Result<Self> {
+    pub fn new<'a>(
+        guest_binary: impl std::borrow::Borrow<GuestBinary<'a>>,
+        cfg: Option<SandboxConfiguration>,
+    ) -> Result<Self> {
         #[cfg(feature = "build-metadata")]
         log_build_details();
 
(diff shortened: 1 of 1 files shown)


dblnz (approved):
Lgtm

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -132,7 +132,7 @@ impl UninitializedSandbox {
         skip(guest_binary),
         parent = Span::current()
     )]
-    pub fn new(guest_binary: GuestBinary, cfg: Option<SandboxConfiguration>) -> Result<Self> {
+    pub fn new(guest_binary: &GuestBinary, cfg: Option<SandboxConfiguration>) -> Result<Self> {
```
jprendes: **suggestion:** Take [`Borrow<GuestBinary>`](GuestBinary), then you can take `GuestBinary` or `&GuestBinary`. That would also help reduce the number of changes in this PR, as it would be backwards compatible.
```suggestion
    pub fn new(guest_binary: impl Borrow<GuestBinary>, cfg: Option<SandboxConfiguration>) -> Result<Self> {
        let guest_binary = guest_binary.borrow();
```

ludfjig: Good idea! Fixed

danbugs (dismissed):

jprendes (dismissed):

ludfjig:
I'm closing this for now, since #605 makes this PR non-trivial. The current state at least avoids copies of the guest binary since #559 has landed (but still requires a new struct for each sandbox - bad!) . Let's re-visit when we redesign the API.

PR #572, opened 2025-06-05
Remove support for building PE files from hyperlight-guest-bin build.rs

simongdavies:
This pull request simplifies the build process for the `hyperlight_guest_bin` by removing support for Windows-specific tools and configurations since we no longer support PE guests.

### DIFF
--- a/src/hyperlight_guest_bin/.cargo/config.toml
+++ b/src/hyperlight_guest_bin/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+target = "x86_64-unknown-none"
--- a/src/hyperlight_guest_bin/build.rs
+++ b/src/hyperlight_guest_bin/build.rs
@@ -62,8 +62,6 @@ fn cargo_main() {
             .include("third_party/musl/arch/x86_64");
     }
 
-    let is_pe = env::var("CARGO_CFG_WINDOWS").is_ok();
-
     if cfg!(any(feature = "printf", feature = "libc")) {
         cfg.define("HYPERLIGHT", None); // used in certain musl files for conditional compilation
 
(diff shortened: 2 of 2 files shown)


src/hyperlight_guest_bin/build.rs
```diff
@@ -78,37 +76,26 @@ fn cargo_main() {
...
+        cfg.flag("--target=x86_64-unknown-linux-none");
+
+        // We don't support stack protectors at the moment, but Arch Linux clang
+        // auto-enables them for -linux platforms, so explicitly disable them.
+        cfg.flag("-fno-stack-protector");
+        cfg.flag("-fstack-clash-protection");
+        cfg.flag("-mstack-probe-size=4096");
+        cfg.compiler("clang");
+
+        unsafe { env::set_var("AR_x86_64_unknown_none", "llvm-ar") };
```
ludfjig: is this correct? I think we might want to keep the if/else block here

ludfjig (approved):
Looks reasonable, maybe @syntactically could have a look as she wrote the code

src/hyperlight_guest_bin/build.rs
simongdavies: fixed

src/hyperlight_guest_bin/build.rs
```diff
@@ -183,29 +172,20 @@ fn cargo_main() {
...
-    Ml64,
     Clang,
-    ClangCl,
 }
 impl From<&std::ffi::OsStr> for Tool {
     fn from(x: &std::ffi::OsStr) -> Tool {
-        if x == "ml64.exe" {
-            Tool::Ml64
-        } else if x == "clang" || x == "clang.exe" {
+        if x == "clang" {
```
jprendes: I think we still need to check for "clang.exe", right?

src/hyperlight_guest_bin/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[build]
+target = "x86_64-unknown-none"
```
jprendes, line 2: Does this let us simplify some Justfiles / scripts?

syntactically: I think we already got rid of most of it, but it's probably worth double-checking.

src/hyperlight_guest_bin/build.rs
syntactically: Agreed, that's probably important for building on Windows?

src/hyperlight_guest_bin/build.rs
```diff
@@ -252,44 +232,24 @@ fn main() -> std::process::ExitCode {
     let mut args = env::args();
     args.next(); // ignore the exe name
     let include_dir = <String as AsRef<Path>>::as_ref(&out_dir).join("include");
-    match tool {
```
syntactically, line 255: Since we still have the `Tool` enum, I would keep the `match` even though it only has one arm so that it's clearer how to add a tool in the future if we ever need to.

syntactically (commented):
Mostly LGTM except for a couple of little things :)

src/hyperlight_guest_bin/.cargo/config.toml
simongdavies: It just makes it possible to run cargo build in the crate directory, I also think it documents what the build target is so I think its worth leaving there

src/hyperlight_guest_bin/build.rs
simongdavies: It seemed to work without but I put it back in case its needed in some cases

src/hyperlight_guest_bin/build.rs
simongdavies: Fixed

simongdavies:
@syntactically @jprendes please take another look when you have a moment

PR #573, opened 2025-06-05
Prevent openat from trapping on seccomp thread, by making it return EACCES instead

### LINKED ISSUE #175: CI instability: openat(2) where we don't expect it
See: https://github.com/hyperlight-dev/hyperlight/actions/runs/13000789626/job/36258850656?pr=157

ludfjig:
We have seen CI being flaky due to a sporatic `openat` syscall. Turns out (g)libc can try to open `/proc/sys/vm/overcommit_memory` during a call to `free()`. 

This PRs makes `openat` return `EACCES` rather than trap.

An alternative solution I tried is to "prime" malloc/free before applying the seccompfilter, but that didn't work.

Closes #175

### DIFF
--- a/src/hyperlight_host/src/seccomp/guest.rs
+++ b/src/hyperlight_host/src/seccomp/guest.rs
@@ -93,20 +93,28 @@ pub(crate) fn get_seccomp_filter_for_host_function_worker_thread(
 
     let arch: TargetArch = std::env::consts::ARCH.try_into()?;
 
-    // First: Filter that forces `openat` to return EACCES
-    let errno_on_openat = SeccompFilter::new(
-        [(libc::SYS_openat, vec![])].into_iter().collect(),
-        SeccompAction::Allow,
-        SeccompAction::Errno(libc::EACCES.try_into()?),
-        arch,
-    )?
-    .try_into()?;
-
-    // Second: Allowlist filter that traps on unknown syscalls
+    // Allowlist filter that traps on unknown syscalls
     let allowlist = SeccompFilter::new(
         allowed_syscalls.into_iter().collect(),
         SeccompAction::Trap,
         SeccompAction::Allow,
+        arch,
+    )?
+    .try_into()?;
+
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/seccomp/guest.rs
```diff
@@ -87,11 +91,42 @@ pub(crate) fn get_seccomp_filter_for_host_function_worker_thread(
...
+    //    SECCOMP_RET_KILL.  This is done to simplify the kernel code and to
+    //    provide a tiny speed-up in the execution of sets of filters by
+    //    avoiding a check for this uncommon case.)  The return value for
+    //    the evaluation of a given system call is the first-seen action
+    //    value of highest precedence (along with its accompanying data)
+    //    returned by execution of all of the filters.
+    //
+    //  (https://man7.org/linux/man-pages/man2/seccomp.2.html).
+    //
+    Ok(vec![errno_on_openat, allowlist])
```
jprendes, line 139: Would this unconditionally block the openat syscalls (with EACCES) for all host functions?
If I wanted to use openat in my host function, how would I allow it?

jprendes: Can we maybe inspect the `allowed_syscalls` list, and if it contains `SYS_openat` then we don't install the `errno_on_openat` filter? something like...
```rust
if !allowed_syscalls.contains(SYS_openat) {
    // SYS_openat is not allowed, but instead of disallowing and trapping, we allow it and return EACCESS
    let errno_on_openat = ..;
    filters.push(errno_on_openat);
    allowed_syscalls.push(SYS_openat);
}
let allowlist = ..;
filters.push(allowlist);
```

ludfjig: yes good catch, I will fix this. Although `allowed_syscall` will unconditionally contain `SYS_openat`, so should probably check `extra_allowed_syscalls` instead

jprendes (approved):
LGTM

PR #588, opened 2025-06-09
Remove version number from hyperlight-testing crate dependency

simongdavies:
In #580 a version number was adding to the hyperlight-testing dependency. This caused the Cargo pubish job for the last release to [fail](https://github.com/hyperlight-dev/hyperlight/actions/runs/15502276649/job/43652946703).

This PR removes the version number so we don't have to publish the hyperlight-testing crate


jprendes:
I think we also need to remove the
```toml
version.workspace = true
```
from `src/hyperlight_testing/Cargo.toml`

simongdavies:
> I think we also need to remove the
> 
> ```toml
> version.workspace = true
> ```
> 
> from `src/hyperlight_testing/Cargo.toml`

How come, I did a dry run publish and it worked

jprendes:
#580 added the version to `src/hyperlight_testing/Cargo.toml` as well.
Removing it from the workspace might be enough to fix the issue, but I think we should also remove it from the crate's toml file.
If dry run is happy, then it's just a nit. I let you decide if you want to remove it now or not :-)

jprendes (dismissed):
LGTM

jprendes (approved):
LGTM

PR #600, opened 2025-06-11
Bring back HostFunctionDefinitions Region

danbugs:
This memory region removed because it wasn't used by Hyperlight Core, but it is used for type introspection in Hyperlight Wasm when running Wasm modules, so this PR reverts the work of #453.


src/hyperlight_guest/src/host_functions.rs
```diff
@@ -0,0 +1,112 @@
+/*
```
jprendes: This file is not used by the parent module.
In `src/hyperlight_guest/src/lib.rs`
```rust
pub mod host_functions;
```

src/hyperlight_guest/src/host_functions.rs
```diff
@@ -0,0 +1,112 @@
...
+use alloc::vec::Vec;
+use core::slice::from_raw_parts;
+
+use hyperlight_common::flatbuffer_wrappers::function_call::FunctionCall;
+use hyperlight_common::flatbuffer_wrappers::function_types::ParameterType;
+use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
+use hyperlight_common::flatbuffer_wrappers::host_function_details::HostFunctionDetails;
+
+use crate::error::{HyperlightGuestError, Result};
+use crate::P_PEB;
```
jprendes: P_PEB doesn't exist anymore.

src/hyperlight_guest/src/host_functions.rs
```diff
@@ -0,0 +1,112 @@
...
+    if host_function_details.host_functions.is_none() {
+        return Err(HyperlightGuestError::new(
+            ErrorCode::GuestError,
+            "No host functions found".to_string(),
+        ));
+    }
+
+    // check if function w/ given name exists
+    let host_function = if let Some(host_function) =
+        host_function_details.find_by_function_name(&function_call.function_name)
```
jprendes: the implementation of `find_by_function_name` is missing

src/hyperlight_guest/src/host_functions.rs
danbugs: Didn't mean to include this w/ the revert, changed.

src/hyperlight_guest/src/host_functions.rs
danbugs: This whole module shouldn't be here. Removed.

src/hyperlight_guest/src/host_functions.rs
danbugs: This whole module shouldn't be here. Removed.

PR #601, opened 2025-06-11
Allow hyperlight-host to build with x86_64-unknown-linux-musl target

simongdavies:
This pull request introduces support for building with the `musl` libc target, updates syscall handling for compatibility, and adds conditional compilation for thread ID initialization. Below is a summary of the most important changes:

### Build System Enhancements:
* Added a new `build-with-musl-libc` target in the `Justfile` to enable building with the `musl` libc target (`x86_64-unknown-linux-musl`). This includes adding the required Rust target and building with the appropriate profile.

### Conditional Compilation for `musl`:
* Updated the `LinuxInterruptHandle` initialization in `HypervLinuxDriver` to conditionally define the `tid` field based on whether the target is `musl` or not. This ensures compatibility with the `musl` libc environment.
* Applied the same conditional logic for the `tid` field in the `KVMDriver` to maintain consistency across hypervisor implementations.

### Syscall Handling:
* Modified the `syscalls_allowlist` function to use `try_into()` for converting `libc::TCGETS` to ensure type safety and compatibility with stricter environments.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -24,7 +24,12 @@ alias cg := build-and-move-c-guests
 
 # build host library
 build target=default-target:
-    cargo build --profile={{ if target == "debug" { "dev" } else { target } }}
+    cargo build --profile={{ if target == "debug" { "dev" } else { target } }} 
+
+# build host library
+build-with-musl-libc target=default-target:
+    cargo build --profile={{ if target == "debug" { "dev" } else { target } }} --target x86_64-unknown-linux-musl
+
 
 # build testing guest binaries
 guests: build-and-move-rust-guests build-and-move-c-guests
--- a/src/hyperlight_host/src/hypervisor/kvm.rs
+++ b/src/hyperlight_host/src/hypervisor/kvm.rs
@@ -360,6 +360,19 @@ impl KVMDriver {
             interrupt_handle: Arc::new(LinuxInterruptHandle {
                 running: AtomicU64::new(0),
                 cancel_requested: AtomicBool::new(false),
+                #[cfg(all(
+                    target_arch = "x86_64",
+                    target_vendor = "unknown",
+                    target_os = "linux",
+                    target_env = "musl"
+                ))]
+                tid: AtomicU64::new(unsafe { libc::pthread_self() as u64 }),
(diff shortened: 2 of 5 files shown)


Justfile
```diff
@@ -24,7 +24,13 @@ alias cg := build-and-move-c-guests
 
 # build host library
 build target=default-target:
-    cargo build --profile={{ if target == "debug" { "dev" } else { target } }}
+    cargo build --profile={{ if target == "debug" { "dev" } else { target } }} 
+
+# build host library
+build-with-musl-libc target=default-target:
+    rustup target add x86_64-unknown-linux-musl
```
ludfjig: What do you think about adding this to `rusttoolchain.toml` instead? That way it would get automatically installed when needed

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -360,6 +360,19 @@ impl KVMDriver {
             interrupt_handle: Arc::new(LinuxInterruptHandle {
                 running: AtomicU64::new(0),
                 cancel_requested: AtomicBool::new(false),
+                #[cfg(all(
+                    target_arch = "x86_64",
+                    target_vendor = "unknown",
+                    target_os = "linux",
+                    target_env = "musl"
+                ))]
+                tid: AtomicU64::new(unsafe { libc::pthread_self() as u64 }),
```
ludfjig, line 369: How about an unconditional cast `as u64`, together with an `allow/expect(unnecessary_cast)`? That would simplify this a bit

ludfjig (dismissed):

Justfile
simongdavies: OK, yes good idea

src/hyperlight_host/src/hypervisor/kvm.rs
simongdavies: Hmmm , then I feel like we have to write a comment saying why its there and then I feel like if we make the code do it like this then its obvious (if a bit annoying)

dblnz (approved):
LGTM

PR #605, opened 2025-06-12
Added capability to load extra blob data in sandbox

danbugs:
This PR contains the following changes:
**(1)** Made a non-breaking change to the signature of the `UninitializedSandbox::new(...)` function. Now, instead of a `GuestBinary`, it takes a `GuestEnvironment`, which contains the usual binary + optional blob data.
**(2)** Created new user memory region to store the extra blob data and added some extra config options for it.
**(3)** Added info of the new memory region to the PEB.
**(4)** Added tests verifying added blob data.

Reviewing commit-by-commit might be best 👍

### DIFF
--- a/src/hyperlight_guest/src/guest_handle/host_comm.rs
+++ b/src/hyperlight_guest/src/guest_handle/host_comm.rs
@@ -34,6 +34,29 @@ use crate::error::{HyperlightGuestError, Result};
 use crate::exit::out32;
 
 impl GuestHandle {
+    /// Get user memory region as bytes.
+    pub fn read_n_bytes_from_user_memory(&self, num: u64) -> Result<Vec<u8>> {
+        let peb_ptr = self.peb().unwrap();
+        let user_memory_region_ptr = unsafe { (*peb_ptr).init_data.ptr as *mut u8 };
+        let user_memory_region_size = unsafe { (*peb_ptr).init_data.size };
+
+        if num > user_memory_region_size {
+            return Err(HyperlightGuestError::new(
+                ErrorCode::GuestError,
+                format!(
+                    "Requested {} bytes from user memory, but only {} bytes are available",
+                    num, user_memory_region_size
+                ),
+            ));
+        } else {
+            let user_memory_region_slice =
+                unsafe { core::slice::from_raw_parts(user_memory_region_ptr, num as usize) };
+            let user_memory_region_bytes = user_memory_region_slice.to_vec();
+
+            Ok(user_memory_region_bytes)
+        }
+    }
+
     /// Get a return value from a host function call.
     /// This usually requires a host function to be called first using
     /// `call_host_function_internal`.
--- a/src/hyperlight_guest_bin/src/host_comm.rs
+++ b/src/hyperlight_guest_bin/src/host_comm.rs
@@ -65,6 +65,11 @@ pub fn get_host_function_details() -> HostFunctionDetails {
     handle.get_host_function_details()
 }
 
+pub fn read_n_bytes_from_user_memory(num: u64) -> Result<Vec<u8>> {
+    let handle = unsafe { GUEST_HANDLE };
+    handle.read_n_bytes_from_user_memory(num)
+}
+
 /// Print a message using the host's print function.
 ///
 /// This function requires memory to be setup to be used. In particular, the
--- a/src/hyperlight_host/src/mem/layout.rs
+++ b/src/hyperlight_host/src/mem/layout.rs
@@ -21,15 +21,20 @@ use rand::{RngCore, rng};
 use tracing::{Span, instrument};
 
 use super::memory_region::MemoryRegionType::{
-    Code, GuardPage, Heap, HostFunctionDefinitions, InputData, OutputData, PageTables, Peb, Stack,
+    Code, GuardPage, Heap, HostFunctionDefinitions, InitData, InputData, OutputData, PageTables,
+    Peb, Stack,
 };
-use super::memory_region::{MemoryRegion, MemoryRegionFlags, MemoryRegionVecBuilder};
+use super::memory_region::{
+    DEFAULT_GUEST_BLOB_MEM_FLAGS, MemoryRegion, MemoryRegionFlags, MemoryRegionVecBuilder,
+};
 use super::mgr::AMOUNT_OF_MEMORY_PER_PT;
 use super::shared_mem::{ExclusiveSharedMemory, GuestSharedMemory, SharedMemory};
 use crate::error::HyperlightError::{GuestOffsetIsInvalid, MemoryRequestTooBig};
 use crate::sandbox::SandboxConfiguration;
 use crate::{Result, new_error};
 
+// +-------------------------------------------+
+// |              Init Data                    | (GuestBlob size)
 // +-------------------------------------------+
 // |             Guest (User) Stack            |
 // +-------------------------------------------+
@@ -56,6 +61,8 @@ use crate::{Result, new_error};
 // |                   PML4                    |
 // +-------------------------------------------+ 0x0_000
 
+/// - `InitData` - some extra data that can be loaded onto the sandbox during
+///     initialization.
 ///
 /// - `HostDefinitions` - the length of this is the `HostFunctionDefinitionSize`
 ///   field from `SandboxConfiguration`
@@ -82,6 +89,7 @@ pub(crate) struct SandboxMemoryLayout {
     pub(super) stack_size: usize,
     /// The heap size of this sandbox.
     pub(super) heap_size: usize,
+    init_data_size: usize,
 
     /// The following fields are offsets to the actual PEB struct fields.
     /// They are used when writing the PEB struct itself
@@ -103,6 +111,7 @@ pub(crate) struct SandboxMemoryLayout {
     guest_heap_buffer_offset: usize,
     guard_page_offset: usize,
     guest_user_stack_buffer_offset: usize, // the lowest address of the user stack
+    init_data_offset: usize,
 
     // other
     pub(crate) peb_address: usize,
@@ -111,6 +120,7 @@ pub(crate) struct SandboxMemoryLayout {
     total_page_table_size: usize,
     // The offset in the sandbox memory where the code starts
     guest_code_offset: usize,
+    pub(crate) init_data_permissions: Option<MemoryRegionFlags>,
 }
 
 impl Debug for SandboxMemoryLayout {
@@ -122,6 +132,10 @@ impl Debug for SandboxMemoryLayout {
             )
             .field("Stack Size", &format_args!("{:#x}", self.stack_size))
             .field("Heap Size", &format_args!("{:#x}", self.heap_size))
+            .field(
+                "Init Data Size",
+                &format_args!("{:#x}", self.init_data_size),
+            )
             .field("PEB Address", &format_args!("{:#x}", self.peb_address))
             .field("PEB Offset", &format_args!("{:#x}", self.peb_offset))
             .field("Code Size", &format_args!("{:#x}", self.code_size))
@@ -182,6 +196,10 @@ impl Debug for SandboxMemoryLayout {
                 &format_args!("{:#x}", self.guest_user_stack_buffer_offset),
             )
             .field(
+                "Init Data Offset",
+                &format_args!("{:#x}", self.init_data_offset),
+            )
+            .field(
                 "Page Table Size",
                 &format_args!("{:#x}", self.total_page_table_size),
             )
@@ -231,6 +249,8 @@ impl SandboxMemoryLayout {
         code_size: usize,
         stack_size: usize,
         heap_size: usize,
+        init_data_size: usize,
+        init_data_permissions: Option<MemoryRegionFlags>,
     ) -> Result<Self> {
         let total_page_table_size =
             Self::get_total_page_table_size(cfg, code_size, stack_size, heap_size);
@@ -275,6 +295,7 @@ impl SandboxMemoryLayout {
         let guest_user_stack_buffer_offset = guard_page_offset + PAGE_SIZE_USIZE;
         // round up stack size to page size. This is needed for MemoryRegion
         let stack_size_rounded = round_up_to(stack_size, PAGE_SIZE_USIZE);
+        let init_data_offset = guest_user_stack_buffer_offset + stack_size_rounded;
 
         Ok(Self {
             peb_offset,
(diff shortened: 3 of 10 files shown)


src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -123,15 +124,98 @@ impl
     }
 }
 
-/// A `GuestBinary` is either a buffer containing the binary or a path to the binary
-#[derive(Debug)]
-pub enum GuestBinary<'a> {
```
ludfjig, line 128: I'm not a huge fan of this rename, as consumers would now be loading GuestBlob  instead of GuestBinary, which I think is incorrect/confusing when it's actually an ELF binary

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -142,31 +226,22 @@ impl UninitializedSandbox {
     /// The err attribute is used to emit an error should the Result be an error, it uses the std::`fmt::Debug trait` to print the error.
     #[instrument(
         err(Debug),
-        skip(guest_binary),
+        skip(env),
         parent = Span::current()
     )]
-    pub fn new(guest_binary: GuestBinary, cfg: Option<SandboxConfiguration>) -> Result<Self> {
+    pub fn new<'a>(
```
ludfjig: We might want to update the docs for this func as well

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -123,15 +124,98 @@ impl
...
+    /// A buffer containing the guest blob
     Buffer(&'a [u8]),
-    /// A path to the guest binary
+    /// A path to the guest blob
     FilePath(String),
 }
 
+impl<'a> GuestBlob<'a> {
+    /// Validates the `GuestBlob`, ensuring that if it is a file path, it exists and can be converted to a `String`.
+    pub fn validated(&'a self) -> Result<Cow<'a, GuestBlob<'a>>> {
```
ludfjig: There will always be a delay from when you call this, to when we load the file, so maybe we don't need this function?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -123,15 +124,98 @@ impl
...
+    fn from(blob: GuestBlob<'a>) -> Self {
+        Cow::Owned(blob)
+    }
+}
+
+/// A `GuestEnvironment` is a structure that contains the guest binary and an optional guest blob.
+#[derive(Debug, Clone)]
+pub struct GuestEnvironment<'a> {
+    /// The guest binary, which can be a file path or a buffer.
+    pub guest_binary: Cow<'a, GuestBlob<'a>>,
```
ludfjig: Do we need all the Cow[s] introduced by this PR?

jprendes:
IIRC, @simongdavies had an idea where the host could share host memory with the VM by adding pages to the VM's page table. Would something like that help in this case?

danbugs:
> The PR comment states that the this is a non-breaking change, but the rename GuestBinary -> GuestBlob is breaking, right? Also I do think the renaming is not intuitive for users who want to load a guest binary, when it's now called GuestBlob

On the breaking-change point, you're right. Guess what I meant to say is that you don't necessarily have to pass a `GuestEnvironment` to create a new uninit sandbox because you can still create it off a binary alone. 

> Not strictly related to your PR, but I think we should consider not allowing passing blobs/binaries by "path", which would force the consumers to do their own reading of files, etc, and instead only take slices of u8.

If so, I can change `GuestBlob` to be just bytes and revert the re-name of the GuestBinary. I opted to re-name because they were the same and I felt like `GuestBlob` also encompassed a binary.

> I haven't fully thought about it, but maybe Uninitialized::new should take a trait instead?

It does currently take a trait: `env: impl Into<GuestEnvironment<'a>>,`

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Reverted the rename ✅

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Checked the docs, and I think we're good. Let me know if I missed anything 👍

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: I reverted the addition of this method—but the same logic exists in the body of `UninitializedSandbox::new(...)` from before this PR.

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Removed the `Cow`s. 'Twas just an optimization I figured we could add in the same vein of #560 .

ludfjig:
> > I haven't fully thought about it, but maybe Uninitialized::new should take a trait instead?
> 
> It does currently take a trait: `env: impl Into<GuestEnvironment<'a>>,`

What do you think about letting `Uninitialized::new` take something like a `&impl GuestEnvironment`

```rust 
pub trait GuestEnvironment {
    /// Get the guest binary
    fn guest_binary(&self) -> &[u8];

    /// Get optional memory blobs
    fn blobs(&self) -> Option<&[&[u8]]>;
}
```

Then we could implement it for exisiting GuestBinary, and also a new GuestBinaryWithBlob, etc, etc.

danbugs:
> > > I haven't fully thought about it, but maybe Uninitialized::new should take a trait instead?
> > 
> > 
> > It does currently take a trait: `env: impl Into<GuestEnvironment<'a>>,`
> 
> What do you think about letting `Uninitialized::new` take something like a `&impl GuestEnvironment`
> 
> ```rust
> pub trait GuestEnvironment {
>     /// Get the guest binary as a `GuestBinary`.
>     fn guest_binary(&self) -> GuestBinary;
> 
>     /// Get optional memory blobs
>     fn blob(&self) -> Option<&[&[u8]]>;
> }
> ```
> 
> Then we could implement it for exisiting GuestBinary, and also a new GuestBinaryWithBlob, etc, etc.

I think I like `impl Into<GuestEnvironment<'a>>` a little better so we don't have to write empty implementations (e.g., `fn blob` for `GuestBinary`) and grow the trait and its deps if we want to support extra items in a guest environment, but I'm not strongly opinionated here. Open to change if you think this has other pros I might be missing 🙂

ludfjig:
> > > > I haven't fully thought about it, but maybe Uninitialized::new should take a trait instead?
> > > 
> > > 
> > > It does currently take a trait: `env: impl Into<GuestEnvironment<'a>>,`
> > 
> > 
> > What do you think about letting `Uninitialized::new` take something like a `&impl GuestEnvironment`
> > ```rust
> > pub trait GuestEnvironment {
> >     /// Get the guest binary as a `GuestBinary`.
> >     fn guest_binary(&self) -> GuestBinary;
> > 
> >     /// Get optional memory blobs
> >     fn blob(&self) -> Option<&[&[u8]]>;
> > }
> > ```
> > Then we could implement it for exisiting GuestBinary, and also a new GuestBinaryWithBlob, etc, etc.
> 
> I think I like `impl Into<GuestEnvironment<'a>>` a little better so we don't have to write empty implementations (e.g., `fn blob` for `GuestBinary`) and grow the trait and its deps if we want to support extra items in a guest environment, but I'm not strongly opinionated here. Open to change if you think this has other pros I might be missing 🙂


@simongdavies @jprendes @dblnz @andreiltd @syntactically, any opinions on the API of this PR?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -31,6 +31,8 @@ use crate::sandbox::SandboxConfiguration;
 use crate::{Result, new_error};
 
 // +-------------------------------------------+
+// |              User Memory                  | (get_total_memory - size_a)
```
simongdavies: Do you think we should call this Init Data? I think its clearer , we should probably use the same term all the way through from the GuestEnvironment enum to here?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -682,12 +695,34 @@ impl SandboxMemoryLayout {
...
+                user_memory_offset
+            ));
+        }
+
+        let user_region_size = self.get_memory_size()? - user_memory_offset;
+
+        let final_offset = if user_region_size > 0 {
+            builder.push_page_aligned(
+                user_region_size,
+                MemoryRegionFlags::READ | MemoryRegionFlags::WRITE | MemoryRegionFlags::EXECUTE,
```
simongdavies: Do you think we should allow the user to specify what flags they want when they pass this data? We could default to the common case not sure if Nanvix wants or needs it to be executable?

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -262,6 +265,16 @@ impl SandboxConfiguration {
         (self.heap_size_override > 0).then_some(self.heap_size_override)
     }
 
+    /// Set the total memory size.
```
simongdavies: I'm not clear on why we need to have the user tell us what the size is? Isn't it the sum of the previous calculation plus an aligned sized buffer to fit the blob? What happens if they get the size wrong (i.e. too small?)

simongdavies (commented):
Looks good just a few questions

simongdavies:
> IIRC, @simongdavies had an idea where the host could share host memory with the VM by adding pages to the VM's page table. Would something like that help in this case?

I don't recall what that might have been, but I think if @danbugs takes up my feedback about allowing the data to have RWX flags associated with it then if its R we could do an optimisation where its only allocated in the host once, that would require us to change the way we bounds check though so I think out of scope here.

We also talked about allowing a user to pass an vector of (address,len) into a sandbox for mapping, that may be a better way to do something like this, and it may fit better in a configuration than as a argument to the new function

simongdavies:
> Not strictly related to your PR, but I think we should consider not allowing passing blobs/binaries by "path", which would force the consumers to do their own reading of files, etc, and instead only take slices of u8.

I think this all stemmed from us having in-proc and more specifically loadlib support in Windows, I think you are correct, I don't think we need it any longer ,not sure if now is the right time though...

simongdavies:
> any opinions on the API of this PR?

I quite like it as it is but not really that fussed either way, eventually we should bury all this behind a builder, that might be the right time to get rid of the paths

src/hyperlight_host/src/mem/layout.rs
danbugs: Changed the mem region name to Init Data and also changed the previous mentions to mirror that ✅

src/hyperlight_host/src/mem/layout.rs
danbugs: Added memory flag options according to our discussed design ✅

src/hyperlight_host/src/sandbox/config.rs
danbugs: Removed this configuration option. I added this as a convenience thing for future work, but I agree the interaction w/ the init data region is a bit clunky because of it ✅

danbugs:
> I don't recall what that might have been, but I think if @danbugs takes up my feedback about allowing the data to have RWX flags associated with it then if its R we could do an optimisation where its only allocated in the host once, that would require us to change the way we bounds check though so I think out of scope here.

Added configurable flags to the new mem region 👍

danbugs:
> I think this all stemmed from us having in-proc and more specifically loadlib support in Windows, I think you are correct, I don't think we need it any longer ,not sure if now is the right time though...

https://github.com/hyperlight-dev/hyperlight/issues/614 ~ made an issue for this, @ludfjig / @simongdavies

src/hyperlight_guest/src/guest_handle/host_comm.rs
```diff
@@ -34,6 +34,29 @@ use crate::error::{HyperlightGuestError, Result};
 use crate::exit::out32;
 
 impl GuestHandle {
+    /// Get user memory region as bytes.
+    pub fn read_n_bytes_from_user_memory(&self, num: u64) -> Result<Vec<u8>> {
```
ludfjig, line 38: Have you considered whether a user only wants the slice, and not copy it to vec?

src/hyperlight_guest_bin/src/host_comm.rs
```diff
@@ -65,6 +65,11 @@ pub fn get_host_function_details() -> HostFunctionDetails {
     handle.get_host_function_details()
 }
 
+pub fn read_n_bytes_from_user_memory(num: u64) -> Result<Vec<u8>> {
```
ludfjig, line 68: Is there a reason this func exist both on a handle, and here? They are both `pub`

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -60,6 +64,32 @@ bitflags! {
     }
 }
 
+impl MemoryRegionFlags {
+    pub(crate) fn translate_flags(&self) -> u64 {
+        let mut page_flags = 0;
+
+        if self.contains(MemoryRegionFlags::READ) {
+            page_flags |= PAGE_PRESENT; // Mark page as present
```
ludfjig: Why does READ correspond to PAGE_PRESENT? Maybe it would be better to unconditionally include PAGE_PRESENT

ludfjig (approved):

src/hyperlight_guest_bin/src/host_comm.rs
danbugs: The guest library defines functions that can be called on a handle. The guest_bin has a global handle we can call functions on. It's the same for all other functions too 👍

src/hyperlight_guest/src/guest_handle/host_comm.rs
danbugs: No. To be honest, I added this function more so to test the functionality. I think we can improve it once we get someone using it and they have different requirements. This is not used by Nanvix.

src/hyperlight_host/src/mem/memory_region.rs
danbugs: I made this mapping based off this comment:
```
pub(crate) const PAGE_RW: u64 = 1 << 1; // Page is Read/Write (if not set page is read only so long as the WP bit in CR0 is set to 1 - which it is in Hyperlight)
```

If we only have `PAGE_PRESENT`, that equates to read permissions, so it felt appropriate. But, yeah, I guess it is a bit misleading. Will change.

danbugs: Done ✅

ludfjig (approved):

PR #624, opened 2025-06-13
make socket component example work

jprendes:
This PR removes the `Clone` and `PartialEq` requirement from `variant` and `record` wit types.
This change makes the `hyperlight-wasm-sockets-example` example work.

### DIFF
--- a/src/hyperlight_component_util/src/rtypes.rs
+++ b/src/hyperlight_component_util/src/rtypes.rs
@@ -386,7 +386,7 @@ fn emit_value_toplevel(s: &mut State, v: Option<u32>, id: Ident, vt: &Value) -> 
             };
             quote! {
                 #derives
-                #[derive(Debug, Clone, PartialEq)]
+                #[derive(Debug)]
                 pub struct #id #vs { #toks }
             }
         }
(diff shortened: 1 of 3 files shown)


jprendes:
It's worth noting that the branch that hyperlight-wasm was using until now didn't derive those traits either:
https://github.com/hyperlight-dev/hyperlight/blob/hyperlight-wasm-macro/src/hyperlight_component_util/src/rtypes.rs#L371-L384

jsturtevant:
> It's worth noting that the branch that hyperlight-wasm was using until now didn't derive those traits either: https://github.com/hyperlight-dev/hyperlight/blob/hyperlight-wasm-macro/src/hyperlight_component_util/src/rtypes.rs#L371-L384

This seems reasonable to me.  We may want to look at how we handle testing variants with resources in them in the future to cover our bases.  This is the wit that generated this issue:

```
package wasi:io@0.2.3 {
  interface error {
    resource error;
  }
  interface poll {
    resource pollable {
      block: func();
    }
  }
  interface streams {
    use error.{error};

    resource output-stream {
      check-write: func() -> result<u64, stream-error>;
      write: func(contents: list<u8>) -> result<_, stream-error>;
      blocking-write-and-flush: func(contents: list<u8>) -> result<_, stream-error>;
      blocking-flush: func() -> result<_, stream-error>;
    }

    variant stream-error {
      last-operation-failed(error),
      closed,
    }

    resource input-stream;
  }
}
```

jsturtevant (approved):

ludfjig (approved):

PR #636, opened 2025-06-17
Allow hyperlight_guest_bin to compile with windows target.

ludfjig:
`.type x,@function` directives are unsupported on windows toolchain and assembler.

Patch similar to this one has been done before https://github.com/hyperlight-dev/hyperlight/pull/104/commits/42ee008bb0df9f4ce5d380f1eafa79e61a2233c8, but was not applied when setjmp was brought back in https://github.com/hyperlight-dev/hyperlight/pull/499

This was never caught before because hyperlight_guest_bin is used with a windows toolchain. I only noticed it because rust-analyzer complains when I use my windows machine.


simongdavies:
I just updated #572 so that running `cargo build` in the root directory of `hyperlight_guest_bin` works on windows by adding a cargo.toml file

jsturtevant:
looks like this was resolved in https://github.com/hyperlight-dev/hyperlight/pull/572

can we close this?

jsturtevant:
#572 went in

PR #637, opened 2025-06-17
Let windows decide at which address to map shared memory in surrogate process

### LINKED ISSUE #579: CI Instability: MapViewOfFileNuma2 sporadic errors
CI often fails benchmarks on windows, with error

 Error("MapViewOfFileNuma2 failed with error code: WIN32_ERROR(487) for mem address 0x21ccd6e0000 ")

It's not clear what is causing it, perhaps OOM? Not sure, but we should fix it

ludfjig:
Previously we assumed that the host process's shared memory address is available, which might not always be the case. I believe this is what caused the CI to fail sporadically.

Also removes some fields relating to size/addresses that we don't need

closes #579

### DIFF
--- a/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
+++ b/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
@@ -144,7 +144,6 @@ impl SurrogateProcessManager {
     pub(super) fn get_surrogate_process(
         &self,
         raw_size: usize,
-        raw_source_address: *const c_void,
         mmap_file_handle: HandleWrapper,
     ) -> Result<SurrogateProcess> {
         let surrogate_process_handle: HANDLE = self.process_receiver.recv()?.into();
@@ -160,7 +159,7 @@ impl SurrogateProcessManager {
                 mapping_file_handle,
                 surrogate_process_handle,
                 0,
-                Some(raw_source_address),
+                None,
                 raw_size,
                 0,
                 PAGE_READWRITE.0,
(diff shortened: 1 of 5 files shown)


src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -160,7 +159,7 @@ impl SurrogateProcessManager {
                 mapping_file_handle,
                 surrogate_process_handle,
                 0,
-                Some(raw_source_address),
+                None,
```
ludfjig, line 162: Probably change raw_size below here to 0

ludfjig: Probably not thinking more about it... We use it to set guard pages, so we should probably use it here explicitly anyway.

src/hyperlight_host/src/hypervisor/windows_hypervisor_platform.rs
```diff
@@ -89,9 +90,28 @@ impl VMPartition {
...
+        let surrogate_address = surrogate_process.allocated_address as usize + PAGE_SIZE_USIZE;
+        if regions.is_empty() {
+            return Err(new_error!("No memory regions to map"));
+        }
+        // this is the address in the main process where the shared memory starts
+        let host_address = regions[0].host_region.start;
+
+        let surrogate_address_bigger;
+        // offset between the surrogate process and the host process address of start of shared memory
+        let offset = if surrogate_address > host_address {
```
simongdavies: Can you just create a signed result and then add it below? i.e. the offset can be negative or positive?

ludfjig: I did it this way to be extra cautious because I think it's possible the result won't fit in an `isize`

ludfjig: fixed since should fit in isize

dblnz (approved):

PR #639, opened 2025-06-17
[host] adds init-paging feature

danbugs:
This is an extra feature flag that allows disabling our paging setup (when removed). This is needed to further enable custom guests and should not impact any of our current workloads. init-paging is on by default.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -75,8 +75,8 @@ test-like-ci config=default-target hypervisor="kvm":
     @# with default features
     just test {{config}} {{ if hypervisor == "mshv3" {"mshv3"} else {""} }}
 
-    @# with only one driver enabled + seccomp
-    just test {{config}} seccomp,build-metadata,{{ if hypervisor == "mshv" {"mshv2"} else if hypervisor == "mshv3" {"mshv3"} else {"kvm"} }}
+    @# with only one driver enabled + seccomp + build-metadata + init-paging
+    just test {{config}} seccomp,build-metadata,init-paging,{{ if hypervisor == "mshv" {"mshv2"} else if hypervisor == "mshv3" {"mshv3"} else {"kvm"} }}
 
     @# make sure certain cargo features compile
     cargo check -p hyperlight-host --features crashdump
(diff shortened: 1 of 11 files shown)


ludfjig:
Nice! Have you considered a runtime config vs cargo feature?

Justfile
```diff
@@ -119,7 +119,7 @@ test-integration guest target=default-target features="":
 test-seccomp target=default-target features="":
     @# run seccomp test with feature "seccomp" on and off
     cargo test --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host test_violate_seccomp_filters --lib {{ if features =="" {''} else { "--features " + features } }} -- --ignored
-    cargo test --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host test_violate_seccomp_filters --no-default-features {{ if features =~"mshv3" {"--features mshv3"} else {"--features mshv2,kvm" } }} --lib -- --ignored
+    cargo test --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host test_violate_seccomp_filters --no-default-features {{ if features =~"mshv3" {"--features mshv3"} else {"--features mshv2,init-paging,kvm" } }} --lib -- --ignored
```
simongdavies:
```suggestion
    cargo test --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host test_violate_seccomp_filters --no-default-features {{ if features =~"mshv3" {"--features mshv3,init-paging"} else {"--features mshv2,init-paging,kvm" } }} --lib -- --ignored
```

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -436,11 +436,12 @@ impl HypervLinuxDriver {
     }
 
     #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
-    fn setup_initial_sregs(vcpu: &mut VcpuFd, pml4_addr: u64) -> Result<()> {
+    fn setup_initial_sregs(vcpu: &mut VcpuFd, _pml4_addr: u64) -> Result<()> {
```
simongdavies, line 439: why the change to parameter name is it because the param is unused in the case where init-paging is not enabled, could we have 2 different funcs in that case , i think it would be cleaner? That way you can also comment about what the function does/why so that then documents the feature

danbugs:
> Nice! Have you considered a runtime config vs cargo feature?

Yes, but for this use-case I think a cargo feature is preferred. Users that are disabling paging will most probably be wanting run a specific type of binary (e.g., Nanvix) where you either want paging or not, so I don't think they'll want to mix-and-match different configs there at runtime. Plus, having to bubble that config option all the way to `setup_initial_sregs` would be a bit messy, imo.

ludfjig:
> > Nice! Have you considered a runtime config vs cargo feature?
> 
> Yes, but for this use-case I think a cargo feature is preferred. Users that are disabling paging will most probably be wanting run a specific type of binary (e.g., Nanvix) where you either want paging or not, so I don't think they'll want to mix-and-match different configs there at runtime. Plus, having to bubble that config option all the way to `setup_initial_sregs` would be a bit messy, imo.

I think I personally would prefer a runtime config (opt out), to avoid a large feature matrix.

danbugs:
> > > Nice! Have you considered a runtime config vs cargo feature?
> > 
> > 
> > Yes, but for this use-case I think a cargo feature is preferred. Users that are disabling paging will most probably be wanting run a specific type of binary (e.g., Nanvix) where you either want paging or not, so I don't think they'll want to mix-and-match different configs there at runtime. Plus, having to bubble that config option all the way to `setup_initial_sregs` would be a bit messy, imo.
> 
> I think I personally would prefer a runtime config (opt out), to avoid a large feature matrix.

But then you'd just have a larger runtime config matrix? Not sure if there's an 100% right option here. Are you ok w/ proceeding w/ a feature at first? 

cc: @simongdavies ?*

Justfile
danbugs: Fixed ✅

simongdavies:
> > > Nice! Have you considered a runtime config vs cargo feature?
> > 
> > 
> > Yes, but for this use-case I think a cargo feature is preferred. Users that are disabling paging will most probably be wanting run a specific type of binary (e.g., Nanvix) where you either want paging or not, so I don't think they'll want to mix-and-match different configs there at runtime. Plus, having to bubble that config option all the way to `setup_initial_sregs` would be a bit messy, imo.
> 
> I think I personally would prefer a runtime config (opt out), to avoid a large feature matrix.

I don't think it makes sense to have this as a runtime config, since this is something that's either required statically or not, I understand the concern about proliferation of features but this one really should be a feature not a configuration option IMO

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
danbugs: W/ changing this, the `pml4_addr` ends up being unused in the `new` function, so I think I'd opt it to leave it as is rn and maybe improve the API later as part of the VM abstraction refactor.

simongdavies (approved):

simongdavies (dismissed):

PR #672, opened 2025-06-27
Clippy updates

### LINKED ISSUE #549: Review `just clippy` command
When we run just clippy we run the command  cargo clippy --all-targets --all-features --profile=dev -- -D warnings which results in the equivalent of --features "fuzzing,gdb,kvm,mshv" being passed to cargo clippy.

This means that certain issues can be missed (see https://github.com/hyperlight-dev/hyperlight/pull/548) when running clippy.

marosset:
Fixes #549 

WIP - I plan to move the clippy and fmt check out of `dep_rust` so we can a) run the in parallel, and b) not need to run them for each arch/hypervisor/target matrix. Right now I just want to test that we've caught all the clippy issues

### DIFF
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -31,6 +31,49 @@ defaults:
     shell: bash
 
 jobs:
+  code-checks:
+    if: ${{ inputs.docs_only == 'false' }}
+    timeout-minutes: 60
+    strategy:
+      fail-fast: true
+      matrix:
+        hypervisor: ['hyperv-ws2025', kvm]
+        config: [debug, release]
+    runs-on: ${{ fromJson(
+      format('["self-hosted", "{0}", "X64", "1ES.Pool=hld-{1}-amd"]', 
+        (matrix.hypervisor == 'hyperv-ws2025') && 'Windows' || 'Linux', 
+        matrix.hypervisor == 'hyperv-ws2025' && 'win2025' || 'kvm')) }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: hyperlight-dev/ci-setup-workflow@v1.5.0
+        with:
+          rust-toolchain: "1.85" 
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # Does not check for updated Cargo.lock files for test rust guests as this causes an issue with this checkwhen deoendabot updates dependencies in common crates
+      - name: Ensure up-to-date Cargo.lock
+        run: |
+          cargo fetch --locked
+
+      - name: fmt
+        run: just fmt-check
+
+      - name: clippy
+        if: ${{ (runner.os == 'Windows' )}}
+        run: |
+          just clippy ${{ matrix.config }}
+          just clippy-guests ${{ matrix.config }}
+
+      - name: clippy exhaustive check
+        if: ${{ (runner.os == 'Linux' )}}
+        run: |
+          just clippy-exhaustive ${{ matrix.config }}
+
+      - name: Verify MSRV
+        run: ./dev/verify-msrv.sh hyperlight-host hyperlight-guest hyperlight-guest-bin hyperlight-common
+
   build:
     if: ${{ inputs.docs_only == 'false' }}
     timeout-minutes: 60
@@ -61,19 +104,6 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: fmt
-        run: just fmt-check
-
-      - name: clippy
-        run: |
-          just clippy ${{ matrix.config }}
-          just clippy-guests ${{ matrix.config }}
-
-      # Does not check for updated Cargo.lock files for test rust guests as this causes an issue with this checkwhen deoendabot updates dependencies in common crates
-      - name: Ensure up-to-date Cargo.lock
-        run: |
-          cargo fetch --locked
-
       - name: Get gh action service name
         if: ${{ (runner.os == 'Windows' )}}
         run: (Get-Service actions.runner.*) | Foreach { $_.Name, $_.UserName, $_.ServiceType }
@@ -92,9 +122,6 @@ jobs:
           just move-c-guests ${{ matrix.config }}
       - name: Build
         run: just build ${{ matrix.config }}
-
-      - name: Verify MSRV
-        run: ./dev/verify-msrv.sh hyperlight-host hyperlight-guest hyperlight-guest-bin hyperlight-common
 
       - name: Run Rust tests
         env:
--- a/hack/clippy-package-features.sh
+++ b/hack/clippy-package-features.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+# Check for required arguments
+if [[ $# -lt 2 ]]; then
+    echo "Usage: $0 <package> <target>" >&2
+    echo "Example: $0 hyperlight-host debug" >&2
+    exit 1
+fi
+
+PACKAGE="$1"
+TARGET="$2"
+
+# Convert target for cargo profile
+PROFILE=$([ "$TARGET" = "debug" ] && echo "dev" || echo "$TARGET")
+
+# Required features needed so the rust packages can compile
+if [[ "$PACKAGE" == "hyperlight-host" ]]; then
+    REQUIRED_FEATURES=("kvm" "mshv3")
+elif [[ "$PACKAGE" == "hyperlight-guest-bin" ]]; then
+    REQUIRED_FEATURES=("printf")
+else 
+    REQUIRED_FEATURES=()
+fi
+
+# Get all features for the package (excluding default and required features)
+# Always exclude "default", and exclude any required features using jq
+features=$(cargo metadata --format-version 1 --no-deps | jq -r --arg pkg "$PACKAGE" '.packages[] | select(.name == $pkg) | .features | keys[] | select(. != "default" and (IN($ARGS.positional[])|not))' --args "${REQUIRED_FEATURES[@]}" || true)
+
+# Convert required features array to comma-separated string for cargo
+if [[ ${#REQUIRED_FEATURES[@]} -gt 0 ]]; then
+    required_features_str=$(IFS=,; echo "${REQUIRED_FEATURES[*]}")
+else
+    required_features_str=""
+fi
+
+# Test with minimal features
+if [[ ${#REQUIRED_FEATURES[@]} -gt 0 ]]; then
+    echo "Testing $PACKAGE with required features only ($required_features_str)..."
+    (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$required_features_str" --profile="$PROFILE" -- -D warnings)
+else
+    echo "Testing $PACKAGE with no features..."
+    (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --profile="$PROFILE" -- -D warnings)
+fi
+
+echo "Testing $PACKAGE with default features..."
+(set -x; cargo clippy -p "$PACKAGE" --all-targets --profile="$PROFILE" -- -D warnings)
+
+# Test each additional feature individually
+for feature in $features; do
+    if [[ ${#REQUIRED_FEATURES[@]} -gt 0 ]]; then
+        echo "Testing $PACKAGE with feature: $required_features_str,$feature"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$required_features_str,$feature" --profile="$PROFILE" -- -D warnings)
+    else
+        echo "Testing $PACKAGE with feature: $feature"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$feature" --profile="$PROFILE" -- -D warnings)
+    fi
+done
+
+# Test all features together
+if [[ -n "$features" ]]; then
+    all_features=$(echo $features | tr '\n' ',' | sed 's/,$//')
+    if [[ ${#REQUIRED_FEATURES[@]} -gt 0 ]]; then
+        echo "Testing $PACKAGE with all features: $required_features_str,$all_features"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$required_features_str,$all_features" --profile="$PROFILE" -- -D warnings)
+    else
+        echo "Testing $PACKAGE with all features: $all_features"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$all_features" --profile="$PROFILE" -- -D warnings)
+    fi
+fi
--- a/src/hyperlight_guest/src/guest_handle/handle.rs
+++ b/src/hyperlight_guest/src/guest_handle/handle.rs
@@ -25,7 +25,7 @@ use hyperlight_common::mem::HyperlightPEB;
 ///
 /// Guests are expected to initialize this and store it. For example, you
 /// could store it in a global variable.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, Default)]
 pub struct GuestHandle {
     peb: Option<*mut HyperlightPEB>,
 }
--- a/src/hyperlight_guest_bin/src/exceptions/handler.rs
+++ b/src/hyperlight_guest_bin/src/exceptions/handler.rs
@@ -21,8 +21,6 @@ use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
 use hyperlight_common::outb::Exception;
 use hyperlight_guest::exit::abort_with_code_and_message;
 
-use crate::paging;
-
 /// See AMD64 Architecture Programmer's Manual, Volume 2
 ///     §8.9.3 Interrupt Stack Frame, pp. 283--284
 ///       Figure 8-14: Long-Mode Stack After Interrupt---Same Privilege,
@@ -56,9 +54,9 @@ const _: () = assert!(size_of::<Context>() == 152 + 512);
 
 // TODO: This will eventually need to end up in a per-thread context,
 // when there are threads.
-pub static handlers: [core::sync::atomic::AtomicU64; 31] =
+pub static HANDLERS: [core::sync::atomic::AtomicU64; 31] =
     [const { core::sync::atomic::AtomicU64::new(0) }; 31];
-type handler_t = fn(n: u64, info: *mut ExceptionInfo, ctx: *mut Context, pf_addr: u64) -> bool;
+pub type HandlerT = fn(n: u64, info: *mut ExceptionInfo, ctx: *mut Context, pf_addr: u64) -> bool;
 
 /// Exception handler
 #[unsafe(no_mangle)]
@@ -89,15 +87,12 @@ pub extern "C" fn hl_exception_handler(
     // vectors (0-31)
     if exception_number < 31 {
         let handler =
-            handlers[exception_number as usize].load(core::sync::atomic::Ordering::Acquire);
+            HANDLERS[exception_number as usize].load(core::sync::atomic::Ordering::Acquire);
         if handler != 0
             && unsafe {
-                core::mem::transmute::<_, handler_t>(handler)(
-                    exception_number,
-                    exn_info,
-                    ctx,
-                    page_fault_address,
-                )
+                core::mem::transmute::<u64, fn(u64, *mut ExceptionInfo, *mut Context, u64) -> bool>(
+                    handler,
+                )(exception_number, exn_info, ctx, page_fault_address)
             }
         {
             return;
--- a/src/hyperlight_guest_bin/src/paging.rs
+++ b/src/hyperlight_guest_bin/src/paging.rs
@@ -53,8 +53,16 @@ struct MapResponse {
 }
 
 /// Assumption: all are page-aligned
+/// # Safety
+/// This function modifies pages backing a virtual memory range which is inherently unsafe w.r.t.
+/// the Rust memory model.
+/// When using this function note:
+/// - No locking is performed before touching page table data structures,
+///   as such do not use concurrently with any other page table operations
+/// - TLB invalidation is not performed,
+///   if previously-unmapped ranges are not being mapped, TLB invalidation may need to be performed afterwards.
 pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
-    let mut pml4_base: u64 = 0;
+    let mut pml4_base: u64;
     unsafe {
         asm!("mov {}, cr3", out(reg) pml4_base);
     }
@@ -71,12 +79,18 @@ pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
     .map(|r| unsafe { alloc_pte_if_needed(r) })
     .flat_map(modify_ptes::<20, 12>)
     .map(|r| map_normal(phys_base, virt_base, r))
-    .collect::<()>();
+    .for_each(drop);
 }
 
 #[allow(unused)]
 /// This function is not presently used for anything, but is useful
 /// for debugging
+/// # Safety
+/// This function traverses page table data structures, and should not be called concurrently
+/// with any other operations that modify the page table.
+/// # Panics
+/// This function will panic if:
+/// - A page map request resolves to multiple page table entries
 pub unsafe fn dbg_print_address_pte(address: u64) -> u64 {
     let mut pml4_base: u64 = 0;
     unsafe {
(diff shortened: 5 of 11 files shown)


hack/clippy-package-features.sh
```diff
@@ -0,0 +1,66 @@
...
+TARGET="$2"
+
+# Convert target for cargo profile
+PROFILE=$([ "$TARGET" = "debug" ] && echo "dev" || echo "$TARGET")
+
+# Required features for compilation (only for hyperlight-host)
+if [[ "$PACKAGE" == "hyperlight-host" ]]; then
+    REQUIRED_FEATURES="kvm,mshv3"
+    # Get all features for the package (excluding default and required features)
+    features=$(cargo metadata --format-version 1 --no-deps | jq -r ".packages[] | select(.name == \"$PACKAGE\") | .features | keys[]" | grep -v -E "^(default|kvm|mshv3)$" || true)
```
syntactically: Might want to use a jq variable for $PACKAGE to avoid any escaping issues

syntactically: Also, we could perhaps do the filtering in jq as well, and pass in the required_features so that we don't need to duplicate them? (Or add a comment mentioning that those probably need to be updated in lockstep?)

src/hyperlight_guest/src/guest_handle/handle.rs
```diff
@@ -30,6 +30,12 @@ pub struct GuestHandle {
     peb: Option<*mut HyperlightPEB>,
 }
 
+impl Default for GuestHandle {
```
syntactically: Did clippy ask for this impl? Not sure if it is being used anywhere...

marosset: it did

ludfjig: Could it be derived instead, and have ::new return default?

marosset: let me see, i just did what clippy suggested :D

marosset: updated

hack/clippy-package-features.sh
marosset: updated! thanks for the suggestions

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,72 @@
...
+elif [[ "$PACKAGE" == "hyperlight-guest-bin" ]]; then
+    REQUIRED_FEATURES="printf"
+else 
+    REQUIRED_FEATURES=""
+fi
+
+# Build grep exclusion pattern
+if [[ -n "$REQUIRED_FEATURES" ]]; then
+    # Convert comma-separated features to grep pattern: "kvm,mshv3" -> "^(default|kvm|mshv3)$"
+    required_pattern=$(echo "$REQUIRED_FEATURES" | sed 's/,/|/g')
```
syntactically: If you do keep this, since this is bash, you can do `sed 's/,/|/g' <<< "$REQUIRED_FEATURES"`. But you also might want to use `\n` as a delimiter, because then you can use `grep -xF` and not worry about escaping regexp meta-characters.

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,72 @@
...
+    grep_pattern="^(default|${required_pattern})$"
+else
+    grep_pattern="^default$"
+fi
+
+# Get all features for the package (excluding default and required features)
+features=$(cargo metadata --format-version 1 --no-deps | jq -r --arg pkg "$PACKAGE" '.packages[] | select(.name == $pkg) | .features | keys[]' | grep -v -E "$grep_pattern" || true)
```
syntactically: Instead of using a grep pattern like this, you can do it all in jq without worrying about escaping. With the array representation for `REQUIRED_FEATURES` suggested above, that's something like `jq -r --arg pkg "$PACKAGE" '.packages[] | select(.name == $pkg) | .features | keys[] | select(IN($ARGS.positional[])|not)' --args "${REQUIRED_FEATURES[@]}"`

You could also keep the features structured with something like `readarray -d $'\0' features < <(cargo metadata --format-version 1 --no-deps | jq --raw-output0 --arg pkg "$PACKAGE" '.packages[] | select(.name == $pkg) | .features | keys[] | select(IN($ARGS.positional[])|not)' --args "${REQUIRED_FEATURES[@]}")`.

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,72 @@
...
+
+PACKAGE="$1"
+TARGET="$2"
+
+# Convert target for cargo profile
+PROFILE=$([ "$TARGET" = "debug" ] && echo "dev" || echo "$TARGET")
+
+# Required features needed so the rust packages can compile
+if [[ "$PACKAGE" == "hyperlight-host" ]]; then
+    REQUIRED_FEATURES="kvm,mshv3"
```
syntactically: An array might be cleaner than having to reserve a character for delimiter?

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,72 @@
...
+fi
+
+echo "Testing $PACKAGE with default features..."
+(set -x; cargo clippy -p "$PACKAGE" --all-targets --profile="$PROFILE" -- -D warnings)
+
+# Test each additional feature individually
+for feature in $features; do
+    if [[ -n "$REQUIRED_FEATURES" ]]; then
+        echo "Testing $PACKAGE with feature: $REQUIRED_FEATURES,$feature"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$REQUIRED_FEATURES,$feature" --profile="$PROFILE" -- -D warnings)
```
syntactically: If you do use an array for REQUIRED_FEATURES, you can recover the comma-separated representation with `(IFS=','; echo -n "${REQUIRED_FEATURES[*]}"` (of course, unlike all the other escaping-avoiding tricks, this has the obvious caveat with commas in the input)

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,72 @@
...
+else
+    echo "Testing $PACKAGE with no features..."
+    (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --profile="$PROFILE" -- -D warnings)
+fi
+
+echo "Testing $PACKAGE with default features..."
+(set -x; cargo clippy -p "$PACKAGE" --all-targets --profile="$PROFILE" -- -D warnings)
+
+# Test each additional feature individually
+for feature in $features; do
```
syntactically, line 51: If you do have features as an array, you can iterate through it without worrying about escaping with `for feature in "${features[@]}"`

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,72 @@
...
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$REQUIRED_FEATURES,$feature" --profile="$PROFILE" -- -D warnings)
+    else
+        echo "Testing $PACKAGE with feature: $feature"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$feature" --profile="$PROFILE" -- -D warnings)
+    fi
+done
+
+# Test all features together
+if [[ -n "$features" ]]; then
+    all_features=$(echo $features | tr '\n' ',' | sed 's/,$//')
```
syntactically, line 63: With the array representation, this would be easiest to do with `features+=("${REQUIRED_FEATURES[@]}")`, but `all_features=("${REQUIRED_FEATURES[@]}" "${features[@]}")` also works

syntactically (commented):
This looks good overall. I have left inline a bunch of opinionated tips for defensive shell coding, which mostly don't matter here due to the fact that cargo features have a fairly limited character set. Feel free to ignore them, but I thought you might find them interesting/useful---unless otherwise noted, the style that I suggest generally deals nicely with pretty much any special characters in names of things by being careful to control quoting and word splitting. The one exception is NULs, which cannot be dealt with in bash unless you go to a great deal of effort to never read anything into a bash string (which is at heart a C string, and can't contain NULs).

hack/clippy-package-features.sh
marosset: updated to use features, thanks!

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,71 @@
...
+else 
+    REQUIRED_FEATURES=()
+fi
+
+# Get all features for the package (excluding default and required features)
+# Always exclude "default", and exclude any required features using jq
+features=$(cargo metadata --format-version 1 --no-deps | jq -r --arg pkg "$PACKAGE" '.packages[] | select(.name == $pkg) | .features | keys[] | select(. != "default" and (IN($ARGS.positional[])|not))' --args "${REQUIRED_FEATURES[@]}" || true)
+
+# Convert required features array to comma-separated string for cargo
+if [[ ${#REQUIRED_FEATURES[@]} -gt 0 ]]; then
```
syntactically, line 32: Does this conditional need to be here?

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,71 @@
...
+    required_features_str=$(IFS=,; echo "${REQUIRED_FEATURES[*]}")
```
syntactically, line 33: I would suggest using quotes around the $() unless you explicitly want word-splitting to occur

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,71 @@
...
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$required_features_str,$feature" --profile="$PROFILE" -- -D warnings)
+    else
+        echo "Testing $PACKAGE with feature: $feature"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$feature" --profile="$PROFILE" -- -D warnings)
+    fi
+done
+
+# Test all features together
+if [[ -n "$features" ]]; then
+    all_features=$(echo $features | tr '\n' ',' | sed 's/,$//')
```
syntactically, line 63: If there are no blank lines & so only one trailing newline, this tr | sed is  a lot like `paste -sd,`.

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,71 @@
...
+    REQUIRED_FEATURES=("kvm" "mshv3")
+elif [[ "$PACKAGE" == "hyperlight-guest-bin" ]]; then
+    REQUIRED_FEATURES=("printf")
+else 
+    REQUIRED_FEATURES=()
+fi
+
+# Get all features for the package (excluding default and required features)
+# Always exclude "default", and exclude any required features using jq
+features=$(cargo metadata --format-version 1 --no-deps | jq -r --arg pkg "$PACKAGE" '.packages[] | select(.name == $pkg) | .features | keys[] | select(. != "default" and (IN($ARGS.positional[])|not))' --args "${REQUIRED_FEATURES[@]}" || true)
```
syntactically, line 29: Any particular reason to keep this newline-separated rather than readarray it? It feels a bit odd to have required features as an array but available features only as a string (and means you end up relying on word splitting again later in the for loop).

hack/clippy-package-features.sh
```diff
@@ -0,0 +1,71 @@
...
+    else
+        echo "Testing $PACKAGE with feature: $feature"
+        (set -x; cargo clippy -p "$PACKAGE" --all-targets --no-default-features --features "$feature" --profile="$PROFILE" -- -D warnings)
+    fi
+done
+
+# Test all features together
+if [[ -n "$features" ]]; then
+    all_features=$(echo $features | tr '\n' ',' | sed 's/,$//')
+    if [[ ${#REQUIRED_FEATURES[@]} -gt 0 ]]; then
```
syntactically, line 64: Concatenating all_features and required_features as structured arrays before having the shell combine them back into a comma-delimited string also gets rid of the need for this conditional, I think.

syntactically (commented):
Again, feel free to ignore shell style nits

marosset:
It looks like some more clippy issues got introduced by https://github.com/hyperlight-dev/hyperlight/pull/696
I'll fix those us as part of this PR, then let's merge this (even with some bash/shell nits :P)

marosset:
@syntactically - Can you review the new updates I pushed (https://github.com/hyperlight-dev/hyperlight/pull/672/commits/50b08fedaae626b8f914f70f84331ce66448d194) which fixes some clippy issues in hyperlight_guest_bin introduced by a recent PR?
Thanks!

.github/workflows/dep_rust.yml
```diff
@@ -31,6 +31,51 @@ defaults:
...
+      - name: clippy
+        if: ${{ (runner.os == 'Windows' )}}
+        run: |
+          just clippy ${{ matrix.config }}
+          just clippy-guests ${{ matrix.config }}
+
+      - name: clippy exhaustive check
+        if: ${{ (runner.os == 'Linux' )}}
+        run: |
+          sudo apt-get install -y jq || true
```
ludfjig: isn't jq installed in our runners already?

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -105,11 +113,18 @@ pub unsafe fn dbg_print_address_pte(address: u64) -> u64 {
...
+    addrs[0]
 }
 
 /// Allocate n contiguous physical pages and return the physical
 /// addresses of the pages in question.
+/// # Safety
+/// This function uses raw pointer conversion to obtain a physical memory address
+/// of allocated memory.
+/// # Panics
+/// This fcuntion will panic if:
```
ludfjig: typo :P

ludfjig (approved):
LGTM, ignoring any potential bash hygiene issues

syntactically:
> Btw, if mshv3 is a required feature, will mshv2 ever be checked? Since mshv3 "overwrites" mshv2

I think we are planning to get rid of mshv2 very shortly.

.github/workflows/dep_rust.yml
marosset: It should be now but wasn't when I first started working on the changes
I'll remove it

src/hyperlight_guest_bin/src/paging.rs
marosset: :-/

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -38,22 +38,25 @@ pub fn ptov(x: u64) -> *mut u8 {
...
+pub struct MapResponse {
     entry_ptr: *mut u64,
     vmin: *mut u8,
     len: u64,
 }
 
 /// Assumption: all are page-aligned
+/// # Safety
+/// This function executes inlined assembly to set a register.
 pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
```
dblnz, line 64: I am getting the same `clippy` warning for the `src/hyperlight_host/src/sandbox/initialized_multi_use.rs:197:5` file as well.
I don't know how the CI hasn't complained, but for me when I run `just clippy debug` it fails.

src/hyperlight_guest_bin/src/exceptions/handler.rs
```diff
@@ -56,9 +54,8 @@ const _: () = assert!(size_of::<Context>() == 152 + 512);
 
 // TODO: This will eventually need to end up in a per-thread context,
 // when there are threads.
-pub static handlers: [core::sync::atomic::AtomicU64; 31] =
+pub static HANDLERS: [core::sync::atomic::AtomicU64; 31] =
     [const { core::sync::atomic::AtomicU64::new(0) }; 31];
-type handler_t = fn(n: u64, info: *mut ExceptionInfo, ctx: *mut Context, pf_addr: u64) -> bool;
```
syntactically, line 61: I think the type alias is useful; I'd rather change it to get rid of the warning (is it just name case? something like ExnHandler would be fine as well) but keep it.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -38,22 +38,25 @@ pub fn ptov(x: u64) -> *mut u8 {
 
 /// A helper structure indicating a mapping operation that needs to be
 /// performed
-struct MapRequest {
+pub struct MapRequest {
```
syntactically: Are these two structs now pub just because of `alloc_ptes_if_needed`? If it's just that one, probably better to make it all private---these are pretty much internal implementation details, which I don't think are being used in hyperlight-wasm right now, and I would like to do a better (and less architecture-specific) proper API rather than expose these.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -38,22 +38,25 @@ pub fn ptov(x: u64) -> *mut u8 {
...
-struct MapResponse {
+pub struct MapResponse {
     entry_ptr: *mut u64,
     vmin: *mut u8,
     len: u64,
 }
 
 /// Assumption: all are page-aligned
+/// # Safety
+/// This function executes inlined assembly to set a register.
```
syntactically: This comment is not accurate: the reason why this function is marked unsafe is that it changes the pages backing a virtual memory range. This is of course quite dangerous to the Rust memory model: among other things, I imagine that there should be no Rust references pointing into the region being remapped, for instance. Rather than trying to enumerate everything, I think a comment that just reiterates what the function does is probably about the best you can do for a safety comment here, maybe with some special notes like:
- This does not do any locking before touching the page-table data structures, and must not be concurrent with or reentrant from another page table modification operation
- This does not do TLB invalidation, so if you are not mapping a previously-unmapped range, you may need to do that yourself.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -38,22 +38,25 @@ pub fn ptov(x: u64) -> *mut u8 {
...
 pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
+    #[allow(unused_assignments)]
```
syntactically: I just checked the Rust reference and it might actually be safe to get rid of the initializing assignments on the ones where the variable is only used as an out() operand in assembly. Sorry!

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -71,12 +74,17 @@ pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
...
     .map(|r| map_normal(phys_base, virt_base, r))
-    .collect::<()>();
+    .for_each(drop);
 }
 
 #[allow(unused)]
 /// This function is not presently used for anything, but is useful
 /// for debugging
+/// # Safety
+/// This function executes inlined assembly to set a register.
```
syntactically: I would just note that this traverses the page table data structures, and should not be concurrent with any page table modification operations.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -105,11 +113,18 @@ pub unsafe fn dbg_print_address_pte(address: u64) -> u64 {
     if addrs.len() != 1 {
         panic!("impossible: 1 page map request resolved to multiple PTEs");
     }
-    return addrs[0];
+    addrs[0]
 }
 
 /// Allocate n contiguous physical pages and return the physical
 /// addresses of the pages in question.
+/// # Safety
```
syntactically, line 127: I put the unsafe on because I think in the future with a real physical page allocator this will be unsafe in some ways (probably including e.g. concurrency/reentrancy), but I don't know if there is anything actually unsafe about this implementation right now. It might make sense to have a more *gestures vaguely* and then details TBD comment here.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -125,7 +140,10 @@ pub unsafe fn alloc_phys_pages(n: u64) -> u64 {
     }
 }
 
+/// # Safety
+/// This function executes inlined assembly.
```
syntactically: Similar comment to dbg_print_address_pte above on this can't be concurrent/reentrant.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -141,8 +159,11 @@ pub unsafe fn require_pte_exist(x: MapResponse) -> MapRequest {
     }
 }
 
-/// Page-mapping callback to allocate a next-level page table if necessary
+/// Page-mapping callback to allocate a next-level page table if necessary.
+/// # Safety
```
syntactically, line 169: Similar comment to map_region above on "this modifies the page tables, which is fundamentally unsafe" + also can't be concurrent/reentrant.

syntactically (commented):
Thanks for catching these! Sorry, I guess our previous clippy configuration was missing this crate & I didn't think to run it manually. I have a bunch of comments on the safety notes; I can also do a commit to write them myself if you would like.

src/hyperlight_guest_bin/src/exceptions/handler.rs
marosset: Ok, let me add this back

src/hyperlight_guest_bin/src/paging.rs
marosset: yes, I'll make everything private

marosset:
> Thanks for catching these! Sorry, I guess our previous clippy configuration was missing this crate & I didn't think to run it manually. I have a bunch of comments on the safety notes; I can also do a commit to write them myself if you would like.

No problem. I think I addressed all the comments. PTAL

hack/clippy-package-features.sh
marosset: will address in the future if needed

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -53,8 +53,16 @@ struct MapResponse {
 }
 
 /// Assumption: all are page-aligned
+/// # Safety
+/// This function modifies pages backing a virtual memory range which is inherntiy unsafe w.r.t.
```
syntactically: A few typos here (apologies if my fault), but otherwise this looks like a reasonably safety comment.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -71,12 +79,18 @@ pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
...
-    .collect::<()>();
+    .for_each(drop);
 }
 
 #[allow(unused)]
 /// This function is not presently used for anything, but is useful
 /// for debugging
+/// # Safety
+/// This function traverses page table data structures, and should not be called concurrently
+/// any other operations that modify the page table.
```
syntactically: Missing "with" I think

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -105,11 +119,18 @@ pub unsafe fn dbg_print_address_pte(address: u64) -> u64 {
...
         panic!("impossible: 1 page map request resolved to multiple PTEs");
     }
-    return addrs[0];
+    addrs[0]
 }
 
 /// Allocate n contiguous physical pages and return the physical
 /// addresses of the pages in question.
+/// # Safety
+/// This function is not inhernently unsafe but will likely become so in the future
```
syntactically: "inherently" typo

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -125,8 +146,11 @@ pub unsafe fn alloc_phys_pages(n: u64) -> u64 {
     }
 }
 
-pub unsafe fn require_pte_exist(x: MapResponse) -> MapRequest {
-    let mut pte: u64 = 0;
+/// # Safety
+/// This function traverses page table data structures, and should not be called concurrently
```
syntactically, line 150: Missing "with" again.

syntactically (dismissed):
These look much better, thanks! Commented on a few typos I saw but other than that looks good :)

marosset:
> These look much better, thanks! Commented on a few typos I saw but other than that looks good :)

/sigh, let me fix those

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -53,8 +53,16 @@ struct MapResponse {
 }
 
 /// Assumption: all are page-aligned
+/// # Safety
+/// This function modifies pages backing a virtual memory range which is inherently unsafe w.r.t.
+/// the Rust memory model.
+/// When using this function note:
+/// - No locking is performed before toching page table data structures,
```
ludfjig: typo `toching`,up to you if you wanna fix

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -71,12 +79,18 @@ pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
     .map(|r| unsafe { alloc_pte_if_needed(r) })
     .flat_map(modify_ptes::<20, 12>)
     .map(|r| map_normal(phys_base, virt_base, r))
-    .collect::<()>();
+    .for_each(drop);
```
ludfjig, line 82: is this needed?

ludfjig (dismissed):

src/hyperlight_guest_bin/src/paging.rs
marosset: This was flagged by clippy

marosset: the intention for the clippy warning was to make it clear that the collection is only being iterated over for its side effects

ludfjig (approved):

PR #678, opened 2025-06-30
gdb: fix issue "Debug not enabled" when `gdb` feature was enabled

dblnz:
## Description
This pull request addresses an issue with the `gdb` feature.

**_NOTE_**: Additionally, I have modified the `guest-debugging` example to use two sandboxes(one with debugging enabled and one without) to help find similar issues in the future.

## Behavior
When the `gdb` feature is turned on for `hyperlight`  and a sandbox is not configured to enable debugging, it receives `Debug not enabled` error.

## Root cause
This happens because of a message that is sent, during the Hypervisor construction phase, from the Hypervisor to the `gdb` thread.
The problem is that the signal is sent even if the sandbox has no debug enabled.
The transmission fails because there is no `gdb` thread to receive it and returns this error.

## Fix
Add a condition before sending the signal to ensure the sandbox has debugging enabled


syntactically (approved):
LGTM

PR #683, opened 2025-07-01
WIP: Use page tracking for snapshot and restore

simongdavies:


src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -95,6 +96,7 @@ pub struct HostMapping {
     size: usize,
     #[cfg(target_os = "windows")]
     handle: HANDLE,
+    dirty_pages: Mutex<Option<Vec<usize>>>,
```
ludfjig: This field looks unused except when starting/stopping tracking. Can it be removed?

src/hyperlight_host/src/mem/linux_dirty_page_tracker.rs
```diff
@@ -0,0 +1,1288 @@
...
+    }
+
+    #[test]
+    fn test_write_protection_active() {
+        let (memory_ptr, memory_size) = create_aligned_memory(PAGE_SIZE);
+        let addr = memory_ptr as usize;
+
+        let tracker = create_test_tracker(addr, memory_size).unwrap();
+
+        // Memory should be write-protected initially
```
ludfjig: Comment states that memory should be write-protected initially, but that doens't seem to be true when looking at `create_aligned_memory`

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -719,6 +723,45 @@ impl Hypervisor for KVMDriver {
...
+                        // Get the dirty bitmap for the memory region
+                        self.vm_fd
+                            .get_dirty_log(i as u32, mem_region.guest_region.len())?
+                    }
+                }
+            };
+            for page_idx in bit_index_iterator(&bitmap) {
+                page_indices.push(current_page + page_idx);
+            }
+            current_page += num_pages;
```
ludfjig, line 867: I think we are making an assumption here that all memory regions are consecutive with no gaps. I think `Vec<MemoryRegion>` has this invariant, but not sure if documented anywhere

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -262,48 +263,108 @@ where
...
 
-    /// this function will create a memory snapshot and push it onto the stack of snapshots
-    /// It should be used when you want to save the state of the memory, for example, when evolving a sandbox to a new state
-    pub(crate) fn push_state(&mut self) -> Result<()> {
-        let snapshot = SharedMemorySnapshot::new(&mut self.shared_mem)?;
-        self.snapshots
+    /// this function will create an initial snapshot and then create the SnapshotManager
+    pub(crate) fn create_initial_snapshot(
+        &mut self,
+        dirty_bitmap: Option<&Vec<u64>>,
```
ludfjig: Can we remove this `Option` as I think it's always `Some`? nit: and then also might want to make it &[u64] instead of &Vec<u64>

src/hyperlight_host/src/mem/shared_memory_snapshot_manager.rs
```diff
@@ -0,0 +1,1354 @@
...
+        let total_size = page_count * PAGE_SIZE_USIZE;
+        let mut buffer = vec![0u8; total_size];
+
+        // if the total size is equal to the shared memory size, we can optimize the copy
+        if total_size == shared_mem.mem_size() {
+            // Copy the entire memory region in one go
+            shared_mem.with_exclusivity(|e| e.copy_to_slice(&mut buffer, 0))??;
+        } else {
+            // Sort pages for deterministic ordering and to enable consecutive page optimization
+            dirty_pages.sort_unstable();
```
ludfjig: I believe this sort is redundant

src/hyperlight_host/src/mem/page_snapshot.rs
```diff
@@ -0,0 +1,81 @@
...
+    /// Create a new empty snapshot
+    pub(super) fn new() -> Self {
+        Self {
+            page_index: HashMap::new(),
+            buffer: Vec::new(),
+        }
+    }
+
+    /// Create a snapshot from a list of page numbers with pre-allocated buffer
+    pub(super) fn with_pages_and_buffer(page_numbers: Vec<usize>, buffer: Vec<u8>) -> Self {
```
ludfjig: to be defensive, should we make sure that page_numbers.len() == buffer.len() * page_size?

src/hyperlight_host/src/mem/shared_memory_snapshot_manager.rs
```diff
@@ -0,0 +1,1354 @@
...
+        // check the each index in the dirty bitmap and restore only the corresponding pages from the snapshots vector
+        // starting at the last snapshot look for the page in each snapshot if it exists and restore it
+        // if it does not exist set the page to zero
+        if self.snapshots.is_empty() {
+            return Err(crate::HyperlightError::NoMemorySnapshot);
+        }
+
+        // Collect dirty pages and sort them for consecutive page optimization
+        let mut dirty_pages: Vec<usize> = bit_index_iterator(dirty_bitmap).collect();
+        dirty_pages.sort_unstable();
```
ludfjig: I believe this sort is redundant

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -262,48 +263,108 @@ where
...
+        match snapshot_manager.as_mut() {
+            None => {
+                log_then_return!("Snapshot manager not initialized");
+            }
+            Some(snapshot_manager) => snapshot_manager
+                .pop_and_restore_state_from_snapshot(&mut self.shared_mem, dirty_bitmap),
+        }
+    }
+
+    pub(crate) fn push_state(&mut self, dirty_bitmap: Option<&Vec<u64>>) -> Result<()> {
```
ludfjig: I think(?) this option can also be removed

src/hyperlight_host/src/mem/shared_memory_snapshot_manager.rs
```diff
@@ -0,0 +1,1354 @@
...
+        // Create the snapshot with the pre-allocated buffer
+        let snapshot = PageSnapshot::with_pages_and_buffer(dirty_pages, buffer);
+
+        Ok(snapshot)
+    }
+
+    pub(super) fn create_new_snapshot<S: SharedMemory>(
+        &mut self,
+        shared_mem: &mut S,
+        dirty_page_map: Option<&Vec<u64>>,
```
ludfjig: I think(?) this option can also be removed

src/hyperlight_host/src/mem/shared_memory_snapshot_manager.rs
```diff
@@ -0,0 +1,1354 @@
...
+            input_data_size,
+            output_data_size,
+            output_data_buffer_offset,
+            input_data_buffer_offset,
+        })
+    }
+
+    fn build_snapshot_from_dirty_page_map<S: SharedMemory>(
+        shared_mem: &mut S,
+        dirty_page_map: Option<&Vec<u64>>,
```
ludfjig: I think (?) this option can also be removed

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -606,6 +607,12 @@ impl Hypervisor for HypervWindowsDriver {
         Ok(())
     }
 
+    fn get_and_clear_dirty_pages(&mut self) -> Result<Vec<u64>> {
+        // For now we just mark all pages dirty which is the equivalent of taking a full snapshot
+        let total_size = self.mem_regions.iter().map(|r| r.guest_region.len()).sum();
```
ludfjig, line 644: Can we use https://learn.microsoft.com/en-us/virtualization/api/hypervisor-platform/funcs/whvquerygparangedirtybitmap ?

ludfjig (commented):
Looks mostly good to me. The signal handling of SIGSEGV makes me a little nervous, and there seems to be an assumption that `Vec<MemoryRegions>` are always consecutive in memory, which i think we might break in the future. I also think a refactor of our memory management code would be nice 😅  .

I also remember I had a bunch of tests for evolve/devolve in my previous dirty-pages PR that maybe could be useful to test logic. Those tests were the ones that allowed me to catch the bug in mshv's implementation :P

Also, these regression seems pretty big (compared to main branch)

image

image

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -391,13 +398,31 @@ impl HypervLinuxDriver {
...
 
+        let mut base_pfn = u64::MAX;
+        let mut total_size: usize = 0;
+
         mem_regions.iter().try_for_each(|region| {
-            let mshv_region = region.to_owned().into();
+            let mshv_region: mshv_user_mem_region = region.to_owned().into();
+            if base_pfn == u64::MAX {
+                base_pfn = mshv_region.guest_pfn;
+            }
```
jprendes, line 431: This is effectively obtaining the guest_pfn of the first region.
Is there a guarantee on the order of the regions?
Why does this correspond to the guest page number?

Should this maybe be the smallest guest_pfn? (although I don't understand why 😅)
```rust
base_pfn = base_pfn.min(mshv_region.guest_pfn);
```

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -391,13 +398,31 @@ impl HypervLinuxDriver {
...
+            total_size += mshv_region.size as usize;
```
jprendes, line 432: This assumes that all regions are contiguous.
I think that's true at this point in time (before any mmap happens).
Should this maybe be computed as the largest region end?
```rust
total_size = total_size.max(region.guest_region.end);
```

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -428,6 +453,7 @@ impl HypervLinuxDriver {
             page_size: 0,
             vm_fd,
             vcpu_fd,
+            n_initial_regions: mem_regions.len(),
```
jprendes: ❤️
I was doing the same in my code changes to support snapshotting mmap regions!

Although, I think it might be better to have mem_regions split in two: maybe `mem_regions` and `mmap_regions`?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -885,6 +911,69 @@ impl Hypervisor for HypervLinuxDriver {
...
+            .first()
+            .ok_or(new_error!(
+                "tried to get dirty page bitmap of 0-sized region"
+            ))?
+            .to_owned()
+            .into();
+
+        let n_contiguous = self
+            .mem_regions
+            .windows(2)
```
jprendes: ❤️
It's actually the first time I see `windows` being used other than the doc! nice!

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -885,6 +911,69 @@ impl Hypervisor for HypervLinuxDriver {
...
+        let first_mshv_region: mshv_user_mem_region = self
+            .mem_regions
+            .first()
+            .ok_or(new_error!(
+                "tried to get dirty page bitmap of 0-sized region"
+            ))?
+            .to_owned()
+            .into();
+
+        let n_contiguous = self
```
jprendes: Do we really need this check?
We don't check for this when we do `HypervLinuxDriver::new`, so we are assuming that's true there.
And we do not expect the `mem_regions` below `n_initial_regions` to change either, so we can continue assuming that.
Also, I think we don't need to compute the `sandbox_total_size` again, we can just store it when we do `HypervLinuxDriver::new`, just like we do with `n_initial_regions`, right?

I think splitting `mem_regions` in two: `mem_regions` and `mmap_regions` would help clarify what we can assume.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -885,6 +911,69 @@ impl Hypervisor for HypervLinuxDriver {
...
+            .mem_regions
+            .windows(2)
+            .take_while(|window| window[0].guest_region.end == window[1].guest_region.start)
+            .count()
+            + 1; // +1 because windows(2) gives us n-1 pairs for n regions
+
+        if n_contiguous != self.n_initial_regions {
```
jprendes: This would fail if we mmap something just after (continuous with) the main memory, right?
But I think it shouldn't.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -885,6 +911,69 @@ impl Hypervisor for HypervLinuxDriver {
...
+            let last_block_start_page = last_block_idx * 64;
+            let last_block_end_page = last_block_start_page + 64;
+
+            // If the last block extends beyond the sandbox, clear the invalid bits
+            if last_block_end_page > sandbox_pages {
+                let valid_bits_in_last_block = sandbox_pages - last_block_start_page;
+                let mask = (1u64 << valid_bits_in_last_block) - 1;
+                *last_block &= mask;
+            }
+        }
```
jprendes:
```suggestion
        if let Some(last_block) = sandbox_dirty_pages.last_mut() {
            let mask = match sandbox_pages % 64 {
                0 => u64::MAX,
                tail_bits => (1u64 << tail_bits) - 1,
            };
            *last_block &= mask;
        }
```

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -750,6 +754,61 @@ impl Hypervisor for KVMDriver {
...
+            .count()
+            + 1; // +1 because windows(2) gives us n-1 pairs for n regions
+
+        if n_contiguous != self.n_initial_regions {
+            return Err(new_error!(
+                "get_and_clear_dirty_pages: not all regions are contiguous, expected {} but got {}",
+                self.n_initial_regions,
+                n_contiguous
+            ));
+        }
```
jprendes: Same comment as for mshv.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: No, there is no guarantee of the order currently, but it happens to be true that they are in order and are contiguous because of the way we build them. It's probably something we want to enforce at some point

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: I like this idea, it would make things easier I think

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Yes true. I think splitting the sandbox regions and the later mmaped regions should help with this.

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -370,30 +378,48 @@ impl ExclusiveSharedMemory {
...
     }
 
+    /// Gets the dirty bitmap and then clears it in self.
+    pub(crate) fn get_and_clear_dirty_pages(&mut self) -> Result<Vec<u64>> {
+        let mut guard = self
+            .dirty_page_tracker
+            .try_lock()
+            .map_err(|_| new_error!("Failed to acquire lock on dirty page tracker"))?;
+        let bitmap = guard.clone();
+        guard.fill(0);
```
jprendes: **nit:** we can save a copy:
```suggestion
        let mut bitmap = vec![0; guard.len()];
        core::mem::swap(&mut bitmap, &mut *guard);
```

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -979,12 +1054,31 @@ impl HostSharedMemory {
 
         Ok(to_return)
     }
+
+    /// Marks pages that cover bytes [offset, offset + size) as dirty
+    pub(super) fn mark_pages_dirty(&mut self, offset: usize, size: usize) -> Result<()> {
```
jprendes: why does this need `&mut self` instead of just `&self`?

jprendes (commented):
Some more comments.

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -750,6 +754,61 @@ impl Hypervisor for KVMDriver {
...
+                        //  Trying to get dirty pages for a guard page region results in a VMMSysError(2)
+                        new_page_bitmap(mem_region.guest_region.len(), false)?
+                    } else {
+                        // Get the dirty bitmap for the memory region
+                        self.vm_fd
+                            .get_dirty_log(i as u32, mem_region.guest_region.len())?
+                    }
+                }
+            };
+            for page_idx in bit_index_iterator(&bitmap) {
```
jprendes, line 864: I think it would be nice to use a dependency like [bitvec](https://crates.io/crates/bitvec).
Then we can do something like this:
```rust
let mut sandbox_dirty_pages = BitVec::new();
for (i, mem_region) = self.mem_regions.iter().take(n_contiguos).enumerate() {
    let num_pages = mem_region.guest_region.len() / PAGE_SIZE_USIZE;
    let bitmap = ...;
    let mut bitmap = BitVec::from_vec(bitmap);
    bitmap.truncate(num_pages);
    dirty.extend_from_bitslice(bitmap.as_bitslice());
}

Ok(sandbox_dirty_pages)
```

src/hyperlight_host/src/mem/bitmap.rs
```diff
@@ -103,6 +105,107 @@ impl Iterator for SetBitIndices<'_> {
     }
 }
 
+// Unused but useful for debugging
+// Prints the dirty bitmap in a human-readable format, coloring each page according to its region
+// NOTE: Might need to be updated if the memory layout changes
+#[allow(dead_code)]
+pub(crate) fn print_dirty_bitmap(bitmap: &[u64], layout: &SandboxMemoryLayout) {
```
jprendes, line 102: nice! ❤️

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -675,6 +675,30 @@ impl ExclusiveSharedMemory {
         Ok(())
     }
 
+    /// Same as `copy_from_slice` but doesn't dirty the pages.
+    /// # Safety
+    /// This function is unsafe because it does not mark the pages as dirty.
+    /// Only use this if you are certain that the pages do not need to be marked as dirty.
+    #[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
+    pub unsafe fn restore_copy_from_slice(&mut self, src: &[u8], offset: usize) -> Result<()> {
```
jprendes: **suggestion:** what do you think of `copy_from_slice_no_dirty` and `zero_fill_no_dirty`?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -95,17 +96,39 @@ impl MultiUseSandbox {
...
     pub fn snapshot(&mut self) -> Result<Snapshot> {
-        let snapshot = self.mem_mgr.unwrap_mgr_mut().snapshot()?;
+        let host_dirty_pages = self
+            .get_mgr_wrapper_mut()
+            .unwrap_mgr_mut()
+            .get_shared_mem_mut()
+            .with_exclusivity(|e| e.get_and_clear_dirty_pages())??;
+        let vm_dirty_pages = self.vm.get_and_clear_dirty_pages()?;
+
+        let dirty_pages_bitmap = bitmap_union(&vm_dirty_pages, &host_dirty_pages);
```
jprendes: IIRC, bitmap_union extends the result to whichever input is larger.
Do we ever expect one of these bitmaps to be larger than the other?
I think that's more or an error condition than something we should just workaround at runtime (i.e., I would prefer it to fail with a useful error message / stack trace, than having to debug and find out that this is the cause)

src/hyperlight_host/src/sandbox/snapshot.rs
```diff
@@ -14,11 +14,12 @@ See the License for the specific language governing permissions and
...
 */
 
+use std::sync::Arc;
+
 use crate::mem::shared_mem_snapshot::SharedMemorySnapshot;
 
 /// A snapshot capturing the state of the memory in a `MultiUseSandbox`.
 #[derive(Clone)]
 pub struct Snapshot {
-    /// TODO: Use Arc<SharedMemorySnapshot>
```
jprendes: ❤️ thanks!

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
```diff
@@ -14,94 +14,644 @@ See the License for the specific language governing permissions and
...
+        parent: Option<Arc<SharedMemorySnapshot>>,
+    ) -> Result<Self> {
+        let data = shared_mem.with_exclusivity(|e| -> Result<HashMap<usize, Vec<u8>>> {
+            let mut snapshot = HashMap::new();
+            bit_index_iterator(dirty_pages_bitmap).try_for_each(|idx| {
+                let mut page = vec![0u8; PAGE_SIZE_USIZE];
+                e.copy_to_slice(&mut page, idx * PAGE_SIZE_USIZE)?;
+                snapshot.insert(idx, page);
+                crate::Result::Ok(())
+            })?;
```
jprendes, line 60: **thought:** It would be great if we could coalesce consecutive dirty pages into only one entry in the hashmap. But doesn't have to be now :-)

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
```diff
@@ -14,94 +14,644 @@ See the License for the specific language governing permissions and
...
+    /// assuming `most_recent_snapshot` is the most recent snapshot `shared_mem` was
+    /// in and `current_dirty_pages` are the pages that have been modified since.
     #[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
+    pub(super) fn restore_from_snapshot<S: SharedMemory>(
+        self: &Arc<Self>,
+        shared_mem: &mut S,
+        current_dirty_pages: &[u64],
+        most_recent_snapshot: &Option<Arc<SharedMemorySnapshot>>,
+    ) -> Result<u64> {
+        if self.sandbox_size() != shared_mem.mem_size() {
```
jprendes: hmm, I think we need to check more than just the sandbox size. It has to have the same layout.
What happens if we have a sandbox with 1M heap, 1k io, and 4k stack, and restore on it a sandbox with 1M stack, 4k io, and 1M heap? the main memory size would be the same, but then we would use the wrong Layout, right?

I think this was instroduced by the new snapshot API, we should create an issue to fix it.

jprendes: Ah, I think the fact that we don't support restoring out of branch is the key that avoid this issue for now.

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
```diff
@@ -14,94 +14,644 @@ See the License for the specific language governing permissions and
...
+                    if Arc::ptr_eq(snapshot, self) {
+                        break;
+                    }
+
+                    for page_num in snapshot.data.keys() {
+                        pages_to_restore.insert(*page_num);
+                    }
+
+                    current_snapshot = snapshot.parent.as_ref();
+                }
```
jprendes: can this be factored out and reused in the next if branch?

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
ludfjig: Actually you're right, this case is possible if we're restoring a brand new sandbox, with no previous snapshot, to a given snapshot.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Good catch! I took this from my old PR. Fixed

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Fixed

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Fixed

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: nice, fixed

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Fixed

src/hyperlight_host/src/mem/shared_mem.rs
ludfjig: Fixed

src/hyperlight_host/src/mem/shared_mem.rs
ludfjig: Fixed

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
ludfjig: fixed

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
ludfjig: Agree. I'll make a note/issue to think about this after PR is merged

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -306,8 +306,12 @@ pub(crate) struct HypervLinuxDriver {
     vm_fd: VmFd,
     vcpu_fd: VcpuFd,
     entrypoint: u64,
+    // Regions part of the original sandbox
     mem_regions: Vec<MemoryRegion>,
-    n_initial_regions: usize,
+    // Size of the (contioous) sandbox mem_regions
```
jprendes:
```suggestion
    // Size of the (contiguous) sandbox mem_regions
```

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
jprendes: Would it be an option not supporting that case, just like we do with out of branch snapshots?
Or would things fundamentally break? like... not being able to restore the first ever snapshot

If it's an option, I'm happy for us to create an issue and defer fixing it.

src/hyperlight_host/src/mem/shared_mem_snapshot.rs
jprendes: I'm thinking that this would make the traversal you do to find the correct page during the restoring step much harder, for probably not much gain.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: It'd say no, because then we'd need to change it if the guest_memory does not start at 0

src/hyperlight_host/benches/benchmarks.rs
```diff
@@ -138,9 +214,98 @@ fn sandbox_benchmark(c: &mut Criterion) {
     group.finish();
 }
 
+// Sandbox creation with different heap sizes
```
simongdavies, line 217: I think we should reorientate the benchmarks, we should have a set of tests that we perform on sandboxes of differing sizes , for each sized sandbox we should:
measure the create time 
measure the drop time
have guest calls with a varying number of parameters that does not reset the sandbox
have guest calls with a varying number of parameters that does reset the sandbox.
repeat the above where there are 1-n host function calls made by the guest function
measure the cost of snapshotting  when modifying varying proportions of the sandbox memory
measure the cost of restoring when modifying varying proportions of the sandbox memory.
measure the cost of calling with the largest parameters that sized sandbox can support.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
simongdavies: Should this be contiguous not continuous?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
simongdavies: Don't we use a constant else where to specify the base address that we map to? In which case can we not compute the PFN based on the constant value (I think it may actually be zero) but the point is that if we base it on the constant and that ever changes or gets removed this code that depends on that will be impacted.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -391,13 +402,31 @@ impl HypervLinuxDriver {
...
+        // if we dont clear them then we end up taking a complete snapsot of memory page by page which gets
+        // progressively slower as the sandbox size increases
+        // the downside of doing this here is that the call to get_dirty_log will takes longer as the number of pages increase
+        // but for larger sandboxes its easily cheaper than copying all the pages
+
+        #[cfg(mshv2)]
+        vm_fd.get_dirty_log(base_pfn, total_size, CLEAR_DIRTY_BIT_FLAG)?;
+        #[cfg(mshv3)]
+        vm_fd.get_dirty_log(base_pfn, total_size, MSHV_GPAP_ACCESS_OP_CLEAR as u8)?;
+
```
simongdavies, line 448: Having looked at the perf here I wonder if we should at least measure what it would look like to track dirty pages in the guest, I think this should be reasonably doable now that the foundation has been done in the guest.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -885,6 +914,42 @@ impl Hypervisor for HypervLinuxDriver {
         self.interrupt_handle.clone()
     }
 
+    // TODO: Implement getting additional host-mapped dirty pages.
```
simongdavies, line 993: What does this TODO mean?

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
simongdavies: This require `WHvMapGpaRangeFlagTrackDirtyPages` to be specified when calling `WHvMapGpaRange` since we are using `WHvMapGpaRange2` then we just need to make sure that this flag works with that API.

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -750,6 +757,47 @@ impl Hypervisor for KVMDriver {
         self.interrupt_handle.clone()
     }
 
+    // TODO: Implement getting additional host-mapped dirty pages.
```
simongdavies, line 840: What does this TODO mean?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -196,6 +196,12 @@ pub(crate) trait Hypervisor: Debug + Sync + Send {
         None
     }
 
+    /// Get dirty pages as a bitmap (Vec<u64>).
+    /// Each bit in a u64 represents a page.
+    /// This also clears the bitflags, marking the pages as non-dirty.
+    /// TODO: Implement getting additional host-mapped dirty pages.
```
simongdavies, line 196: TODO?

simongdavies: If this is about the additonal mmaped regions then I think at the moment these are read only so what is needed is metadata to recreate them?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -576,7 +576,10 @@ impl ExclusiveSharedMemory {
     ///   the safety documentation of pointer::offset.
     ///
     ///   This is ensured by a check in ::new()
-    pub(super) fn as_mut_slice(&mut self) -> &mut [u8] {
+    ///
+    /// Additionally, writes to the returned slice will not mark pages as dirty.
+    /// User must call `mark_pages_dirty` manually to mark pages as dirty.
+    fn as_mut_slice(&mut self) -> &mut [u8] {
```
simongdavies, line 614: Is it possible to get rid of this? This is one of the reasons that I preferred the abstraction that signal handling gave, I think that this has the potential for creating difficult to find bugs?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -624,6 +665,40 @@ impl ExclusiveSharedMemory {
...
+        data[offset..offset + len].fill(0);
+        self.mark_pages_dirty(offset, len)?;
+        Ok(())
+    }
+
+    /// Same as `copy_from_slice` but doesn't dirty the pages.
+    /// # Safety
+    /// This function is unsafe because it does not mark the pages as dirty.
+    /// Only use this if you are certain that the pages do not need to be marked as dirty.
+    #[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
```
simongdavies, line 685: What circumstances is this used under?

src/hyperlight_host/src/mem/bitmap.rs
```diff
@@ -93,6 +95,107 @@ impl Iterator for SetBitIndices<'_> {
...
```
simongdavies, line 102: Any reason we cant have this as a Debug trait impl?

ludfjig: We cannot implement Debug on Vec<u64> or slice. Would need a wrapper type. Also this takes a &SandboxMemoryLayout

src/hyperlight_host/src/mem/shared_mem.rs
ludfjig: We need these two for restoring a snapshot. Restoring a sandbox to a snapshotted state should not itself dirty the pages

src/hyperlight_host/src/mem/shared_mem.rs
ludfjig: No, most other methods on SharedMemory use this to write memory. At least it's private now

src/hyperlight_host/benches/benchmarks.rs
ludfjig: Filed #722 for this

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: I put this here so we won't forget about it once additionally mmaped memory is not read-only, and we potentially want to be able to reset it as well

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: I put this here so we won't forget about it once additionally mmaped memory is not read-only, and we potentially want to be able to reset it as well

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: I put this here so we won't forget about it once additionally mmaped memory is not read-only, and we potentially want to be able to reset it as well

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: The first guest region's guest address is 0 (BASE_ADDRESS). The page frame number is obtained by right-shifting the guest address by 12. Note 1 << 12 == page size. I think it is better to calculate it directly like this, rather than assume that it is derived from BASE_ADDRESS, which may or may not be the case.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
jprendes: Fixed the comment :-)

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
syntactically: @simongdavies I am already working on (a superset of)) this

syntactically: In the meantime, I wonder if a useful perf optimization is a handle to disable dirty page tracking because we expect that this sandbox will only ever be restored to a totally different one. Not sure if that would be useful ATM, but depending on how exactly we do snapshotting for AFD it could make sense.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
syntactically: I would suggest this comment be made a little bit more explicit about e.g. "TODO: When regions of memory other than the main guest memory can be writable, ensure we don't miss any dirty pages in them" or something like that?

src/hyperlight_host/src/mem/shared_mem.rs
syntactically: Maybe mark it as `unsafe` since there is an internal invariant that it enforces, to make callers read the docs, etc? Not sure if it is strictly unsafe in the Rust sense, but I suppose if you wrote other code that depended on the snapshot behaviour having what it ought to you could perhaps construct something.

syntactically: Or, replace it with a version that takes offset/len and does the dirty marking itself... but that's probably not super ideal? Idk.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
syntactically: > @simongdavies I am already working on (a superset of)) this

To clarify: in order to achieve our security properties (i.e. an untrusted guest cannot corrupt the execution of or learn information from another guest, even one restored into the same sandbox), we can't trust the guest for dirty page tracking unless the guest is also managing copy-on-write with a hypervisor-enforced readonly underlying store. Therefore, I am looking at making the latter happen.

syntactically:
We may still do more or less this, but it will be implemented differently.

PR #686, opened 2025-07-02
[WIP] Add SIMD support

ludfjig:
KVM done, TODO windows & mshv

### DIFF
--- a/src/hyperlight_host/src/hypervisor/kvm.rs
+++ b/src/hyperlight_host/src/hypervisor/kvm.rs
@@ -21,7 +21,7 @@ use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 
-use kvm_bindings::{KVM_MEM_READONLY, kvm_fpu, kvm_regs, kvm_userspace_memory_region};
+use kvm_bindings::{KVM_MEM_READONLY, kvm_fpu, kvm_regs, kvm_userspace_memory_region, kvm_xcrs};
 use kvm_ioctls::Cap::UserMemory;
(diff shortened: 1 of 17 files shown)


src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -419,6 +420,120 @@ impl KVMDriver {
...
+        // avx is EAX=1, ECX bit 28
+        if !entries
+            .get(1)
+            .map(|entry| entry.ecx & (1 << 28) != 0)
+            .unwrap_or(false)
+        {
+            return Err(new_error!("AVX support not detected on the host machine"));
+        }
+        // avx2 is EAX=7, EBX bit 5
+        if !entries
```
syntactically, line 525: Why do we error on all of these, rather than mirroring the host cpuid & being careful to only configure the features that actually exist on this platform?

ludfjig: What if a guest is compiled with for example `+avx`, but the host doesn't support it? We won't be able to run it so I thought it'd be better to error early than get InvalidOp exception in guest.

syntactically: I feel like that might just need to be a "you have to be careful about this when using simd if not using feature detection" thing? (a lot of software nowdays has feature detection for advanced simd anyway). I feel it probably doesn't make sense to statically decide in Hyperlight what simd features will be supported; this both rules out using hyperlight on older hardware & prevents using newer simd features on newer hardware.

syntactically:
Now that we have exception handlers that actually return to the guest
(added in #696), doing this might require a little bit of
investigation to make sure that context save/restore works
properly. See also: #711.

syntactically:
Superceded by custom target SIMD

PR #695, opened 2025-07-07
[WIP] Enable guest tracing 

dblnz:
# Description
This PR update the work @syntactically has done in #103 to latest main and extends it to support tracing the guest for performance.

## How it works
It defines a `TraceBuffer` in the guest that holds `TraceRecord`s. These records are sent to the host using an `Out` instruction and providing the pointer to an array of `TraceRecords`. The host accesses the guest memory, reads these records and writes them to the trace file.

Note: For now, the timestamp of the traces is not calculated correctly. It needs to be calculated relative to the start of the guest and base of on the invariant TSC frequency.


dblnz:
The output data should be changed to something that other tools can open #704.

syntactically:
> Does not report an invariant TSC. I am currently investigating if I can get an Azure VM that reports supporting an invariant TSC.

Just to document this: it was confirmed that the host under a nested MSHV will incorrectly not see the invariant TSC bit, and this should be ignored; on the azure platforms, at least, the TSC is in fact invariant.

docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -91,3 +91,72 @@ NOTE: when running this on windows that this is a linux container, so you will n
 ```
 
 Once the container or the exe is running, the trace output can be viewed in the jaeger UI at [http://localhost:16686/search](http://localhost:16686/search).
+
+## Guest Tracing, Unwinding, and Memory Profiling
+
+Hyperlight provides advanced observability features for guest code running inside micro virtual machines. You can enable guest-side tracing, stack unwinding, and memory profiling using the `trace_guest`, `unwind_guest`, and `mem_profile` features. This section explains how to build, run, and inspect guest traces.
+
+### Building a Guest with Tracing Support
```
jsturtevant, line 104: This section is focused on how to use the development tools to test from a hyperlight developers perspective but it would be nice to think about it from a user of hyper light.  For instance, if I am writing my own guest what do I need to do?  I can of course look at these samples but having some docs about it would be useful here.

src/hyperlight_common/Cargo.toml
```diff
@@ -25,11 +25,14 @@ spin = "0.10.0"
 [features]
 default = ["tracing"]
 fuzzing = ["dep:arbitrary"]
+trace_guest = []
+unwind_guest = []
+mem_profile = []
```
jsturtevant, line 30: What is the reasoning for three separate features? From a usability standpoint it seems like a lot of things to keep track of. 

From a maintainer standpoint it also feels like a large additional matrix of things to test

src/hyperlight_guest/src/exit.rs
```diff
@@ -21,17 +21,22 @@ use hyperlight_common::outb::OutBAction;
...
 /// Exits the VM with an Abort OUT action and code 0.
 #[unsafe(no_mangle)]
+#[hyperlight_guest_tracing_macro::trace_function]
 pub extern "C" fn abort() -> ! {
     abort_with_code(&[0, 0xFF])
 }
 
 /// Exits the VM with an Abort OUT action and a specific code.
+#[hyperlight_guest_tracing_macro::trace_function]
 pub fn abort_with_code(code: &[u8]) -> ! {
```
jsturtevant, line 40: do you need to flush on an abort too?

src/hyperlight_guest/src/exit.rs
```diff
@@ -62,7 +68,10 @@ pub unsafe fn abort_with_code_and_message(code: &[u8], message_ptr: *const c_cha
 }
 
 /// OUT bytes to the host through multiple exits.
+#[hyperlight_guest_tracing_macro::trace_function]
 pub(crate) fn outb(port: u16, data: &[u8]) {
+    // Ensure all tracing data is flushed before sending OUT bytes
+    hyperlight_guest_tracing_macro::flush!();
```
jsturtevant, line 74: In the abort section this means this gets called twice, is that expected?

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -60,6 +61,7 @@ pub(crate) fn call_guest_function(function_call: FunctionCall) -> Result<Vec<u8>
             core::mem::transmute::<usize, GuestFunc>(function_pointer)
         };
 
+        hyperlight_guest_tracing_macro::trace!("Calling guest function");
```
jsturtevant: Did you need a corresponding "guest function completed"?

If so this seems like a common pattern we could include in the macro

as is done in the io.rs:

```
   hyperlight_guest_tracing_macro::trace!("Start copy of data");
   odb[stack_ptr_rel as usize..stack_ptr_rel as usize + data.len()].copy_from_slice(&data);
   hyperlight_guest_tracing_macro::trace!("Finish copy of data");
```

src/hyperlight_guest_tracing_macro/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+#[proc_macro]
+pub fn trace(input: TokenStream) -> TokenStream {
+    // Convert to proc_macro2::TokenStream for parsing
+    let input2: proc_macro2::TokenStream = input.clone().into();
+
+    // Try to parse as message
+    if let Ok(parsed) = syn::parse2::<TraceInput>(input2) {
+        let trace_message = match parsed.message {
+            Lit::Str(lit_str) => lit_str.value(),
+            _ => "expression".to_string(),
```
jsturtevant: Should this fail or print a message that indicates something when wrong using the marco?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -616,6 +617,7 @@ impl ExclusiveSharedMemory {
     /// Return the address of memory at an offset to this `SharedMemory` checking
     /// that the memory is within the bounds of the `SharedMemory`.
     #[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
+    #[allow(dead_code)]
```
jsturtevant, line 620: Why all the deadcode allows?

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -180,6 +297,93 @@ fn handle_outb_impl(
...
+            let mut buffer = vec![0u8; len as usize * std::mem::size_of::<TraceRecord>()];
+            let buffer = &mut buffer[..];
+
+            // Read the trace records from the guest memory
+            // TODO: maybe this can be done without copying?
+            mem_mgr
+                .as_ref()
+                .shared_mem
+                .copy_to_slice(buffer, ptr as usize - SandboxMemoryLayout::BASE_ADDRESS)
+                // .read::<u64>((addr - SandboxMemoryLayout::BASE_ADDRESS as u64) as usize)
```
jsturtevant: nit: remove this commented line

src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -86,6 +93,102 @@ pub fn is_hypervisor_present() -> bool {
     hypervisor::get_available_hypervisor().is_some()
 }
 
+#[cfg(feature = "trace_guest")]
+#[derive(Clone)]
+/// The information that trace collection requires in order to write
+/// an accurate trace.
+pub(crate) struct TraceInfo {
```
jsturtevant, line 103: How is this related to tracerecord in the `hyperlight_guest_tracing` library?  This looks to be new functionality that is only related to tracing, could it be moved into the `hyperlight_guest_tracing` library? and then referenced where needed when the feature is on?

src/hyperlight_common/Cargo.toml
jsturtevant: I think part of the confusion for me is that these seem like three related (as in use some base functionality) but different features but are wrapped up into a single PR.

src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -86,6 +93,102 @@ pub fn is_hypervisor_present() -> bool {
...
+    fn calculate_tsc_freq() -> crate::Result<u64> {
+        if !hyperlight_guest_tracing::invariant_tsc::has_invariant_tsc() {
+            return Err(crate::new_error!(
+                "Invariant TSC is not supported on this platform"
+            ));
+        }
+        let start = hyperlight_guest_tracing::invariant_tsc::read_tsc();
+        let start_time = std::time::Instant::now();
+        // Sleep for 1 second to get a good sample
+        std::thread::sleep(std::time::Duration::from_secs(1));
```
jsturtevant: Will this cause a delay when running with tracing feature?

jsturtevant (commented):
Great work!  Could we add a step where the test binaries are executing in the CI pipeline?  Right now it looks like none of this code is being exercised since it is all behind feature flag.

src/hyperlight_common/Cargo.toml
syntactically: @jsturtevant They kind of are, since Doru built this on top of my old memory profiling PR from way back when. The idea with that was that the base tracing feature that lets you generate a file of fine-grained events quickly is useful for other things (and indeed it is being used it here for the performance events), while the full memory profiling build is _extraordinarily_ slow, since it generates a trace event on every allocation/deallocation. Carrying around the registers to unwind the guest stack also seemed like something that would be useful in other contexts but a different amount of overhead & not always needed, e.g. the performance events here don't I believe use it.

src/hyperlight_host/src/mem/shared_mem.rs
syntactically: I think these are to get rid of clippy warnings when you build with the basic trace_guest infrastructure support but not with any of the things that use it. Maybe another option here is to replace the feature with a cfg item that is any(features_which_use_this)

src/hyperlight_host/src/sandbox/mod.rs
syntactically: This is in the host, not the guest.

jsturtevant: ah, yes still forming that mental model. This makes more sense now.

src/hyperlight_common/Cargo.toml
jsturtevant: Makes sense that you may want to turn those on/off due to overhead. I think it would make for an easier review if these were split out, but I am fine with it if others are.

syntactically: I recommend reviewing commit-by-commit; I put a lot of effort into making sure that each commit itself is small/atomic/reviewable.  That means that we can have a PR like this one where we only merge things after they are useful (& consequently we know that the design is usable), but reviewers can understand and review each step along that path in isolation.

docs/hyperlight-metrics-logs-and-traces.md
dblnz: You are right, some more documentation is always good.
However, I would add that, at this time, this feature is not yet "friendly" enough for a user to utilize within the guest. This is mostly because of the output artifact which is not really a standardized format.
Additionally, we do not publish these new tracing crates yet, for the users to consume in their guests.
Furthermore, I believe that this PR addresses enablement of guest tracing from the perspective of hyperlight developers so that we can have some idea about what and where improvements can be done.
Subsequent work will focus on enabling this for users and providing the best experience we can.

src/hyperlight_common/Cargo.toml
dblnz: @jsturtevant I agree that this is a lot of work to review and it would be simpler to have it split in multiple reviews.
What I can do towards this goal is have this work split in two PRs. One with the updates I've made to Lucy's work to get it up to date with Hyperlight, and another work with the tracing I've added.
However, I still think these feature separation would be needed, because they enable different kinds of measurements which target different scenarios.

src/hyperlight_guest/src/exit.rs
dblnz: This does not flush, it just records the calling of the `abort_with_code` function.

src/hyperlight_guest/src/exit.rs
dblnz: It does have a performance impact, but the `outb` function is used on the error path, so it is not very critical.
Additionally, I wanted to flush on all use of the `outb` function, as it is public for the crate. I could only call flush on abort/abort_with_code, but I wanted to make sure that before leaving the guest, the traces are sent to the host.
Additionally, having `flush` called at this level, ensures that whoever uses `outb` and does not take `tracing` into consideration would not have problems.

src/hyperlight_guest_bin/src/guest_function/call.rs
dblnz: Yeah, I'll add this

src/hyperlight_guest_tracing_macro/src/lib.rs
dblnz: The thing is, being on the guest side here, we do not have a `println`.
Maybe I can make it fail at compile time, let me look into it.

src/hyperlight_host/src/mem/shared_mem.rs
dblnz: Actually, this is something weird for which I spent some time and I wasn't able to find out why it happened.
The commit in question is : `[hyperlight_host/trace] Support collecting guest stacktraces` after which all these clippy warnings appear. The weird thing is I couldn't find where this commit actually stopped using these functions.
I'll have another look at it.

src/hyperlight_host/src/sandbox/mod.rs
dblnz: Yes.
I think this could be improved. The linux kernel prints to `dmesg` the same frequency we calculate here, we could take it from there, but we need permissions.

src/hyperlight_common/Cargo.toml
jsturtevant: I'm not too picky here, if others are happy we can keep as is.

src/hyperlight_guest_tracing_macro/src/lib.rs
jsturtevant: Even just something other than `"expression"` might help debugging down the line

src/hyperlight_guest_bin/src/guest_function/call.rs
dblnz: This does not work so smoothly as it does for the `[proc_macro_attribute]`. This is the reason I haven't added it so far.

I could add something like:
```
trace!("copy data",
    odb[stack_ptr_rel as usize..stack_ptr_rel as usize + data.len()].copy_from_slice(&data);
);
```
that translates to:
```
create_trace_record("> copy data");
    odb[stack_ptr_rel as usize..stack_ptr_rel as usize + data.len()].copy_from_slice(&data);
create_trace_record("< copy data");
```
But there would be an issue with return types if someone would invoke it for a statement/expression that returns something.

The `#[proc_macro]` cannot check the return type. What I could do, is to run the statement/expression as a closure, but that could mess up scopes and in case there is an early return, it would not compile.

```
trace!("test", {
    if x > 0 {
        call_function_x()
    } else {
        return Err(...)
    }
}
);
```

The only thing I can do is:
```
create_trace_record("> copy data");
let res = || { 
    if x > 0 {
        call_function_x()
    } else {
        return Err(...)
    }
}();
create_trace_record("< copy data");
res
```
This last example won't have the desired functionality of returning from the function, it would return from the closure.

jsturtevant: This makes sense.  Thanks for looking into it.

docs/hyperlight-metrics-logs-and-traces.md
jsturtevant: That's fair for now, we can address it in follow-ups, particularly https://github.com/hyperlight-dev/hyperlight/issues/704?

src/hyperlight_guest_tracing_macro/src/lib.rs
syntactically: This code is running in the proc macro isn't it? So this is at compile time and has a full std I thought?

src/hyperlight_guest_bin/src/guest_function/call.rs
syntactically: Spoke with Doru offline, I think we found a solution to this?

src/hyperlight_host/src/sandbox/mod.rs
dblnz: Removed this delay. Now the frequency is calculated using two readings of TSC on the host.
One when the guest is started, the other when the first trace records are received.

src/hyperlight_guest_tracing_macro/src/lib.rs
dblnz: I changed this to fail if no Str is provided.

src/hyperlight_guest_bin/src/exceptions/handler.rs
```diff
@@ -65,6 +65,10 @@ pub extern "C" fn hl_exception_handler(
     exception_number: u64,
     page_fault_address: u64,
 ) {
+    // This function does not return, so wrapping it in a tracing macro
+    // issues a warning of unreachable code.
+    hyperlight_guest_tracing_macro::trace!("> hl_exception_handler");
```
syntactically, line 72: This actually does now return when the exception is handled.

.github/workflows/dep_rust.yml
```diff
@@ -169,6 +170,13 @@ jobs:
           RUST_LOG: debug
         run: just test-rust-crashdump ${{ matrix.config }} ${{ matrix.hypervisor == 'mshv' && 'mshv2' || ''}}
 
+      - name: Run Rust Tracing tests - linux
+        if: runner.os == 'Linux'
+        env:
+          CARGO_TERM_COLOR: always
+          RUST_LOG: debug
+        run: just test-rust-tracing ${{ matrix.config }} ${{ matrix.hypervisor == 'mshv3' && 'mshv3' || ''}}
```
syntactically: I think the way this chooses whether to run mshv2 or mshv3 might need to be updated after #703?

src/hyperlight_guest/src/exit.rs
syntactically, line 74: Again, not something to worry about before merging this, but since the two flushes in this file are right before we cause a vmexit for some other reason (outb, hlt) it would be neat to think about changing the exit ABI so that when tracing is enabled we always bring along a trace structure in some extra registers and can coalesce the exits.

src/hyperlight_guest_tracing_macro/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+/// feature is enabled.
+///
+/// The trace record will contain the function name as a string.
+/// Note: This macro is intended to be used with the `hyperlight_guest_tracing` crate.
+#[proc_macro_attribute]
+pub fn trace_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
+    let input_fn = parse_macro_input!(item as ItemFn);
+
+    let fn_name = &input_fn.sig.ident;
+    let fn_name_str = fn_name.to_string();
```
syntactically, line 40: Should this make sure that the strings built from `fn_name_str` don't exceed `MAX_TRACE_MSG_LEN`?

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+/// Maximum length of a trace message in bytes.
+const MAX_TRACE_MSG_LEN: usize = 64;
+
+#[derive(Debug, Copy, Clone)]
+/// Represents a trace record of a guest with a number of cycles and a message.
+pub struct TraceRecord {
+    /// The number of CPU cycles returned by the invariant TSC.
+    pub cycles: u64,
+    /// The length of the message in bytes.
+    pub msg_len: usize,
```
syntactically, line 45: Instead of storing an actual string here, which has some extra overhead to copy and takes up a significant amount of space in the trace buffer, could it possibly make sense to have the tracing macros choose a smaller representation (e.g. a single u32, plus data, if any, that actually varies) for each call site, produce a separate file with call site -> string message mappings, and only record the call site number in the buffer?

docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -91,3 +91,72 @@ NOTE: when running this on windows that this is a linux container, so you will n
...
+
+This command will list the stack frames and tracing information captured during guest execution, allowing you to analyze guest behavior, stack traces, and memory usage.
+
+#### Example
+
+```bash
+cargo run -p trace_dump ./src/tests/rust_guests/bin/debug/simpleguest ./trace/<UUID>.trace list_frames
+```
+
+You can use additional features such as `unwind_guest` and `mem_profile` by enabling them during the build and run steps. Refer to the guest and host documentation for more details on these features.
```
syntactically: Does this documentation actually exist? (Probably fine for now if not)

Justfile
```diff
@@ -141,6 +145,25 @@ test-rust-gdb-debugging target=default-target features="":
...
+    cargo test -p hyperlight-guest-tracing --profile={{ if target == "debug" { "dev" } else { target } }}
+    cargo test -p hyperlight-guest-tracing-macro --profile={{ if target == "debug" { "dev" } else { target } }}
+
+    # Prepare the tracing guest for testing
+    just build-rust-guests {{ target }} trace_guest
+    just move-rust-guests {{ target }}
+    # Run hello-world example with tracing enabled to get the trace output
+    # Capture the trace file path and print use it afterwards to run cargo run -p trace_dump
+    cargo run --profile={{ if target == "debug" { "dev" } else { target } }} --example hello-world --features {{ if features =="" {'trace_guest'} else { "trace_guest," + features } }} \
+        | sed -n 's/.*Creating trace file at: \(.*\)/\1/p' \
```
syntactically, line 160: Should this be checking that the dumped frames match what is expected or anything?

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+    /// The message associated with the trace record.
+    pub msg: [u8; MAX_TRACE_MSG_LEN],
+}
+
+impl TryFrom<&str> for TraceRecord {
+    type Error = &'static str;
+
+    fn try_from(msg: &str) -> Result<Self, Self::Error> {
+        if msg.len() > MAX_TRACE_MSG_LEN {
+            return Err("Message too long");
```
syntactically: Would it be more useful to truncate here?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -626,6 +643,15 @@ impl Hypervisor for KVMDriver {
         {
             Err(kvm_ioctls::Error::new(libc::EINTR))
         } else {
+            #[cfg(feature = "trace_guest")]
+            if self.trace_info.guest_start_epoch.is_none() {
```
syntactically, line 647: Is it important that this be super close to the actual VM entry? If not, some duplication could perhaps be removed by moving it to `VirtualCPU::run()` instead of needing to do this in each hypervisor driver?

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+}
+
+/// A buffer for storing trace records.
+struct TraceBuffer {
+    /// The entries in the trace buffer.
+    entries: [TraceRecord; MAX_NO_OF_ENTRIES],
+    /// The index where the next entry will be written.
+    write_index: usize,
+    /// Function to send the trace records to the host.
+    send_to_host: SendToHostFn,
```
syntactically: Is this generic over the sending function purely for unit testing? Is there any performance impact to the extra indirection/lack of inlining?

src/hyperlight_guest_tracing_macro/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+    };
+
+    TokenStream::from(expanded)
+}
+
+// Input structure for the trace macro
+struct TraceMacroInput {
+    message: syn::Lit,
```
syntactically, line 96: Since this is a literal string, can we make sure that it does not overflow `MAX_TRACE_MSG_LEN` and warn/error at compile time if it does?

src/trace_dump/main.rs
```diff
@@ -0,0 +1,927 @@
...
+fn dump_trace_record(
+    _state: &mut State,
+    _rs: &mut (),
+    indent: &mut u64,
+    now: Duration,
+    msg: Rc<[u8]>,
+) -> Option<()> {
+    let msg = String::from_utf8_lossy(&msg);
+
+    // Pritty printing of trace records to show indentation based on the trace depth.
```
syntactically: Typo: s/Pritty printing/Pretty-printing/

src/trace_dump/main.rs
```diff
@@ -0,0 +1,927 @@
...
+    // Indentation is increased for messages starting with `>`, and decreased for messages starting
```
syntactically, line 118: Is it intentional that this doesn't also do the same for the +/- messages from tracing an expression?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -1078,6 +1121,26 @@ impl Hypervisor for HypervLinuxDriver {
 
         Ok(())
     }
+
+    #[cfg(feature = "trace_guest")]
+    fn read_trace_reg(&self, reg: TraceRegister) -> Result<u64> {
```
syntactically, line 1126: Not a priority for merging this PR, but now that we have the interface that GDB can use for reading debug registers, it would be nice to unify that and this, or at least use the same datastructure to represent a register name.

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -180,6 +302,105 @@ fn handle_outb_impl(
...
+            };
+            let Ok(ptr) = _hv.read_trace_reg(crate::hypervisor::TraceRegister::RCX) else {
+                return Ok(());
+            };
+            let mut buffer = vec![0u8; len as usize * std::mem::size_of::<TraceRecord>()];
+            let buffer = &mut buffer[..];
+
+            // Read the trace records from the guest memory
+            // TODO: maybe this can be done without copying?
+            mem_mgr
```
syntactically, line 356: Avoiding TOCTOU without copying here is probably more effort than its worth?

src/hyperlight_guest_tracing_macro/src/lib.rs
```diff
@@ -0,0 +1,225 @@
...
+/// Usage:
+/// ```rust
+/// hyperlight_guest_tracing_macro::flush!();
+/// ```
+#[proc_macro]
+pub fn flush(_input: TokenStream) -> TokenStream {
+    let expanded = quote! {
+        {
+            #[cfg(feature = "trace_guest")]
+            hyperlight_guest_tracing::flush_trace_buffer();
```
syntactically, line 245: What's the benefit of reexporting this as a macro? Just that you don't need to use two crates?

src/trace_dump/main.rs
```diff
@@ -0,0 +1,927 @@
...
+            .sites
+            .get_mut(frame)
+            .expect("free of un-allocated site") -= amt;
+    }
+    state.traces.apply_path(trace.iter().rev(), |t| *t -= amt);
+    render_state(state, rs, now)?;
+    Some(())
+}
+
+fn render_trace_record(
```
syntactically, line 605: Is there a reason this is pulled out to a named function used in one place, even though the other occurrences are just anonymous lambdas?

syntactically (commented):
This looks great! I left a few relatively minor comments/questions inline.

I had one other question about a commit message, so I couldn't put it in the diff comments: in d3c416c2a70dd10a875cc0e25ae614e631cc06e4, what does the commit message mean by "the Hypervisor checks for the frame Id, to verify what type of exit it is"?

src/hyperlight_guest_bin/src/exceptions/handler.rs
dblnz: In the last week's version of main I had, I was getting an `unreachable` warning.
I'll modify to capture the `hl_exception_handler`.

.github/workflows/dep_rust.yml
dblnz: Yes, I'll address this

src/hyperlight_guest/src/exit.rs
dblnz: That would be better, indeed, good suggestion.

src/hyperlight_guest_tracing/src/lib.rs
dblnz: That would be awesome to have, however I think it can be done as an improvement step, because for now we have some numbers to help us with perf improvements.
I agree this would have an even smaller impact on the measurements.

Justfile
dblnz: A simple check to verify at least `hyperlight_main` got called could be worth doing, although if anything fails, the example should report an error, so the check might be redundant.

src/hyperlight_host/src/hypervisor/kvm.rs
dblnz: It is important to have it as close to the vm entry so that we have a reliable `Instant` regarding the start of the vm.
This way, we can calculate, how much time it takes until vm entry.
Additionally, all the timestamps from the guest are calculated relative to this saved `Instant`.
To be honest, at this time, it does not bring a lot of value, however, if we want to have a complete view of `Host` and `Guest` tracing we would need to do this at some point in time.

Furthermore, we have some work in progress to refactor the way the Hypervisor drivers are implemented and reduce the duplicated code around registers, debugging. After that is finalized, this will only be stored once.

src/hyperlight_guest_tracing/src/lib.rs
dblnz: This was done only for testing. I haven't observed noticeable performance impact, I'll have another look.

src/trace_dump/main.rs
dblnz: Yes, I wanted to have a discrepancy between `>/<`  and `+/-` to emphasize the fact that one is actually used to trace a call into a function (that creates a new stack frame), and the other is an expression that we want to trace.

src/trace_dump/main.rs
dblnz: This might be some dev artifact, I'll change to anon lambdas

src/hyperlight_guest_tracing_macro/src/lib.rs
dblnz: That was the intention, so you don't need to use one crate for tracing and another for flushing when they should come from the same place, however you still need to `use hyperlight_guest_tracing;` so that the generated code could compile.
So I am not sure how much value it brings, other than only having`hyperlight_guest_tracing_macro::{trace, flush}` in your code.

dblnz:
> This looks great! I left a few relatively minor comments/questions inline.
> 
> I had one other question about a commit message, so I couldn't put it in the diff comments: in [d3c416c](https://github.com/hyperlight-dev/hyperlight/commit/d3c416c2a70dd10a875cc0e25ae614e631cc06e4), what does the commit message mean by "the Hypervisor checks for the frame Id, to verify what type of exit it is"?

My phrasing could be better, but I was referring to the `OutBAction` type, that can be a `TraceRecord`, `TraceMemoryFree`, `TraceMemoryAlloc` or a `TraceRecordStack` in the context of the tracing.
I'll rephrase when I address the comments you left

src/hyperlight_guest_bin/src/exceptions/handler.rs
dblnz: The issue is that if I'm to use the `trace_function` macro, that would add a `create_trace_record` at the end of the function, which is unreachable because of the `abort_with_code_and_message` call.
I'll update the comment.

src/hyperlight_guest/src/exit.rs
dblnz: I'll add this as an improvement to another ticket

src/hyperlight_guest_tracing_macro/src/lib.rs
syntactically: It feels a bit off to me to have a macro that doesn't actually do any meaningful syntax manipulation and expands to just a function call, but I guess I can't think of any reason why it would actually be a problem, and it does save typing out the cfg attribute, so fair enough.

Justfile
syntactically: It does seem like it would be nice for the test to have an end-to-end test that some meaningful trace records have actually made it out of the guest, rather than only checking that the guest doesn't break with tracing enabled.

src/hyperlight_guest_bin/src/exceptions/handler.rs
syntactically: In the version on HEAD, there is a return path which doesn't abort through this function (see https://github.com/hyperlight-dev/hyperlight/blob/97ee46e8251472687bce30cad639689c4e652402/src/hyperlight_guest_bin/src/exceptions/handler.rs#L98 ), so I think it is worth adding the exit trace point for that, or else everything after an exception will show up as being inside the handler.  However, using the attribute macro might still not work because of the `extern "C"`?

src/trace_dump/main.rs
dblnz: It follows the pattern already used. Some of the other functions are used in the same way: `render_ident`, `render_unwind`

src/hyperlight_guest_bin/src/exceptions/handler.rs
dblnz: I have, I just pushed a new version.
I am not sure, the tracing attribute is set for the `entrypoint` function and it works

src/hyperlight_guest_tracing_macro/src/lib.rs
dblnz: I addressed this

docs/hyperlight-metrics-logs-and-traces.md
dblnz: I have updated the docs

Justfile
dblnz: I'll implement this in the improvement ticket also.

src/hyperlight_guest_tracing/src/lib.rs
dblnz: Done

src/hyperlight_guest_tracing/src/lib.rs
dblnz: I have added a generic parameter to the struct to address this

src/hyperlight_guest_tracing_macro/src/lib.rs
dblnz: I have added a check to avoid overflow

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
dblnz: I'll address this in the improvement ticket.

src/hyperlight_guest_tracing_macro/src/lib.rs
dblnz: This would go away when I implement the suggestion above which automatically sends traces in a register for each OutB if tracing is enabled.

src/hyperlight_guest_bin/src/exceptions/handler.rs
syntactically: Yeah I guess syn handles it nicely---it just ends up in `fn_sig.abi`.  Does just using the `#[trace_function]` work here then?

src/trace_dump/main.rs
syntactically: Oh, so it does. My apologies!

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -0,0 +1,277 @@
...
+extern crate alloc;
+
+use core::mem::MaybeUninit;
+
+use hyperlight_common::outb::OutBAction;
+use spin::Mutex;
+
+/// Type alias for the function that sends trace records to the host.
+type SendToHostFn = fn(u64, &[TraceRecord]);
+
```
syntactically, line 28: Since `TraceBuffer` here has the generic argument instantiated with the `fn(u64, &[TraceRecord])` function pointer type, rather than with a specific closure type, I am not sure if the added generic parameter actually helps with inlining?  I want to be careful to not add the extra complexity of the generic parameter unless we are sure that it actually buys us something.

syntactically (dismissed):
This is looking really good, thanks!

Please squash the fixup commits (e825f6e0195de462071697ff2009e86fb135a87f onwards) into the original commits for the relevant things before merging.

src/hyperlight_guest_tracing/src/lib.rs
dblnz: I'll take care of this in #723

src/hyperlight_guest_bin/src/exceptions/handler.rs
dblnz: No, it complains that the ending trace is unreachable

PR #696, opened 2025-07-08
Support mapping host memory into the guest

### LINKED ISSUE #666: Add support for mapping a host memory region into a guest physical memory slot
This is a missing primitive that will be useful for zero-copy file I/O into the sandbox.

syntactically:
Fixes #666


src/hyperlight_guest_bin/src/exceptions/handler.rs
```diff
@@ -21,20 +21,86 @@ use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
...
-        page_fault_address, stack_pointer
+        "Exception vector: {:#}\n\
+         Faulting Instruction: {:#x}\n\
+         Page Fault Address: {:#x}\n\
+         Error code: {:#x}\n\
+         Stack Pointer: {:#x}",
+        exception_number, saved_rip, page_fault_address, error_code, stack_pointer
     );
 
+    if exception_number < 31 {
```
jprendes, line 90: Could you add a small comment on what `< 31` means? and why `<handler_t>(...) == true` causes an early return?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -173,6 +181,81 @@ impl MultiUseSandbox {
...
+        unsafe { self.vm.map_region(rgn) }?;
+        self.mem_mgr.unwrap_mgr_mut().mapped_rgns += 1;
+        Ok(())
+    }
+
+    /// Map the contents of a file into the guest at a particular address
+    ///
+    /// Returns the length of the mapping
+    #[instrument(err(Debug), skip(self, _fp, _guest_base), parent = Span::current())]
+    pub(crate) fn map_file_cow(&mut self, _fp: &Path, _guest_base: u64) -> Result<u64> {
```
dblnz, line 218: This will be used only for already parsed files, right?

dblnz (commented):
Overall it looks good to me, but I think someone with more expertise on the guest paging setup should have a look.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: What do you mean by "parsed"?

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -0,0 +1,233 @@
...
+/// A helper structure indicating that a particular PTE needs to be
+/// modified
+struct MapResponse {
+    entry_ptr: *mut u64,
+    vmin: *mut u8,
+    len: u64,
+}
+
+/// Assumption: all are page-aligned
+pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
```
jprendes, line 56: I don't think I know enough to review this file, but from afar it LGTM.
Plus, tests pass, so... there's that! :-D

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -493,6 +493,21 @@ impl Hypervisor for KVMDriver {
         Ok(())
     }
 
+    #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
+    unsafe fn map_region(&mut self, _rgn: &MemoryRegion) -> Result<()> {
+        log_then_return!("Mapping host memory into the guest not yet supported on this platform");
```
jprendes, line 498: We should add a TODO here, or even better, create a GH issue :-)

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -173,6 +181,81 @@ impl MultiUseSandbox {
...
+                libc::O_RDWR,
+            );
+            if fd < 0 {
+                log_then_return!("open error: {:?}", std::io::Error::last_os_error());
+            }
+            let mut st = MaybeUninit::<libc::stat>::uninit();
+            libc::fstat(fd, st.as_mut_ptr());
+            let st = st.assume_init();
+            let page_size = page_size::get();
+            let size = (st.st_size as usize).div_ceil(page_size) * page_size;
```
jprendes: Can we rust's `std::fs::File`, [`AsFd::as_fd`](https://doc.rust-lang.org/std/os/fd/trait.AsFd.html#tymethod.as_fd), and [`MetadataExt::st_size`](https://doc.rust-lang.org/std/os/linux/fs/trait.MetadataExt.html#tymethod.st_size) instead?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -173,6 +181,81 @@ impl MultiUseSandbox {
...
+            let base = libc::mmap(
+                std::ptr::null_mut(),
+                size,
+                libc::PROT_READ | libc::PROT_WRITE | libc::PROT_EXEC,
+                libc::MAP_PRIVATE,
+                fd,
+                0,
+            );
+            if base == libc::MAP_FAILED {
+                log_then_return!("mmap error: {:?}", std::io::Error::last_os_error());
```
jprendes, line 236: This is leaking the `fd` in case of failure, right?

jprendes (approved):
LGTM.
Don't know if you want to wait for more reviews or not.
AFAICT this only adds functionality and doesn't change any existing behaviour, so I'm ok if you want to merge it.
It gives me confidence that all tests are passing :-)

Could you create an issue to add tests for it?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -173,6 +181,81 @@ impl MultiUseSandbox {
...
+    ///
+    /// `rgn.region_type` is ignored, since guest PTEs are not created
+    /// for the new memory.
+    ///
+    /// It is the caller's responsibility to ensure that the host side
+    /// of the region remains intact and is not written to until this
+    /// mapping is removed, either due to the destruction of the
+    /// sandbox or due to a state rollback
+    #[instrument(err(Debug), skip(self, rgn), parent = Span::current())]
+    pub unsafe fn map_region(&mut self, rgn: &MemoryRegion) -> Result<()> {
```
jprendes, line 197: I'm wondering if this is the right API.
Should the sandbox own the mapped memory? Maybe taking it as `Vec<u8>` or `&[u8]`?
Happy to go with this version, just leaving the questions here, we can always change the API in a different PR.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
dblnz: Do we want to allow files being mapped directly into the guest without them being checked?
I don't think it is the case, but for the guest binary provided as path, that would not be ok to be mapped directly without it being parsed.
Or maybe I don't understand the usecase?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: > Should the sandbox own the mapped memory?

The main point of this API is to let us choose to expose some memory from somewhere else into the sandbox, thereby achieving zerocopy!

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: Good catch. I think there is another problem here actually where even in the success case the mapped region is never unmapped, but this API isn't actually all that useful (at least yet) due to the performance implications of using a new mmap each time on mshv, so I am just going to create an issue to remind myself about that and will deal with it in a follow up in the next week or two.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: Done

src/hyperlight_host/src/hypervisor/kvm.rs
syntactically: Yes, I was planning to create an issue.

src/hyperlight_guest_bin/src/exceptions/handler.rs
syntactically: Done

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -173,6 +181,75 @@ impl MultiUseSandbox {
...
+    /// Map a region of host memory into the sandbox.
+    ///
+    /// Depending on the host platform, there are likely alignment
+    /// requirements of at least one page for base and len.
+    ///
+    /// `rgn.region_type` is ignored, since guest PTEs are not created
+    /// for the new memory.
+    ///
+    /// It is the caller's responsibility to ensure that the host side
+    /// of the region remains intact and is not written to until this
```
qbradley, line 193: Is the requirement that the memory "is not written to" because this could introduce race conditions between host and guest?  Or is there something about the mapping that can be damaged by writing to the memory?

syntactically: I wrote that partly due to the race issue and partly because I haven't thought out the case with writes in full detail since it isn't necessary for the first use. It definitely might be possible to relax that language in the future.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -0,0 +1,252 @@
...
+    MapRequest {
+        table_base: page_addr,
+        vmin: x.vmin,
+        len: x.len,
+    }
+}
+
+/// Map a normal memory page
+///
+/// TODO: support permissions; currently mapping is always RWX
```
jsturtevant, line 179: Should we be allow all permissions by default?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: On MSHV, the mappings feel a little delicate to me because they are all `pin_user_pages()`'d. Right now, they are always pinned with `FOLL_WRITE`, which probably means that it's acceptable to write to them from userspace even if the MSHV mapping is readonly, but I'm not 100% sure if that is always true/will always stay that way.

jprendes (approved):
LGTM

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: This isn't used for the guest binaries (yet, although hopefully we can use a similar-ish mechanism for snapshots). This is just used for mapping bulk data in zerocopy readonly, like the wasm binary which is parsed and set up by wasmtime inside the sandbox.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: Marking this as resolved so that this can merge, but feel free to reach out!

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: Did we create an issue for this?  What is the correct way to make sure the mapped regions aren't leaked?

syntactically: Yes, there is an issue #700

src/hyperlight_guest_bin/src/paging.rs
syntactically: I think that the only users of this don't benefit from that right now, but I think it would be good to change that at the same time as #701 is fixed.

syntactically:
> It would be nice to see some tests or examples to verify the use of this new functionality

See #698, #699, and hyperlight-dev/hyperlight-wasm#102

src/hyperlight_guest_bin/src/paging.rs
jsturtevant: making a not for myself as well, this is on the guest side

PR #697, opened 2025-07-08
Remove sandbox evolving and devolving and replace it with snapshotting API.

jprendes:
This PR introduces a first version of the snapshot API.

The current implementation doesn't play very nicely with memory mapping. This is because restoring a snapshot will not restore the mapped memory. As this PR already introduces many other changes, I would prefer to keep fixing the interaction with memory mapping to a different PR.

I recommend reviewing it commit by commit. Each commit should be fairly self contained.

This PR introduces a public snapshot API.
As a result, a sandbox is not automatically snapshotted and restored anymore. It has to be triggered by the user instead.
Before we were snapshotting a sandbox when evolving it, and restoring it after each guest call.
As a result, we do not need the evolve / devolve mechanism anymore.
Evolving allowed us to persist changes in a sandbox, now this happens with all guest calls.
Devolving was not being used anywhere AFAIK.
As the evolving/devolving API is removed, we don't need the transition metadata anymore (which was used by evolving), and neither the noop transition.
We can also remove the Sandbox trait that by this point has lost all but one of its methods, which can be easily inlined in the only place where it was used.
We don't need the call context anymore either.

### DIFF
--- a/src/hyperlight_host/examples/func_ctx/main.rs
+++ b/src/hyperlight_host/examples/func_ctx/main.rs
@@ -14,39 +14,28 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use hyperlight_host::sandbox::{MultiUseSandbox, UninitializedSandbox};
-use hyperlight_host::{GuestBinary, Result};
+use hyperlight_host::GuestBinary;
+use hyperlight_host::sandbox::UninitializedSandbox;
 use hyperlight_testing::simple_guest_as_string;
 
 fn main() {
     // create a new `MultiUseSandbox` configured to run the `simpleguest.exe`
     // test guest binary
-    let mut sbox1: MultiUseSandbox = {
-        let path = simple_guest_as_string().unwrap();
-        let u_sbox = UninitializedSandbox::new(GuestBinary::FilePath(path), None).unwrap();
-        u_sbox.evolve()
-    }
-    .unwrap();
+    let path = simple_guest_as_string().unwrap();
+    let mut sbox = UninitializedSandbox::new(GuestBinary::FilePath(path), None)
+        .unwrap()
+        .evolve()
+        .unwrap();
 
-    // create a new call context from the sandbox, then do some calls with it.
-    do_calls(&mut sbox1).unwrap();
-
-    // create a new call context from the returned sandbox, then do some calls
-    // with that one
-    do_calls(&mut sbox1).unwrap();
-}
-
-/// Given a `MultiUseGuestCallContext` derived from an existing
-/// `MultiUseSandbox` configured to run the `simpleguest.exe` test guest
-/// binary, do several calls against that binary, print their results, then
-/// call `ctx.finish()` and return the resulting `MultiUseSandbox`. Return an `Err`
-/// if anything failed.
-fn do_calls(sbox: &mut MultiUseSandbox) -> Result<()> {
-    let res: String = sbox.call_guest_function_by_name("Echo", "hello".to_string())?;
+    // Do several calls against a sandbox running the `simpleguest.exe` binary,
+    // and print their results
+    let res: String = sbox
+        .call_guest_function_by_name("Echo", "hello".to_string())
+        .unwrap();
     println!("got Echo res: {res}");
 
-    let res: i32 = sbox.call_guest_function_by_name("CallMalloc", 200_i32)?;
+    let res: i32 = sbox
+        .call_guest_function_by_name("CallMalloc", 200_i32)
+        .unwrap();
     println!("got CallMalloc res: {res}");
-
-    Ok(())
 }
--- a/src/hyperlight_host/src/mem/mgr.rs
+++ b/src/hyperlight_host/src/mem/mgr.rs
@@ -263,7 +263,7 @@ where
         SharedMemorySnapshot::new(&mut self.shared_mem, self.mapped_rgns)
     }
 
-    /// This function restores a memory snapshot the given snapshot.
+    /// This function restores a memory snapshot from a given snapshot.
     ///
     /// Returns the number of memory regions mapped into the sandbox
     /// that need to be unmapped in order for the restore to be


src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -460,18 +447,13 @@ mod tests {
 
         let snapshot = sbox.snapshot().unwrap();
 
-        let func = Box::new(|call_ctx: &mut MultiUseGuestCallContext| {
-            call_ctx.call::<i32>("AddToStatic", 5i32)?;
-            Ok(())
-        });
-        let transition_func = MultiUseContextCallback::from(func);
-        let mut sbox = sbox.evolve(transition_func).unwrap();
+        sbox.persist_call_guest_function_by_name::<i32>("AddToStatic", 5i32).unwrap();
```
dblnz: Nit:
```suggestion
let _ = sbox.persist_call_guest_function_by_name::<i32>("AddToStatic", 5i32).unwrap();
```

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -213,7 +195,7 @@ impl MultiUseSandbox {
                 args.into_value(),
             );
             let ret = Output::from_value(ret?);
-            self.mem_mgr.unwrap_mgr_mut().push_state()?;
+            self.mem_mgr.unwrap_mgr_mut().push_state().unwrap();
```
dblnz: Do we really want to panic here in case of an Err?
Shouldn't a `?` suffice?
```suggestion
            self.mem_mgr.unwrap_mgr_mut().push_state()?;
```

src/hyperlight_host/tests/common/mod.rs
```diff
@@ -34,9 +35,12 @@ pub fn new_uninit() -> Result<UninitializedSandbox> {
 
 /// Use this instead of the `new_uninit` if you want your test to only run with the rust guest, not the c guest
 pub fn new_uninit_rust() -> Result<UninitializedSandbox> {
+    let mut cfg = SandboxConfiguration::default();
+    cfg.set_input_data_size(112 * 500 + 7); // 1 MiB
```
dblnz: Why do we need this set input data size?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -318,35 +300,43 @@ impl MultiUseSandbox {
...
+            check_for_guest_error(self.get_mgr_wrapper_mut())?;
+
+            self
+                .get_mgr_wrapper_mut()
+                .as_mut()
+                .get_guest_function_call_result()
+        })();
+
+        // TODO: Do we want to allow re-entrant guest function calls?
+        self.get_mgr_wrapper_mut().as_mut().clear_io_buffers();
```
dblnz, line 253: Why is this needed?
Shouldn't all the data from the IO buffers have already been retrieved? This is for sending parameters/results over.
Maybe I'm missing something

dblnz (commented):
Great work, @jprendes!
I like when I see more code removed than it was added 😃 
The PR description is a bit confusing, mentioning the removal of evolve/devolve APIs, but there is still `evolve`ing involved 😄  

I left some comments here and there where I didn't understand something.

**NOTE**: Some of the comments may be outdated as I reviewed each commit.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: > Shouldn't all the data from the IO buffers have already been retrieved?

Normally yes, but in the case of an abnormal termination (e.g., host killing the guest), that's not the case.
Before it was not an issue because we always restored the guest memory, effectively resetting the I/O buffer.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: I think that was a leftover from some debugging.
That line is removed by a later commit anyway.
Do you have a strong preference on fixing the intermediate commit?

jprendes:
> but there is still `evolve`ing involved

Haha, yes, indeed. At lest the traits are not there anymore.
I would like to change that method name to `initialize`, but that's used all over the place.
Because of that I think that one should go on it's own PR.

src/hyperlight_host/tests/common/mod.rs
jprendes: Good catch, let me look into it.

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -265,56 +259,28 @@ where
         }
     }
 
-    /// this function will create a memory snapshot and push it onto the stack of snapshots
```
ludfjig, line 268: Love this simplification!

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -555,4 +519,25 @@ impl SandboxMemoryManager<HostSharedMemory> {
             self.layout.sandbox_memory_config.get_output_data_size(),
         )
     }
+
+    pub(crate) fn clear_io_buffers(&mut self) {
```
ludfjig, line 523: 👍

ludfjig (approved):
LGTM

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -265,10 +265,33 @@ where
         }
     }
 
+    pub(crate) fn snapshot(&mut self) -> Result<SharedMemorySnapshot> {
+        SharedMemorySnapshot::new(&mut self.shared_mem, self.mapped_rgns)
+    }
+
+    
+    /// This function restores a memory snapshot the given snapshot.
```
jsturtevant: ```/// This function restores a memory snapshot from a given snapshot.```

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: Fixed

src/hyperlight_host/tests/common/mod.rs
jprendes: IIRC, this is a left over from debugging the issue that led me to introduce the `clear_io_buffers`.
Basically, if the guest function call ends abnormally, the io buffer doesn't get completely flushed.
The `interrupt_same_thread` test repeatedly calls a guest function that gets killed from the host, and eventually filled up the io buffer.
I'll revert this change.

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -688,22 +686,10 @@ fn execute_on_heap() {
     }
 }
 
-#[test]
-fn memory_resets_after_failed_guestcall() {
-    let mut sbox1 = new_uninit_rust().unwrap().evolve(Noop::default()).unwrap();
```
ludfjig, line 693: Is this test removed because memory doesn't reset anymore?

ludfjig (dismissed):

src/hyperlight_host/examples/func_ctx/main.rs
```diff
@@ -24,32 +23,32 @@ use hyperlight_testing::simple_guest_as_string;
...
     }
     .unwrap();
 
     // create a new call context from the sandbox, then do some calls with it.
-    let ctx1 = sbox1.new_call_context();
-    let sbox2 = do_calls(ctx1).unwrap();
+    do_calls(&mut sbox1).unwrap();
+
     // create a new call context from the returned sandbox, then do some calls
     // with that one
```
jsturtevant: is this comment old?  What was this test doing previously?  It looks like now we just call the same sandbox twice, Did we need to verify anything in between or after?

src/hyperlight_host/examples/func_ctx/main.rs
```diff
@@ -24,32 +23,32 @@ use hyperlight_testing::simple_guest_as_string;
...
-    let ctx2 = sbox2.new_call_context();
-    do_calls(ctx2).unwrap();
+    do_calls(&mut sbox1).unwrap();
 }
 
 /// Given a `MultiUseGuestCallContext` derived from an existing
```
jsturtevant: update this comment

src/hyperlight_host/tests/integration_test.rs
jprendes: Correct

ludfjig (approved):

src/hyperlight_host/src/func/call_ctx.rs
```diff
@@ -1,235 +0,0 @@
...
-    /// (AKA: producers) to make batches of calls.
-    ///
-    /// Upon receipt of a message to execute a batch, a new
-    /// `MultiUseGuestCallContext` is created in the receiver thread from the
-    /// existing `MultiUseSandbox`, and the batch is executed.
-    ///
-    /// After the batch is complete, the `MultiUseGuestCallContext` is done
-    /// and it is converted back to the underlying `MultiUseSandbox`
-    #[test]
-    fn test_multi_call_multi_thread() {
```
jsturtevant, line 169: do we need some tests that cover multi threading? Maybe not with the multiusecallcontext but in general?

src/hyperlight_host/benches/benchmarks.rs
```diff
@@ -43,13 +43,15 @@ fn guest_call_benchmark(c: &mut Criterion) {
...
+    group.bench_function("guest_call_with_restore", |b| {
+        let mut sbox = create_multiuse_sandbox();
+        let snapshot = sbox.snapshot().unwrap();
 
         b.iter(|| {
-            sandbox
-                .call_guest_function_by_name::<String>("Echo", "hello\n".to_string())
-                .unwrap()
+            sbox
+                .call::<String>("Echo", "hello\n".to_string())
```
jsturtevant: Whats the difference between `call` and `call_guest_function_by_name`?  I see them both being used and its not clear to me when I would use one over the other

src/hyperlight_host/tests/integration_test.rs
jsturtevant: do we have a test that makes sure the memory is reset when manually done?

src/hyperlight_host/benches/benchmarks.rs
```diff
@@ -38,24 +36,20 @@ fn guest_call_benchmark(c: &mut Criterion) {
...
+    group.bench_function("guest_call_with_restore", |b| {
+        let mut sbox = create_multiuse_sandbox();
+        let snapshot = sbox.snapshot().unwrap();
 
         b.iter(|| {
-            sandbox
-                .call_guest_function_by_name::<String>("Echo", "hello\n".to_string())
-                .unwrap()
+            sbox.call::<String>("Echo", "hello\n".to_string()).unwrap();
+            sbox.restore(&snapshot).unwrap();
```
jsturtevant, line 52: How do we ensure when someone upgrades to this new api that they don't assume that memory was already restored when it actually wasn't?  Is that a concern?

src/hyperlight_host/benches/benchmarks.rs
jprendes: `call` is the uses the `Callable` trait, and just calls `call_guest_function_by_name`.
Eventually I would like to remove that trait, and rename `call_guest_function_by_name` -> `call`.
But right now the `Callable` trait is used (don't remember exactly where), so I'd prefer leaving it for now.

jprendes: In short, if `Callable` is in scope, you can you `call` or `call_guest_function_by_name` interchangeably.

src/hyperlight_host/benches/benchmarks.rs
jprendes: It is a concern.
I think we need good communication.
Also renaming the method would help bring the change to attention.

src/hyperlight_host/examples/func_ctx/main.rs
jprendes: Not sure what was the actual point of this test before.
But I agree it doesn't seem to be very useful.
I think we can safely simplify this to run `do_calls` only once, and even inline it.

src/hyperlight_host/examples/func_ctx/main.rs
jprendes: done

src/hyperlight_host/src/func/call_ctx.rs
jprendes: The removed test wasn't even testing multi threading on the sandbox.
All requests were queued and handled from only one thread.
I don't think we need any specific thread for multithreading right now other than for interrupt_handler, which we do have.
I say this because the VM is single threaded, and we can only interact with it from a single thread at a time.

src/hyperlight_host/benches/benchmarks.rs
jprendes: Maybe we can leave a function called `call_guest_function_by_name` that does do snapshotting and restoring, and mark it as deprecated with intruction to move to `new_less_fancy_call_guest_function_by_name` (or maybe just `call_guest`)

src/hyperlight_host/benches/benchmarks.rs
jsturtevant: Thanks that helps!

src/hyperlight_host/benches/benchmarks.rs
jsturtevant: leaving the existing function and adding a new one that doesn't do this sounds reasonable to me.  I don't know that we would want/need to depreciate the old one unless we find its not useful. 

Clear rust docs here will help when someone is trying to decide between the two.  Having tests showing difference will help us make sure we are keeping the same behavior.

syntactically: Especially with the changes I am looking at, taking a snapshot can be _quite_ expensive and restoring a snapshot, while cheap (especially for a commonly-used snapshot), is still not free. I also think very few usage patterns will actually want to bundle together the "take snapshot/call function/restore snapshot" sequence in this exact way: especially given how expensive sandboxes are to create right now, the common usage pattern seems to be to keep around a pool of sandboxes and then whenever you need to run something for a tenant, restore that tenant's snapshot image into it, run one or possibly several function calls, and then put the sandbox back into the pool ready to have something else restored into it. None of those function calls benefit from an unconditional snapshot/restore around them---you end up snapshotting either "the state you just restored to" (not very useful) or "some random intermediate state that you will throw away after this request/batch of requests", and you end up restoring to a tenant's "ready to handle a new request" configuration just before putting the sandbox in the pool and then restoring it again from a different tenant's image.

I think that leaving this around with the "obvious" name but a function that usually isn't what you want will be quite a trap for people, and so I would really favour getting rid of the old function and encouraging people to move fully to the new API. (To the extent that I might even prefer the old function just error unconditionally for a couple of releases over just marking it as deprecated).

src/hyperlight_host/benches/benchmarks.rs
syntactically: @jprendes The Callable is used in the macros to abstract over registering functions on a core hyperlight sandbox vs on a Wasm sandbox (or in the future some other guest-specific sandbox type), since the host side of the macros is presently the exact same.

src/hyperlight_host/benches/benchmarks.rs
jprendes: I created an issue so that we address this before the next release: #714

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: Is it worth only doing this on an Err return?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -260,56 +207,52 @@ impl MultiUseSandbox {
...
-            .as_mut()
-            .write_guest_function_call(&buffer)?;
-
-        self.vm.dispatch_call_from_host(
-            self.dispatch_ptr.clone(),
-            self.out_hdl.clone(),
-            self.mem_hdl.clone(),
-            #[cfg(gdb)]
-            self.dbg_mem_access_fn.clone(),
-        )?;
```
syntactically, line 305: Why is this closure necessary?

syntactically (approved):
Looks good to me apart from some very minor comments.

I agree with @jsturtevant that there is some worry about whether code will continue compiling but do something different, which would be extremely unfortunate.  However, I would like to get this merged so that it can unblock other work that would ideally be (re)based on top of this, so I personally at least am happy to leave working out the exact details of what should be done there to a follow-up issue that we consider a blocker for release.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: Good idea. I will create an issue.

jprendes: #715

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: I want to make sure we call the cleanup code before returning and:
* try blocks are still not a thing
* a struct with a Drop impl would need to keep a reference to self while the code needs to use &mut self.

I want try blocks 😕

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
dblnz: I don't have strong preferences about this. I just commented as I reviewed it.
Good thing that it got solved.

PR #703, opened 2025-07-09
Making mshv3 the default mshv hypervisor in hyperlight_host (instead of mshv2)

marosset:
Setting `mshv3` as the default mshv hypervisor (instead of `mshv2`) in preparation for deprecating/removing mshv2 support.

### DIFF
--- a/.github/workflows/Benchmarks.yml
+++ b/.github/workflows/Benchmarks.yml
@@ -57,7 +57,7 @@ jobs:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Run Benchmarks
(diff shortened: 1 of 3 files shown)


syntactically:
If you are doing this without removing the mshv2 code, and want to leave in the mshv2 CI checks for now, I think you need to change the cfg aliases [here](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_host/build.rs#L102) to make it so that having both mshv2 and mshv3 enabled results in mshv2 support instead of mshv3 support, or you need to build with no-default-features any time you want mshv2 support (which sounds a bit unpleasant).

syntactically (approved):
LGTM, if it works everywhere. This will probably need dependent changes in all the downstream repos.

ludfjig (approved):

PR #709, opened 2025-07-10
Add memory mapping support with KVM

jprendes:
This PR adds memory mapping with KVM.
This means that we are now only missing support for memory mapping on Windows.

It also adds a smoke test for the supported platforms.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/kvm.rs
+++ b/src/hyperlight_host/src/hypervisor/kvm.rs
@@ -21,7 +21,7 @@ use std::sync::Arc;
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 
-use kvm_bindings::{KVM_MEM_READONLY, kvm_fpu, kvm_regs, kvm_userspace_memory_region};
+use kvm_bindings::{kvm_fpu, kvm_regs, kvm_userspace_memory_region};
 use kvm_ioctls::Cap::UserMemory;
 use kvm_ioctls::{Kvm, VcpuExit, VcpuFd, VmFd};
 use log::LevelFilter;
@@ -284,7 +284,8 @@ mod debug {
 /// A Hypervisor driver for KVM on Linux
 pub(crate) struct KVMDriver {
     _kvm: Kvm,
-    _vm_fd: VmFd,
+    vm_fd: VmFd,
+    page_size: usize,
     vcpu_fd: VcpuFd,
     entrypoint: u64,
     orig_rsp: GuestPtr,
@@ -317,21 +318,9 @@ impl KVMDriver {
 
         let vm_fd = kvm.create_vm_with_type(0)?;
 
-        let perm_flags =
-            MemoryRegionFlags::READ | MemoryRegionFlags::WRITE | MemoryRegionFlags::EXECUTE;
-
         mem_regions.iter().enumerate().try_for_each(|(i, region)| {
-            let perm_flags = perm_flags.intersection(region.flags);
-            let kvm_region = kvm_userspace_memory_region {
-                slot: i as u32,
-                guest_phys_addr: region.guest_region.start as u64,
-                memory_size: (region.guest_region.end - region.guest_region.start) as u64,
-                userspace_addr: region.host_region.start as u64,
-                flags: match perm_flags {
-                    MemoryRegionFlags::READ => KVM_MEM_READONLY,
-                    _ => 0, // normal, RWX
-                },
-            };
+            let mut kvm_region: kvm_userspace_memory_region = region.clone().into();
+            kvm_region.slot = i as u32;
             unsafe { vm_fd.set_user_memory_region(kvm_region) }
         })?;
 
@@ -378,7 +367,8 @@ impl KVMDriver {
         #[allow(unused_mut)]
         let mut hv = Self {
             _kvm: kvm,
-            _vm_fd: vm_fd,
+            vm_fd,
+            page_size: 0,
             vcpu_fd,
             entrypoint,
             orig_rsp: rsp_gp,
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -317,21 +318,8 @@ impl KVMDriver {
...
-                slot: i as u32,
-                guest_phys_addr: region.guest_region.start as u64,
-                memory_size: (region.guest_region.end - region.guest_region.start) as u64,
-                userspace_addr: region.host_region.start as u64,
-                flags: match perm_flags {
-                    MemoryRegionFlags::READ => KVM_MEM_READONLY,
-                    _ => 0, // normal, RWX
-                },
-            };
+            let kvm_region = mem_region_to_kvm_region(region, i);
```
syntactically: For mshv, we have an `impl From<MemoryRegion> for mshv_user_mem_region` in memory_region.rs. Not a big thing, but it would be nice to harmonize the kvm/mshv approaches, and either do that here as well or change the mshv one to be a similar named function.

jprendes:
I tried making the test a bit more complex, where the guest tries to access an mmap buffer that hasn't been mapped by the host.
That guest function call fails with `Err(Error("MMIO access address 0x100000000"))` as expected, but then the sandbox seem to be left in a bad state even after the state is restored from the last snapshot.

Subsequent call (a simple PrintOutput call in this case) fails with
```
Err(GuestAborted(8, "panicked at /home/jprendes/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/flatbuffers-25.2.10/src/endian_scalar.rs:142:5:\ninsufficient capacity for emplace_scalar, needed 1 got 0"))
```

I think there must be some sandbox state we aren't correctly restoring after a guest call fails with certain modes.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -693,4 +695,60 @@ mod tests {
...
+        let ptr = unsafe { std::alloc::alloc(layout) };
+
+        let mut vec = unsafe { std::vec::Vec::from_raw_parts(ptr, len, len) };
+        vec[0..src.len()].copy_from_slice(src);
+
+        vec
+    }
+
+    #[cfg(target_os = "linux")]
+    fn region_for_buffer(buff: &[u8], guest_base: usize) -> MemoryRegion {
```
syntactically: The specifics of how the memory model interacts with threads outside it are a little bit unclear, but I don't know that it's generally safe to have this reference last less long than the memory region is in the sandbox, since the current way things are the host can construct an &mut that refers to the region whilst it is mapped into the sandbox and being read there

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -693,4 +695,60 @@ mod tests {
...
+                "ReadMappedBuffer",
+                (guest_base as u64, expected.len() as u64),
+            )
+            .unwrap();
+
+        assert_eq!(data, expected);
+    }
+
+    #[cfg(target_os = "linux")]
+    fn page_aligned_buffer(src: &[u8]) -> Vec<u8> {
```
syntactically: Is this really safe? Vec is free to reallocate its buffer essentially whenever it pleases, isn't it?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: See  the comments in the shared memory file about why we try to avoid this generally

jprendes: Fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: Fixed

src/hyperlight_host/src/hypervisor/kvm.rs
jprendes: Done

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -693,4 +697,60 @@ mod tests {
...
+        let expected = b"hello world";
+        let data = page_aligned_memory(expected);
+        let guest_base = 0x1_0000_0000; // Arbitrary guest base address
+
+        unsafe {
+            sbox.map_region(&region_for_memory(&data, guest_base))
+                .unwrap();
+        }
+
+        let data: Vec<u8> = sbox
```
syntactically: As per the comment in shared_memory.rs, I think you need to have a `data.lock.try_read()?` guard live while in the hypervisor.

syntactically: Actually, it looks like this guard in other places was accidentally dropped in #533, so we need to fix that too (but that can be separate)

jprendes: Oh my, this is easy to misuse x_x
Fixed

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -308,3 +310,24 @@ impl From<MemoryRegion> for mshv_user_mem_region {
...
+
+        let perm_flags = perm_flags.intersection(region.flags);
+
+        kvm_userspace_memory_region {
+            slot: 0,
+            guest_phys_addr: region.guest_region.start as u64,
+            memory_size: (region.guest_region.end - region.guest_region.start) as u64,
+            userspace_addr: region.host_region.start as u64,
+            flags: match perm_flags {
+                MemoryRegionFlags::READ => KVM_MEM_READONLY,
```
syntactically, line 328: KVM_MEM_READONLY might be RX not R? But checking that can be a followup, since I think that predates this PR.

syntactically (approved):
LGTM

src/hyperlight_host/src/mem/memory_region.rs
jprendes: looks like there are only two possible flags:
```
/* for kvm_userspace_memory_region::flags */
#define KVM_MEM_LOG_DIRTY_PAGES       (1UL << 0)
#define KVM_MEM_READONLY      (1UL << 1)
```

so, IIUC, looks like there's no way of restricting execution.

syntactically:
> I think there must be some sandbox state we aren't correctly restoring after a guest call fails with certain modes.

I think that we never actually tested restore after an abort? Tracked in #710

PR #712, opened 2025-07-11
Fix Windows build with `--no-default-features`

### LINKED ISSUE #685: hyperlight_host doesn't build on Windows without init-paging feature enabled
### What happened?

Sandbox creation code for WHP has a build error if init-paging feature is not enabled

Sandbox creation for WHP should build with all optional features disabled

include hyperlight_host v0.7.0 in another project and build on Windows

### Hyperlight Version

0.7.0

### OS version

<details>

```console
On Linux:
$ cat /etc/os-release
# paste output here

$ uname -a
# paste output here

On Windows:
C:\> cmd /c ver
# paste output here

````
</details>

### Additional Information

mgr.shared_mem.raw_mem_size() is referenced at https://github.com/hyperlight-dev/hyperlight/blob/c725c8bdbdb575fd97e269647e6f3ab05684a213/src/hyperlight_host/src/sandbox/uninitialized_evolve.rs#L245-L255 when target_os = Windows

BUT

crate::mem::shared_mem::SharedMemory is only used when init-paging feature is enabled https://github.com/hyperlight-dev/hyperlight/blob/c725c8bdbdb575fd97e269647e6f3ab05684a213/src/hyperlight_host/src/sandbox/uninitialized_evolve.rs#L36-L37

This causes build issues when `hyperlight_host`'s build.rs script runs when the crate is included in other projects (like hyperlight-wasm) AND init-paging feature is not enabled.

### PR DESCRIPTION
**Pull request by danbugs**
Fixes #685 and updates CI to ensure Windows builds successfully without default features.

---

Btw, to validate this fix, I tested running Nanvix (which uses `--no-default-features`) on Windows as a guest. This revealed that additional register setup was needed in `hyperv_windows.rs`:
- Mirror the CS register setup used with default features, except without the L bit
- Explicitly set the `Limit` parameter for CS to `0xFFFF` (required when paging isn't initialized)

These changes ensure Windows Hyper-V properly initializes guest VMs that don't use the init-paging feature. Same idea for Linux Hyper-V too.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
@@ -479,6 +479,28 @@ impl HypervLinuxDriver {
             cs: SegmentRegister {
                 base: 0,
                 selector: 0,
+                limit: 0xFFFF,
+                type_: 11,
+                present: 1,
+                s: 1,
+                ..Default::default()
+            },
+            ds: SegmentRegister {
+                base: 0,
+                selector: 0,
+                limit: 0xFFFF,
+                type_: 3,
+                present: 1,
+                s: 1,
+                ..Default::default()
+            },
+            tr: SegmentRegister {
+                base: 0,
+                selector: 0,
+                limit: 0xFFFF,
+                type_: 11,
+                present: 1,
+                s: 0,
                 ..Default::default()
             },
             ..Default::default()

### CODE REVIEW

**Review by ludfjig — DISMISSED** [review]
Looks good! I think we should strive to set the same registers on all hypervisors, so we don't rely on any inherent defaults of a particular hv. Thoughts?

**Comment by danbugs** [comment]
> Looks good! I think we should strive to set the same registers on all hypervisors, so we don't rely on any inherent defaults of a particular hv. Thoughts?

Probably a good idea—we should also avoid relying on `Default` and be a little more explicit. For example, on KVM, the default CS register (excluding base & selector fields) looks like this:
```
CS: limit=0xffff, type=0xa, s=1, dpl=0, p=1, avl=0, l=0, db=0, g=0
```

This is exactly the setup we needed w/o `default-features` to run a workload like Nanvix and why it worked from the get go. But, on WHP, `Default` just zeroes out everything :|

Might be a good work item for #469?

File: src/hyperlight_host/src/hypervisor/hyperv_windows.rs [hunk 1]
```diff
@@ -399,16 +402,22 @@ impl HypervWindowsDriver {
         ])?;
 
         #[cfg(not(feature = "init-paging"))]
-        proc.set_registers(&[(
```
syntactically, line 402: Does this change need to be mirrored on the Linux hypervisor drivers?

Justfile
```diff
@@ -121,10 +121,12 @@ test-seccomp target=default-target features="":
     cargo test --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host test_violate_seccomp_filters --lib {{ if features =="" {''} else { "--features " + features } }} -- --ignored
     cargo test --profile={{ if target == "debug" { "dev" } else { target } }} -p hyperlight-host test_violate_seccomp_filters --no-default-features {{ if features =~"mshv3" {"--features init-paging,mshv3"} else {"--features mshv2,init-paging,kvm" } }} --lib -- --ignored
 
-# runs tests that ensure compilation fails when it should
-test-compilation-fail target=default-target:
-    @# the following should fail on linux because one of kvm, mshv, or mshv3 feature must be specified, which is why the exit code is inverted with an !.
-    {{ if os() == "linux" { "! cargo check -p hyperlight-host --no-default-features 2> /dev/null"} else { "" } }}
+# tests compilation with no default features on different platforms
+test-compilation-no-default-features target=default-target:
+    @# Linux should fail without a hypervisor feature (kvm, mshv, or mshv3)
```
syntactically, line 126: Should we also be checking that compilation with a hypervisor driver but without init-paging succeeds on Linux?

dblnz (approved):
LGTM.

I agree that we need a common way of to set the register, independent of the features enabled.

marosset (approved):

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
danbugs: Yep—also had to change the sregs setup for hyperv_linux (done in [97b3e43](https://github.com/hyperlight-dev/hyperlight/pull/712/commits/97b3e4380ef19fb839f8d3bd35bcc08f7023766b)). Good catch!

Justfile
danbugs: Done ✅

dblnz (approved):

PR #713, opened 2025-07-14
Add a scratch-backed user data region for host/guest byte exchange

qbradley:
Problem:
Hyperlight embedders that move larger byte payloads across the host/guest boundary currently have to choose between two poor fits. The normal function-call path serializes parameters and returns through FlatBuffers-backed input/output stacks, which is a good contract for structured arguments but adds serialization/deserialization and stack-buffer overhead for large byte arrays. The init-data/user-memory path is also not appropriate as a mutable per-call exchange buffer: init data belongs to snapshot-backed guest state, and after copy-on-write snapshots the live guest-written bytes may no longer be readable by a simple host memcpy from the original shared region.

The missing primitive was an opt-in, bounded, host-addressable scratch buffer whose lifecycle matches transient per-call data: the host can write bytes, the guest can read or mutate them in place, the host can read the result, and restore clears the buffer because scratch is not snapshot state.

What changed:
This change adds a User Data region as a fixed-capacity slab in scratch memory, placed after the existing input and output buffers and before the copied page-table / allocator area. The capacity is configured with `SandboxConfiguration::set_user_data_size`, defaults to zero, participates in scratch minimum-size checks, and is included in layout compatibility so snapshots cannot be restored into a sandbox with a mismatched user-data capacity.

The region is advertised to guests through the PEB as a `GuestMemoryRegion`. Host code gets `MultiUseSandbox::user_data_size`, `write_user_data`, and `read_user_data`. Rust guests get capacity/pointer helpers plus bounded read/write helpers and non-allocating borrowed access helpers for larger payloads. C guests get `hl_user_data_size()` and `hl_user_data_ptr()`, intentionally exposing pointer-and-capacity metadata for caller-bounded raw access.

The implementation preserves the default behavior for existing users: capacity zero means no usable region, no additional public action is required, and the input/output buffer layout is unchanged except for the appended zero-sized user-data descriptor. When a positive capacity is configured, scratch sizing must be large enough to hold input, output, user data, fixed scratch overhead, and copied page tables.

Lifecycle and API contract:
The supported exchange pattern is: host writes bytes with `write_user_data`, guest code reads or mutates the region during a mutating/non-restoring call, then the host reads bytes with `read_user_data` before any restore. Restore clears scratch, so user data is cleared both on explicit restore and through convenience call paths that snapshot/restore around the call.

Host reads and writes start at offset zero and are whole-buffer operations from the caller's perspective. Oversized reads/writes fail before accessing adjacent scratch memory. Successful shorter writes intentionally preserve bytes after `data.len()` rather than clearing the tail. This keeps the region a raw application-managed buffer; callers that use variable-length payloads must track logical length or explicitly clear unused bytes when stale tail bytes would matter.

`MultiUseSandbox::from_snapshot` treats user-data capacity as a layout compatibility field rather than as an ignored runtime override. If the caller supplies a config whose user-data capacity differs from the snapshot layout, construction fails with the existing snapshot-layout mismatch error. This matches restore behavior and avoids surprising reuse of snapshots under incompatible raw-buffer contracts.

Security considerations and justification:
The feature stays within Hyperlight's hypervisor isolation model: the guest still only sees memory already mapped into the VM, and the host APIs bound all host copies by configured capacity. Rust guest helpers also check the PEB-advertised capacity before producing borrowed or owned access. The C API is deliberately lower-level: it exposes a raw pointer plus capacity and documents that C callers must bound their own memory operations.

The important security model change is that user data is a raw byte channel, not a FlatBuffers-verified call/return message. Bytes read by the host from user data must be treated as guest-controlled and validated by the embedder's application protocol before influencing memory access, host function calls, authorization decisions, or other security-sensitive behavior. The security docs and developer guidance now call this out explicitly.

Because short writes preserve the tail, stale bytes can leak between logical requests if an embedder reuses a sandbox and reads beyond the current logical payload length. This is an intentional raw-buffer semantic, but it is documented and tested. Embedders with tenant/request boundaries should restore, clear the full region, or maintain an application-level length field and never expose bytes past it.

The feature does not add VM/page-level guard isolation around user data. That was considered, but it would require separate mappings or page-permission changes rather than reusing the existing scratch slot. For the first release, the chosen security boundary is helper/API-level bounds over scratch memory, matching Hyperlight's existing scratch-buffer model.

Performance comparison:
A sibling sample-workload benchmark was added for the existing 24 KiB input / 8 KiB output workload. The original benchmark passes a 24 KiB byte vector through the FlatBuffers call path and returns an 8 KiB byte vector. The new sibling benchmark writes the 24 KiB input through `write_user_data`, calls a guest function that reads the region in place and reports an 8 KiB logical output, then reads the 8 KiB result with `read_user_data`.

A quick local Criterion run with `cargo bench -p hyperlight-host --bench benchmarks -- sample_workloads --warm-up-time 1 --measurement-time 2 --sample-size 10` showed:

    - FlatBuffers C guest: about 89.9 microseconds per call
    - FlatBuffers Rust guest: about 90.5 microseconds per call
    - User data C guest: about 69.2 microseconds per call
    - User data Rust guest: about 72.4 microseconds per call

That run shows the user-data path roughly 20-23 microseconds faster for this sample workload. The exact number is environment-dependent, but the direction matches the design goal: avoid FlatBuffers serialization of large byte payloads while keeping the unavoidable VM call overhead.

Alternatives considered:
1. Keep using init data with CoW-aware host reads. This was rejected because init data is snapshot-backed guest state, not transient per-call I/O. Making host reads follow CoW page-table state would add page walking and CR3/cache complexity, and would continue to blur the semantic boundary between snapshot state and scratch exchange data.
2. Exclude or special-case GuestBlob/init data from snapshots. This was rejected as too invasive: init data lives in the primary snapshot region, and carving out a writable non-snapshotted subregion would affect memory layout, snapshot construction, restore behavior, compatibility checks, and VM mappings for all users.
3. Use writable `map_region` or a separate mapping slot. This was rejected for the first cut because writable mappings are not the existing supported pattern, require separate lifecycle and memory-slot handling, and are heavier than reusing scratch memory that is already host-addressable and guest-visible.
4. Split the feature into two directional raw regions: host-write/guest-read and guest-write/host-read. This was considered in planning and rejected for this change because the primary use case is in-place transformation of a shared byte buffer. The split design gives clearer ownership for request/response protocols but doubles configuration, layout, PEB, helper, compatibility, and test surface area; it also still would not provide VM-enforced directional permissions without a larger mapping redesign.
5. Zero-fill the tail on every successful short host write. This was rejected in favor of raw buffer semantics. Preserving the tail avoids hidden work and lets applications manage subregion layout themselves. The trade-off is documented clearly: callers must track logical length or explicitly clear when stale bytes matter.
6. Provide only an allocating Rust guest read helper. This was rejected for large/hot paths after review. The final Rust guest surface includes non-allocating closure-based accessors so guests can inspect or mutate large configured regions without allocating a full `Vec` in the guest heap; the allocating helper remains as a convenience for smaller reads.

Validation:
The implementation includes configuration and layout tests for zero, one-byte, non-page-aligned 4097-byte, 64 KiB, and 1 MiB capacities. Host memory-manager tests cover fresh-zero reads, exact read/write, oversized read/write rejection, atomic oversized-write failure, and preserved tail behavior after short writes. Integration tests exercise Rust and C guests for capacity discovery, guest-visible zeroes, host-to-guest-to-host mutation, half-capacity mutation, capacity+1 rejection, and failed-call semantics. Restore tests cover explicit restore clearing, deprecated convenience-call restore clearing, capacity mismatch rejection, and `from_snapshot` behavior.

Verification performed during the change included formatting with the repository rustfmt toolchain, debug and release builds, all-target/all-feature clippy, Rust and C guest builds in debug and release, targeted debug and release user-data tests, i686 checks, and a final Society-of-Thought review with follow-up fixes for snapshot compatibility, stale-tail documentation, guest capacity+1 coverage, non-allocating guest helpers, scratch-sizing documentation, and security guidance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Prerelease] - Unreleased
 
+### Added
+* User data region APIs for transient host/guest byte exchange, including host read/write methods and Rust/C guest discovery helpers.
+
 ## [v0.15.0] - 2026-05-06
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ fn say_hello(name: String) -> Result<String> {
 }
 ```
 
-To get started, see the [Getting Started](./docs/getting-started.md) guide. For more details on writing guests, see [How to build a Hyperlight guest binary](./docs/how-to-build-a-hyperlight-guest-binary.md).
+To get started, see the [Getting Started](./docs/getting-started.md) guide. For more details on writing guests, see [How to build a Hyperlight guest binary](./docs/how-to-build-a-hyperlight-guest-binary.md). For execution internals, including the user data region, see [How code gets executed in a VM](./docs/hyperlight-execution-details.md).
 
 ## When to use Hyperlight
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -26,7 +26,7 @@ This project is composed internally of several components, depicted in the below
 
 * [Getting Started](./getting-started.md)
 * [Glossary](./glossary.md)
-* [How code gets executed in a VM](./hyperlight-execution-details.md)
+* [How code gets executed in a VM](./hyperlight-execution-details.md), including the user data region
 * [How to build a Hyperlight guest binary](./how-to-build-a-hyperlight-guest-binary.md)
 * [Security considerations](./security.md)
 * [Technical requirements document](./technical-requirements-document.md)
--- a/docs/hyperlight-execution-details.md
+++ b/docs/hyperlight-execution-details.md
@@ -35,6 +35,18 @@ At the highest level, Hyperlight takes roughly the following steps to create and
    1. In the former case, exit successfully
    2. In any of the latter cases, exit with a failure message
 
+## User data region
+
+Sandboxes can optionally reserve a user data region for transient host/guest byte exchange. The default capacity is zero; embedders configure a positive capacity with `SandboxConfiguration::set_user_data_size`. Fresh regions read as zeroes, and larger capacities may require increasing scratch size to satisfy the sandbox's existing scratch memory limits.
+
+When configured, the region is advertised to the guest through the PEB and can be accessed from the host with `MultiUseSandbox::write_user_data`, `MultiUseSandbox::read_user_data`, and `MultiUseSandbox::user_data_size`. Host reads and writes start at the beginning of the region and fail if the caller's buffer exceeds the configured capacity. Successful shorter writes preserve existing tail bytes, so callers should track logical payload length or clear unused bytes explicitly.
+
+The region is scratch memory, so it is not captured in snapshots. A restore clears the region, including restore performed by convenience call paths that restore the sandbox before returning to the host. Snapshot restore compatibility includes the configured capacity and reports the existing layout-mismatch error for capacity mismatches; `MultiUseSandbox::from_snapshot` rejects a caller-supplied configuration whose user data capacity differs from the snapshot layout. The intended exchange pattern is host write, mutating guest call, then host read before restore; after a failed guest call, callers should treat region contents as application-defined unless their protocol defines a successful handoff.
+
+Rust guest helpers expose bounded access to the region, including non-allocating borrowed access for larger payloads. The C guest API exposes `hl_user_data_size()` and `hl_user_data_ptr()`; C callers must use the reported size to bound copies. These APIs are helper-level bounds, not VM guard-page isolation around the region.
+
+Representative tests cover 4097-byte, 64 KiB, and 1 MiB capacities across Rust and C guests. Larger capacities should be measured for the application's expected copy and restore costs.
+
 ---
 
 _<sup>[1]</sup> nearly universal support_
--- a/docs/security-guidance-for-developers.md
+++ b/docs/security-guidance-for-developers.md
@@ -10,6 +10,7 @@ This document discusses the security requirements and best practices for service
 * All host functions that receive parameters from a guest, or operate indirectly on guest data _MUST_ be continuously fuzzed
 * Host functions _MUST NOT_ call APIs or be used to expose functionality deemed risky in a multi-tenant context
 * Guests and host processes _MUST_ use the same version of a FlatBuffer definition
+* Raw user data region bytes _MUST_ be treated as guest-controlled and validated by the host protocol before use
 
 More detailed guidance on the requirements and best practices is detailed below.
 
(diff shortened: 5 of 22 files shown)


syntactically:
Is the performance motivation for this basically due to the facts that (a) function parameters always get copied more than they ought to right now and/or (b) the lifecycle of the function parameters doesn't match the lifecycle of the data that you need, since currently it is tied directly to the call stack? If so, I  would like to take a little bit of a look into whether the same need can be served by removing some of the copies (we have some preliminary work in #444) and changing the call ABI to be more flexible about how function parameters/returns are allocated (roughly: the guest can allocate buffers for these on its own, and manage their lifecycle itself), which is also a requirement for supporting native async in the host/guest ABI. It would be nice to eventually replace the entire special-case "init data" at sandbox initialization with a more generic mix of features (memory mapping for zerocopy readonly data, this kind of thing for longer-lived read-write data where you can hand off ownership between the host and guest, etc).

The whole call context abstraction is hopefully going away in #697, so it also might be worth waiting to move on this until that is merged?

qbradley:
I do not know with certainty that extra copies is the only cause of the overhead, but per byte cost of parameters is currently high, especially on /dev/mshv.  In my microbencharks, I see a guest call with 16 bytes of parameters taking around 100 microseconds, while with 32KB of parameters, it takes over 200 microseconds. However by copying the bulk data using init-data using above, 32KB takes around 112 microseconds.

With the above, I still need an extra copy because on the rust side I prepare a buffer, then copy it into the user-memory area using rewrite_init_data().  In the guest side, I access the pointer to the user-memory to interpret the data. Ideally on the host side I could also prepare the input buffer directly to user-memory rather than using an intermediate buffer.

I could imagine a shared-memory API that allows me to map memory into the host and the guest and then I would not need this, but meanwhile this change gets me a large step in the direction of high throughput input/output through my guest.

syntactically:
> I do not know with certainty that extra copies is the only cause of the overhead, but per byte cost of parameters is currently high, especially on /dev/mshv. In my microbencharks, I see a guest call with 16 bytes of parameters taking around 100 microseconds, while with 32KB of parameters, it takes over 200 microseconds. However by copying the bulk data using init-data using above, 32KB takes around 112 microseconds.

That makes sense. It would be interesting to see how that shakes out with some of the changes to parameters that we are working on right now. In addition to the PR that I linked above, I am looking a little bit at making parameter buffer allocation happen in the guest with a more sensible ABI as part of something else, and I think a general push to improve parameter passing to the point of being useful is underway.

> With the above, I still need an extra copy because on the rust side I prepare a buffer, then copy it into the user-memory area using rewrite_init_data(). In the guest side, I access the pointer to the user-memory to interpret the data. Ideally on the host side I could also prepare the input buffer directly to user-memory rather than using an intermediate buffer.

We would probably need to be a little careful with the API design on this before doing anything like that in the general case, because of the many footguns around toctou, etc. which come from exposing an API that reads incrementally from a region which the guest is still able to mutate. So I would like to look at achieving single-copy first, and then look into ways to add various forms of zerocopy that are a bit safer than just "here's a pointer to a buffer that the guest is mutating, parse it as you will".

qbradley:
For our use case, we do not modify the "shared" memory while the guest is running.  The host writes to the shared memory, then makes a call into the guest who examines the memory and then updates it and returns to the host.  It is my understanding (please correct me if I am wrong) that you no longer use a separate thread to run the guest.  The host then examines the shared memory. For our specific use case, I think this mitigates a lot of the potential problems. I can understand how a general API has to handle everything someone might try to do. Another assumption I have is that a hypercall to the guest should be equivalent to a full fence, like a syscall is, but please also correct me if I am wrong on that one.

syntactically:
> For our use case, we do not modify the "shared" memory while the guest is running. The host writes to the shared memory, then makes a call into the guest who examines the memory and then updates it and returns to the host. It is my understanding (please correct me if I am wrong) that you no longer use a separate thread to run the guest. The host then examines the shared memory. For our specific use case, I think this mitigates a lot of the potential problems. 

In that case, I agree that zerocopy is probably safer, but I just want to be cautious with adding API to make sure we don't do something that encourages unsafe usage in general, beyond this one application.  Out of curiosity, is your data structured in any interesting ways processed by the host side, or is it just bulk data?

> Another assumption I have is that a hypercall to the guest should be equivalent to a full fence, like a syscall is, but please also correct me if I am wrong on that one.

I think we are making the same assumption right now, and it is at least generally true.

qbradley:
It is a moderately complex structure that must be parsed. It is a sequence of variable length structures defined by initial tag byte.

qbradley:
I have pushed a new implementation based on the scratch area and with a more complete API. Your feedback would be appreciated.

PR #732, opened 2025-07-24
Removed the OutBHandler and MemAccessHandler abstractions and related implementations.

### LINKED ISSUE #495: Remove `OutBHandlerWrapper` type

simongdavies:
This pull fixes #495. It removes the OutBHandler and MemAccessHandler related traits, types and structs. 

Since there is only a single implementation the functions that these types abstracted over (`handle_outb` and `handle_mem_access`) the code can be simplified by calling these functions directly.

One implication of this change is that the `Hypervisor` now contains a `MemMgrWrapper<T>` which ultimately contains a `HostMapping` instance. Since the Hypervisor is `Send + Sync` this change would require `HostMapping` to be `Send + Sync` which is not automatically implemented as it contains a `*mut u8 type`. The solution that this PR implements is to remove the `Sync` constraint from the Hypervisor trait since its not possible to access a Sandbox concurrently from multiple threads. 

There is probably some code in `shared_mem` that on the face of it does not seem necessary without multi threaded access but in the future when async guest io is implemented will still be needed so that is the only change made.

### DIFF
--- a/src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
+++ b/src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
@@ -27,7 +27,6 @@ use super::mem_access::dbg_mem_access_handler_wrapper;
 use super::uninitialized::SandboxRuntimeConfig;
 use crate::HyperlightError::NoHypervisorFound;
 use crate::hypervisor::Hypervisor;
-use crate::hypervisor::handlers::{MemAccessHandlerCaller, OutBHandlerCaller};
 use crate::mem::exe::LoadInfo;
 use crate::mem::layout::SandboxMemoryLayout;
 use crate::mem::mgr::SandboxMemoryManager;
@@ -41,8 +40,6 @@ use crate::sandbox::TraceInfo;
 #[cfg(gdb)]
 use crate::sandbox::config::DebugInfo;
 use crate::sandbox::host_funcs::FunctionRegistry;
-use crate::sandbox::mem_access::mem_access_handler_wrapper;
-use crate::sandbox::outb::outb_handler_wrapper;
 use crate::sandbox::{HostSharedMemory, MemMgrWrapper};
 #[cfg(target_os = "linux")]
 use crate::signal_handlers::setup_signal_handlers;
@@ -69,8 +66,6 @@ where
         Arc<Mutex<FunctionRegistry>>,
         MemMgrWrapper<HostSharedMemory>,
         Box<dyn Hypervisor>,
-        Arc<Mutex<dyn OutBHandlerCaller>>,
-        Arc<Mutex<dyn MemAccessHandlerCaller>>,
         RawPtr,
     ) -> Result<ResSandbox>,
 {
@@ -82,7 +77,6 @@ where
         &u_sbox.rt_cfg,
         u_sbox.load_info,
     )?;
-    let outb_hdl = outb_handler_wrapper(hshm.clone(), u_sbox.host_funcs.clone());
 
     let seed = {
         let mut rng = rand::rng();
@@ -94,7 +88,6 @@ where
     };
 
     let page_size = u32::try_from(page_size::get())?;
-    let mem_access_hdl = mem_access_handler_wrapper(hshm.clone());
 
     #[cfg(gdb)]
     let dbg_mem_access_hdl = dbg_mem_access_handler_wrapper(hshm.clone());
@@ -106,8 +99,8 @@ where
         peb_addr,
         seed,
         page_size,
-        outb_hdl.clone(),
-        mem_access_hdl.clone(),
+        hshm.clone(),
+        u_sbox.host_funcs.clone(),
         u_sbox.max_guest_log_level,
         #[cfg(gdb)]
         dbg_mem_access_hdl,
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -601,13 +609,22 @@ impl Hypervisor for HypervLinuxDriver {
...
         self.vcpu_fd.set_regs(&regs)?;
 
-        VirtualCPU::run(
+        // Extract mem_mgr to avoid borrowing conflicts
+        let mem_mgr = self
+            .mem_mgr
+            .take()
+            .ok_or_else(|| new_error!("mem_mgr should be initialized"))?;
+
+        let result = VirtualCPU::run(
```
syntactically, line 618: I wonder if there is a way to simplify the somewhat windy code flow around here and remove the need to take the mem mgr option out, but perhaps that's for another day.

src/hyperlight_host/src/hypervisor/handlers.rs
```diff
@@ -16,118 +16,11 @@ limitations under the License.
...
-        func()
-    }
-}
+use crate::Result;
 
 /// The trait representing custom logic to handle the case when
 /// a Hypervisor's virtual CPU (vCPU) informs Hyperlight a debug memory access
 /// has been requested.
-#[cfg(gdb)]
 pub trait DbgMemAccessHandlerCaller: Send {
```
syntactically, line 24: It looks like this trait also only has one (non-test) `impl`. Does it make sense to do the same thing as with the others and remove it?

Also, why remove the `cfg(gdb)` guards?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -688,22 +703,47 @@
...
-                self,
-                port,
-                val,
-            )?;
+        #[cfg(feature = "trace_guest")]
+        {
+            // We need to handle the borrow checker issue where we need both:
+            // - &mut MemMgrWrapper (from self.mem_mgr.as_mut())
+            // - &mut dyn Hypervisor (from self)
+            // We'll use a temporary approach to extract the mem_mgr temporarily
```
syntactically, line 717: Is there a reason to cfg gate this instead of doing the same thing in both cases (e.g. a measurable performance impact?)

syntactically (approved):
> The solution that this PR implements is to remove the Sync constraint from the Hypervisor trait since its not possible to access a Sandbox concurrently from multiple threads.

That seems like a reasonable constraint, especially given the performance cliffs that we see with accessing a KVM partition from multiple threads (and how easy it seems to be for people to run into that).  Is there a compelling reason not to make the `Hypervisor` trait non-`Send` as well, at least for the temporary future?

I do think that in general, we are going to need to be able to split up the actual hypervisor handle and the access to the host side of the memory mapping, in order to support the async use case, where the hypervisor handler(s) might (at least sometimes) need to be in a different thread running the guest vcpu whilst the host is also accessing memory buffers for async parameter passing.  However, that is something we can address later.

> There is probably some code in shared_mem that on the face of it does not seem necessary without multi threaded access but in the future when async guest io is implemented will still be needed so that is the only change made.

Thanks for leaving this in; indeed some of this was preemptively written.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
simongdavies: Agreed, I think memory management in the host needs a good refactor but I also think we should wait and possibly do this as part of the wider snapshot restore work.

src/hyperlight_host/src/hypervisor/handlers.rs
simongdavies: Yes, I was intending to do this as a separate PR, opened #743.

The `#[cfg(dbg)]` is now on the mod definition in mod.rs as all the code in the module is for dbg only.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
simongdavies: Nothing more than it only needed to be done for that feature.

simongdavies:
> That seems like a reasonable constraint, especially given the performance cliffs that we see with accessing a KVM partition from multiple threads (and how easy it seems to be for people to run into that). Is there a compelling reason not to make the `Hypervisor` trait non-`Send` as well, at least for the temporary future?

This wont solve the KVM perf issue since that occurs when the Sandbox is sent to another thread and then used there. I dont think we can remove that since the pattern that a few folks are using is to create a sandbox outside the hotpath, cache it and then use it on a request (which more than likely is a different thread)

syntactically:
> I dont think we can remove that since the pattern that a few folks are using is to create a sandbox outside the hotpath, cache it and then use it on a request (which more than likely is a different thread)

I think to get good perf with KVM you would need to have one of those caches per thread?

We could make the sandbox !Send only on kvm builds, but that is probably a bit too much of a headache for people who deploy with other hypervisors but develop with kvm (and don't care that much about the development performance), I guess.

PR #734, opened 2025-07-25
Add macros feature to hyperlight_guest_tracing to prepare for publishing the crate

dblnz:
To be able to use the tracing functionality as a dependency for a guest, we need to publish the crates.
It would be best if we could only publish one crate: `hyperlight-guest-tracing`.

This PR adds a `macros` feature to the `hyperlight-guest-tracing` crate that re-exports the macros from `hyperlight-guest-tracing-macro` crate so the users don't need to have two dependencies.

### DIFF
--- a/src/hyperlight_guest_tracing_macro/src/lib.rs
+++ b/src/hyperlight_guest_tracing_macro/src/lib.rs
@@ -48,15 +48,6 @@ pub fn trace_function(_attr: TokenStream, item: TokenStream) -> TokenStream {
     let entry_msg = format!("> {}", fn_name_str);
     let exit_msg = format!("< {}", fn_name_str);
 
-    if entry_msg.len() > hyperlight_guest_tracing::MAX_TRACE_MSG_LEN
-        || exit_msg.len() > hyperlight_guest_tracing::MAX_TRACE_MSG_LEN
-    {
-        panic!(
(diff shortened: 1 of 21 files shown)


simongdavies (commented):
Why is macros optional? What are the circumstances where I would use tracing and not macros?

simongdavies (approved):

ludfjig (approved):

src/hyperlight_guest_tracing_macro/src/lib.rs
```diff
@@ -103,17 +104,6 @@ impl syn::parse::Parse for TraceMacroInput {
         if !matches!(message, syn::Lit::Str(_)) {
             return Err(input.error("first argument to trace! must be a string literal"));
         }
-        if let syn::Lit::Str(ref lit_str) = message {
-            if lit_str.value().is_empty() {
-                return Err(input.error("trace message must not be empty"));
-            }
```
jprendes, line 109: **question:** Where did the "trace message must not be empty" check move to?

jprendes:
> It would be best if we could only publish one crate: hyperlight-guest-tracing.

We would still need to publish both, as `hyperlight-guest-tracing` depends on `hyperlight-guest-tracing-macros` and you can't have unpublished dependencies in crates.io

PR #740, opened 2025-07-29
[fix] Make sure memory is not mapped writeable into sandbox in kvm

ludfjig:


simongdavies (dismissed):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -620,6 +624,47 @@ mod tests {
         assert_eq!(actual, expected);
     }
 
+    // Makes sure MemoryRegionFlags::READ | MemoryRegionFlags::EXECUTE is not writable
```
syntactically: Do you want to also add a test that it is in fact executable? That might be covered by hyperlight-wasm using it, but it might be nice to have confirmation in here.

syntactically (dismissed):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Fixed

syntactically (approved):
LGTM

PR #742, opened 2025-07-30
Make snapshot region aware

ludfjig:


src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -281,6 +259,30 @@ pub(crate) trait Hypervisor: Debug + Sync + Send {
     fn trace_info_as_mut(&mut self) -> &mut TraceInfo;
 }
 
+/// Returns a Some(HyperlightExit::AccessViolation(..)) if the given gpa doesn't have
```
syntactically, line 264: Is this being moved just general cleanup because it wasn't using self?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -95,18 +96,35 @@ impl MultiUseSandbox {
...
     #[instrument(err(Debug), skip_all, parent = Span::current())]
     pub fn restore(&mut self, snapshot: &Snapshot) -> Result<()> {
-        let rgns_to_unmap = self
-            .mem_mgr
+        self.mem_mgr
             .unwrap_mgr_mut()
             .restore_snapshot(&snapshot.inner)?;
-        unsafe { self.vm.unmap_regions(rgns_to_unmap)? };
+
+        let current_regions: HashSet<_> = self.vm.get_mapped_regions().cloned().collect();
```
syntactically, line 107: Does it make sense to just store these as a hashset in the hypervisor trait implementation as well, so that removing one does not require iterating through them all?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -515,25 +523,32 @@ impl Hypervisor for KVMDriver {
...
-    unsafe fn unmap_regions(&mut self, n: u64) -> Result<()> {
-        let n_keep = self.mem_regions.len() - n as usize;
-        for (k, region) in self.mem_regions.split_off(n_keep).iter().enumerate() {
-            let mut kvm_region: kvm_userspace_memory_region = region.clone().into();
-            kvm_region.slot = (n_keep + k) as u32;
+    unsafe fn unmap_region(&mut self, region: &MemoryRegion) -> Result<()> {
+        if let Some(idx) = self.mmap_regions.iter().position(|(r, _)| r == region) {
+            let (region, slot) = self.mmap_regions.remove(idx);
+            let mut kvm_region: kvm_userspace_memory_region = region.into();
+            kvm_region.slot = slot;
```
syntactically, line 553: Do the slot indices need to be reclaimed? I know there is a limit (KVM_CAP_NR_MEMSLOTS) on the maximum id that can be used.

syntactically (dismissed):

simongdavies:
> Will add a follow up pr for making sure snapshots "belong" to a particular sandbox, and can not be used to restore different sandbox.

Can we get this done before the next release?

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Yes, and I wanted it take an iterator which wasn't possible if it's on the trait

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: You're right, this could be in the hot path. Fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I thought about this, but I don't think so, because KVM will need to store <(MemoryRegion, slot)>, so will need to copy regardless to make it into HashSet/Vec of `<MemoryRegion>`. Maybe it's something we can improve later

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -1014,7 +1026,7 @@ impl Hypervisor for HypervLinuxDriver {
             });
 
             Ok(Some(crashdump::CrashDumpContext::new(
-                &self.mem_regions,
+                &self.sandbox_regions,
```
dblnz, line 1014: Should we add the `mmap`ed regions also to the crashdump? I think there's some additional logic needed to make that work

syntactically (approved):
LGTM

dblnz (approved):
LGTM!

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Yes probably, but didn't want to do it in this PR

ludfjig: #747

simongdavies (approved):

PR #748, opened 2025-07-30
Adds blocker to release if any issue has release-blocker label

### LINKED ISSUE #716: Add release blocker tag and block publishing workflows
Some issues, like #714, really should be resolved before we cut a release. We should add a tag that blocks our release workflow to make sure that it's impossible for them to slip through the cracks.

simongdavies:
If any open issue has the label "release-blocker" don't create a release or publish crates.

Fixes #716

### DIFF
--- a/.github/workflows/ReleaseBlockerLabelCleanUp.yml
+++ b/.github/workflows/ReleaseBlockerLabelCleanUp.yml
@@ -1,0 +1,31 @@
+name: Release Blocker Cleanup
+
+on:
+  issues:
+    types: [closed]
+
+permissions:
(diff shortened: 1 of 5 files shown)


.github/workflows/ReleaseBlockerLabelCleanUp.yml
dblnz, line 1: Where is this workflow used?

simongdavies: Its triggered whenever an issue is closed, it doesn't really need to be done since the check only considered for open issues

PR #749, opened 2025-07-31
Fix some cargo docs

ludfjig:
Fixes some outdata docs, adds some docs where it was missing...


src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -154,13 +165,11 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
-    /// Create a new sandbox configured to run the binary at path
-    /// `bin_path`.
+    /// Creates a new uninitialized sandbox for the given guest environment.
     ///
-    /// The instrument attribute is used to generate tracing spans and also to emit an error should the Result be an error.
-    /// The skip attribute is used to skip the guest binary from being printed in the tracing span.
-    /// The name attribute is used to name the tracing span.
-    /// The err attribute is used to emit an error should the Result be an error, it uses the std::`fmt::Debug trait` to print the error.
+    /// The guest binary can be provided as either a file path or memory buffer.
+    /// An optional configuration can customize memory sizes and sandbox settings.
```
simongdavies, line 171: nit: It may be worth saying something about the primary use for this is allow the registration of host functions before initialization?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -327,10 +338,10 @@ impl UninitializedSandbox {
         register_host_function(host_func, self, name.as_ref(), Some(extra_allowed_syscalls))
     }
 
-    /// Register a host function named "HostPrint" that will be called by the guest
-    /// when it wants to print to the console.
-    /// The "HostPrint" host function is kind of special, as we expect it to have the
-    /// `FnMut(String) -> i32` signature.
+    /// Registers the special "HostPrint" function for guest console output.
```
simongdavies: nit: I feel like this should say something about overriding the default behaviour which is to write to stdout or console if a tty is attached to the host process?

simongdavies (dismissed):
@ludfjig , a couple of nits but this is great Thanks!!

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -154,19 +231,16 @@ impl MultiUseSandbox {
...
+    /// page table entries are not created.
     ///
     /// # Safety
-    /// It is the caller's responsibility to ensure that the host side
-    /// of the region remains intact and is not written to until this
-    /// mapping is removed, either due to the destruction of the
-    /// sandbox or due to a state rollback
+    ///
+    /// The caller must ensure the host memory region remains valid and unmodified
+    /// for the lifetime of `self`.
```
syntactically, line 243: I'm not sure that "the lifetime of `self`" is accurate here?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -185,9 +259,9 @@ impl MultiUseSandbox {
         Ok(())
     }
 
-    /// Map the contents of a file into the guest at a particular address
+    /// Maps a file's contents into the guest address space with copy-on-write semantics.
```
syntactically: This does not actually have copy-on-write semantics right now, since it does not allow writable mappings at all. Yes, we should probably have been better about the function name, especially now that it's clear that it will be some time if ever before that actually makes sense to do...

src/hyperlight_host/src/sandbox/uninitialized.rs
syntactically: (tangential) didn't we plan to get rid `HostPrint` entirely some releases ago?

src/hyperlight_host/src/sandbox/uninitialized.rs
ludfjig: fixed

src/hyperlight_host/src/sandbox/uninitialized.rs
ludfjig: Fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I thought this was more accurate given that snapshot restores can remap memory now. I thought "Lifetime of self" was the easiest way to communicate this.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: thanks for catching this. Undid my change

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: Why wouldn't this be accurate?

syntactically: > I thought this was more accurate given that snapshot restores can remap memory now. I thought "Lifetime of self" was the easiest way to communicate this.

I think it should be explicit that it should last until the last time that a snapshot that uses it has been used.

PR #751, opened 2025-07-31
Update Rust version to 1.86

simongdavies:
Updates Rust Version to 1.86

@syntactically there is a reference to 1.85 [here](https://github.com/hyperlight-dev/hyperlight/blob/main/flake.nix#L57) I did not know if this needed to be updated or how?

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -52,7 +52,7 @@ RUN SNIPPET="export PROMPT_COMMAND='history -a' && export HISTFILE=/commandhisto
 
 USER $USER
 
-ARG RUST_TOOLCHAIN=1.85
+ARG RUST_TOOLCHAIN=1.86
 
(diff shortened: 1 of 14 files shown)


syntactically:
@simongdavies That is there to provide the toolchain used by the dev/verify-msrv.sh script, so if all the msrvs have been updated then it should indeed be replaced by 1.86, otherwise there should be a new 1.86 entry added to have both available. If you don't mind the extra effort, it would be great to replace/add as appropriate:
```nix
            "1.86" = {
              date = "2025-04-03";
              channel = "stable";
              sha256 = "sha256-X/4ZBHO3iW0fOenQ3foEvscgAPJYl2abspaBThDOukI=";
            };
```

syntactically:
Oh, I just realised you already merged this, I will do it myself and roll it in to my next PR, sorry for the noise.

PR #753, opened 2025-07-31
fix(guest-bin): align user memory allocations

andreiltd:
The guest allocation wrapper aligned only the total memory block but did not guarantee that the pointer returned to the user was itself properly aligned. This patch attempts to fix that.

It also adds alloc_aligned function which would be tricky to implement outside the guest but will make implementing other posix like allocation API much easier.

### DIFF
--- a/src/hyperlight_guest_bin/src/memory.rs
+++ b/src/hyperlight_guest_bin/src/memory.rs
@@ -50,6 +50,9 @@ struct Header(Layout);
 /// Allocates a block of memory with the given size. The memory is only guaranteed to be initialized to 0s if `zero` is true, otherwise
 /// it may or may not be initialized.
 ///
+/// # Invariants
+/// `alignment` must be non-zero and a power of two
+///
 /// # Safety
 /// The returned pointer must be freed with `memory::free` when it is no longer needed, otherwise memory will leak.
 unsafe fn alloc_helper(size: usize, alignment: usize, zero: bool) -> *mut c_void {
--- a/src/hyperlight_guest_bin/src/memory.rs
+++ b/src/hyperlight_guest_bin/src/memory.rs
@@ -40,36 +40,48 @@ use hyperlight_guest::exit::abort_with_code;
 */
 
 // We assume the maximum alignment for any value is the alignment of u128.
-const MAX_ALIGN: usize = align_of::<u128>();
+const DEFAULT_ALIGN: usize = align_of::<u128>();
+const HEADER_LEN: usize = size_of::<Header>();
+
+#[repr(transparent)]
+// A header that stores the layout information for the allocated memory block.
+struct Header(Layout);
 
 /// Allocates a block of memory with the given size. The memory is only guaranteed to be initialized to 0s if `zero` is true, otherwise
 /// it may or may not be initialized.
 ///
 /// # Safety
 /// The returned pointer must be freed with `memory::free` when it is no longer needed, otherwise memory will leak.
-unsafe fn alloc_helper(size: usize, zero: bool) -> *mut c_void {
+unsafe fn alloc_helper(size: usize, alignment: usize, zero: bool) -> *mut c_void {
     if size == 0 {
         return ptr::null_mut();
     }
 
-    // Allocate a block that includes space for both layout information and data
-    let total_size = size
-        .checked_add(size_of::<Layout>())
-        .expect("data and layout size should not overflow in alloc");
-    let layout = Layout::from_size_align(total_size, MAX_ALIGN).expect("Invalid layout");
+    let actual_align = alignment.max(align_of::<Header>());
+    let data_offset = HEADER_LEN.next_multiple_of(actual_align);
+
+    let Some(total_size) = data_offset.checked_add(size) else {
+        abort_with_code(&[ErrorCode::MallocFailed as u8]);
+    };
+
+    // Create layout for entire allocation
+    let layout =
+        Layout::from_size_align(total_size, actual_align).expect("Invalid layout parameters");
 
     unsafe {
         let raw_ptr = match zero {
             true => alloc::alloc::alloc_zeroed(layout),
             false => alloc::alloc::alloc(layout),
         };
+
         if raw_ptr.is_null() {
             abort_with_code(&[ErrorCode::MallocFailed as u8]);
-        } else {
-            let layout_ptr = raw_ptr as *mut Layout;
-            layout_ptr.write(layout);
-            layout_ptr.add(1) as *mut c_void
         }
+
+        // Place Header immediately before the user data region
+        let header_ptr = raw_ptr.add(data_offset - HEADER_LEN).cast::<Header>();
+        header_ptr.write(Header(layout));
+        raw_ptr.add(data_offset) as *mut c_void
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -25,51 +25,71 @@ use hyperlight_guest::exit::abort_with_code;
...
+    Each memory allocation via `malloc/calloc/realloc` is stored together with a `Header` containing
+    an `alloc::Layout` and offset information. The header is stored just before the actual raw memory
+    returned to the caller.
 
-    Example: A call to malloc(64) will allocate space for both an `alloc::Layout` and 64 bytes of memory:
+    Example: A call to malloc(64) with 16-byte alignment will allocate space for the header and 64 bytes:
 
     ----------------------------------------------------------------------------------------
-    | Layout { size: 64 + size_of::<Layout>(), ... }    |      64 bytes of memory         | ...
+    | allocated memory...     | Header { layout, offset }    |      64 bytes of memory    | ...
```
syntactically: It might be nicer to say something like "padding memory" instead of "allocated memory".

Are we guaranteed to be respecting the alignment of the Header structure itself here?  I guess we are as long as the Header size is a multiple of its alignment, which is a Rust invariant?

src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -25,51 +25,71 @@ use hyperlight_guest::exit::abort_with_code;
...
-                                                        |
-                                                    ptr returned to caller
+    ^                         ^                              ^
+    |                         |                              |
+    raw_ptr                   header_ptr                     ptr returned to caller
+    (from allocator)          (at data_offset - header_size) (at data_offset)
+
+    The header contains:
+    - layout: The Layout used for the entire allocation (total_size, actual_align)
+    - offset: The offset from raw_ptr to the user data (data_offset)
```
syntactically: Is keeping the offset around necessary? Can't it be computed from the alignment in the Layout?

src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -95,8 +115,22 @@ pub unsafe extern "C" fn calloc(nmemb: usize, size: usize) -> *mut c_void {
...
+}
+
+/// Allocates aligned memory.
+///
+/// # Safety
+/// The returned pointer must be freed with `free` when it is no longer needed.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn aligned_alloc(alignment: usize, size: usize) -> *mut c_void {
+    // Validate alignment
+    if alignment == 0 || (alignment & (alignment - 1)) != 0 {
```
syntactically, line 124: Maybe add a comment on alloc_helper detailing this invariant.

src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -105,12 +139,20 @@ pub unsafe extern "C" fn calloc(nmemb: usize, size: usize) -> *mut c_void {
 /// `ptr` must be a pointer to a memory block previously allocated by `memory::malloc`, `memory::calloc`, or `memory::realloc`.
 #[unsafe(no_mangle)]
 pub unsafe extern "C" fn free(ptr: *mut c_void) {
-    if !ptr.is_null() {
-        unsafe {
-            let block_start = (ptr as *const Layout).sub(1);
```
syntactically, line 110: FWIW I think this idiom is safe to continue using as both (a) raw pointers can be unaligned and (b) size is always supposed to be a multiple of alignment. Feel free to ignore this comment if you had a different reason for preferring the other idiom.

src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -25,51 +25,71 @@ use hyperlight_guest::exit::abort_with_code;
...
+        Layout::from_size_align(total_size, actual_align).expect("Invalid layout parameters");
 
     unsafe {
         let raw_ptr = match zero {
             true => alloc::alloc::alloc_zeroed(layout),
             false => alloc::alloc::alloc(layout),
         };
+
         if raw_ptr.is_null() {
             abort_with_code(&[ErrorCode::MallocFailed as u8]);
```
jprendes, line 81: I know it's not part of your changes, but should we about here, or return null?
The aligned_alloc method returns null is the alignment is wrong. Should that also abort?

src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -105,12 +139,20 @@ pub unsafe extern "C" fn calloc(nmemb: usize, size: usize) -> *mut c_void {
...
+
+    let user_ptr = ptr as *const u8;
+
+    unsafe {
+        // Read the Header just before the user data
+        let header_ptr = user_ptr.sub(size_of::<Header>()).cast::<Header>();
+        let Header { layout, offset } = header_ptr.read();
+
+        // Deallocate from the original base pointer
+        let raw_ptr = user_ptr.sub(offset) as *mut u8;
```
jprendes, line 150: could the header just store the raw_ptr instead of the offset?

syntactically: I think you can recompute that just from the layout, so I would be in favour of still only storing the layout I think?

andreiltd: Yes, I rewrote this to recalculate the offset on free.

src/hyperlight_guest_bin/src/memory.rs
andreiltd: Yes, because the actual alignment is: `alignment.max(align_of::<Header>())` which guarantee we can place header just before user ptr safely.

src/hyperlight_guest_bin/src/memory.rs
andreiltd: I rewrote this to recalculate the offset instead in free and realloc.

src/hyperlight_guest_bin/src/memory.rs
andreiltd: I would say aborting is more useful and probably "safer" in memory constrained env, in the same time if we want to be somewhat posix compatible we should return null and set `errno` to `ENOMEM`. 🤷

src/hyperlight_guest_bin/src/memory.rs
```diff
@@ -40,36 +40,51 @@ use hyperlight_guest::exit::abort_with_code;
...
 /// Allocates a block of memory with the given size. The memory is only guaranteed to be initialized to 0s if `zero` is true, otherwise
 /// it may or may not be initialized.
 ///
+/// # Invariants
+/// `alignment` must be non-zero and a power of two
+///
 /// # Safety
 /// The returned pointer must be freed with `memory::free` when it is no longer needed, otherwise memory will leak.
-unsafe fn alloc_helper(size: usize, zero: bool) -> *mut c_void {
+unsafe fn alloc_helper(size: usize, alignment: usize, zero: bool) -> *mut c_void {
```
jsturtevant, line 58: Are there any tests we should be adding to verify this new behavior?

src/hyperlight_guest_bin/src/memory.rs
syntactically: > in the same time if we want to be somewhat posix compatible we should return null and set errno to ENOMEM

But, the posix that everyone cares about is Linux, which refuses to ever do this (it overcommits so it always returns an unmapped page if nothing else is available, and if this bluff is called it is too late to return NULL, so it tries desparately to kill things to free up memory and IIRC panics in the rare occasion that doesn't work), and people famously do not handle the malloc() failure case well for this reason (+ it's very hard to "handle well"). So I think aborting is probably the preferable behaviour.

src/hyperlight_guest_bin/src/memory.rs
andreiltd: Yeah, this is tricky. This crate default target is baremetal and cargo doesn't allow different default targets for build and test: https://github.com/rust-lang/cargo/issues/6784 so `cargo test` won't work.

It really only make sense to write the tests within this module so we have an access to internal memory layout with header and whatnot. So we should probably either use custom test harness or unstable `test_runner` features? 

https://doc.rust-lang.org/beta/unstable-book/language-features/custom-test-frameworks.html

cc: @jprendes

PR #754, opened 2025-07-31
Fix label cleanup

simongdavies:
Fixes the label cleanup job.


.github/workflows/ReleaseBlockerLabelCleanUp.yml
```diff
@@ -12,6 +12,8 @@ jobs:
   remove-release-blocker:
     runs-on: ubuntu-latest
     steps:
+      - name: Checkout code
```
syntactically, line 15: It seems a bit of a shame to have to do this given that we don't actually need the code... but I guess it is required for some reason?

syntactically (approved):

PR #761, opened 2025-08-04
Bring back the previous behavior of `call_guest_function_by_name`

jprendes:
This PR brings back the previous behaviour of `call_guest_function_by_name`, and exposes the new behavioud behind a different name: ~~`run`~~ `call`.

~~Originally I wanted to call the new behaviour `call`, but that method exists as an alias to `call_guest_function_by_name` throught `Callable` trait.~~
~~The risk of using the `call` name is that anyone using the `Callable` trait would get an (almost) silent change in behaviour (almost because there would be a lint about `Callable` being unused).~~
Decided that there's no big risk of anyone using the `Callable` trait directly other than through the WIT macros, so using `call` should be fine.

Let me know if anyone prefers a different name.

### DIFF
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -467,8 +467,7 @@ impl Callable for MultiUseSandbox {
         func_name: &str,
         args: impl ParameterTuple,
     ) -> Result<Output> {
-        #[allow(deprecated)]
-        self.call_guest_function_by_name(func_name, args)
+        self.call(func_name, args)
     }
 }
 


dblnz (approved):
LGTM!

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -67,6 +67,7 @@ pub struct MultiUseSandbox {
     dispatch_ptr: RawPtr,
     #[cfg(gdb)]
     dbg_mem_access_fn: DbgMemAccessHandlerWrapper,
+    snapshot: Option<Snapshot>,
```
ludfjig, line 68: Would you mind adding a comment about what this option(snapshot) represents?

ludfjig (approved):
Looks good to me

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -326,9 +354,33 @@ mod tests {
     use crate::mem::memory_region::{MemoryRegion, MemoryRegionFlags, MemoryRegionType};
     #[cfg(target_os = "linux")]
     use crate::mem::shared_mem::{ExclusiveSharedMemory, GuestSharedMemory, SharedMemory as _};
-    use crate::sandbox::{Callable, SandboxConfiguration};
+    use crate::sandbox::{Callable as _, SandboxConfiguration};
     use crate::{GuestBinary, HyperlightError, MultiUseSandbox, Result, UninitializedSandbox};
 
+    /// Tests that call_guest_function_by_name restores the state correctly
+    #[test]
```
jsturtevant, line 506: Thanks for the test!

jsturtevant (approved):

dblnz (approved):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -133,17 +148,40 @@ impl MultiUseSandbox {
...
         Ok(())
     }
 
     /// Call a guest function by name, with the given return type and arguments.
-    /// The changes made to the sandbox are persisted
+    /// The changes made to the sandbox are **not** persisted
+    #[deprecated(
+        since = "0.8.0",
+        note = "Deprecated in favour or call_guest and snapshot/restore."
+    )]
```
ludfjig: Should this be `of call` ?

ludfjig (dismissed):
LGTM. I guess we decided we don't care about the Callable trait?

jsturtevant:
> LGTM. I guess we decided we don't care about the Callable trait?

It looks like we need to provide a way for this to choose the behavior, or we need to break it/keep current behavior

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -215,12 +220,65 @@ impl MultiUseSandbox {
     /// # Ok(())
     /// # }
     /// ```
+    #[doc(hidden)]
+    #[deprecated(
+        since = "0.8.0",
+        note = "Deprecated in favour or call and snapshot/restore."
```
syntactically: Typo: s/or/of/

Should this say in favour of `run` and snapshot/restore? Or was that intended to be Callable::call?

syntactically (commented):
I am slightly confused by the division of labour between 65ff6ff2927561f6bdf85f782b8ae2613ee1b0c3 and 1d74a976bc28c3eb237e98b79b18379eb0050e9a. It seems like the former has code  that uses the snaphsot member that is perhaps not correct without the changes from the latter?

Is the intention with 796c6c94f8d3c35b6469932b3a6790c589d11018 to make all the tests call through `Callable`?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: Ah, that was during a rebase, when the message was introduced it should have been `run`, and then changed to `call`, in the rebase I used `call` when the message was introduced.
The final result (i.e., all commits together) should be correct anyway (modulo s/or/of/)

jprendes:
> Is the intention with https://github.com/hyperlight-dev/hyperlight/commit/796c6c94f8d3c35b6469932b3a6790c589d11018 to make all the tests call through Callable?

No, I renamed the method from `run` to `call`. Callable is not being used in the tests, other than through the macro.

jprendes:
Let me update the PR description

jsturtevant (approved):

syntactically:
> No, I renamed the method from `run` to `call`. Callable is not being used in the tests, other than through the macro.

How is the name conflict  with `Callable::call` managed?

Do you mind squashing the `run` rename with the places where each bit of code was introduced as well? It seems a bit silly to have one commit renaming things from `call_guest_by_name` to `run` and another doing it back to`call`!

jprendes:
> How is the name conflict with Callable::call managed?

If the trait is not on scope, there's no conflict.
If the trait is on scope, doing `sox.call`, the new method wins over the trait.
If doing `Callable::call(&sbox, ...)` again there's no conflict, it uses the trait method.
If doing `<sbox as Callable>::call`, i ha ent tried, but I guess the trait is used.

jsturtevant (approved):

PR #762, opened 2025-08-04
Rename `MultiUseSandbox` -> `Sandbox`, `UninitializedSandbox::evolve` -> `UninitializedSandbox::init`

jprendes:
Rename `MultiUseSandbox` as `Sandbox`.
Rename `UninitializedSandbox::evolve` as `UninitializedSandbox::init`

### DIFF
--- a/src/hyperlight_component_util/src/host.rs
+++ b/src/hyperlight_component_util/src/host.rs
@@ -359,7 +359,7 @@ fn emit_component<'a, 'b, 'c>(s: &'c mut State<'a, 'b>, wn: WitName, ct: &'c Com
             type Exports<I: #ns::#import_trait + ::std::marker::Send> = #wrapper_name<I, ::hyperlight_host::sandbox::sandbox::Sandbox>;
             fn instantiate<I: #ns::#import_trait + ::std::marker::Send + 'static>(mut self, i: I) -> Self::Exports<I> {
                 let rts = register_host_functions(&mut self, i);
-                let sb = self.evolve().unwrap();
+                let sb = self.init().unwrap();
                 #wrapper_name {
                     sb,
                     rt: rts,
--- a/src/hyperlight_component_util/src/host.rs
+++ b/src/hyperlight_component_util/src/host.rs
@@ -356,7 +356,7 @@ fn emit_component<'a, 'b, 'c>(s: &'c mut State<'a, 'b>, wn: WitName, ct: &'c Com
             #(#exports)*
         }
         impl #ns::#r#trait for ::hyperlight_host::sandbox::UninitializedSandbox {
-            type Exports<I: #ns::#import_trait + ::std::marker::Send> = #wrapper_name<I, ::hyperlight_host::sandbox::initialized_multi_use::MultiUseSandbox>;
+            type Exports<I: #ns::#import_trait + ::std::marker::Send> = #wrapper_name<I, ::hyperlight_host::sandbox::sandbox::Sandbox>;
             fn instantiate<I: #ns::#import_trait + ::std::marker::Send + 'static>(mut self, i: I) -> Self::Exports<I> {
                 let rts = register_host_functions(&mut self, i);
                 let sb = self.evolve().unwrap();
--- a/src/hyperlight_host/src/lib.rs
+++ b/src/hyperlight_host/src/lib.rs
@@ -22,7 +22,7 @@ limitations under the License.
 //! and host-guest communication.
 //!
 //! The primary entry points are [`UninitializedSandbox`] for initial setup and
-//! [`MultiUseSandbox`] for executing guest functions.
+//! [`Sandbox`] for executing guest functions.
 //!
 //! ## Guest Requirements
 //!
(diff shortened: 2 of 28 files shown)


src/hyperlight_host/src/lib.rs
```diff
@@ -74,7 +74,7 @@ pub use error::HyperlightError;
 /// Re-export for `MemMgrWrapper` type
 /// A sandbox that can call be used to make multiple calls to guest functions,
 /// and otherwise reused multiple times
-pub use sandbox::MultiUseSandbox;
```
ludfjig, line 95: Maybe we can take this chance to remove these reexports? I'm not sure there is a reason to keep these?

src/hyperlight_component_util/src/host.rs
```diff
@@ -356,10 +356,10 @@ fn emit_component<'a, 'b, 'c>(s: &'c mut State<'a, 'b>, wn: WitName, ct: &'c Com
             #(#exports)*
         }
         impl #ns::#r#trait for ::hyperlight_host::sandbox::UninitializedSandbox {
-            type Exports<I: #ns::#import_trait + ::std::marker::Send> = #wrapper_name<I, ::hyperlight_host::sandbox::initialized_multi_use::MultiUseSandbox>;
+            type Exports<I: #ns::#import_trait + ::std::marker::Send> = #wrapper_name<I, ::hyperlight_host::sandbox::sandbox::Sandbox>;
```
ludfjig, line 359: `::hyperlight_host::sandbox::sandbox::Sandbox` does a look a little bit suspicious to me. Why is it `...sandbox::sandbox...`?

src/hyperlight_host/src/sandbox/sandbox.rs
```diff
@@ -365,14 +365,14 @@ mod tests {
         }
     }
 
-    /// Tests that evolving from MultiUseSandbox to MultiUseSandbox creates a new state
+    /// Tests that evolving from Sandbox to Sandbox creates a new state
     /// and restoring a snapshot from before evolving restores the previous state
     #[test]
     fn snapshot_evolve_restore_handles_state_correctly() {
```
ludfjig: Can we also rename tests / fix comments that have `evolve` in their name?

src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -54,8 +55,8 @@ pub use callable::Callable;
 pub use config::SandboxConfiguration;
 #[cfg(feature = "unwind_guest")]
 use framehop::Unwinder;
-/// Re-export for the `MultiUseSandbox` type
-pub use initialized_multi_use::MultiUseSandbox;
+/// Re-export for the `Sandbox` type
+pub use sandbox::Sandbox;
```
ludfjig, line 59: Maybe we should think about how we want to structure `Sandbox` on a module level

src/hyperlight_host/src/sandbox/mod.rs
```diff
@@ -20,16 +20,17 @@ pub mod config;
...
 /// Functionality for dealing with memory access from the VM guest
 /// executable
 pub(crate) mod mem_access;
 /// Functionality for interacting with a sandbox's internally-stored
 /// `SandboxMemoryManager`
 pub(crate) mod mem_mgr;
 pub(crate) mod outb;
+/// Functionality for dealing with initialized sandboxes that can
+/// call 0 or more guest functions
+#[allow(clippy::module_inception)]
```
ludfjig, line 32: Do we want this?

ludfjig (dismissed):
Lgtm. I am thinking we might want to take this chance to do other breaking changes such as determining which module `Sandbox` belongs to, and whether or not we want to any re-exports of `Sandbox` at all.

src/hyperlight_component_util/src/host.rs
jprendes: when we used to have different sandbox types, it used to be `sandbox::sandbox_type::SandboxType`, since the MultiUseSandbox is becoming basically our only sandbox type, it is now `sandbox::sandbox::Sandbox`, which I agree is a bit weird. 

As the only public thing in the `sandbox::sandbox` module is the `Sandbox` struct, I propose we make the inner `sandbox` module private, and re-export the `Sandbox` struct to be `hl_host::sandbox::Sandbox`.

We can do further cleanup for `0.9`, I think it's fine to introduce breaking changes while we are in `0.x` releases, we are still learning what ergonomics work best.

src/hyperlight_host/src/sandbox/mod.rs
jprendes: do you mean allowing this clippy lint? Without it we either go back to the previous name, or it would be a bigger refactor.

src/hyperlight_host/src/lib.rs
jprendes: I opened #765 to see what we want to expose as our public API

src/hyperlight_component_util/src/host.rs
jprendes: I opened #765 to see what we want to expose as our public API

src/hyperlight_host/src/sandbox/mod.rs
jprendes: I opened #765 to see what we want to expose as our public API

ludfjig (dismissed):

jsturtevant:
closing this and will re-address as part of https://github.com/hyperlight-dev/hyperlight/issues/765 in move to 1.0

PR #763, opened 2025-08-04
Add a sandbox builder

jprendes:


syntactically:
I would prefer to move to a model where host functions are known at compile time & sandbox configuration is greatly reduced, so basically instead of all this you would just have `create_snapshot_from_binary(binary)`, `sandbox::new()` (where host functions get registered at compile time, so that we don't have to do string comparisons constantly), `load_snapshot_into_sandbox(snapshot, scratch_mem_size)`

jsturtevant:
closing this and will re-address as part of https://github.com/hyperlight-dev/hyperlight/issues/765 in move to 1.0

PR #766, opened 2025-08-05
Give release-blocker-action permission to read github issues

ludfjig:
Fixes insufficent permission, see here https://github.com/hyperlight-dev/hyperlight/actions/runs/16727948409


jsturtevant:
what changed that requires this new permission?

syntactically:
I think this just never worked before @jsturtevant.  @simongdavies added it last week and it has been failing ever since (see https://github.com/hyperlight-dev/hyperlight/actions/workflows/CreateRelease.yml).

PR #778, opened 2025-08-11
Optimize function call serializing

ludfjig:
This PR slightly optimizes serialization of `FunctionCall`s by:

1. Making the serialization call return a &[u8] instead of Vec<u8>, saving a memory allocation.
2. Avoids an unnecessary step where VecBytes(Vec<u8>) were converted to vec through an useless Iterator.

Todo: 

- Guest functions should not need to return a Vec<u8>. Instead they should be passed some kind of `Writer` on which they pass return values.
- FunctionCall should contained a borrowed &str/&[u8] instead of String and Vec


src/hyperlight_common/src/flatbuffer_wrappers/util.rs
```diff
@@ -169,3 +170,350 @@ impl FlatbufferSerializable for bool {
...
+///
+/// The function aims to be lightweight and fast and run in O(1) as long as the number of parameters is limited
+/// (which it is since hyperlight only currently supports up to 12).
+///
+/// Note: This estimates the capacity needed for the inner vec inside a FlatBufferBuilder. It does not
+/// necessarily match the size of the final encoded buffer. The estimation always rounds up to the
+/// nearest power of two to match FlatBufferBuilder's allocation strategy.
+///
+/// The estimations are numbers used are empirically derived based on the tests below and vaguely based
+/// on https://flatbuffers.dev/internals/ and https://github.com/dvidelabs/flatcc/blob/master/doc/binary-format.md#flatbuffers-binary-format
```
jsturtevant: link to a static github version here

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
```diff
@@ -169,3 +170,350 @@ impl FlatbufferSerializable for bool {
...
+            ParameterValue::Int(_) | ParameterValue::UInt(_) => 16,
+            ParameterValue::Long(_) | ParameterValue::ULong(_) => 20,
+            ParameterValue::Float(_) => 16,
+            ParameterValue::Double(_) => 20,
+            ParameterValue::Bool(_) => 12,
+        };
+    }
+
+    // match how vec grows
+    estimated_capacity.next_power_of_two()
```
jsturtevant, line 211: I guess given our limit of 12 args, this would never get to the point where it wraps to 0? or what if we have 12 really big paramaters?  What would happen in the case it does?

jsturtevant: oh, I see that it would return zero then the buffer would just grow like it does now. if we used checked would it allow use to then set the max so we don't encure a cost? I  haven't thought to hard about the likely hood here...

src/hyperlight_guest/Cargo.toml
```diff
@@ -16,6 +16,7 @@ anyhow = { version = "1.0.99", default-features = false }
 serde_json = { version = "1.0", default-features = false, features = ["alloc"] }
 hyperlight-common = { workspace = true }
 hyperlight-guest-tracing = { workspace = true, default-features = false }
+flatbuffers = { version= "25.2.10", default-features = false }
```
jsturtevant, line 19: should this be a workspace version?

src/hyperlight_host/benches/benchmarks.rs
```diff
@@ -133,9 +137,91 @@ fn sandbox_benchmark(c: &mut Criterion) {
...
+            |input| {
+                let ret: Vec<u8> = sandbox.call("24K_in_8K_out", (input,)).unwrap();
+                assert_eq!(ret.len(), 8 * 1024, "Expected output length to be 8K");
+                std::hint::black_box(ret);
+            },
+        );
+    }
+
+    group.bench_function("24K_in_8K_out_c", |b| {
+        bench_24k_in_8k_out(b, c_simple_guest_as_string().unwrap());
```
jsturtevant, line 212: do we want to measure some larger values too?

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
jsturtevant: In either case, it might be useful debug info for sometime down the line

jsturtevant (commented):
LGTM, thanks for the extensive tests!

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
ludfjig: Good point but I don't think this is an issue in practice (overflowing a usize)

src/hyperlight_guest/Cargo.toml
ludfjig: Only hyperlight dependencies are workspace dependencies, so I don't think so

src/hyperlight_host/benches/benchmarks.rs
ludfjig: I added these because it's a particular size we know someone is using... Happy to add more in the future but right now I'll hold off

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
```diff
@@ -169,3 +170,350 @@ impl FlatbufferSerializable for bool {
...
+    for arg in args {
+        estimated_capacity += 16; // Base parameter structure
+        estimated_capacity += match arg {
+            ParameterValue::String(s) => s.len() + 20,
+            ParameterValue::VecBytes(v) => v.len() + 20,
+            ParameterValue::Int(_) | ParameterValue::UInt(_) => 16,
+            ParameterValue::Long(_) | ParameterValue::ULong(_) => 20,
+            ParameterValue::Float(_) => 16,
+            ParameterValue::Double(_) => 20,
+            ParameterValue::Bool(_) => 12,
```
danbugs, line 206: Maybe consider making these values consts to avoid diff magic numbers here and there?

src/tests/c_guests/c_simpleguest/main.c
```diff
@@ -260,6 +267,7 @@ HYPERLIGHT_WRAP_FUNCTION(guest_abort_with_msg, Int, 2, Int, String)
 HYPERLIGHT_WRAP_FUNCTION(guest_abort_with_code, Int, 1, Int)
 HYPERLIGHT_WRAP_FUNCTION(execute_on_stack, Int, 0)
 HYPERLIGHT_WRAP_FUNCTION(log_message, Int, 2, String, Long)
+// HYPERLIGHT_WRAP_FUNCTION(twenty_four_k_in_eight_k_out, VecBytes, 1, VecBytes) is not valid for functions that return VecBytes
```
danbugs, line 270: Did you mean to remove this comment?

src/tests/c_guests/c_simpleguest/main.c
```diff
@@ -295,6 +303,9 @@ void hyperlight_main(void)
     HYPERLIGHT_REGISTER_FUNCTION("GuestAbortWithMessage", guest_abort_with_msg);
     HYPERLIGHT_REGISTER_FUNCTION("ExecuteOnStack", execute_on_stack);
     HYPERLIGHT_REGISTER_FUNCTION("LogMessage", log_message);
+    // HYPERLIGHT_REGISTER_FUNCTION macro does not work for functions that return VecBytes,
```
danbugs, line 306: Did you mean to remove this comment?

danbugs (dismissed):
Nice optimization! Mostly LGTM. Just a couple of questions/nits here and there 👍

src/tests/c_guests/c_simpleguest/main.c
ludfjig: no I left it on purpose to indicate why you cannot use HYPERLIGHT_REGISTER_FUNCTION

src/tests/c_guests/c_simpleguest/main.c
ludfjig: No I left it on purpose to indicate why HYPERLIGHT_WRAP_FUNCTION cannot be used here

jsturtevant (dismissed):
Feel free to address the remaining comments but this LGTM

src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
```diff
@@ -72,214 +72,130 @@ impl FunctionCall {
...
-        let parameters = function_call_fb
-            .parameters()
-            .map(|v| {
-                v.iter()
-                    .map(|p| p.try_into())
-                    .collect::<Result<Vec<ParameterValue>>>()
-            })
-            .transpose()?;
+    /// Encodes self into the given builder and returns the encoded data.
+    pub fn encode<'a>(&self, builder: &'a mut FlatBufferBuilder) -> &'a [u8] {
```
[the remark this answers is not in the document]
jsturtevant: this seems like a reasonable suggestion

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
ludfjig: I think this is fine IMO, shouldn't need tweaking

src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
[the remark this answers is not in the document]
ludfjig: Fixed

Issue #715, opened 2025-07-16
Call clear_io_buffers only on error conditions


jprendes:
Is it worth only doing this on an Err return?

_Originally posted by @syntactically in https://github.com/hyperlight-dev/hyperlight/pull/697#discussion_r2211751907_

syntactically:
In particular, the change that is currently there blocks reentrancy, which I don't necessarily think is a bad idea (we get many benefits in component-land from doing this on purpose in most scenarios), but which is a big break from previous behaviour.  Clearing the I/O buffers only on errors would I think re-enable reentrancy in the common case, and then we could decide if we want to get rid of it later.

PR #811, opened 2025-08-20
Only clear io buffer after unsuccesfull guest call

ludfjig:
We are currently zeroing out some buffers when we do not need to. By removing the extra clear, performance should increase:

Copy-pasted comment from code here

In the happy path we do not need to clear io-buffers from the host because:

- the serialized guest function call is be zeroed out by the guest during deserialization, see call to `try_pop_shared_input_data_into::<FunctionCall>()`
- the serialized guest function result is zeroed out by us (the host) during deserialization, see `get_guest_function_call_result`
- any serialized host function call are zeroed out by us (the host) during deserialization, see `get_host_function_call`
- any serialized host function result is zeroed out by the guest during deserialization, see `get_host_return_value`

~~In addition,this PR fixes an issue when host functions panicked, the associated mutexes were poisoned which prevented making further host calls to the same function.~~ EDIT: this PR does not fix that

Closes https://github.com/hyperlight-dev/hyperlight/issues/715

### DIFF
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -502,6 +502,7 @@ mod tests {
     use std::sync::{Arc, Barrier};
     use std::thread;
 
+    use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
     use hyperlight_testing::simple_guest_as_string;
 
     #[cfg(target_os = "linux")]
@@ -523,10 +524,14 @@ mod tests {
         sandbox.register("HostAdd", |a: i32, b: i32| a + b).unwrap();
         let mut sandbox = sandbox.evolve().unwrap();
 
-        // will exhaust io if leaky
+        // will exhaust io if leaky. Tests both success and error paths
         for _ in 0..1000 {
             let result = sandbox.call::<i32>("Add", (5i32, 10i32)).unwrap();
             assert_eq!(result, 15);
+            let result = sandbox.call::<i32>("AddToStaticAndFail", ()).unwrap_err();
+            assert!(
+                matches!(result, HyperlightError::GuestError (code, msg ) if code == ErrorCode::GuestError && msg == "Crash on purpose")
+            );
         }
     }
 


src/hyperlight_host/src/func/host_functions.rs
```diff
@@ -179,9 +179,21 @@ macro_rules! impl_host_function {
...
-                        func.try_lock()
-                            .map_err(|e| new_error!("Error locking at {}:{}: {}", file!(), line!(), e))?
-                            (args)
+                        match func.try_lock() {
+                            Ok(mut guard) => guard(args),
+                            Err(poison_err) => {
+                                match poison_err {
+                                    // The previous call to this host function panicked, poisoning the lock.
+                                    // We can clear the poison safely.
+                                    std::sync::TryLockError::Poisoned(guard) => {
```
jsturtevant: How do we know it was poisoned due to the panick?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -421,9 +421,14 @@ impl MultiUseSandbox {
                 .get_guest_function_call_result()
         })();
 
-        // TODO: Do we want to allow re-entrant guest function calls?
-        self.get_mgr_wrapper_mut().as_mut().clear_io_buffers();
-
+        // In the happy path we do not need to clear io-buffers from the host because:
+        // - the serialized guest function call is be zeroed out by the guest during deserialization, see call to `try_pop_shared_input_data_into::<FunctionCall>()`
```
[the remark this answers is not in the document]
ludfjig: fixed

src/hyperlight_host/src/func/host_functions.rs
ludfjig: I think that's the only option

src/hyperlight_host/src/seccomp/guest.rs
```diff
@@ -81,6 +81,10 @@ fn syscalls_allowlist() -> Result<Vec<(i64, Vec<SeccompRule>)>> {
         // since it will try to open /proc/sys/vm/overcommit_memory (https://sourceware.org/git/?p=glibc.git;a=blob;f=sysdeps/unix/sysv/linux/malloc-sysdep.h;h=778d8971d53e284397c3a5dcdd923e93be5e4731;hb=HEAD)
         // We have another more restrictive filter for it below so it will return EACCES instead of trap, in which case libc will use the default value
         (libc::SYS_openat, vec![]),
+        // the following 3 are needed by Rust's panic handler
+        (libc::SYS_getcwd, vec![]),
+        (libc::SYS_readlink, vec![]),
+        (libc::SYS_gettid, vec![]),
```
danbugs: Did you figure this out from just trial and error?

danbugs (dismissed):
LGTM

src/hyperlight_host/src/seccomp/guest.rs
ludfjig: correct

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -190,7 +190,18 @@ fn maybe_with_seccomp<T: Send>(
                             return Err(crate::HyperlightError::DisallowedSyscall);
                         }
 
-                        crate::log_then_return!("Host function {} panicked", name);
```
jsturtevant: did we want to continue to log something here?

ludfjig: I removed it because the panic cannot go unnoticed, so I deemed the logging unnecessary

src/hyperlight_host/src/seccomp/guest.rs
ludfjig: No longer need to allow these...

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -504,6 +509,25 @@ mod tests {
     use crate::sandbox::SandboxConfiguration;
     use crate::{GuestBinary, HyperlightError, MultiUseSandbox, Result, UninitializedSandbox};
 
+    /// Make sure input/output buffers are properly reset after guest call (with host call)
+    #[test]
+    fn io_buffer_reset() {
```
syntactically, line 517: Should we also test the case with an error?

ludfjig: Added

PR #818, opened 2025-08-22
Remove Allocations from Panic Handler

### LINKED ISSUE #735: Don't alloc memory in panic handler
If we panic in guest due to running out of memory, allocating inside the panich andler will cause another panic, causing infinite panic recursion until the stack is overflown and we hit a guard page, which exits as a `Stackoverflow` error. 

Our current tests for running out of memory tries to allocate a huge chunk which fails, but the allocation inside the panic handler succeeds because it's way smaller than the first one... :/

adamperlin:
This PR fixes #735. The approach used here is to create a `FixedStringBuf` type which wraps a byte slice and which implements `fmt::Write` by writing into this underlying slice. This static buffer can then be used to format the actual panic message.


src/hyperlight_common/src/fixed_buf.rs
```diff
@@ -0,0 +1,73 @@
+use core::fmt;
```
danbugs: Missing the license header here 👍

src/hyperlight_common/src/fixed_buf.rs
```diff
@@ -0,0 +1,89 @@
...
+*/
+
+use core::fmt;
+use core::result::Result;
+use core::ffi::{CStr, FromBytesUntilNulError};
+
+/// FixedStringBuf is a buffer that can hold a fixed-size string.
+/// It is meant to be used with a slice that the user has pre-allocated
+/// to avoid extra allocations during string formatting.
+pub struct FixedStringBuf<'a> {
```
ludfjig: I think it might be easier to have this own the buffer, like so

```
pub struct FixedStringBuf {
    pub buf: [u8; 512],
    pub pos: usize,
}
```

Then you can create a static one wrapped in a mutex, like before.

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,40 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
+
+#[inline(always)]
+fn _panic_handler(info: &core::panic::PanicInfo) -> ! {
+    let mut panic_buf_guard: MutexGuard<'_, FixedStringBuf<'static>> = PANIC_BUF.lock();
+    let write_res = write!(panic_buf_guard, "{}", info);
+    if let Err(_) = write_res {
+        // reset the buffer to ensure there is space
+        // for the new panic message below
+        panic_buf_guard.reset();
+        panic!("panic: message format failed");
```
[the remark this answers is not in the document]
adamperlin: This should be resolved now.

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,40 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
+
+    // create a CStr from the underlying array in PANIC_BUF using the as_cstr method.
+    // this wraps CStr::from_bytes_until_nul which takes a borrowed byte slice
+    // and does not allocate. 
+    let c_string_res = panic_buf_guard.as_c_str();
+    if let Err(_) = c_string_res {
+        // reset the buffer here as well, to ensure there is space
+        // in the buffer to write the new panic message below.
+        panic_buf_guard.reset();
+        panic!("panic: failed to convert to CStr");
```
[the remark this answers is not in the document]
adamperlin: This is also fixed now.

src/hyperlight_common/src/fixed_buf.rs
adamperlin: My biggest issue with this is that then then the buffer's length is hardcoded for any user of FixedStringBuf. I wonder if we could have a macro that generates a `FixedSizeBufN` struct with an internal capacity of `N`

ludfjig: You might be able parameterize it over the buf size by making it generic
```
pub struct FixedStringBuf<const N: usize> {
    pub buf: [u8; N],
    pub pos: usize,
}
```

adamperlin: Ah perfect, that's exactly what I was looking for 😄

dblnz: Another option could be: https://docs.rs/heapless/0.9.1/heapless/string/type.String.html.
Might be worth checking it out 😄

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -30,11 +31,12 @@ use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
 use hyperlight_common::mem::HyperlightPEB;
 #[cfg(feature = "mem_profile")]
 use hyperlight_common::outb::OutBAction;
+use hyperlight_common::fixed_buf::{FixedStringBuf};
```
dblnz: Nit:
```suggestion
use hyperlight_common::fixed_buf::FixedStringBuf;
```

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -30,11 +31,12 @@ use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
...
 use hyperlight_guest::exit::{abort_with_code_and_message, halt};
 use hyperlight_guest::guest_handle::handle::GuestHandle;
 use hyperlight_guest_tracing::{trace, trace_function};
 use log::LevelFilter;
-use spin::Once;
+use spin::{Once, Mutex, MutexGuard};
```
dblnz: Nit: I think `MutexGuard` might not be needed here?

src/hyperlight_common/src/fixed_buf.rs
```diff
@@ -0,0 +1,93 @@
...
+*/
+
+use core::fmt;
+use core::result::Result;
+
+
+/// FixedStringBuf is a buffer that can hold a fixed-size string of capacity N.
+/// It is meant to be used with a slice that the user has pre-allocated
+/// to avoid extra allocations during string formatting.
+pub struct FixedStringBuf<const N: usize> {
```
dblnz: Another option could be using: https://docs.rs/heapless/0.9.1/heapless/string/type.String.html.
Might be worth checking it out 😄

adamperlin:
Just a note here, this is currently blocked on fuzzing failures due to the memory leak described in #826; currently attempting to use snapshotting in the host call fuzzer to restore back to a clean memory state on each fuzzing iteration, which I will open a separate PR for.

adamperlin:
@ludfjig @dblnz This should be ready for another review!

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,50 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
+        unsafe {
+            abort_with_code_and_message(
+                &[ErrorCode::UnknownError as u8],
+                c"panic: message format failed".as_ptr(),
+            )
+        }
+    }
 
-    unsafe { abort_with_code_and_message(&[ErrorCode::UnknownError as u8], c_string.as_ptr()) }
+    let append_res = panic_buf_guard.push('\0');
```
ludfjig: I think the nullbyte can be combined with the `write!` call above

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,50 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
-    let c_string = alloc::ffi::CString::new(msg)
-        .unwrap_or_else(|_| alloc::ffi::CString::new("panic (invalid utf8)").unwrap());
+    _panic_handler(info)
+}
+
+static PANIC_BUF: Mutex<String<512>> = Mutex::new(String::new());
+
+#[inline(always)]
+fn _panic_handler(info: &core::panic::PanicInfo) -> ! {
+    let mut panic_buf_guard = PANIC_BUF.lock();
```
ludfjig: Is there any chance we can release the guard before exiting the vm? I am worried about what would happen if we panic on next entry, and the lock is already held

ludfjig (changes requested):

src/hyperlight_guest_bin/src/lib.rs
adamperlin: Ah good catch!

src/hyperlight_guest_bin/src/lib.rs
adamperlin: I think so. One option would be to allow the abort sequence to be split into multiple function calls, like
```rust
unsafe {
    send_abort_code(...);
    send_abort_message(...);
    // no need to hold lock since message is no longer borrowed and has been copied via outb
    drop(panic_buf_guard);
    send_abort_terminator();
}
```
A non-thread safe way would be to get the raw pointer first (to get around the borrow checker) then drop, then exit:
```rust
unsafe {
    let msg_ptr = c_str_res.unwrap().as_ptr();
    drop(panic_buf_guard);
    // this would be non-threadsafe since we now have a raw pointer into the
    // unlocked mutex
    abort_with_code_and_message(
        &[ErrorCode::UnknownError as u8],
        msg_ptr,
    )
}
```
Any thoughts?

adamperlin: Per offline conversation, we'll stack allocate the buffer to avoid needing a mutex at all

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,39 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
-    let msg = info.to_string();
-    let c_string = alloc::ffi::CString::new(msg)
-        .unwrap_or_else(|_| alloc::ffi::CString::new("panic (invalid utf8)").unwrap());
+    _panic_handler(info)
+}
+
+#[inline(always)]
+fn _panic_handler(info: &core::panic::PanicInfo) -> ! {
+    // stack allocate a 256-byte message buffer.
+    let mut panic_buf = String::<256>::new();
```
jsturtevant: How did you decide on this size?

jsturtevant: What happens if `write!` puts more in the buffer?

adamperlin: It was originally 512 bytes and I just changed it back. It's fairly arbitrary, just a number of characters that seemed reasonable for path + location + message. If `write!` attempts to write more in the buffer, it will trigger an `fmt::Error` and I've made the error message more descriptive to include the panic buffer length

jsturtevant: Is there away to determine the size required so we don't over/under allocate?

ludfjig: The buffer size is determined at compile time, so I don't think so

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,39 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
 #[allow(dead_code)]
 fn panic(info: &core::panic::PanicInfo) -> ! {
-    let msg = info.to_string();
-    let c_string = alloc::ffi::CString::new(msg)
-        .unwrap_or_else(|_| alloc::ffi::CString::new("panic (invalid utf8)").unwrap());
+    _panic_handler(info)
+}
+
+#[inline(always)]
+fn _panic_handler(info: &core::panic::PanicInfo) -> ! {
```
ludfjig, line 158: I wouldn't mind putting the entire body inside `panic` instead of `_panic_handler`, but maybe it's done like this on purpose?

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,39 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
+    // stack allocate a 512-byte message buffer.
+    let mut panic_buf = String::<512>::new();
```
ludfjig: 512 sound pretty big. Have you tried smaller and checked if messages got cut off?

ludfjig (approved):
Looks good to me. Let's see what others think!

src/hyperlight_guest_bin/src/lib.rs
adamperlin: Did this during dev because of the annoying rust-analyzer warning about duplicate panic handlers. I can definitely move this out!

src/hyperlight_guest_bin/src/lib.rs
adamperlin: I haven't seen any failures at least in our tests with buffer size 256. My main worry is about really long paths in the panic message, though. It would be really nice if we could configure this somehow, maybe as a build parameter?

dblnz (approved):
LGTM!

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -139,11 +141,39 @@ pub static mut OS_PAGE_SIZE: u32 = 0;
...
+    let write_res = write!(panic_buf, "{}\0", info);
```
syntactically: Do we have a guarantee that formatting will not allocate when it's writing to a Write that does not allocate? Skimming through the fmt implementation it looks like that is the case, but there is not an official guarantee anywhere that I can see.

src/hyperlight_guest_bin/src/lib.rs
syntactically: Since performance is not critical here, why don't we just have an `impl Write` that exits once for each chunk of message, so we don't have to worry at all about running out of space?

src/hyperlight_guest_bin/src/lib.rs
adamperlin: I don't think there is any explicit guarantee about this. Looking through the fmt implementation that's also the conclusion I came to, that allocation is dependent on the underlying `Write` implementation. I do have a unit test that reproduces the original issue in that it exhausts the heap and then tries one more allocation leading to panic. So, if the panic handler *were* to allocate at any point it would cause infinite panic recursion + stack overflow in that test which we are not seeing.

src/hyperlight_guest_bin/src/lib.rs
adamperlin: That would be really nice, I just can't see a way to implement this with the `fmt::Write` trait as is. It only allows a binary "write succeeded" or "write failed" result, and we are not fully in control of what is actually being written into the buffer since the formatted string isn't always going to be known until runtime by the actual formatter. If you have any ideas for how to get around this though (or another approach) let me know!

syntactically: Not sure I quite follow. I was thinking basically:
```rust
impl std::fmt::Write for HyperlightAbortWriter {
  fn write_str(&mut self, s: &str) -> Result<(), Error> {
    asm!(
      "mov ds, {}"
      "out ds, eax", 
      const OutBAction::AbortMessage as u16, out("rax") s.as_ptr(), out("rbx") s.len()
    );
    Ok(())
  }
}
```
with appropriate code in the host to reassemble successive calls to this into one string?

adamperlin: Ah got it, thanks for the suggestion! This works, and I have implemented this approach. It turns out there isn't actually any new code needed on the host side to re-assemble since messages are already being chunked anyways. There is still a size limit on the host side but I'm not sure that's something we'd want to remove?

andreiltd:
Sorry, I'm a bit late to the party but I think a similar problem exists in `hyperlight_guest_bin` exception handler, I'm not sure if this is something we want to address as part of this PR or follow up on this. Notably:

https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest_bin/src/exceptions/handler.rs#L82

The additional problem in `hyperlight_guest_bin` is that the code is unsound because the message string is _NOT_ null terminated but it is treated as such later in the code:

https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest_bin/src/exceptions/handler.rs#L112

In result this I belive is UB:
https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest/src/exit.rs#L57

syntactically:
> Sorry, I'm a bit late to the party but I think a similar problem exists in `hyperlight_guest_bin` exception handler, I'm not sure if this is something we want to address as part of this PR or follow up on this. Notably:
> 
> https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest_bin/src/exceptions/handler.rs#L82

I'm aware of that one and am rewriting the whole exn handler at the moment anyway, so I was just going to roll it up in a slightly larger PR with the rest of the architecture stuff. Unless you are running into an urgent issue caused by this?

andreiltd:
Hey @syntactically, nothing urgent, no :) Great to hear you're working on this.

adamperlin:
I'm glad this is being worked on too! Thank you for bringing this up @andreiltd!

@ludfjig @dblnz Could I get one more review on this? I've changed the approach based on @syntactically's suggestion so now there is no fixed-size buffer on the guest side!

ludfjig (dismissed):
Looks good to me, but I am not too familiar with how OUT currently works in the host. Maybe @danbugs could review to make sure this looks correct?

src/hyperlight_guest/src/exit.rs
```diff
@@ -67,6 +67,14 @@ pub unsafe fn abort_with_code_and_message(code: &[u8], message_ptr: *const c_cha
     }
 }
 
+/// This function exists to give the guest more manual control
+/// over the abort sequence. For example, in the panic handler,
```
danbugs:
```suggestion
/// over the abort sequence. For example, in `hyperlight_guest_bin`'s panic handler,
```

nit

danbugs (approved):
LGTM!

PR #819, opened 2025-08-22
Clear mutex poision if previous host function panicked

ludfjig:


jprendes (dismissed):
LGTM

syntactically:
I would like to be careful to avoid adding too much reliance on single-threadedness, since that might change in the future, or at least being extremely clear about where it is and how it would need to change.

I also think that a host function failure should probably poison the sandbox, requiring it to be restored before it can be used again?

PR #827, opened 2025-08-27
Fix unbounded growth of panic hook after each new sandbox

### LINKED ISSUE #816: Memory leak / panic nesting in `setup_signal_handlers`
Each new sandbox calls `setup_signal_handlers` to setup a panic handler which ignores DisallowedSyscall panics. This creates a new closure that captures the previous panic hook, but the old closures never get cleaned up. Thus, the more sandboxes are created, the more closures the panic hook contains.

I believe this is also the cause of the very long panic backtraces we've been seeing, since the more sandboxes you create, the more closures the panic has to go through. 

We can consider seting up the panic hook just once using something like use std::sync::Once, however we need to be careful in still setting up the `vm_kill_signal`, since this still needs to rerun if sandbox has a different `config.get_interrupt_vcpu_sigrtmin_offset()` from previous sandbox

ludfjig:
`setup_signal_handlers` is called once per new sandbox (this is needed due to the possibility of different sandboxes using different signal # to cancel executions). This also caused the panic hook to be set unnecessarily for each new sandbox. This PR fixes it by introducing a simple `Once`.

I confirmed locally that this fixes the long stacktraces we've been seeing, since it's non-trivial to add a test for it.

Closes #816

### DIFF
--- a/src/hyperlight_host/src/signal_handlers/mod.rs
+++ b/src/hyperlight_host/src/signal_handlers/mod.rs
@@ -29,26 +29,31 @@ pub(crate) fn setup_signal_handlers(config: &SandboxConfiguration) -> crate::Res
     // should be safe to call.
     #[cfg(feature = "seccomp")]
     {
+        use std::sync::Once;
+
         vmm_sys_util::signal::register_signal_handler(
             libc::SIGSYS,
             sigsys_signal_handler::handle_sigsys,
         )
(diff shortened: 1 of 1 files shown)


adamperlin (approved):
Someone else should definitely sign off here, but from my reading this seems like a simple and reasonable fix!

PR #829, opened 2025-08-28
Small bugfixes in the component macros

jprendes:
These are the changes required for [hyperlight-wasm-http-example](https://github.com/jprendes/hyperlight-wasm-http-example/) to build and run correctly.

I think that we need better test coverage for the macro, and I believe @syntactically is working on that.
However, I would like to get these changes in for the 0.9.0 release.

The change in `resolve_tv`, in the case of a `TypeBound::SubResource` the function was returning a relative index instead of an absolute index. In the other non-recursive branch, `TypeBound::Eq`, we need to return the relative index. I believe @syntactically is also working on a more sane refactor of this method.

The change in the generated code, when the host sends some resource to the guest, it needs to insert them in the resource table, for that the code in `#marshal` assumes that a variable `rts` is in scope, which was not the case. Also, `rts` is behind a mutex that we need to release before doing the actual guest function call, so creation of `args` is moved forward.

### DIFF
--- a/src/hyperlight_component_util/src/emit.rs
+++ b/src/hyperlight_component_util/src/emit.rs
@@ -589,12 +589,13 @@ impl<'a, 'b> State<'a, 'b> {
     /// ends up with a resource type, in which case we return the
     /// resource index
     pub fn resolve_tv(&self, n: u32) -> (u32, Option<Defined<'b>>) {
-        match &self.bound_vars[self.var_offset + n as usize].bound {
+        let m = n + self.var_offset as u32;
+        match &self.bound_vars[m as usize].bound {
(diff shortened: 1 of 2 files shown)


syntactically:
> In the other non-recursive branch, TypeBound::Eq, we need to return the relative index

Yes, this is because it is used to find a type name as well as the actual type in some of the marshalling code, I think. I have a version of this is that has a bit clearer of a comment and uses a little structure for the return value of the method to make it much more clear what is going on.

I am doing the final bit of getting tests to be more complete, fixing some other issues that popped up as well, checking if a name collision issue we also saw is real, and disentangling from other aarch64 changes I was working on, so my slightly larger PR should be up within a couple of hours (within plenty of time for the release)

Issue #282, opened 2025-02-24
Non-standard libc headers


ludfjig:
For example, hyperlight-guest has "printf.h", but usually the printf-type functions are defined in "stdio.h". This makes it harder to port existing programs to run with hyperlight's libc, as it requires more manual patching of code. We should aim to make it as frictionless as possible to run existing code inside hyperlight, and one way we can do this is aim to have more standard libc header files

PR #831, opened 2025-08-28
feat(guest): replace musl with picolibc

andreiltd:
This patch removes custom musl implementation used by the guest and replace it with [picolibc](https://github.com/picolibc/picolibc) v1.8.11.

Note: The picolibc submodule uses the picolibc-bsd fork (https://github.com/hyperlight-dev/picolibc-bsd), which is a redistribution with all copyleft-licensed files removed from the tree and history. Only BSD/MIT/permissive-licensed source files are present.


src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    let bytes = call_host_function::<Vec<u8>>(
+        "HostRead",
+        Some(vec![ParameterValue::ULong(count as u64)]),
+        ReturnType::VecBytes,
+    )
+    .unwrap_or_default();
+
+    let n = bytes.len();
```
jprendes: Do we want to assert `n <= count`?
A wrongly implemented `HostRead` could cause a buffer overflow

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+
+#[unsafe(no_mangle)]
+pub extern "C" fn write(fd: c_int, buf: *const c_void, count: usize) -> isize {
+    // Only stdout (fd=1) and stderr (fd=2)
+    if fd != 1 && fd != 2 {
+        return count as isize;
+    }
+
+    let slice = unsafe { core::slice::from_raw_parts(buf as *const u8, count) };
+    let s = core::str::from_utf8(slice).unwrap_or("<invalid utf8>");
```
jprendes: **question:** what do you think of using `String::from_utf8_lossy` instead? we are doing `s.to_string()` anyway.

src/hyperlight_guest_bin/Cargo.toml
```diff
@@ -14,9 +14,9 @@ and third-party code used by our C-API needed to build a native hyperlight-guest
 """
 
 [features]
-default = ["libc", "printf"]
-libc = [] # compile musl libc
-printf = [ "libc" ] # compile printf
```
jprendes, line 19: **opinion:** I would just put everything under the same libc feature

jprendes (commented):
I love this PR!

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+        core::ptr::copy_nonoverlapping(bytes.as_ptr(), buf as *mut u8, n);
+    }
+    n as isize
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn write(fd: c_int, buf: *const c_void, count: usize) -> isize {
+    // Only stdout (fd=1) and stderr (fd=2)
+    if fd != 1 && fd != 2 {
+        return count as isize;
```
jprendes: **question:** shuoldn't this return -1, or some other indication of error? and set errno?

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,141 @@
...
+    fn host_clock(&self) -> bool {
+        *self.host_clock.get_or_init(|| Self::query("CurrentTime"))
+    }
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn read(fd: c_int, buf: *mut c_void, count: usize) -> isize {
+    // Only stdin (fd=0) and only if HostRead is defined
+    if fd != 0 || !CAPS.host_read() {
+        return 0;
```
jprendes: **question:** should the case of `fd != 0` return -1 and set errno instead?

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: Sure! As long as `from_utf8_lossy` doesn't allocate if string is valid utf-8 which I believe is the case?

src/hyperlight_guest_bin/Cargo.toml
andreiltd: Sure, I guess any code from `libc` if not referenced will be DCE anyway 🤷

src/hyperlight_guest_bin/src/host_bridge.rs
jprendes: It returns a Cow, so, that's correct, but you will need to allocate a string to send it to the host function anyway... so... ¯\\\_(ツ)\_/¯

jsturtevant:
will help with https://github.com/hyperlight-dev/hyperlight/issues/282

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: Right, I think that in order to support this properly we would need to generate the bindings to libc headers and use that to reference `errno`, `EBADF` etc. But then, it might be better to just have `picolibc-sys` crate. Anyway I agree we should report an error here somehow.

src/hyperlight_host/tests/sandbox_host_tests.rs
```diff
@@ -240,16 +241,15 @@ fn iostack_is_working() {
...
-            .try_lock()
-            .map_err(|_| new_error!("Error locking"))
-            .unwrap();
-        lock.push(msg);
-        len as i32
+        let (lock, cvar) = &*pair_clone;
+        let mut vec = lock.lock().unwrap();
+        vec.push(msg);
+        cvar.notify_one();
+        vec.len() as i32
```
jprendes:
```suggestion
        vec.push(msg.clone());
        cvar.notify_one();
        msg.len() as i32
```

raftario:
Is this still on the table ?

jsturtevant:
Yes but need approval on licensing before being able to move forward: https://github.com/cncf/foundation/issues/1117

src/hyperlight_guest_bin/.cargo/config.toml
```diff
@@ -1,2 +1,3 @@
 [build]
 target = "x86_64-unknown-none"
+rustflags = ["-Zdirect-access-external-data"]
```
jprendes: Where does this come from?

andreiltd: IIRC rustc generated unsupported reallocations without this flag

jprendes: This is a nightly flag, right?
Also, this will not apply to comsumers of hyperlight-guest-bin, right?

andreiltd: I think it's worth to recheck and try to remove it.

src/hyperlight_guest_bin/c/clock.c
```diff
@@ -0,0 +1,38 @@
...
+
+#include <sys/time.h>
+#include <time.h>
+#include <errno.h>
+#include <stddef.h>
+#include <stdint.h>
+
+extern void _current_time(uint64_t *ts);
+
+int gettimeofday(struct timeval *__restrict tv, void *__restrict __tz) {
```
jprendes: **Thought:** If we were running bindgen on picolib headers, then we could easily implement this one in rust instead of C. IIUC, the main issue is the definition of `timeval`, `clockid_t`, and `timespec` in Rust, that we don't want to get wrong.

src/hyperlight_guest_bin/.cargo/config.toml
andreiltd: Everything seems to work without the flag, I remove it for now 🤷‍♂️

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,125 @@
...
+        return -1;
+    }
+
+    match call_host_function::<Vec<u8>>(
+        "HostRead",
+        Some(vec![ParameterValue::ULong(count as u64)]),
+        ReturnType::VecBytes,
+    ) {
+        Ok(bytes) => {
+            let n = bytes.len();
```
jprendes: maybe check that `bytes.lenth() <= count`?

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,125 @@
...
+            unsafe {
```
jprendes: Check that `n <= count`

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,125 @@
...
+
+#[unsafe(no_mangle)]
+pub extern "C" fn write(fd: c_int, buf: *const c_void, count: usize) -> isize {
+    if fd != 1 && fd != 2 {
+        set_errno(EBADF);
+        return -1;
+    }
+
+    let slice = unsafe { core::slice::from_raw_parts(buf as *const u8, count) };
+    let s = core::str::from_utf8(slice).unwrap_or("<invalid utf8>");
```
jprendes: I think I would like to insist on the `from_utf8_lossy`, as mentioned we are allocationg anyway when doing `s.to_string()`

src/hyperlight_guest_bin/build.rs
```diff
@@ -14,69 +14,222 @@ See the License for the specific language governing permissions and
...
-            .flag("-Wno-string-plus-int");
-
-        cfg.flag("-fPIC");
+    let code = r#"_Complex double x = 1.0 + 2.0*I; int main() { return 0; }"#;
+    let has = if detect_cc_feature(code)? { 1 } else { 0 };
+    writeln!(file, "#define __HAVE_COMPLEX {has}")?;
+
+    let code = r#"long long x = 0; int main() { return 0; }"#;
+    let has = if detect_cc_feature(code)? { 1 } else { 0 };
+    writeln!(file, "#define __IO_LONG_LONG {has}")?;
```
jsturtevant: this gets `  writeln!(file, "#undef __IO_LONG_LONG")?; // minimal format` alittle while later (178). Do we need this here?

src/hyperlight_guest_bin/c/clock.c
jsturtevant, line 1: did we write this file? is there attribution that is needed?

src/hyperlight_guest_bin/c/clock.c
```diff
@@ -0,0 +1,38 @@
...
+    uint64_t current_time[2];
+    (void)__tz;
+
+    _current_time(current_time);
+    tv->tv_sec = current_time[0];
```
jsturtevant: do we need to check for null here?

src/hyperlight_guest_bin/c/clock.c
```diff
@@ -0,0 +1,38 @@
...
+    return 0;
+}
+
+int clock_gettime(clockid_t clk_id, struct timespec *tp) {
+    uint64_t current_time[2];
+    switch (clk_id) {
+        case CLOCK_REALTIME:
+        case CLOCK_MONOTONIC:
+            _current_time(current_time);
+            tp->tv_sec = current_time[0];
```
jsturtevant: same here?

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    match call_host_function::<Vec<u8>>(
+        "HostRead",
+        Some(vec![ParameterValue::ULong(count as u64)]),
+        ReturnType::VecBytes,
+    ) {
+        Ok(bytes) => {
+            let n = bytes.len();
+            unsafe {
+                core::ptr::copy_nonoverlapping(bytes.as_ptr(), buf as *mut u8, n);
```
jsturtevant: should not rely on the host provided length. nonoverlapping could write past the buffer lenghth. We are in the guest here so it would just cause corruption but might be tough to track down I think

src/hyperlight_guest_bin/src/libc.rs
```diff
@@ -0,0 +1,125 @@
...
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn write(fd: c_int, buf: *const c_void, count: usize) -> isize {
+    if fd != 1 && fd != 2 {
+        set_errno(EBADF);
+        return -1;
+    }
+
+    let slice = unsafe { core::slice::from_raw_parts(buf as *const u8, count) };
```
jsturtevant, line 95: do we need a null check on the buf here? https://doc.rust-lang.org/core/slice/fn.from_raw_parts.html#safety

>data must be non-null and aligned even for zero-length slices

NOTICE.txt
```diff
@@ -2,227 +2,60 @@ NOTICES
...
-
-This file previously contained text expressing a belief that most of
-the files covered by the above exception were sufficiently trivial not
-to be subject to copyright, resulting in confusion over whether it
-negated the permissions granted in the license. In the spirit of
-permissive licensing, and of not having licensing issues being an
-obstacle to adoption, that text has been removed.
+Note: The picolibc submodule uses sparse checkout to exclude
+GPL/AGPL-licensed test and script files that are not needed for
+building. Only BSD/MIT/permissive-licensed source files are included.
```
jsturtevant: Just wanted to surface this.  We should include it in the PR description I think

src/hyperlight_guest_bin/c/clock.c
dblnz: If this is really necessary, I think @jprendes just wrote a Rust implementation for this in `hyperlight-js`. maybe we could use it?

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -51,6 +51,9 @@ pub mod host_comm;
 pub mod memory;
 pub mod paging;
 
+#[cfg(feature = "libc")]
+mod host_bridge;
```
dblnz: Is there a place we document or mention the functions needed to be implemented by the host for the guest to have `libc` work?
I am referring to `HostRead` and the others from `host_bridge.rs`

src/hyperlight_guest_bin/build.rs
```diff
@@ -14,69 +14,222 @@ See the License for the specific language governing permissions and
...
+            r#"
+            #if __has_builtin({test})
+            int main() {{ return 0; }}
+            #else
+            #error "Builtin not supported"
+            #endif
+        "#
+        );
+
+        let has = if detect_cc_feature(&code)? { 1 } else { 0 };
```
dblnz, line 148: These here are quite a lot of invocations of `clang`. Do you think this can be done in a single call? Something like printing an integer for each builtin and then parsing the output to identify which is supported.

src/hyperlight_guest_bin/build.rs
```diff
@@ -14,69 +14,222 @@ See the License for the specific language governing permissions and
...
+        .flag("-Wold-style-definition")
+        .flag("-Wno-implicit-int")
+        .flag("-Wno-missing-braces")
+        .flag("-Wno-return-type")
+        // We don't support stack protectors at the moment, but Arch Linux clang
+        // auto-enables them for -linux platforms, so explicitly disable them.
+        .flag("-fno-stack-protector")
+        .flag("-fstack-clash-protection")
+        .flag("-mstack-probe-size=4096")
+        // We don't use a different stack for all interrupts, so there
```
dblnz: I think we have a different stack for the Exceptions.

https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest_bin/src/arch/amd64/init.rs#L95

src/hyperlight_guest_bin/build_files.rs
dblnz, line 1: How do you see the process of updating to a new `picolibc` version?
Checkout the new version, verify copyright headers, check the checked out files against the ones in this file,  update this file in case files were removed or added?

dblnz: Maybe it's worth mentioning in the readme, to have some instructions

Justfile
```diff
@@ -43,7 +43,7 @@ build target=default-target:
 guests: build-and-move-rust-guests build-and-move-c-guests
 
 ensure-cargo-hyperlight:
-    command -v cargo-hyperlight >/dev/null 2>&1 || cargo install --locked cargo-hyperlight
+    cargo install --locked --force --git https://github.com/hyperlight-dev/cargo-hyperlight --branch picolibc cargo-hyperlight
```
dblnz: This changes to the "official version of the tool" before merging this PR, right?

dblnz (commented):
Great work, Tomasz!! This is one hell of a diff and I like it 😆

src/hyperlight_guest_bin/build.rs
andreiltd: I measured this and a single invocation on my computer takes 25ms so it's roughly 500ms here. I think it's negligible and complexity of batching the checks is not worth it. I' pretty sure build systems like meson are doing the same thing for feature discovery.

src/hyperlight_guest_bin/build.rs
andreiltd: This comment is carried over from https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest_bin/build.rs#L91, I'm happy to change it, any suggestion?

Justfile
andreiltd: Yes!

syntactically:
Is the licensing approval finally done for this, making it ready for review/merge?

src/hyperlight_guest_bin/build.rs
syntactically: Oh, indeed, that comment was made incorrect by #1205. I think there may still be some issues with enabling a redzone in general, perhaps due to e.g. interrupt reentrancy, and I (at least) don't feel like figuring out if it's worthwhile to enable it. Let's just change this to something more generic---maybe we could say something like "Hyperlight's stack management was not designed with a redzone in mind, so we leave it disabled for now" ?

NOTICE.txt
syntactically: Where is the sparsity pattern that makes this actually happen?

src/tests/c_guests/c_simpleguest/main.c
```diff
@@ -1,16 +1,22 @@
...
 #include "stdlib.h"
 #include "assert.h"
-// Included from hyperlight_guest_bin/third_party/printf
-#include "printf.h"
+#include "stdio.h"
 
 #define GUEST_SCRATCH_SIZE (0x40000) // default scratch size
 #define MAX_BUFFER_SIZE (1024)
 
+#define printf_f(fmt, ...)                          \
```
syntactically, line 13: Does it make sense to see if we can either build picolibc in a configuration without stdio buffers, or to automatically do a flush on the guest-function-call-complete path?

I feel like the buffered-I/O-by-default regime will be very confusing to users---where, for example, if you are doing the "call a function and then restore the sandbox" pattern might often not see any output at all.

src/hyperlight_guest_bin/src/host_bridge.rs
```diff
@@ -0,0 +1,125 @@
...
+        Err(_) => {
+            set_errno(EIO);
+            -1
+        }
+    }
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn _current_time(ts: *mut u64) -> c_int {
+    let bytes = call_host_function::<Vec<u8>>("CurrentTime", Some(vec![]), ReturnType::VecBytes)
```
syntactically: Similarly, I think I might like to see this fail at least by default? As with my comments on #1173, I'm against exposing a clock, especially a wall clock, to the guest by default.

src/hyperlight_guest_bin/src/libc.rs
```diff
@@ -0,0 +1,125 @@
...
+
+// POSIX errno values (matching picolibc sys/errno.h)
+const EIO: c_int = 5;
+const EBADF: c_int = 9;
+const ENOSYS: c_int = 88;
+
+static CURRENT_TIME: AtomicU64 = AtomicU64::new(0);
+
+#[unsafe(no_mangle)]
+pub extern "C" fn read(fd: c_int, buf: *mut c_void, count: usize) -> isize {
```
syntactically, line 69: Where are the fds referenced in these functions allocated? I don't see an `open`, so are these always either 0/1/2 for stdin/out/err? I think it might make sense to make the behaviour a bit more limited (maybe reads from fd0 always return eof; writes to fd1 are `HostPrint`, and writes to fd2 are `debug_print`, or something like that?)

NOTICE.txt
jprendes: That would be here:
https://github.com/hyperlight-dev/hyperlight/pull/831/changes#diff-99cd76bae1f874496398d422f5fcc07ece8599f9f26cc81bb1cb154d0cf0003fR338

src/hyperlight_guest_bin/src/libc.rs
jprendes: That's almost what you have here, right? except that:
* fd2 also goes to `HostPrint` instead of `debug_print` (I didn't know `debug_print` was a thing!).
* fd0 first tries `HostRead`, if that's not present returns `-1` and sets `errno = EIO` (instead of EOF, which would be returning `0` and no `errno`)
* any other fd is `EBADF`

So, do you suggest these changes?
* fd2 to use `debug_print` instead of `HostPrint`
* fd0 to return `0` (and untouched `errno`) in case of no `HostRead`

src/hyperlight_guest_bin/c/clock.c
andreiltd: Yes, those are ours. We cannot use the implementation from hyperlight-js directly, it serves a bit different purpose. That being said I rewrote picolibc stubs to rust. The tradeoff is that some of the types and defines that you normally get from libc headers have to be redifined in Rust with compatible ABI. I think that is OK -- they have to be backwards compatible and I don't see them change anytime soon.

src/hyperlight_guest_bin/c/clock.c
andreiltd: This file is no longer in the diff.

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: Yeah, I think the timing is a bit unfortunate, could you check if your comments are resolved in the newest diff?

src/hyperlight_guest_bin/build_files.rs
andreiltd: I've added docs/picolibc.md file and put update instructions there.

src/hyperlight_guest_bin/src/libc.rs
andreiltd: It's now checked in Rust implementation.

src/hyperlight_guest_bin/c/clock.c
andreiltd: It's now checked in Rust implementation.

src/hyperlight_guest_bin/c/clock.c
andreiltd: It's now checked in Rust implementation.

src/hyperlight_guest_bin/src/libc.rs
syntactically: I don't care much about the fd2---it was just the first thing that sprang to mind.

Unless we have a compelling use case, I think I would prefer for read from fd0 to unconditionally be eof, without probing for a host function which seems a bit surprising and magical.  If we ever have C code that needs to be able to read() from stdin, we could add some hyperlight_guest_bin API to let a little Rust wrapper code install its own read function.

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: This is fixed

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: This is fixed

src/tests/c_guests/c_simpleguest/main.c
andreiltd: Yeah, I think we can disable posix console feature: https://github.com/picolibc/picolibc/blob/main/doc/build.md#stdio-options

Fuzzy on the details but IIRC that will change to char by char sending to host which I thought was insane.

src/hyperlight_guest_bin/c/clock.c
andreiltd: > If we were running bindgen on picolib headers, then we could easily implement this one in rust

That would be correct yes, but I would rather much have a single c file defining the interface or redefine types in Rust with correct layout. It's hard to justify running the whole bindgen machinery for few u64.

src/hyperlight_guest_bin/build.rs
andreiltd: Fixed as suggested.

src/hyperlight_guest_bin/src/libc.rs
andreiltd: I'm happy to go with that logic. @jprendes any objections?

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: This is very common dependency, for example spidermonkey crashes immediately if we don't provide some time interface. Do we really want to error or at least return an unix epoch?

src/tests/c_guests/c_simpleguest/main.c
syntactically: Yeah, that makes sense... char-by-char is definitely unfortunate. Flush-on-guest-function-return would be nice if we could coalesce the flush exit and the function-return exit, but I'm not sure how easy that is?

src/hyperlight_guest_bin/src/host_bridge.rs
syntactically: Interesting... I didn't realise so many things were using wall-clock time. Do you know what we were doing for spidermonkey before? I don't think we did actually have the real time?

I wouldn't be against having it always be the epoch personally (at least by default), but I'll let others chime in here as well.

jprendes: at some point we had a fallback where we took some epoch (2000?) and added 1s or 1ms for each call

src/hyperlight_guest_bin/src/libc.rs
jprendes: you can call `fflush(NULL)` and that flushes all open FDs.
Can we leave buffering and add that on VM exit (or similar) in an upcoming PR?

jprendes: also I think our current _putchar impl can go away, right?

src/hyperlight_guest_bin/build.rs
dblnz: Sounds reasonable

src/hyperlight_guest_bin/build_files.rs
dblnz: Thanks!

docs/picolibc.md
dblnz, line 1: This looks great! Thanks for adding this description!

dblnz (dismissed):
Great work, Tomasz!
I have nothing else to add.

src/hyperlight_guest_bin/src/libc.rs
andreiltd: I got rid of function probing and replaced `current_time` with unix epoch increasing 1s each call -- this should be enough to get some sort of ordering working. The read implementation now return eof immediately. The behavior is documented in docs/picolibc.md

As Lucy said if we need to give user control over this we can expose hooks in the guest-bin crate.

src/tests/c_guests/c_simpleguest/main.c
andreiltd: As Jorge suggested: https://github.com/hyperlight-dev/hyperlight/pull/831#discussion_r2891131121 we can perhaps add flushing in next PR?

src/hyperlight_guest_bin/build.rs
```diff
@@ -86,46 +234,153 @@ fn cargo_main() {
...
+        .flag("-U_FORTIFY_SOURCE")
+        .define("ABORT_PROVIDED", "1")
+        .define("DEFINE_MEMALIGN", "1")
+        .define("DEFINE_POSIX_MEMALIGN", "1")
+        .define("_LIBC", None)
+        .define("_FILE_OFFSET_BITS", "64");
 
-        if cfg!(windows) {
-            unsafe { env::set_var("AR_x86_64_unknown_none", "llvm-ar") };
+    // TODO: this is wrong, it should generate config file to out_dir/include
```
ludfjig, line 255: Do we have a plan to fix this?

src/hyperlight_guest_bin/build.rs
```diff
@@ -86,46 +234,153 @@ fn cargo_main() {
...
+            "-C",
+            "third_party/picolibc",
+            "sparse-checkout",
+            "set",
+            "**",
+            "!test/**",
+            "!scripts/**",
+            "!COPYING.GPL2",
+        ])
+        .status()?;
```
ludfjig: I think we should check `.success()` on these commands too

src/hyperlight_guest_bin/src/libc.rs
```diff
@@ -0,0 +1,173 @@
...
+        Ok(_) => count as isize,
+        Err(_) => {
+            set_errno(EIO);
+            -1
+        }
+    }
+}
+
+#[unsafe(no_mangle)]
+pub extern "C" fn _current_time(ts: *mut u64) -> c_int {
```
ludfjig: does this need a null check too given below `unsafe`?

src/hyperlight_guest_bin/build.rs
```diff
@@ -14,69 +14,217 @@ See the License for the specific language governing permissions and
...
-            .flag("-Wno-unused-but-set-variable")
-            .flag("-Wno-unused-parameter")
-            .flag("-Wno-string-plus-int");
-
-        cfg.flag("-fPIC");
+    let code = r#"_Complex double x = 1.0 + 2.0*I; int main() { return 0; }"#;
+    let has = if detect_cc_feature(code)? { 1 } else { 0 };
+    writeln!(file, "#define __HAVE_COMPLEX {has}")?;
+
+    // Static undefs
```
ludfjig, line 160: How did you decide whether to enable/disable some of these? I'm thinking mostly about

__IO_C99_FORMATS disabled, __IO_WCHAR enabled, __OBSOLETE_MATH disabled, __IO_LONG_LONG disabled

ludfjig: And if a particular guest wants different picolibc settings, is that possible? I can imagine that some settings are specific to hyperlight itself, and are not allowed to be changed by guest, but maybe certain guests need more control?

Justfile
```diff
@@ -44,10 +44,10 @@ build target=default-target:
 guests: build-and-move-rust-guests build-and-move-c-guests
 
 ensure-cargo-hyperlight:
-    {{ if os() == "windows" { "if (-not (Get-Command cargo-hyperlight -ErrorAction SilentlyContinue)) { cargo install --locked cargo-hyperlight }" } else { "command -v cargo-hyperlight >/dev/null 2>&1 || cargo install --locked cargo-hyperlight" } }}
+    cargo install --locked cargo-hyperlight
```
ludfjig, line 47: Was removal of the windows specific code here on purpose?

ludfjig (dismissed):
I think this is awesome!

src/hyperlight_guest_bin/build.rs
andreiltd: `picolibc` is using meson as build system -- most of those defines are default settings that meson generates. The defines are gathered from generated by meson picolibc.h file and compile_command.json file. You can have a look at all the switches that we enable/disable for picolibc here:

https://github.com/andreiltd/picoforge/blob/main/picolibc.just

and the docs for all the options:

https://github.com/picolibc/picolibc/blob/main/doc/build.md

Justfile
andreiltd: IIUC this code it checks if `cargo-hyperlight` command exists and if so skips the installation. I believe this is redundant because cargo install is already skipping if binary is already installed unless you pass `--force`. The problem with the command in its current form is that it will never update `cargo-hyperlight` becuase it is cached.

Separately, I think it should be installed as part of .ci setup as well to ensure we control the version that ci is using, then the if-exists checks have more sense.

src/hyperlight_guest_bin/build.rs
ludfjig: I understand, I was more curious if there some specific reason for these enables/disables since I think some are not default and since it's not immediately clear to me what the correct option should be. In addition, would it be valuable to allow guests to override these for example via environment variables?

andreiltd: ﻿The non-defaults serve the hyperlight guest model: single-threaded, no TLS, no semihosting, no libc malloc since we provide rust allocator, verbose asserts for better debuggability, and some I/O performance tweaks discussed above like io buffering.

There are some knobs like wchar support that were needed to allow building baremetal libc++ against picolibc.

We can expose build settings but I would rather do that trough cargo features not env variables. That being said I'm not sure if we should proactively expose all the build tweaks or wait and see if any of this turns out to be problematic?

src/hyperlight_guest_bin/build.rs
andreiltd: I'm not sure how we fix this. Polluting source tree with generated files is obviously bad but we need a stable location for c guests that have no access to hashed rust target dir. 

One way of improving this would be to use the existing `HYPERLIGHT_GUEST_TOOLCHAIN_ROOT` mechanism for c guests. We could copy generated headers to ﻿`HYPERLIGHT_GUEST_TOOLCHAIN_ROOT/include` and set the env when building c guests. Then we can prepend the headers from that dir to clang calls.

I believe a proper fix would be to replace just script and cargo with polyglot build system.

src/hyperlight_guest_bin/build.rs
andreiltd: These are no longer in the patch.

src/hyperlight_guest_bin/src/host_bridge.rs
andreiltd: I've added fallback that ensures values are monotonically increased from epoch.

NOTICE.txt
andreiltd: Switched to using fork.

src/hyperlight_guest_bin/build.rs
jprendes: Is this something that cargo-hyperlight could help with?

src/hyperlight_guest_bin/src/libc.rs
```diff
@@ -0,0 +1,178 @@
...
+        return -1;
+    }
+
+    if fd != 0 {
+        set_errno(EBADF);
+        return -1;
+    }
+
+    // Return EOF immediately — no host round-trip required.
+    let _ = (buf, count);
```
jprendes: do we need this for some kind of lint?

src/hyperlight_guest_bin/src/libc.rs
```diff
@@ -0,0 +1,178 @@
...
+        set_errno(EINVAL);
+        return -1;
+    }
+
+    let (secs, nanos) = current_time();
+    let ts = unsafe { core::slice::from_raw_parts_mut(ts, 2) };
+    ts[0] = secs;
+    ts[1] = nanos;
+    0
+}
```
jprendes: I think this one is not needed

docs/picolibc.md
```diff
@@ -0,0 +1,98 @@
...
+### Write (stdout / stderr)
+
+The `write()` stub delegates to the `HostPrint` host function. Only stdout
+(fd 1) and stderr (fd 2) are supported; both map to the same `HostPrint`
+call, which accepts a `String` parameter and returns an `Int`. Other file
+descriptors return `EBADF`.
+
+### Time
+
+The `clock_gettime()`, `gettimeofday()`, and `_current_time()` stubs do
```
jprendes: I think _current_time is not needed, right?

src/hyperlight_guest_bin/src/libc.rs
```diff
@@ -0,0 +1,178 @@
...
+unsafe extern "C" {
+    static mut errno: c_int;
+}
+
+fn set_errno(val: c_int) {
+    // SAFETY: single-threaded guest, errno is a global int (__GLOBAL_ERRNO)
+    unsafe { errno = val };
+}
+
+// POSIX errno values (matching picolibc sys/errno.h)
```
jprendes, line 35: I think there is some value to running bindgen on picolibc, we could export the bindings in this module, and users of rust code could have access to the underlying libc, like the libc crate does in normal rust.

src/hyperlight_guest_bin/build.rs
```diff
@@ -14,126 +14,348 @@ See the License for the specific language governing permissions and
...
+
     let include_dir = PathBuf::from(&out_dir).join("include");
     fs::create_dir_all(&include_dir)
-        .unwrap_or_else(|e| panic!("Could not create include dir {:?}: {}", &include_dir, e));
-    if is_x86_64 {
-        if cfg!(feature = "printf") {
-            copy_includes(&include_dir, "third_party/printf/");
+        .with_context(|| format!("Could not create include dir {include_dir:?}"))?;
+
+    if cfg!(feature = "libc") {
```
jprendes, line 335: this should be `std::env::var_os("CARGO_FEATURE_LIBC").is_some()`

src/hyperlight_guest_bin/build.rs
```diff
@@ -14,126 +14,348 @@ See the License for the specific language governing permissions and
...
-            copy_includes(&include_dir, "third_party/musl/arch/x86_64");
-            copy_includes(&include_dir, "third_party/musl/src/internal");
+
+        let picolibc_dir = PathBuf::from("third_party/picolibc");
+        let mut build = cc_build(&picolibc_dir, &target)?;
+
+        add_libc(&mut build, &picolibc_dir, &target)?;
+        add_libm(&mut build, &picolibc_dir, &target)?;
+
+        if cfg!(windows) {
```
jprendes, line 347: This should be `std::env::var_os("CARGO_CFG_WINDOWS").is_some()`

jprendes (commented):
I would like the libc stuff to go in its own crate, like `hyperlight-libc`.
Just my opinion, and if you agree, it doesn't need to be in this PR either, I'd rather have this in than delaying due to splitting in another crate.

src/hyperlight_guest_bin/build.rs
andreiltd: Could you elaborate why this would be needed here?

src/hyperlight_guest_bin/build.rs
andreiltd: Wouldn't this check the target and we want to probe host OS here?

jprendes: You are correct, sorry.

andreiltd: > Also note that CARGO_CFG_* variables should be used instead of the cfg! macro or #[cfg] attribute in build scripts, those check the host platform, not the target.

https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

src/hyperlight_guest_bin/build.rs
jprendes: You want to probe if the lib is buit with that feature, not if the build script binary is built with that feature. AFAIK, using CARGO_FEATURE_XXX is the documented way to do this.
Here https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts

src/hyperlight_guest_bin/src/libc.rs
syntactically: I think the normal libc crate can also be used nostd with a custom libc?

jprendes: In unsupported platforms you get an empty crate
https://github.com/rust-lang/libc/blob/088a28455a81c6f9cf770cea42b94380d3cc9b05/src/lib.rs#L156

src/hyperlight_guest_bin/build.rs
jprendes: I just ran a test and either approach works ok, I'll resolve this issue

jprendes (approved):
LGTM

src/hyperlight_guest_bin/src/libc.rs
jprendes: Opened #1398 to track exposing the libc bindings

src/hyperlight_guest_bin/build.rs
jprendes: Opened #1399 to track exposing selected picolibc options as crate features.

src/hyperlight_guest_bin/src/libc.rs
jprendes: Opened #1400 to track flushing of stdio streams.
I believe the other concerns have already been addressed.

src/hyperlight_guest_bin/src/host_bridge.rs
jprendes: I believe this comment is resolved as we are using a deterministic and reproducible time now.

src/tests/c_guests/c_simpleguest/main.c
jprendes: Opened #1400 to track flushing on VM exit.

src/hyperlight_guest_bin/c/clock.c
jprendes: Opened #1398 to track exposing the libc bindings to rust guest.
If we do that when we can also use those bindings for the stubs.

src/hyperlight_guest_bin/build.rs
jprendes: I opened #1401 to track this issue.

src/hyperlight_guest_bin/c/clock.c
jprendes: I believe this comment is resolved.
I opened #1398 to track the libc bindings is rust.

syntactically (approved):

raftario:
Thank you so much for all the work on this !

PR #839, opened 2025-08-28
Make the component macros support passing host resources to guests

syntactically:
This should fix a bunch of issues including those that #829 was getting at.


jprendes:
I ran the updated code against the wasi:http demo and it works.
https://github.com/jprendes/hyperlight-wasm-http-example

jprendes (approved):
LGTM

Issue #826, opened 2025-08-27
Memory Leak in Guest on Error Calling Host Function


adamperlin:
### What happened?

This leak was uncovered in fuzzing. When a guest calls a host function, if there is an error calling the host function itself -- for example, the host function does not exist -- the error propagates on the host side and we never return back to the guest to allow it to unwind and drop any allocated resources. In the `fuzz_host_call` target, we are currently running into memory exhaustion because we try to execute many iterations of such faulty host calls on the same sandbox in sequence. 

### What did you expect to happen?

If there is any error produced in attempting to call a host function in the guest, we need to yield control back to the guest and trigger some kind of unwind.

### Steps to reproduce the behavior

1. Checkout the branch used in #818
2. Run `just guests; just fuzz fuzz_host_call`
3. Fuzzing should fail with a memory allocation error almost immediately

### Hyperlight Version

0.8.0

### OS version

<details>

```console
On Linux:
$ cat /etc/os-release

PRETTY_NAME="Ubuntu 22.04.5 LTS"
NAME="Ubuntu"
VERSION_ID="22.04"
VERSION="22.04.5 LTS (Jammy Jellyfish)"
VERSION_CODENAME=jammy
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=jammy

$ uname -a
Linux user 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux
````
</details>

PR #840, opened 2025-08-28
Fixes to Host Call Fuzzing

adamperlin:
This PR prevents `fuzz_host_call` from hitting a memory leak upon encountering host function calling errors. It restores to a known clean snapshot on each iteration. This fix should not be needed once #826 is fixed. 

 Snapshot restore initially wasn't working in the fuzzing case due to a bug discovered by @ludfjig in `call_type_erased_guest_function_by_name` (snapshot wasn't being set to None) so this bug has been fixed. 

This PR also more explicitly ignores some expected errors that may come up from host call fuzzing.

### DIFF
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -381,6 +381,8 @@ impl MultiUseSandbox {
         ret_type: ReturnType,
         args: Vec<ParameterValue>,
     ) -> Result<ReturnValue> {
+        // Reset snapshot since we are mutating the sandbox state
+        self.snapshot = None;
         maybe_time_and_emit_guest_call(func_name, || {
             self.call_guest_function_by_name_no_reset(func_name, ret_type, args)
         })
--- a/fuzz/fuzz_targets/host_call.rs
+++ b/fuzz/fuzz_targets/host_call.rs
@@ -19,34 +19,57 @@ limitations under the License.
 use std::sync::{Mutex, OnceLock};
 
 use hyperlight_host::func::{ParameterValue, ReturnType};
+use hyperlight_host::sandbox::SandboxConfiguration;
+use hyperlight_host::sandbox::snapshot::Snapshot;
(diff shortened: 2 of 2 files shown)


ludfjig (approved):
Nice, lgtm

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -381,6 +381,8 @@ impl MultiUseSandbox {
         ret_type: ReturnType,
         args: Vec<ParameterValue>,
     ) -> Result<ReturnValue> {
+        // Reset snapshot since we are mutating the sandbox state
+        self.snapshot = None;
```
jsturtevant, line 385: Why are we clearing this here?  Shouldn't the restore call above take care of this?

jsturtevant: otherwise I would have expected it too look like https://github.com/hyperlight-dev/hyperlight/blob/828a529a054ae865efb7a5f480c897bb32cf800d/src/hyperlight_host/src/sandbox/initialized_multi_use.rs#L243

adamperlin: I think this is a bug: since this function is wrapping `call_guest_function_by_name_no_reset` and we're mutating the guest sandbox, the attached snapshot will be invalidated here, right? If we don't clear the snapshot, the restore call will hit the "snapshot exists" case and won't do the restore.

adamperlin: For the moment we could change the behavior to do the `snapshot` and `restore` in the `call_type_erased_guest_function_by_name` though, but since it seemed more clear to do it in the fuzzing code!

jsturtevant: Oh, I think I miss read this. I think this is fine and mimics the 'call func'

adamperlin: Ah ok, I'm glad it looks fine!

Issue #704, opened 2025-07-09
Update guest traces to standard format


dblnz:
To be able to better visualize the guest traces it would be helpful if this data is represented in a standard format that can be consumed by other tools (OTEL).

syntactically:
There might also be some overlap with using the same mechanism to achieve #318 ?

dblnz:
This is blocked by #375

dblnz:
This has been addressed by https://github.com/hyperlight-dev/hyperlight/pull/844

Issue #318, opened 2025-03-04
Investigate making trace spans flow into the hyperlight guest


simongdavies:
Hyperlight makes use of [Rust tracing Spans](https://docs.rs/tracing/latest/tracing/span/index.html) to trace threads of execution through a Sandbox, this is done primarily throught the [instrument macro](https://docs.rs/tracing/latest/tracing/attr.instrument.html).

We should explicitly enable a span to continue across a guest->host->guest call chain.

It may be possible to do this using Trace Context Propogator, see [here](https://github.com/Mossaka/otel-condvar/blob/main/src/main.rs) for an example of this

dblnz:
This has been addressed by #844 which combines host and guest traces to achieve a complete trace

PR #844, opened 2025-08-31
Tracing improvements

### LINKED ISSUE #723: Improve guest tracing
The following issue is a follow-up ticket to track some remaining comments from the PR #695 
 
- [it would be neat to think about changing the exit ABI so that when tracing is enabled we always bring along a trace structure in some extra registers and can coalesce the exits.](https://github.com/hyperlight-dev/hyperlight/pull/695#discussion_r2220497987)
- [produce a separate file with call site -> string message mappings, and only record the call site number in the buffer](https://github.com/hyperlight-dev/hyperlight/pull/695#discussion_r2220497992)
- [check tracing output for expected values](https://github.com/hyperlight-dev/hyperlight/pull/695#discussion_r2220497997)
- [reconsider the `flush` macro](https://github.com/hyperlight-dev/hyperlight/pull/695#discussion_r2220498021)
- [Unify the hypervisors registers with the `read_trace_req` to not have yet another type that represents registers](https://github.com/hyperlight-dev/hyperlight/pull/695#discussion_r2220498017)

dblnz:
## Description
This PR closes #723, #704 and partially addresses #318.
These changes change the way we perform guest tracing to using the `tracing` crate and its macros (`instrument`, trace).

### Guest
What makes this possible is the implementation of the `Subscriber` trait in the `hyperlight-guest-tracing` crate. By implementing it, we can now handle the capturing of spans and events and choose how to store them and when to export them to the host.

The `GuestSubscriber` type that implements `Subscriber` keeps an internal `TraceState` that holds all the needed information.
Whenever a new span is created, entered or exited, a callback on the subscriber is called so that we can handle the functionality. The same happens with the events also.

Each time a new span or event is added to the internal state, we check whether the buffer got full and send them to the host to process.

### Host
When the host detects a VM exit from the guest, it checks whether it contains tracing information in the `OutB` instruction.
When tracing information is found, the host starts going through it and check against the local storage of spans.
- If the spans have previously been created, just update the end timestamp (if present) and add new events (if any).
- If they haven't been created, create and store them.

The spans parents are set based on the information got from the host.

## TODO
- [ ]  The current issue is with the guest calls that end up calling back into the host.
These do not correctly set the parents of the spans created in the host to the last one created in the guest before doing the VM exit
I need to find a way to propagate the context into the guest and back whenever it is needed. But using the Opentelemetry propagators needs `std` support which we do not have in the guest.
- [ ] There is a corner case with calculating the timestamp for the span that is open when a VM exit is done

<details>
<summary>Jaeger picture of a Guest call that calls back into the host </summary>
image
</details>


src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -1094,15 +1095,9 @@ impl Hypervisor for HypervWindowsDriver {
     }
 
     #[cfg(feature = "trace_guest")]
-    fn read_trace_reg(&self, reg: TraceRegister) -> Result<u64> {
+    fn read_trace_reg(&self) -> Result<X86_64Regs> {
```
jprendes: should this be `read_regs(&self)`?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -1149,11 +1149,7 @@ impl Hypervisor for HypervLinuxDriver {
     }
 
     #[cfg(feature = "trace_guest")]
-    fn trace_info_as_ref(&self) -> &TraceInfo {
-        &self.trace_info
-    }
-    #[cfg(feature = "trace_guest")]
-    fn trace_info_as_mut(&mut self) -> &mut TraceInfo {
+    fn trace_info_mut(&mut self) -> &mut TraceInfo {
```
jprendes: \<rant\>
Ah, I never understand this pattern, why isn't trace_info directly a public member?
But we have that all over the codebase, so... it's ok here
\</rant\>

Cargo.toml
```diff
@@ -11,7 +11,6 @@ members = [
     "src/hyperlight_host",
     "src/hyperlight_guest_capi",
     "src/hyperlight_guest_tracing",
-    "src/hyperlight_guest_tracing_macro",
```
jprendes, line 14: Should we keep it around, but as deprecated, shimming the new API? so that we don't break anyone who was already using this?

src/hyperlight_common/src/outb.rs
```diff
@@ -114,8 +111,6 @@ impl TryFrom<u16> for OutBAction {
             101 => Ok(OutBAction::CallFunction),
             102 => Ok(OutBAction::Abort),
             103 => Ok(OutBAction::DebugPrint),
-            #[cfg(feature = "trace_guest")]
-            104 => Ok(OutBAction::TraceRecord),
```
jprendes: I haven't though about this before, but, does this mean that a guest compiled with an older hyperlight_guest doesn't run in a newer hyperlight_host?
We should think about this more carefully, not particularly in this PR, but I think in general we haven't been too diligent on paying attention to that.

jprendes: I've created #845 to track that.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -923,6 +925,21 @@ impl Hypervisor for HypervLinuxDriver {
...
             },
         };
+
+        // If trace is enabled, process the trace batch
+        #[cfg(feature = "trace_guest")]
+        match result {
+            HyperlightExit::Halt()
+            | HyperlightExit::IoOut(_, _, _, _)
+            | HyperlightExit::Mmio(_) => {
+                // If the result is not a halt, io out, mmio or debug exit, we need to process the trace batch
```
jprendes: Is the comment wrong? the logic wrong? or my understanding wrong?
```suggestion
                // If the result is a halt, io out, mmio or debug exit, we need to process the trace batch
```

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -875,6 +877,20 @@ impl Hypervisor for HypervWindowsDriver {
             }
         };
 
+        // If trace is enabled, process the trace batch
+        #[cfg(feature = "trace_guest")]
+        match result {
+            HyperlightExit::Halt()
+            | HyperlightExit::IoOut(_, _, _, _)
+            | HyperlightExit::Mmio(_) => {
+                // If the result is not a halt, io out, mmio or debug exit, we need to process the trace batch
```
jprendes: Same here

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -809,6 +811,21 @@ impl Hypervisor for KVMDriver {
...
             }
         };
+
+        // If trace is enabled, process the trace batch
+        #[cfg(feature = "trace_guest")]
+        match result {
+            HyperlightExit::Halt()
+            | HyperlightExit::IoOut(_, _, _, _)
+            | HyperlightExit::Mmio(_) => {
+                // If the result is not a halt, io out, mmio or debug exit, we need to process the trace batch
```
jprendes: Same here

jprendes (commented):
I made a high level review, and what I've seen looks good :-)

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
dblnz: Because in this case, I need to retrieve the `TraceInfo` through the `&dyn Hypervisor` which doesn't know about the `KvmDriver`'s member field.

I don't like it either, to be honest, I think this will look a lot better when we unify the Hypervisors into a single struct. A lot of duplicated code will go also

Cargo.toml
dblnz: Yes, I think this is the way to go. But I don't think we'll ever use it again

src/hyperlight_common/src/outb.rs
dblnz: Yes, and in this case, Ideally we should keep that `104` value reserved

src/hyperlight_common/Cargo.toml
```diff
@@ -26,7 +26,6 @@ spin = "0.10.0"
 default = ["tracing"]
 fuzzing = ["dep:arbitrary"]
 trace_guest = []
-unwind_guest = []
```
ludfjig, line 29: What happened to this feature?

src/hyperlight_guest/src/exit.rs
```diff
@@ -18,26 +18,53 @@ use core::arch::asm;
 use core::ffi::{CStr, c_char};
 
 use hyperlight_common::outb::OutBAction;
+use tracing::{Span, instrument};
 
 /// Halt the execution of the guest and returns control to the host.
 #[inline(never)]
-#[hyperlight_guest_tracing::trace_function]
+#[instrument(skip_all, parent = Span::current(), level= "Trace")]
```
ludfjig: I know we have `parent = Span::current()` in host as well, but I was wondering if we can leave this out if it's already the default?

src/hyperlight_guest/src/guest_handle/host_comm.rs
```diff
@@ -175,24 +176,46 @@ impl GuestHandle {
...
+                caller = caller,
+                source_file = source_file,
+                code.lineno = line,
+            );
+        } else {
+            send_to_host();
+        }
+        #[cfg(not(feature = "trace_guest"))]
+        {
+            send_to_host();
```
ludfjig, line 205: Is this logic correct? Would there be any differences in logging when tracing is enabled and using `trace!` vs disabled?

src/hyperlight_guest_tracing/src/invariant_tsc.rs
```diff
@@ -0,0 +1,47 @@
...
+
+/// Read the timestamp counter
+///
+/// This function provides a high-performance timestamp by reading the TSC.
+/// Should only be used when invariant TSC is supported for reliable timing.
+///
+/// # Safety
+/// This function uses unsafe assembly instructions but is safe to call.
+/// However, the resulting timestamp is only meaningful if invariant TSC is supported.
+pub fn read_tsc() -> u64 {
```
ludfjig, line 45: Maybe we don't need this safety comment if it's safe?

jsturtevant:
> I am curious what the performance looks like with tracing vs without, maybe we could add a benchmark or something for this?

+1

src/hyperlight_common/Cargo.toml
dblnz: I merged it into the `mem_profile` one as it is only used there and it increases the number of combinations we need to test in CI

src/hyperlight_guest/src/exit.rs
dblnz: We could do that

src/hyperlight_guest/src/guest_handle/host_comm.rs
dblnz: Yes.
So the current mechanism we have is doing a `VM Exit` for each log we emit on the guest so that the host receives the info and does the emitting on behalf of the guest. We do not want those kind of interruptions if we can batch the traces, right?

dblnz:
> First round review looks good to me. I am curious what the performance looks like with tracing vs without, maybe we could add a benchmark or something for this?

Ok, I can do that.

> Also, is there any possibility that we don't flush spans/records after exiting the guest, and that some end up not being emitted?

Hmm, in my limited testing I haven't seen this case, but I wouldn't exclude the possibility.
One scenario that comes to mind is cancellation, where the vCPU forced to stop, so no `out` instruction to deliver the info.
Any ideas how we can treat these scenarios?

> Another thing to consider is log crate vs tracing crate. Should we ditch one? Or is there some mechanism that allows regular logs to be consumed by tracing crate: And should we expose any of these in guest C-api?

I am not sure about the best approach is.
The solution I added for guest spans/events only works with opentelemetry Subscribers on the host, so if no opentelemetry subscribers, the they won't be captured (haven't tried it yet, but this is how it should work)

> Also is the tracing buffer sizes configurable? Maybe it should be if it isn't, so users can tweak it in case it affects performance.

The tracing buffer is compile time configurable, which I agree is not ideal for customers.
But I needed a fixed size buffer so that I could reliably give the pointer to the host to access memory (current approach, I don't know if it is the best, but it is certainly faster than using the input/output buffer which copies data twice. This way we only copy once, on the host).

dblnz:
> First round review looks good to me. I am curious what the performance looks like with tracing vs without, maybe we could add a benchmark or something for this?

I've run some benchmarks locally and here are the results.
Relative to the work done in the dummy guest, the tracing logic could look like it takes a lot, but we need a realistic guest scenario to correctly assess how relevant the numbers are.

| Runtime                  | Strategy | Flavour| RPS      | p50 (s) | p95 (s) | p99 (s) | p99.99 (s) | Peak RSS (MB) |
| ------------------------ | -------- | ------ | -------- | ------- | ------- | ------- | ---------- | ------------- |
| hyperlight-dummy         | new      |        | 69.96    | 0.7065  | 1.1569  | 1.2509  | 1.3412     | 12.79         |
| hyperlight-dummy-tracing | new      | nolog  | 70.92    | 0.6728  | 1.1247  | 1.2065  | 1.3029     | 20.21         |
| hyperlight-dummy-tracing | new      | log    | 70.14    | 0.7354  | 1.0132  | 1.0953  | 1.1922     | 18.54         |
| hyperlight-dummy         | reload   |        | 93506.17 | 0.0004  | 0.0011  | 0.0018  | 0.0067     | 15.00         |
| hyperlight-dummy-tracing | reload   | nolog  | 56152.01 | 0.0006  | 0.0023  | 0.0049  | 0.0575     | 23.42         |
| hyperlight-dummy-tracing | reload   | log    | 38585.95 | 0.0009  | 0.0035  | 0.0069  | 0.0507     | 21.90         |
| hyperlight-dummy         | reuse    |        | 80821.05 | 0.0005  | 0.0016  | 0.0026  | 0.0065     | 14.85         |
| hyperlight-dummy-tracing | reuse    | nolog  | 54745.61 | 0.0007  | 0.0020  | 0.0049  | 0.0581     | 23.28         |
| hyperlight-dummy-tracing | reuse    | log    | 41301.24 | 0.0009  | 0.0028  | 0.0056  | 0.1097     | 21.24         |

Runtimes:
- hyperlight-dummy - Hyperlight Sandbox running a dummy guest (no TracingProvider)
- hyperlight-dummy-tracing - Hyperlight Sandbox running a dummy guest with `trace_guest` enabled on both guest and host and a TracingProvider instantiated
    - nolog - the max log level is `none`, which means the guest doesn't send tracing info, but the Host logic to handle them is enabled
    - log - the max log level is `trace` for the guest, which makes the guest send tracing info and the host to handle them

Some thoughts:
- This feature started as a `development` phase improvement to help figure out where we have bottlenecks on the guest.
This is a huge improvement over the previous solution we had.
- At this point, I do not think this should be used in production by default because the performance is not ideal. It could be useful on an error path, when something doesn't work as expected, it could be turned on
- These numbers do not perfectly reflect only the `tracing` logic comparison because there are multiple variables such as: the communication with the Jaeger Collector to send the tracing data (when the `TracingProvider` is installed).
- Some user feedback would be perfect
- This would benefit from other generic performance improvements, such as (sharing memory with the guest in such a way that doesn't imply copying memory - guest immutable references).
 I think we should first discuss where we stand in terms of these big changes in Hyperlight, which may also positively impact the tracing performance.

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -232,9 +234,13 @@ pub extern "C" fn entrypoint(peb_address: u64, seed: u64, ops: u64, max_log_leve
                 .expect("Invalid log level");
             init_logger(max_log_level);
 
-            trace!("hyperlight_main",
-                hyperlight_main();
-            );
+            // It is important that all the tracing events are produced after the tracing is initialized.
+            #[cfg(feature = "trace_guest")]
+            if max_log_level == LevelFilter::Trace {
```
ludfjig: why only init if `trace` level?

docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -110,19 +109,57 @@ just build-rust-guests debug trace_guest
...
+
+When the guest starts executing the `entrypoint` function, it receives a `max_log_level` parameter that tells the guest what kind of logging level is expected from it.
+
+The `trace_guest` logic takes advantage of this parameter and when the `max_log_level` is `trace`, it allocates a custom made `GuestSubscriber` that implements the `Subscriber`
+trait from `tracing_core` that allows defining a subscriber for the `tracing` crate to handle new spans and events.
+
+This custom subscriber stores the spans and events in a buffer initialized only when tracing is enabled. For each new span and event, a method is called on the custom subscriber which
+not only stores the data, but also keeps track of the hierarchy and dependencies between the other spans/events.
+
+When the storage space is filled, the guest triggers a VM Exit that sends the guest pointers to the host. The host can access the guest memory, get the data and parse it to create the `spans` and `events` using the `opentelemetry` crate which allows specifying the starting and ending timestamps
```
ludfjig, line 148: Maybe an option could be to only emit logs during regular VMexits in order to minimize context switches. Or maybe another option could be to discard logs when the buffer is full in order to not incur extra context switches

dblnz: Yes, this is another option.
I prioritized not losing any info, but we might not care about that. A ringbuffer would do the trick here

src/hyperlight_guest_bin/src/lib.rs
dblnz: This is to avoid the performance penalty if it is not really needed.
This provides a way to selectively enable it at runtime for development purposes.

I wouldn't say it is yet ready for production, but it definitely is a step in the right direction, starting from what we had (something custom that needs maintenance), to something that uses the `tracing` and `opentelemetry` crates.

docs/hyperlight-metrics-logs-and-traces.md
ludfjig: yes. If this could be configurable that would be ideal i think

dblnz: That would be a nice feature, but at this moment, the logic on the host relies on the spans not being lost so that it knows how to `arrange` them correctly.
Each span contains info about it's parent span, if one of those is lost, then there's a gap.

A ringbuffer would work well with logs, but traces, I am not sure.

dblnz:
> This looks good to me in general, but I think maybe we should consider what tunables we need

I see this version using opentelemetry as an incremental step, after which additional tunables can be added to improve usability.
Unless there is any fundamental issue with this approach, I think this can be merged and improved after.

Improvements:
- Switch to using a circular buffer for traces/events to avoid context switches
- Provide a mechanism on the guest to enable/disable the tracing support, configure the memory used and log level at runtime:

src/hyperlight_guest_bin/src/lib.rs
jsturtevant: What if I wanted to turn on tracing for  LevelFilter::DEBUG or LevelFilter::INFO

jsturtevant: Should this be `!= LevelFilter::Off` https://docs.rs/tracing/latest/tracing/level_filters/struct.LevelFilter.html#impl-LevelFilter

src/hyperlight_guest_tracing/src/visitor.rs
```diff
@@ -0,0 +1,56 @@
...
+        let _ = k.push_str(&field.name()[..usize::min(field.name().len(), k.capacity())]);
+        let _ =
+            val.push_str(&alloc::format!("{value:?}")[..usize::min(value.len(), val.capacity())]);
+        let _ = self.out.push((k, val));
+    }
+    fn record_str(&mut self, f: &Field, v: &str) {
+        let mut k = heapless::String::<FK>::new();
+        let mut val = heapless::String::<FV>::new();
+        // Shorten key and value if they are bigger than the space allocated
+        let _ = k.push_str(&f.name()[..usize::min(f.name().len(), k.capacity())]);
```
jsturtevant, line 53: is there a chance the key gets reduced to something that would overwrite and corrupt data?

src/hyperlight_host/src/hypervisor/arch/mod.rs
jsturtevant, line 1: is this the similiar to what is in https://github.com/hyperlight-dev/hyperlight/pull/907?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -602,7 +602,7 @@ mod tests {
     fn test_with_small_stack_and_heap() {
         let mut cfg = SandboxConfiguration::default();
         cfg.set_heap_size(20 * 1024);
-        cfg.set_stack_size(16 * 1024);
+        cfg.set_stack_size(18 * 1024);
```
jsturtevant, line 605: is this for the extra space needed with traces?

src/hyperlight_host/src/hypervisor/arch/mod.rs
dblnz: Yes, if that gets merged before this PR, I'll update to use it

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
dblnz: Yes, I discovered lately that this test failed after I rebased to latest main. After some investigation, it happens because of the traces

src/hyperlight_guest_tracing/src/visitor.rs
dblnz: I don't see how, it is trimmed to the minimum between its length and the capacity of the buffer, so it shouldn't go beyond the capacity of the container.

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -231,9 +233,13 @@ pub extern "C" fn entrypoint(peb_address: u64, seed: u64, ops: u64, max_log_leve
                 .expect("Invalid log level");
             init_logger(max_log_level);
```
simongdavies, line 234: Should we only call `init_logger` if either `#[not(cfg(feature = "trace_guest"))]` or if this is set as an else to `if max_log_level != LevelFilter::Off`?

I don't think we should log if tracing is on (we should redirect and `log::<level>!` macros to tracing (will this just happen by default?). Conversely if `trace_guest` is off we should capture any `trace::<level>!` macros as logs (see https://docs.rs/tracing/latest/tracing/#emitting-log-records). 

Maybe the way we should deal with this is to always emit traces from the guest (and get rid of the logging functionality in the guest (which makes the comment above redundant). Likewise in the host we could always use tracing, that way a consumer would have full control over if they saw logs or traces?

docs/hyperlight-metrics-logs-and-traces.md
```diff
@@ -110,19 +109,57 @@ just build-rust-guests debug trace_guest
 just move-rust-guests debug
 ```
 
-This will build the guest binaries with the `trace_guest` feature enabled and move them to the appropriate location for use by the host.
+This builds the guest binaries with the `trace_guest` feature enabled and move them to the appropriate location for use by the host.
+
+**NOTE**: To enable the tracing in your application you need to use the `trace_guest` feature on the `hyperlight-guest-bin` and `hyperlight-guest` crates.
```
simongdavies, line 114: I don't think this is strictly true? I can trace the host without this feature?.

Ideally it would be nice to get rid of this feature entirely and have it the functionality enabled by the presence of a subscriber in the host but that is maybe something we can think about in the future

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -13,364 +13,212 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+#[cfg(feature = "trace")]
+pub use state::TraceBatchInfo;
+#[cfg(feature = "trace")]
+pub use trace::{
+    clean_trace_state, end_trace, guest_trace_info, init_guest_tracing, is_trace_enabled,
+    set_start_tsc,
+};
+
+/// Maximum number of spans that the guest can store
+const MAX_NO_OF_SPANS: usize = 10;
```
simongdavies, line 46: Just wondering if rather than having fixed sizes for these we could just allocate on the guest heap and then have them processed each time we have a natural VM Exit, probably not an issue for this PR, but if we were to unify all the guest logging into traces (see my other comment about this) then this would probably be a better approach.

docs/hyperlight-metrics-logs-and-traces.md
dblnz: We need a way to capture, store and communicate to the host the captured traces.
By having this custom Subscriber, we have that logic defined by us, so we know the format to expect from the guest.
If we find a way to address this, then we don't need that feature

src/hyperlight_guest_tracing/src/lib.rs
dblnz: I agree with this, I would like to expose a way to configure the space used.
So the guest can say what memory it can spare

src/hyperlight_guest_tracing/src/visitor.rs
jsturtevant: What if two fields shorten to the same name?  I was thinking these are keys in a dictionary or something similar?

dblnz: Oh, no, it is a Vec<(String, String)>, so it doesn't check this. For each (k, v), a new (k, v) is pushed on that vec. It depends on what the `tracing` crate does on the `host` when the fields are set.

jsturtevant: Is it possible that the host is using the key as a unique key?

dblnz: It might be possible, it depends on the implementation of host Subscriber.

I'll double check, but from what I remember, I am provinding the vec of (k, v) in the host, the same way it was peovided in the guest. So I don't do any assumption about what the keys are or represent.

Let me double check this to make sure

src/hyperlight_guest_bin/src/lib.rs
dblnz: This sounds good, we should have a unified solution for logging/tracing guests.
I'll make this change as an improvement after, as it needs some testing to ensure the functionality.

dblnz: I've modified the logger to emit traces when tracing is enabled. So this call to `init_logger` is aware of tracing.
But I plan on changing it a bit to be more reliable.
I think, a unified solution to both logging and tracing would be best.

I'll do that as a separate issue.

src/hyperlight_guest_tracing/src/lib.rs
dblnz: I created #964 to address this.

src/hyperlight_guest_bin/src/lib.rs
dblnz: I created https://github.com/hyperlight-dev/hyperlight/issues/964 to address this.

src/hyperlight_guest_bin/src/lib.rs
dblnz: I changed this

docs/hyperlight-metrics-logs-and-traces.md
dblnz: I created https://github.com/hyperlight-dev/hyperlight/issues/964 to address this.

src/hyperlight_guest_tracing/src/visitor.rs
dblnz: So, I've had another look at the codebase to see if there's any issue. I found the following:
- When a new span is created, to extract the `record` fields and store them, we call the `record` method that uses the logic above ([Source](https://github.com/dblnz/hyperlight/blob/tracing-improvements/src/hyperlight_guest_tracing/src/state.rs#L191))
- When the host processes those fields, one by one, it calls `set_attribute` on the span to update the attributes ([Source](https://github.com/dblnz/hyperlight/blob/tracing-improvements/src/hyperlight_host/src/sandbox/trace/context.rs#L260))

It is as if one would add/call `set_attribute` for each field on the guest, so however the `set_attribute` method handles using the same `key`, that is the behaviour.
Let me know if that address your concern or whether I understood your concern correctly.

jsturtevant: Thanks for looking, it seems mostly not an issue since it is just some attribute data.  It could overwrite some data which might be confusing when debugging an issue later.  We might want to add a note in the readme about this truncation as a warning, or maybe log a warning here.

dblnz: I've documented this in the docs and on the methods.
Anyway, this is going to go away, I think, with the next improvements.
We'll use heap and maybe this truncation won't be needed

src/hyperlight_common/src/outb.rs
```diff
@@ -90,23 +90,20 @@ impl TryFrom<u8> for Exception {
 /// - CallFunction: makes a call to a host function,
 /// - Abort: aborts the execution of the guest,
 /// - DebugPrint: prints a message to the host
-/// - TraceRecordStack: records the stack trace of the guest
+/// - TraceBatch: reports a batch of spans and vents from the guest
```
ludfjig: nit: `vents` -> `events`

ludfjig (approved):
I'm a big fan of this! Btw what happens when buffer in guest is full and an event/span comes in?  Also question: is the `parent=Span::current` required inside all the 
`#[instrument(skip_all, parent = Span::current(), level= "Trace")]`? Or is it the default

dblnz:
The buffer length is verified every time after a new event/span is pushed. So if it got full, it is emptied afterwards.

The ´parent = Span::current()´ is not needed. I am planning to change that afterwards with some other small things.

Issue #606, opened 2025-06-12
Support for `x86_64-unknown-linux-musl` target


simongdavies:
#601 adds a couple of changes so that a use cab successfully build a hyperlight-host with the ` x86_64-unknown-linux-musl` target.

We need to either test for this target or switch to it in our CI, if we decide to test for it we should probably do this periodcially and consider moving some of the tests that we do on each PR to a periodic basis (i.e. test against less targets during PR validation).

danbugs:
Support exists, but we need a way to validate we don't regress it (e.g., add it in CI?). We'll discuss in v0.8.0.

jsturtevant:
We should also ensure this builds for projects like https://github.com/hyperlight-dev/hyperlight-wasm

I did run a quick test and it did work.

Issue #626, opened 2025-06-13
CI instability on `test_gdb_end_to_end `


ludfjig:
GDB test related failed

https://github.com/hyperlight-dev/hyperlight/actions/runs/15645064712/job/44080943114?pr=624

dblnz:
Output saved here:
```
cargo test --profile=dev --example guest-debugging --features gdb
   Compiling managed v0.8.0
   Compiling hyperlight-host v0.6.1 (/mnt/vss/_work/hyperlight/hyperlight/src/hyperlight_host)
   Compiling gdbstub v0.7.6
   Compiling gdbstub_arch v0.3.2
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 5.94s
     Running unittests examples/guest-debugging/main.rs (target/debug/examples/guest_debugging-37b3b6b63a06b41c)

running 1 test
   Compiling hyperlight-host v0.6.1 (/mnt/vss/_work/hyperlight/hyperlight/src/hyperlight_host)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 2.89s
     Running `/mnt/vss/_work/hyperlight/hyperlight/target/debug/examples/guest-debugging`
test tests::test_gdb_end_to_end ... FAILED

failures:

---- tests::test_gdb_end_to_end stdout ----

thread 'tests::test_gdb_end_to_end' panicked at src/hyperlight_host/examples/guest-debugging/main.rs:237:9:
Failed to remove gdb.output file: No such file or directory (os error 2)
stack backtrace:
   0:     0x5ba76d983e8a - std::backtrace_rs::backtrace::libunwind::trace::h2740d05102fd9881
                               at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:     0x5ba76d983e8a - std::backtrace_rs::backtrace::trace_unsynchronized::h93ae2edf130065c8
                               at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x5ba76d983e8a - std::sys::backtrace::_print_fmt::h7660a544e6110dd9
                               at /rustc/4eb161250e340c8f48f66e2b929ef4a5bed7c181/library/std/src/sys/backtrace.rs:66:9
[...]

PR #849, opened 2025-09-02
Add a periodic job for building and test musl

jsturtevant:
fixes: #606

This pull request refactors the CI workflows and `Justfile` to improve support for musl target.  It refactors the rust build job so it can be re-used across multiple targets and other github workflows. This doesn't currently reduce the number of jobs we run on PR's but sets the foundation for being able to reduce the number of items run.

When enabling musl I found a few items which will need follow up (will create issues):

- Building tracy-client (used in one of the tests) [didn't build on musl cleanly](https://github.com/houseabsolute/actions-rust-cross/issues/22).  I was able to get it build using cross but there was more extensive changes require to get cross working so disabled it for now, the rest of the other examples work.
- The seccomp tests fail with `signal: 11, SIGSEGV: invalid memory reference` when executing host functions that violate a seccomp rule.  We've seen a few other issues around this code so to get basic coverage up I've disabled this for now and plan to revisit enabling this feature.
- Includes a fix for https://github.com/hyperlight-dev/hyperlight/issues/626. Not sure why this was easier to reproduce with musl builds but I ran into this race condition a couple times while testing.

This does require installing some tools on the host VM's so I've marked this in draft until those tools are enabled.

### DIFF
--- a/.github/workflows/RustNightly.yml
+++ b/.github/workflows/RustNightly.yml
@@ -1,0 +1,25 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Nightly
+
+on:
+  workflow_dispatch:
+
+jobs:
+  musl:
+    strategy:
+      fail-fast: true
+      matrix:
+        hypervisor: [kvm]
+        cpu: [amd, intel]
+        config: [debug, release]
+    uses: ./.github/workflows/dep_rust.yml
+    secrets: inherit
+    with: 
+      hypervisor: ${{ matrix.hypervisor }}
+      cpu: ${{ matrix.cpu }}
+      config: ${{ matrix.config }}
+      target_triple: x86_64-unknown-linux-musl
+
+# TODO: using notification script
+
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -11,6 +11,26 @@ on:
         required: false
         type: string
         default: "false"
+      target_triple:
+        description: Target triple for cross-compilation
+        required: false
+        type: string
+        default: ""
+      hypervisor:
+        description: Hypervisor for this run (passed from caller matrix)
+        required: false
+        type: string
+        default: "kvm"
+      config:
+        description: Build configuration for this run (passed from caller matrix)
+        required: false
+        type: string
+        default: "debug"
+      cpu:
+        description: CPU architecture for the build (passed from caller matrix)
+        required: false
+        type: string
+        default: "amd"
 
 env:
   CARGO_TERM_COLOR: always
@@ -32,13 +52,8 @@ defaults:
 
 jobs:
   code-checks:
-    if: ${{ inputs.docs_only == 'false' }}
+    if: ${{ inputs.docs_only == 'false' && (inputs.hypervisor == 'hyperv-ws2025' || inputs.hypervisor == 'kvm') }}
     timeout-minutes: 60
-    strategy:
-      fail-fast: true
-      matrix:
-        hypervisor: ['hyperv-ws2025', kvm]
-        config: [debug, release]
     runs-on: ${{ fromJson(
       format('["self-hosted", "{0}", "X64", "1ES.Pool=hld-{1}-amd"]', 
         (matrix.hypervisor == 'hyperv-ws2025') && 'Windows' || 'Linux', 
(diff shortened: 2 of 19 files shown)


simongdavies:
Was it expected that the code checks got skipped?

.github/workflows/RustNightly.yml
```diff
@@ -0,0 +1,39 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Nightly
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: '0 0 */2 * *'
```
ludfjig, line 9: mind adding a comment here about what period this is?

.github/workflows/dep_rust.yml
```diff
@@ -11,6 +11,26 @@ on:
         required: false
         type: string
         default: "false"
+      target_triple:
+        description: Target triple for cross-compilation
+        required: false
+        type: string
+        default: ""
```
ludfjig, line 18: If target_triple is not passed, will it still work with default value ""?

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -199,7 +199,7 @@ fn maybe_with_seccomp<T: Send>(
     })
 }
 
-#[cfg(not(all(feature = "seccomp", target_os = "linux")))]
+#[cfg(not(all(seccomp)))]
```
ludfjig: nit: remove `all`?

jsturtevant:
> Do we want to add a recipe to the Justfile similar to like-ci that also includes the nightly stuff.

The nice thing about the way this works now is that the `just like-ci` stays the same.  For the musl you can run:

```
TARGET_TRIPLE=x86_64-unknown-linux-musl just like-ci debug "features"
```

or 

can use https://just.systems/man/en/setting-variables-from-the-command-line.html
```
just target-triple=x86_64-unknown-linux-musl like-ci debug "features"
```

> Should we move some of the stuff we are doing on each PR to the nightly job (e.g. don't think we need both the Windowws Server versions on each PR)

I was thinking we should do this, I didn't know if we wanted to do that here or as a separate PR.

> Was it expected that the code checks got skipped?

Yes we weren't running the code checks for all platforms. When I extracted the logic out in the `rust_dep.yaml` I added a skip (`if`) to those so we didn't start running extra, which is why they show up as skips explicitly.  

We can either:

- pull the `checks` out to their own job and only run it stand alone for the platforms we want it run on 
- leave it as is with skips, or as we move things over we wouldn't need the skip (`if`) since we would run less on PR's and run the other platforms on the cron job

.github/workflows/dep_rust.yml
jsturtevant: Yes, the PR workflow doesn't set it.

.github/workflows/RustNightly.yml
jsturtevant: updated

src/hyperlight_host/src/sandbox/host_funcs.rs
jsturtevant: updated

.github/workflows/RustNightly.yml
```diff
@@ -0,0 +1,40 @@
...
+    needs: musl
+    if: always() && needs.musl.result == 'failure'
+    permissions:
+      issues: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v5
+
+      - name: Notify Nightly Failure
+        run: ./dev/notify-ci-failure.sh --title="Nightly musl Failure - ${{ github.run_number }}" --labels="area/ci-periodics,area/testing,kind/bug,lifecycle/needs-review"
```
simongdavies: Do you think we should add the release blocker label here? that way we cannot release without this job passing(or removing the label from the issue?)

jsturtevant: great idea

.github/workflows/dep_rust.yml
```diff
@@ -112,79 +124,86 @@ jobs:
...
           CARGO_TERM_COLOR: always
           RUST_LOG: debug
-        run: just run-rust-examples ${{ matrix.config }}
+          TARGET_TRIPLE: ${{ inputs.target_triple }}
+        run: just run-rust-examples ${{ inputs.config }}
+
 
       - name: Run Rust examples - linux
-        if: ${{ (runner.os != 'Windows') }}
+        if: false
```
ludfjig: is this intended?

ludfjig (approved):

.github/workflows/dep_rust.yml
jsturtevant: no 🤦, fixed

PR #851, opened 2025-09-02
Guest function improvements and macros

jprendes:


src/hyperlight_host/src/func/host_functions.rs
```diff
@@ -87,7 +88,7 @@ where
     // Use Arc in here instead of Box because it's useful in tests and
     // presumably in other places to be able to clone a HostFunction and
     // use it across different sandboxes.
-    func: Arc<dyn Fn(Args) -> Result<Output> + Send + Sync + 'static>,
+    func: Arc<dyn Function<Output, Args, HyperlightError> + Send + Sync + 'static>,
```
jprendes, line 91: Update doc comment

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -450,7 +450,8 @@ impl MultiUseSandbox {
                 Output::TYPE,
                 args.into_value(),
             );
-            Output::from_value(ret?)
+            let ret = Output::from_value(ret?)?;
```
jprendes, line 455: Explain why the `?` here

src/hyperlight_guest_bin/src/guest_function/definition.rs
```diff
@@ -35,6 +41,104 @@ pub struct GuestFunctionDefinition {
...
+            R: ResultType<HyperlightGuestError>,
+        {
+            #[doc(hidden)]
+            const ASSERT_ZERO_SIZED: () = const {
+                assert!(core::mem::size_of::<Self>() == 0)
+            };
+
+            fn into_guest_function(self) -> fn(&FunctionCall) -> Result<Vec<u8>> {
+                |fc: &FunctionCall| {
+                    // SAFETY: This is safe because:
```
jprendes, line 129: Expand the explanation here

jprendes: Check if we can implement IntoGuestFunction directly on `fn() -> ...`

src/hyperlight_common/src/func/ret_type.rs
```diff
@@ -70,28 +70,47 @@ macro_rules! impl_supported_return_type {
...
+    E: core::fmt::Debug,
+{
     type ReturnType = T;
 
     fn into_result(self) -> Result<Self::ReturnType, E> {
         Ok(self)
     }
+
+    fn from_result(res: Result<Self::ReturnType, E>) -> Self {
+        res.unwrap()
```
jprendes: clarify why unwrap is fine here

src/hyperlight_guest_macro/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        return Error::new(
+            fn_declaration.sig.asyncness.span(),
+            "Async functions are not allowed in guest functions",
+        )
+        .to_compile_error()
+        .into();
+    }
+
+    let output = quote! {
+        #fn_declaration
```
jprendes, line 143: Consider mentioning that you can end up exposing functions that the rust typesystem wouldn't let you expose otherwise.

src/hyperlight_guest_macro/src/lib.rs
```diff
@@ -0,0 +1,204 @@
...
+        let name: LitStr = input.parse()?;
+        if !input.is_empty() {
+            return Err(Error::new(input.span(), "expected a single identifier"));
+        }
+        Ok(NameArg::Name(name))
+    }
+}
+
+#[proc_macro_attribute]
+pub fn guest_function(attr: TokenStream, item: TokenStream) -> TokenStream {
```
jprendes, line 89: Add comments explaining what's going on here

src/tests/rust_guests/simpleguest/src/main.rs
```diff
@@ -664,1004 +367,269 @@ fn execute_on_heap(_function_call: &FunctionCall) -> Result<Vec<u8>> {
...
-    );
-    register_function(simple_print_output_def);
-
     let print_output_def = GuestFunctionDefinition::new(
         "PrintOutputWithHostPrint".to_string(),
         Vec::from(&[ParameterType::String]),
         ReturnType::Int,
         print_output_with_host_print as usize,
     );
     register_function(print_output_def);
```
jsturtevant, line 540: Why is this one kept as a explicit registration?

src/tests/rust_guests/simpleguest/src/main.rs
```diff
@@ -664,1004 +367,269 @@ fn execute_on_heap(_function_call: &FunctionCall) -> Result<Vec<u8>> {
...
+    assert_eq!(
+        e.message,
+        format!("HostFunction {hostfuncname} was not found")
+    );
+    Ok(())
 }
 
 #[no_mangle]
 #[instrument(skip_all, parent = Span::current(), level= "Trace")]
 pub extern "C" fn hyperlight_main() {
```
jsturtevant, line 533: We should update the read me with this new way to register functions. If there is there any reason to register functions manually, we should make a note of it.  

Do we want to keep a few basic manually registrations around as test cases?

src/hyperlight_guest_bin/src/guest_function/definition.rs
jprendes: > Check if we can implement IntoGuestFunction directly on `fn() -> ...`

nop, we can't

src/hyperlight_common/src/func/ret_type.rs
jprendes: Moved this bit to hyperlight-guest-bin, so there will be no risk of accidentally unwrapping in the host.

src/tests/rust_guests/simpleguest/src/main.rs
jprendes: `print_output_with_host_print` is defined in `hyperlight_guest_bin` with a signature of `fn(&FunctionCall) -> Result<Vec<u8>>`.
This is not a signature that the `guest_function` macro can work with as it's missing all the type information.

src/tests/rust_guests/simpleguest/src/main.rs
jprendes: Would the `print_output_with_host_print` registration be enough?

jprendes: Updated the README.md

README.md
```diff
@@ -74,54 +74,33 @@ fn main() -> hyperlight_host::Result<()> {
...
-            ErrorCode::GuestFunctionParameterTypeMismatch,
-            "Invalid parameters passed to simple_print_output".to_string(),
-        ))
-    }
+
+use hyperlight_guest::bail;
+use hyperlight_guest::error::Result;
+use hyperlight_guest_bin::{guest_function, host_function};
+
+#[host_function("HostPrint")]
```
ludfjig, line 86: What happens if two macros have the same arg to host_function(...)? Does it compile? Also, maybe we could consider making the arg optional, defaulting to the name of the function declaration (not sure if this is a good idea or not?)

ludfjig (commented):
Looks good to me. Btw, how does linkme work? xD

README.md
simongdavies: I asked both these questions when Jorge and I discussed this PR yesterday, I dont recall the answer for the first one, the second IIRC is supported.

jprendes: The argument is optional and does default to the function name, its just that simpleguest uses CamelCase instead of snake_case for all but 1 function.

Using the same ame twice does whatever it qould have done registering 2 functions with the same name before this PR, I don't know if its replaced or a runtime error.

We could look at making it a compile error as a follow up, or rather a linktime error (and that would only work for the functions using the macro, if you do the manual registration there's not much we can do)

jprendes: Sorry, the one that would make sense to generate an error is guest_function, host_function using the same name should be fine

simongdavies (dismissed):

jprendes:
> Looks good to me. Btw, how does linkme work? xD

This is the best explanation I found, other than inspecting the source code :-)
https://internals.rust-lang.org/t/from-life-before-main-to-common-life-in-main/16006/24

src/tests/rust_guests/simpleguest/src/main.rs
jsturtevant: Yea i think so, I believe we are mostly covered since we are generating the same code, but want to make sure that can still use the old way and we don't introduce subtle bugs in the two different ways of registering functions.

jsturtevant (approved):

dblnz (commented):
This looks great! Good work Jorge!
I didn't get to look very thoroughly, but overall it looks good

Issue #343, opened 2025-03-11
Change our no_std allocator


ludfjig:
We are currently using buddy-system-allocator inside guests. We should consider switching this out to a more performance allocator. 

Some options:

- mimalloc (if possible?)
- Talc https://github.com/SFBdragon/talc

Maybe more?

prydt:
Hi, I was wondering if I could take this issue.

If so, I was wondering if there are any good benchmarks for hyperlight I should use to compare different allocators.

jsturtevant:
> mimalloc (if possible?)

Just noting that there was a brief investigation into mimalloc and it was found that it needed 64Mb of memory for the initial arena which ruled it out.  Not sure if that is still the case.

PR #867, opened 2025-09-05
Use mimalloc as a guest allocator

andreiltd:
Depends on: https://github.com/hyperlight-dev/hyperlight/pull/831


ludfjig:
@andreiltd should we try this again now that we have picolibc and mimalloc v3 is out?

syntactically:
Related issue: #343

jsturtevant:
We will need to move to mimalloc v3 to use this with piclibc.  Might be relevant for parralelism.  Should use issue https://github.com/hyperlight-dev/hyperlight/issues/343 to track this.

PR #868, opened 2025-09-05
Restructure guest/host error handling

### LINKED ISSUE #497: Confusing ERROR tracing when not actually an error
https://github.com/ludfjig/hyperlight/blob/vm_trait/src/hyperlight_common/src/flatbuffer_wrappers/guest_error.rs#L199

causes any failures to be logged with ERROR level. However, we use this function unconditionally to check whether an error occured or not here https://github.com/ludfjig/hyperlight/blob/vm_trait/src/hyperlight_host/src/func/guest_err.rs#L28

Thus if the function fails, it is not an error

### LINKED ISSUE #826: Memory Leak in Guest on Error Calling Host Function

ludfjig:
This PR improves error handling between host and guest functions to prevent memory leaks and ensure more reliable deserialization.

Previously, when a host function (invoked from the guest) returned an error, the error was reported immediately without unwinding the guest stack. This left guest-side allocations in an inconsistent state, leading to memory leaks on subsequent entries into the guest.

In addition, error reporting relied on a fragile mechanism: guest errors were manually serialized into a buffer, and the host would attempt to detect them by trying to deserialize an error. If deserialization succeeded, an error was assumed to have occurred. This approach is risky because there was nothing preventing GuestError and FunctionCallResult from possibly having the same serialized format since they are completely separate.

Changes

- Guest/host function calls now always return a FunctionCallResult, which explicitly represents either Ok or Err.
- If host function returns an error, it's serialized back into the guest, and the guest will properly unwind, and report it back to the host, fixing a memory leak.

TODO: 

For C guests: If host function returns an error, guest will panic when trying to read an expected good value (because there is only an error instead), and leak memory as a result. Will fix this in follow up PR

Closes #826 
Closes #497

### DIFF
--- a/docs/how-to-use-flatbuffers.md
+++ b/docs/how-to-use-flatbuffers.md
@@ -1,6 +1,6 @@
 # How to use FlatBuffers
 
-> Note: the last generation of the flatbuffer code was with done with flatc version 25.2.10 (i.e., the last version as of May 1st, 2025).
+> Note: the last generation of the flatbuffer code was with done with flatc version 25.9.23 (i.e., the last version as of Oct 2nd, 2025).
 
 Flatbuffers is used to serialize and deserialize some data structures.
 


src/hyperlight_host/src/func/guest_err.rs
```diff
@@ -1,48 +0,0 @@
...
-
-/// Check for a guest error and return an `Err` if one was found,
-/// and `Ok` if one was not found.
-pub(crate) fn check_for_guest_error(mgr: &mut MemMgrWrapper<HostSharedMemory>) -> Result<()> {
-    let guest_err = mgr.as_mut().get_guest_error().ok();
-    let Some(guest_err) = guest_err else {
-        return Ok(());
-    };
-
-    metrics::counter!(
```
jsturtevant, line 35: did we replace this metric?

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -81,10 +82,12 @@ pub(crate) fn call_guest_function(function_call: FunctionCall) -> Result<Vec<u8>
 
 // This function is marked as no_mangle/inline to prevent the compiler from inlining it , if its inlined the epilogue will not be called
 // and we will leak memory as the epilogue will not be called as halt() is not going to return.
+//
+// This function may panic, as we have no other ways of dealing with errors at this level
```
jsturtevant, line 86: This seems like a tricky change, although the previous version looks to have just ignored errors?

src/hyperlight_host/tests/sandbox_host_tests.rs
```diff
@@ -374,6 +374,7 @@ fn host_function_error() -> Result<()> {
             assert!(
                 matches!(&res, HyperlightError::GuestError(_, msg) if msg == "Host function error!") // rust guest
             || matches!(&res, HyperlightError::GuestAborted(_, msg) if msg.contains("Host function error!")) // c guest
+            || matches!(&res, HyperlightError::StackOverflow()) // c guest. TODO fix this. C guest leaks when host func returns error guest panics.
```
jsturtevant, line 362: do we have an issue for this? was this the case prior?

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
```diff
@@ -588,123 +738,179 @@ impl TryFrom<FbFunctionCallResult<'_>> for ReturnValue {
...
-}
+// impl TryFrom<&[u8]> for ReturnValue {
+//     type Error = Error;
+//     #[cfg_attr(feature = "tracing", instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace"))]
+//     fn try_from(value: &[u8]) -> Result<Self> {
+//         let function_call_result_fb = size_prefixed_root::<FbFunctionCallResult>(value)
+//             .map_err(|e| anyhow!("Failed to get ReturnValue from bytes: {:?}", e))?;
+//         function_call_result_fb.try_into()
+//     }
+// }
```
jsturtevant: remove this?

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
```diff
@@ -22,15 +22,166 @@ use flatbuffers::size_prefixed_root;
...
+
+impl FunctionCallResult {
+    /// Encodes self into the given builder and returns the encoded data.
+    ///
+    /// # Notes
+    ///
+    /// The builder should not be reused after a call to encode, since this function
+    /// does not reset the state of the builder. If you want to reuse the builder,
+    /// you'll need to reset it first.
+    pub fn encode<'a>(&self, builder: &'a mut flatbuffers::FlatBufferBuilder) -> &'a [u8] {
```
jsturtevant, line 46: should we have some sanity tests like we do with https://github.com/hyperlight-dev/hyperlight/blob/b419bde3cec2bb2db1b0258891f52f620e1d3ca5/src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs#L283?

src/hyperlight_host/tests/sandbox_host_tests.rs
```diff
@@ -366,10 +366,17 @@ fn host_function_error() -> Result<()> {
...
         let mut init_sandbox: MultiUseSandbox = sandbox.evolve()?;
         let msg = "Hello world";
-        let res = init_sandbox
-            .call::<i32>("GuestMethod1", msg.to_string())
-            .unwrap_err();
-        assert!(matches!(res, HyperlightError::Error(msg) if msg == "Host function error!"));
+
+        for _ in 0..1000 {
+            let res = init_sandbox
+                .call::<i32>("GuestMethod1", msg.to_string())
```
jsturtevant, line 357: could we also add a test that demonstrates how a guest would successfully handle a host function and continue processing with this new type? Do we have that already?

src/hyperlight_host/src/func/guest_err.rs
jsturtevant: I was reviewing commit by commit, I saw this was moved.

src/hyperlight_guest_bin/src/guest_function/call.rs
ludfjig: correct

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
ludfjig: fixed

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
ludfjig: added

src/hyperlight_host/tests/sandbox_host_tests.rs
ludfjig: added

src/hyperlight_host/tests/sandbox_host_tests.rs
ludfjig: This was the case prior. Feedback appreciated how to make this a nicer C-API (for all types, not just i32):
```
#[unsafe(no_mangle)]
pub extern "C" fn hl_get_host_return_value_as_Int() -> i32 {
    get_host_return_value().expect("Unable to get host return value as int")
}

```

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -493,6 +510,46 @@ mod tests {
     use crate::sandbox::SandboxConfiguration;
     use crate::{GuestBinary, HyperlightError, MultiUseSandbox, Result, UninitializedSandbox};
 
+    /// Make sure input/output buffers are properly reset after guest call (with host call)
+    #[test]
+    #[ignore = "added this test before fixing bug"]
```
jsturtevant: should remove this?

ludfjig: whoops

src/hyperlight_host/tests/sandbox_host_tests.rs
jsturtevant: could we mimic something like wasmtime does? https://github.com/bytecodealliance/wasmtime/blob/7380932631f7784d944cb0326a6ffaaf5dac29fc/crates/c-api/src/component/val.rs#L180-L185

I don't think this needs to be a blocker as we are in this state today

jsturtevant (dismissed):

docs/how-to-use-flatbuffers.md
```diff
@@ -19,9 +19,3 @@ We recommend building `flatc` from source. To generate rust code, use
 ```console
 just gen-all-fbs-rust-code
 ```
-
```
danbugs, line 22: Can you update the "last version as of ..." date at the top of this file?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -493,6 +510,43 @@ mod tests {
     use crate::sandbox::SandboxConfiguration;
     use crate::{GuestBinary, HyperlightError, MultiUseSandbox, Result, UninitializedSandbox};
 
+    /// Make sure input/output buffers are properly reset after guest call (with host call)
+    #[test]
+    fn host_func_error() {
```
danbugs, line 515: nit: a better name might be `test_host_function_error_no_resource_leak`.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -493,6 +510,43 @@ mod tests {
...
+            let result = sandbox
+                .call::<i64>(
+                    "CallGivenParamlessHostFuncThatReturnsI64",
+                    "HostError".to_string(),
+                )
+                .unwrap_err();
+
+            assert!(
+                matches!(result, HyperlightError::GuestError(code, msg) if code == ErrorCode::HostFunctionError && msg == "hi"),
+            );
```
danbugs, line 536:
```suggestion
assert!(
    matches!(result, HyperlightError::GuestError(code, msg) 
        if code == ErrorCode::HostFunctionError && msg == "hi"),
    "Expected host error to be wrapped as GuestError::HostFunctionError, got {:?}", result
);
```

danbugs (commented):
Awesome work. Mostly LGTM, just got some minor nits.

docs/how-to-use-flatbuffers.md
ludfjig: done

adamperlin (dismissed):
This looks good to me! It's really great that this fixed an additional fragile error handling case as well with error deserialization. One question:  have you tried reverting the behavior of host call fuzzing back to how it was before we discovered this issue, i.e., re-using the same sandbox state for each fuzzed call? Maybe we can open a follow up issue to do this?

adamperlin:
> This looks good to me! It's really great that this fixed an additional fragile error handling case as well with error deserialization. One question: have you tried reverting the behavior of host call fuzzing back to how it was before we discovered this issue, i.e., re-using the same sandbox state for each fuzzed call? Maybe we can open a follow up issue to do this?

Added #933 to track this

src/hyperlight_host/tests/sandbox_host_tests.rs
simongdavies: +1 on creating an issue for this

simongdavies (approved):
LGTM , this is a great fix, just +1 on creating an issue for the TODO

PR #875, opened 2025-09-09
Update readme.md

simongdavies:
Update pre-requisites on Linux


ludfjig (approved):
LGTM. These requirements are also specified in .devcontainer/Dockerfile. Should we update them there too?

PR #879, opened 2025-09-09
Remove old code used for removed in-process feature

ludfjig:
These are all unused but seems we forgot to remove them when removing the inprocess feature

### DIFF
--- a/src/hyperlight_host/src/mem/layout.rs
+++ b/src/hyperlight_host/src/mem/layout.rs
@@ -402,9 +402,6 @@ impl SandboxMemoryLayout {
     }
 
     /// Get the offset in guest memory to the start of output data.
-    ///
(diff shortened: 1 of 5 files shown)


src/hyperlight_host/src/mem/layout.rs
```diff
@@ -424,7 +406,7 @@ impl SandboxMemoryLayout {
     /// This function exists to accommodate the macro that generates C API
```
simongdavies: Is this comment not true? the #[cfg(test)] seems to suggest not?

simongdavies (approved):
One minor comment

src/hyperlight_host/src/mem/layout.rs
ludfjig: correct. I will update the comment

PR #883, opened 2025-09-11
Update rust to 1.89

simongdavies:


flake.nix
```diff
@@ -64,6 +64,11 @@
               channel = "stable";
               sha256 = "sha256-X/4ZBHO3iW0fOenQ3foEvscgAPJYl2abspaBThDOukI=";
             };
+            "1.89" = {
+              date = "2025-04-03";
+              channel = "stable";
+              sha256 = "sha256-X/4ZBHO3iW0fOenQ3foEvscgAPJYl2abspaBThDOukI=";
```
ludfjig: this sha and date are the same as 1.86?

src/hyperlight_common/src/clippy.toml
```diff
@@ -2,6 +2,4 @@ disallowed-macros = [
     { path = "std::assert", reason = "no asserts in release builds" },
     { path = "std::assert_eq", reason = "no asserts in release builds" },
     { path = "std::assert_ne", reason = "no asserts in release builds" },
-    { path = "std::assert_true", reason = "no asserts in release builds" },
-    { path = "std::assert_false", reason = "no asserts in release builds" },
```
ludfjig, line 6: what happened to these

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -267,16 +267,15 @@ pub(crate) fn get_memory_access_violation<'a>(
...
-            || region.flags.contains(MemoryRegionFlags::STACK_GUARD)
-        {
-            return Some(HyperlightExit::AccessViolation(
-                gpa as u64,
-                access_info,
-                region.flags,
-            ));
-        }
+    if let Some(region) = region
+        && (!region.flags.contains(access_info)
```
ludfjig, line 271: love the new syntax that 1.89 finally allows!

trace/97b8e643-c45d-4d03-a33d-d19abbc0fae9.trace
ludfjig, line 1: did you mean to include this file?

Cargo.toml
```diff
@@ -29,7 +29,7 @@ exclude = [
 [workspace.package]
 version = "0.9.0"
 edition = "2024"
-rust-version = "1.86"
+rust-version = "1.89"
```
jprendes: Do we want to bump the minimum supported rust version? Is there anything we NEED from 1.89?

flake.nix
simongdavies: Will delete it , I was going to add the new data but I didnt do the work to figure out what the values should be

src/hyperlight_common/src/clippy.toml
simongdavies: Apparently they don't exist so they don't do anything, there is now a lint that warns about this

trace/97b8e643-c45d-4d03-a33d-d19abbc0fae9.trace
simongdavies: No, dont know where it came from, will add trace to gitignore

Cargo.toml
simongdavies: I don't believe we need anything, I can check

.devcontainer/Dockerfile
```diff
@@ -48,13 +48,15 @@ RUN SNIPPET="export PROMPT_COMMAND='history -a' && export HISTFILE=/commandhisto
...
+ARG RUST_TOOLCHAIN=1.89
 
 # Install rust
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y \
     && rustup default ${RUST_TOOLCHAIN} \
     && rustup target add x86_64-unknown-linux-gnu \
+    && rustup target add x86_64-unknown-linux-musl \
     && rustup target add x86_64-unknown-none      \
     && rustup toolchain add nightly-x86_64-unknown-linux-gnu \
+    && rustup toolchain add nightly-x86_64-unknown-linux-musl \
```
jsturtevant: Thanks, I guess I could also add cross-rs to my other PR to this docker file

src/hyperlight_host/src/hypervisor/windows_hypervisor_platform.rs
```diff
@@ -213,7 +213,7 @@ impl VMPartition {
 // with an error about a missing entrypoint
 // This function should always succeed since before we get here we have already checked that the hypervisor is present and
 // that we are on a supported version of windows.
-type WHvMapGpaRange2Func = unsafe extern "cdecl" fn(
+type WHvMapGpaRange2Func = unsafe extern "system" fn(
```
jsturtevant, line 216: this changes the ABI? was it wrong previously?

Cargo.toml
simongdavies: OK, yes we need to up this because the lint `#[allow(mismatched_lifetime_syntaxes)]` is new and we need that to make sure that the fb generated code does not trigger the lint. We could exclude all the fb generated code from linting but we would still need to increase msrv to 1.88 because of https://github.com/rust-lang/rust/issues/53667.

So excluding fb code from clippy would allow us to reduce this to 1.88?

.devcontainer/Dockerfile
simongdavies: Dont mind, this was one of those I added so I didnt forget later

src/hyperlight_host/src/hypervisor/windows_hypervisor_platform.rs
simongdavies: Yes , apparently so, new lint said it would be a hard error soon

jsturtevant: ok, Seems like tests are passing so the change didn't tangibly make a difference...

src/hyperlight_common/src/flatbuffers/mod.rs
```diff
@@ -2,6 +2,7 @@
 // @generated
 pub mod hyperlight {
     use super::*;
+    #[allow(mismatched_lifetime_syntaxes)]
```
ludfjig, line 5: nit: we have a bunch of other `allow` specified for the entire flatbuffer mod in src/hyperlight_common/src/lib.rs. Maybe consider moving this `allow` over there so it won't be removed next time we regenerate these.

.devcontainer/Dockerfile
jsturtevant: added cross-rs to my pr

src/hyperlight_common/src/flatbuffers/mod.rs
simongdavies: Fixed

Cargo.toml
simongdavies: OK I reduced it to 1.88 by amending the clippy rule for the flatbuffers generated code

PR #885, opened 2025-09-12
Fix flaky test test_drop

ludfjig:
This test was observed to be flaky, I believe due to race condition where other tests mmap allocate the same address (reused) after the memory is dropped, but before proocmaps are read. Fix by increasing memory size alot to prevent reuse of same address (hopefully). Also now runs on windows and no need to #[ignore]. If this doesn't fix it, we can try some other strategy or remove the test


ludfjig:
Closing because this didn't help... Let's consider removing this test if it is a problem in the future, because the test just tests a drop impl which i think we already are very confident are working

PR #895, opened 2025-09-18
Refactor evolve code to be simpler

ludfjig:
Evolving involves a transformation closure and whatnot, which just add complexity. It is not needed anymore since we only have 1 type of sandbox


jsturtevant (commented):
LGTM, but I don't have a ton on history on this file

jsturtevant (approved):

PR #903, opened 2025-09-22
Closes #274: Adding FFI function for types bool, float, String, vecbytes

### LINKED ISSUE #274: Missing `hl_get_host_return_value_as_*`
Missing bool, double, string, vec, etc..

vshailesh:
Added functions for #274


ludfjig:
Thanks for this. Would you mind adding a corresponding test for these types in simpleguest.c ( a guest function that calls a host function) and call them from either integration_test.rs or sandbox_host_test.rs or initialized_multi_use.rs?

vshailesh:
I will add tests

ludfjig:
Would you mind adding test for each of the new types, like you've done for bool? Also, a StackOverflow error is unexpected here, and I think you're forgetting to register a hostfunction on the sandbox before calling `evolve` on it. See here  for an example: https://github.com/ludfjig/hyperlight/blob/ae355d7e33a53c5d7e224edf2428e8251a10a62e/src/hyperlight_host/src/sandbox/initialized_multi_use.rs#L573

src/tests/c_guests/c_simpleguest/main.c
```diff
@@ -256,6 +256,24 @@ int guest_function(const char *from_host) {
   return 0;
 }
 
+
+bool host_returns_bool_value(const char *from_host) {
```
jsturtevant: not taking a string as parameter might make this simpler to test

src/tests/c_guests/c_simpleguest/main.c
```diff
@@ -325,6 +344,9 @@ void hyperlight_main(void)
     // HYPERLIGHT_REGISTER_FUNCTION macro does not work for functions that return VecBytes,
     // so we use hl_register_function_definition directly
     hl_register_function_definition("24K_in_8K_out", twenty_four_k_in_eight_k_out, 1, (hl_ParameterType[]){hl_ParameterType_VecBytes}, hl_ReturnType_VecBytes);
+    // HYPERLIGHT_REGISTER_FUNCTION macro does not work for functions that return VecBytes,
+    // so we use hl_register_function_definition directly
+    hl_register_function_definition("24K_in_8K_out", twenty_four_k_in_eight_k_out, 1, (hl_ParameterType[]){hl_ParameterType_VecBytes}, hl_ReturnType_VecBytes);
```
jsturtevant: looks like a copy paste of the function above?

vshailesh: I am still working on it, apologies, I was testing something for myself.
this is not final PR. I will tag someone when I have a final version

src/tests/c_guests/c_simpleguest/main.c
vshailesh: thank you I will look at doing it without string input.

Issue #905, opened 2025-09-23
KVM set/read the mxcsr value


jsturtevant:
How could we change the code to set/read the mxcsr value?

_Originally posted by @dblnz in https://github.com/hyperlight-dev/hyperlight/pull/817#discussion_r2297780516_

jsturtevant:
completed with https://github.com/hyperlight-dev/hyperlight/pull/1120

Issue #791, opened 2025-08-14
Ensure that system registers don't leak information between uses of a sandbox


syntactically:
Because most hypervisors treat setting up or destroying a partition as a heavy operation, reusing a Hyperlight sandbox for a different workload does not create a new partition. Traditionally, we tried to reset the state of the pre-existing partition with the evolve/devolve abstraction, which has now been replaced by snapshots.

Both approaches have focused mostly on ensuring the state of memory is reset to the expected new state, and have left resetting register state to the entry sequence for the next sandbox function that is called. On amd64, we happen to use calls that do in fact reset all general-purpose and floating point registers.  However, we do not reset all system registers/MSRs/etc.  Unfortunately, this means that  co-operating guests (or e.g. a compromised guest and a malicious guest) could pass data between uses of a sandbox by leaving it in otherwise-unused system registers across a snapshot restore. We should (a) insofar as this is possible, disable access to system registers that we are not using in the guest and (b) ensure that every register that is available to a guest is backed up during snapshotting and restored during restore.

jsturtevant:
progress on this was made in https://github.com/hyperlight-dev/hyperlight/pull/1120 and we have the remaining work in https://github.com/hyperlight-dev/hyperlight/pull/991

ludfjig:
Upcoming `HVCALL_SCRUB_PARTITION` should help with this (on hyperv)

simongdavies:
Wait for mshv crate support or add ourselves

syntactically:
If we are planning on scrub on mshv and msr filtering on KVM, do we have a plan of record for WHP?

ludfjig:
> If we are planning on scrub on mshv and msr filtering on KVM, do we have a plan of record for WHP?

Deleting the vcpu and recreating it should be possible (and fast) on WHP, and cover more than just MSRs. I'm not sure if there is state in the VM too that needs to be reset, would need to check

ludfjig:
there's also this internal api https://docs.rs/windows-sys/latest/windows_sys/Win32/System/Hypervisor/fn.WHvResetPartition.html, which is slower than delete + recreate vcpu

syntactically:
@ludfjig @jsturtevant we should be able to close this now that #991 is in?

PR #906, opened 2025-09-23
Guest Function reset tests for registers

### LINKED ISSUE #904: FPU Tag Word reset to wrong value
We currently use `pub(crate) const FP_TAG_WORD_DEFAULT: u8 = 0xff; // each 8 of x87 fpu registers is empty` which is incorrect (at least for kvm), which uses abridged format (FXSAVE, not FSAVE), so we should set this to 0x00 to indicate empty, instead of 0xff.

See:
https://docs.kernel.org/virt/kvm/api.html#kvm-get-fpu
https://www.felixcloutier.com/x86/fxsave

jsturtevant:
Fixes #904 #905

This pull request improves the correctness and test coverage of FPU and MXCSR register handling in the hypervisor and guest code. It fixes how the x87 FPU tag word is initialized, ensures MXCSR is properly reset for each guest call, and adds comprehensive tests (including integration and guest-side changes) to verify that FPU and MXCSR state is correctly reset and restored.

I did run benchmarks against using xsave on each call and had no noticeable different when using it.  There is enough noise in a single get call that any change wasn't tractable.

**FPU and MXCSR register initialization and handling:**

* Corrected the default value for `FP_TAG_WORD_DEFAULT` to `0x00` to match the FXSAVE format, where a two-bit value of 11 (encoded as 0) indicates an empty x87 FPU register.
* Updated the hypervisor's KVM driver to explicitly reset MXCSR using XSAVE/XRSTOR, since KVM's `set_fpu` does not handle MXCSR. This ensures the guest's MXCSR is properly initialized. [[1]](diffhunk://#diff-3b6dcdd84f2f74e34c585b325896c07d83437608e9b245549892c69da3040681L582-R584) [[2]](diffhunk://#diff-3b6dcdd84f2f74e34c585b325896c07d83437608e9b245549892c69da3040681R593-R611)

**Testing and validation:**

* Added new unit and integration tests to verify that FPU control word, tag word, and MXCSR registers are reset between guest calls and correctly restored after snapshot/restore operations. [[1]](diffhunk://#diff-ff2b13b76470ee68e295ce3277b29d2557ac2f403f696f5eec5adb12af3c27f6R1025-R1064) [[2]](diffhunk://#diff-0653ec70e87c8e9f58e66156f5d49924fe8ffe4501aafe6b99de943adb77363cR751-R783)
* Added guest-side functions (`ModifyFpuMxcsr`, `ReadFpuMxcsr`) and registered them, allowing host tests to manipulate and inspect FPU/MXCSR state inside the guest. [[1]](diffhunk://#diff-905dd6f4d540bd294c7d5b61266909e69bc0ba69c693939498db15e903b96023R933-R1037) [[2]](diffhunk://#diff-905dd6f4d540bd294c7d5b61266909e69bc0ba69c693939498db15e903b96023R1585-R1601)

### DIFF
--- a/typos.toml
+++ b/typos.toml
@@ -8,3 +8,4 @@ extend-exclude = ["**/*.patch", "src/hyperlight_guest_bin/third_party/**/*", "NO
 # typ is used for field name as type is a reserved keyword
 typ="typ"
 mmaped="mmapped"
+FPR="FPR"


src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -579,7 +579,9 @@ impl Hypervisor for KVMDriver {
         };
         self.vcpu_fd.set_regs(&regs)?;
 
-        // reset fpu state
+        // Note kvm set_fpu doesn't actually set or read the mxcsr value
+        // We reset it in a call below
+        // https://elixir.bootlin.com/linux/v6.16/source/arch/x86/kvm/x86.c#L12229
         let fpu = kvm_fpu {
             fcw: FP_CONTROL_WORD_DEFAULT,
             ftwx: FP_TAG_WORD_DEFAULT,
```
ludfjig, line 587: nit: Since it's 0 we can just remove it from here and let the ..Default::default() zero it. We can probably remove the constant completely in this case unless it's used for something else

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -588,6 +590,25 @@ impl Hypervisor for KVMDriver {
         };
         self.vcpu_fd.set_fpu(&fpu)?;
 
+        // Set MXCSR from XSAVE
+        // Locations are from
+        // AMD64 Architecture Programmer's Manual, Volume 2
+        // 11.5.10 Mode-Specific XSAVE/XRSTOR State Management
+        // MXCSR is at byte offset 24 -> u32 index 6
+        let mut xsave = match self.vcpu_fd.get_xsave() {
```
ludfjig, line 598: If we're doing xsave, can we omit the existing `self.vcpu_fd.set_fpu(&fpu)?;` call? Do they overlap in functionality?

Also: Maybe we should save xsave after calling initialize, and restore to that known good state after each to `dispatch_call_from_host`?

jsturtevant: > If we're doing xsave, can we omit the existing self.vcpu_fd.set_fpu(&fpu)?; call?

I considered this but felt the the xsave structure was pretty opaque with lots that could change. It felt clearer/safer to do this and there didn't seem to be a noticeable perf difference.

> Also: Maybe we should save xsave after calling initialize, and restore to that known good state after each to dispatch_call_from_host?

This is a good idea!  Let me take a look at this.

src/hyperlight_host/src/hypervisor/kvm.rs
jsturtevant: Thinking on this a bit and I think keeping it here as a const is a good idea since it allows us to be explicit about using the FXSAVE format (00 vs 11). It shouldn't have any impact but allows us to document that we are using the setting we explicitly want.

src/hyperlight_host/src/hypervisor/kvm.rs
jsturtevant: > Also: Maybe we should save xsave after calling initialize, and restore to that known good state after each to dispatch_call_from_host?
>
> This is a good idea! Let me take a look at this.

I tried this out by initializing the FP registers once and storing them.  I didn't notice any impact on performance (but it's hard to measure at the function call level due to so many variances).  Saving does incur an additional ~4k of memory per vm instance.  I lean towards leaving this approach and revisiting it if required.  what do you think?

ludfjig: I think that's fine. However, I think I am a little skeptical in general about using xsave just for the purpose of resetting just the single MXCSR register. There are plenty other registers we don't reset currently, that we probably should. Maybe let's take the opportunity to reset some more registers? What do you think? 

Also, regarding 
> If we're doing xsave, can we omit the existing self.vcpu_fd.set_fpu(&fpu)?; call?

>> I considered this but felt the the xsave structure was pretty opaque with lots that could change. It felt clearer/safer to do this and there didn't seem to be a noticeable perf difference.

I think xsave should be pretty standardized, at least per-hypervisor, so I do think it's worth looking into this

jsturtevant: > There are plenty other registers we don't reset currently, that we probably should. Maybe let's take the opportunity to reset some more registers? What do you think?

Yes, we have #791 and I started to look into other registers.  I was using this as a way to get a pattern set up without going after everything in one go.

> I think xsave should be pretty standardized, at least per-hypervisor, so I do think it's worth looking into this

xsave does have a specific layout depending on the features enabled. Maybe we could do this after https://github.com/hyperlight-dev/hyperlight/pull/907 and other related changes.

jsturtevant: spoke with @ludfjig offline and this will be integrated into the work with https://github.com/hyperlight-dev/hyperlight/pull/924 and  https://github.com/hyperlight-dev/hyperlight/issues/791

jsturtevant:
Spoke offline with @ludfjig and these changes are going to be picked up in

PR #907, opened 2025-09-23
Unify register representation across hypervisors

### LINKED ISSUE #312: Unify register representations across all Hypervisor implementations

### LINKED ISSUE #908: WHV_REGISTER_VALUE needs 16-alignment
... but has incorrect rust alignment of 8

ludfjig:
This PR is part of my bigger effort to refactor VM trait (PR 1/3). This PR introduces registers structs that abstract over differences between kvm, mshv, and whp, for general registers, special registers, and fpu registers. It also fixes some UB where whp requires register values to be 16-aligned. 

The introduced abstractions are not too useful when just looking at this PR in isolation, but future PR will refactor and deduplicate a bunch of code, where this will be needed.

Closes #908
Closes #312

### DIFF
--- a/src/hyperlight_host/src/hypervisor/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/mod.rs
@@ -19,15 +19,14 @@ use tracing::{Span, instrument};
 
 use crate::HyperlightError::StackOverflow;
 use crate::error::HyperlightError::ExecutionCanceledByHost;
+use crate::hypervisor::regs::{
+    CommonFpu, CommonRegisters, CommonSegmentRegister, CommonSpecialRegisters,
+};
 use crate::mem::memory_region::{MemoryRegion, MemoryRegionFlags};
 use crate::metrics::METRIC_GUEST_CANCELLATION;
 #[cfg(feature = "trace_guest")]
 use crate::sandbox::TraceInfo;
 use crate::{HyperlightError, Result, log_then_return};
-
-/// Util for handling x87 fpu state
-#[cfg(any(kvm, mshv, target_os = "windows"))]
-pub mod fpu;
 
 /// HyperV-on-linux functionality
 #[cfg(mshv)]
@@ -39,6 +38,9 @@ pub(crate) mod hyperv_windows;
 /// GDB debugging support
 #[cfg(gdb)]
 pub(crate) mod gdb;
+
+/// Abstracts over different hypervisor register representations
+pub(crate) mod regs;
 
 #[cfg(kvm)]
 /// Functionality to manipulate KVM-based virtual machines
@@ -116,21 +118,6 @@ pub enum HyperlightExit {
     Retry(),
 }
 
-/// Registers which may be useful for tracing/stack unwinding
-#[cfg(feature = "trace_guest")]
-pub enum TraceRegister {
-    /// RAX
-    RAX,
-    /// RCX
-    RCX,
-    /// RIP
-    RIP,
-    /// RSP
-    RSP,
-    /// RBP
-    RBP,
-}
-
 /// A common set of hypervisor functionality
 pub(crate) trait Hypervisor: Debug + Send {
     /// Initialise the internally stored vCPU with the given PEB address and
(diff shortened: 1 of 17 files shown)


src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -189,6 +176,87 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+    fn sregs(&self) -> Result<CommonSpecialRegisters>;
+    /// Set special regs
+    #[allow(dead_code)]
+    fn set_sregs(&mut self, sregs: &CommonSpecialRegisters) -> Result<()>;
+
+    /// Setup initial special registers for the hypervisor
+    /// This is a default implementation that works for all hypervisors
+    fn setup_initial_sregs(&mut self, _pml4_addr: u64) -> Result<()> {
+        #[cfg(feature = "init-paging")]
+        let sregs = CommonSpecialRegisters {
```
jsturtevant, line 202: Having registers in one location seems like a create improvement as we move forward with adding new features.

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -189,6 +176,87 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+    #[allow(dead_code)]
+    fn fpu(&self) -> Result<CommonFpu>;
+    /// Set fpu regs
+    #[allow(dead_code)]
+    fn set_fpu(&mut self, fpu: &CommonFpu) -> Result<()>;
+    /// Get special regs
+    #[allow(dead_code)]
+    fn sregs(&self) -> Result<CommonSpecialRegisters>;
+    /// Set special regs
+    #[allow(dead_code)]
```
dblnz, line 195: Are this `allow(dead_code)` really needed?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -40,6 +39,9 @@ pub(crate) mod hyperv_windows;
 #[cfg(gdb)]
 pub(crate) mod gdb;
 
+/// Abstracts over different hypervisor register representations
+pub(crate) mod regs;
```
dblnz, line 43: Nit: I think it would be more appropriate to have an `arch` module that abstracts over registers and `architecture` specific logic.
This can be done later anyway.

src/hyperlight_host/src/hypervisor/regs/fpu.rs
```diff
@@ -0,0 +1,466 @@
+/*
+Copyright 2024 The Hyperlight Authors.
```
dblnz, line 2: I am not sure what the guidance is here, but shouldn't we use `2025` if we added the file this year? I'm asking for myself also, to know how I do it.

src/hyperlight_host/src/sandbox/outb.rs
dblnz, line 1: I had done something similar but only for this tracing scope.
The unification you've done is more comprehensive

dblnz (approved):
Great work, @ludfjig!
This makes the code a lot easier to read 😄

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: yes. In next PR I will consolidate everything

src/hyperlight_host/src/hypervisor/regs/fpu.rs
ludfjig: I'm not sure tbh :D

src/hyperlight_host/src/hypervisor/regs/standard_regs.rs
```diff
@@ -0,0 +1,420 @@
...
+    WHvX64RegisterR10,
+    WHvX64RegisterR11,
+    WHvX64RegisterR12,
+    WHvX64RegisterR13,
+    WHvX64RegisterR14,
+    WHvX64RegisterR15,
+    WHvX64RegisterRip,
+    WHvX64RegisterRflags,
+];
+
```
danbugs, line 275: Nit: might be good to add a compile time check that the length of the array matches WHP_REGS_NAMES_LEN?

```
#[cfg(target_os = "windows")]
const _: () = assert!(WHP_REGS_NAMES.len() == WHP_REGS_NAMES_LEN);
```

danbugs (approved):
LGTM! Awesome work!!

src/hyperlight_host/src/hypervisor/regs/standard_regs.rs
ludfjig: That should already be type-checked in the type declaration `pub(crate) const WHP_REGS_NAMES: [WHV_REGISTER_NAME; WHP_REGS_NAMES_LEN] = ....`

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Agree, will fix it up in the future to make it easier for arm64

src/hyperlight_host/src/hypervisor/regs/fpu.rs
simongdavies: For new files , I think definitely, I think i we update files we should probably update it as well?

simongdavies: Co-pilot seems to agree 😄

PR #924, opened 2025-09-30
Introduce new Vm trait

ludfjig:
This is PR 2/3 in a bigger effort to remove duplicate code accross drivers.

This PR introduces

- `Vm` trait. It's a minimal trait for common functionality of a minimal Vm. It abstracts over differences in kvm, mshv, whp. This traits only knows things like set/get registers, run, but nothing about guest functions or hyperlight specifics.
- `HyperlightVm` struct.  This is a struct that contains the `dyn Vm` above, as well as things like guest_ptr, rsp, memory-regions, gdb connections, etc. You can think of this as replacing the previous Hypervisor trait (but now it's just 1 struct to avoid duplicate code).


src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -14,626 +14,414 @@ See the License for the specific language governing permissions and
...
 
-        let is_canceled = exit_context.ExitReason == WHV_RUN_VP_EXIT_REASON(8193i32); // WHvRunVpExitReasonCanceled
+        Ok(())
+    }
 
-        // Check if this was a manual cancellation (vs internal Windows cancellation)
-        let cancel_was_requested_manually =
-            interrupt_handle_internal.is_cancel_requested_for_generation(generation);
+    fn unmap_memory(&mut self, (_slot, _region): (u32, &MemoryRegion)) -> Result<()> {
+        log_then_return!("Mapping host memory into the guest not yet supported on this platform");
```
jsturtevant, line 205: It looks implemented above?

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -14,626 +14,414 @@ See the License for the specific language governing permissions and
...
+            return Err(new_error!(
+                "Failed to get Xsave state: expected {} bytes, got {}",
+                buffer_size_needed,
+                written_bytes
+            ));
         }
 
-        let mut regs = self.regs()?;
-        regs.rip = rip + instruction_length;
-        self.set_regs(&regs)
```
jsturtevant, line 518: I don't see this rip set in the new implementation in `hyperlight_vm`

src/hyperlight_host/src/hypervisor/vm.rs
```diff
@@ -0,0 +1,123 @@
...
+
+    /// Unmap memory region from this VM that has previously been mapped using `map_memory`.
+    fn unmap_memory(&mut self, region: (u32, &MemoryRegion)) -> Result<()>;
+
+    /// Runs the vCPU until it exits
+    fn run_vcpu(&mut self) -> Result<VmExit>;
+
+    // --------------------------
+    // --- DEBUGGING BELOW ------
+    // --------------------------
```
jsturtevant: I think I preferd this being another trait that was implemented as it was before.  Whats the motivation for bringing this in here?

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -287,167 +274,6 @@ pub(crate) enum DebugResponse {
     WriteRegisters,
 }
```
jsturtevant, line 276: couldn't we keep some of this trait here instead of moving it up into the new vm trait?

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -51,61 +52,52 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
-        sw_breakpoints: {:?}
         ",
-        single_step,
         rip,
         dr6,
         entrypoint,
         exception,
-        hw_breakpoints,
-        sw_breakpoints,
     );
```
jsturtevant, line 99: It's unclear why these have been removed from the error log

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: That's just the initial mapping. Subsequent mappings after creating the initial sandbox is not yet supported on windows

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: Since not all drivers require incrementing RIP, it's an implementation detail of each driver to handle it as part of their `run_vcpu`. If you check hyperv_linux/windows.rs, you'll see that it's still done

src/hyperlight_host/src/hypervisor/vm.rs
ludfjig: Generally I thought it made sense for them to belong on the "Vm" and to be able to get rid of the vcpu abstraction, and a lot of these are not needed anymore. But I agree I think maybe it makes sense.... I can implement all the debugging functionality  separately on a supertrait of `Vm` trait, that should be a bit cleaner. In additon, some code such as getting/setting debug registers, etc, will be needed in the future for purposes of resetting vcpu, even when debugging is not enabled.

src/hyperlight_host/src/hypervisor/gdb/mod.rs
ludfjig: We could, I could make it a supertrait of vm

src/hyperlight_host/src/hypervisor/gdb/arch.rs
ludfjig: They are not necessary to determine stop reason so I removed them in order not having to keep unnecessary state around and simplify the code

src/hyperlight_host/src/hypervisor/vm.rs
jsturtevant: Since these are all hidden behind the flag, you could keep `GuestDebug` (modified to what you have here) in the `gdb` folder and `impl GuestDebug for Hyperlight_vm` when the flag is enabled I believe

src/hyperlight_host/src/hypervisor/vm.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// The operation should be retried, for example this can happen on Linux where a call to run the CPU can return EAGAIN
+    #[cfg_attr(
+        target_os = "windows",
+        expect(
+            dead_code,
+            reason = "Retry() is never constructed on Windows, but it is still matched on (which dead_code lint ignores)"
+        )
+    )]
+    Retry(),
+    #[cfg(gdb)]
```
dblnz: Nit: Invert the comment and `#[cfg..`
```suggestion
    /// The vCPU has exited due to a debug event (usually breakpoint)
    #[cfg(gdb)]
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1119 @@
...
+use crate::sandbox::host_funcs::FunctionRegistry;
+use crate::sandbox::hypervisor::{HypervisorType, get_available_hypervisor};
+use crate::sandbox::outb::handle_outb;
+#[cfg(feature = "mem_profile")]
+use crate::sandbox::trace::MemTraceInfo;
+#[cfg(crashdump)]
+use crate::sandbox::uninitialized::SandboxRuntimeConfig;
+use crate::{HyperlightError, Result, log_then_return, new_error};
+
+pub(crate) struct HyperlightVm {
```
dblnz, line 74: Some comments here explaining what this struct wants to achieve and why we need it would be helpful.
I assume this replaces the old Hypervisor

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -51,61 +51,52 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
+) -> Result<VcpuStopReason> {
+    let CommonRegisters { rip, .. } = vm.regs()?;
     if DB_EX_ID == exception {
         // If the BS flag in DR6 register is set, it means a single step
         // instruction triggered the exit
         // Check page 19-4 Vol. 3B of Intel 64 and IA-32
         // Architectures Software Developer's Manual
-        if dr6 & DR6_BS_FLAG_MASK != 0 && single_step {
-            return VcpuStopReason::DoneStep;
+        if dr6 & DR6_BS_FLAG_MASK != 0 {
```
dblnz, line 66: I think it is fine to remove the `single_step` variable here. It was only an additional state for verifying the correct flag was in sync with internal state of debugging

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -51,61 +51,52 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
+                return Ok(VcpuStopReason::EntryPointBp);
             }
-            return VcpuStopReason::HwBp;
+            return Ok(VcpuStopReason::HwBp);
         }
     }
 
-    if BP_EX_ID == exception && sw_breakpoints.contains_key(&rip) {
-        return VcpuStopReason::SwBp;
+    if BP_EX_ID == exception {
```
dblnz, line 83: I think this might remove an option here.

If the vCPU stops because of an issue, not a SW/HW breakpoint set by the debugger, it will not be transmitted as an `unknown` breakpoint.

I am not sure though if this means that the debugger won't show it as an `exception`.

src/hyperlight_host/src/hypervisor/vm.rs
ludfjig: done

src/hyperlight_host/src/hypervisor/gdb/mod.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/gdb/arch.rs
ludfjig: I think #BP is only ever raised by INT3, so checking sw breakpoints is redundant?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1119 @@
...
+        #[cfg(gdb)]
+        let gdb_conn = if let Some(gdb_conn) = gdb_conn {
+            // Add breakpoint to the entry point address
+            vm.set_debug(true)?;
+            vm.add_hw_breakpoint(entrypoint)?;
+
+            Some(gdb_conn)
+        } else {
+            None
+        };
```
jprendes:
```suggestion
        if gdb_conn.is_some() {
            // Add breakpoint to the entry point address
            vm.set_debug(true)?;
            vm.add_hw_breakpoint(entrypoint)?;
        }
```

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -73,990 +51,208 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
-            region.host_region.end,
-        ]
-        .iter()
-        .any(|x| x % self.page_size != 0)
-        {
-            log_then_return!(
-                "region is not page-aligned {:x}, {region:?}",
-                self.page_size
-            );
-        }
```
jprendes, line 480: Are we getting rid of this check?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1119 @@
...
+            self.freed_slots.push(slot);
+            self.vm.unmap_memory((slot, region))?;
+        } else {
+            return Err(new_error!("Region not found in mapped regions"));
+        }
+
+        Ok(())
+    }
+
+    pub(crate) fn get_mapped_regions(&self) -> &[(u32, MemoryRegion)] {
```
jprendes: do we want to give access to the slot? it's an implementation detail, right?

jprendes:
```suggestion
    pub(crate) fn get_mapped_regions(&self) -> impl Iterator<Item = &MemoryRegion> {
```

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -611,10 +594,6 @@ impl MultiUseSandbox {
         if self.poisoned {
             return Err(crate::HyperlightError::PoisonedSandbox);
         }
-        // Mark that a guest function call is now active
-        // (This also increments the generation counter internally)
-        // The guard will automatically clear call_active when dropped
-        let _guard = CallActiveGuard::new(self.vm.interrupt_handle())?;
```
jsturtevant: Where did this go to?  Is call_active still being used?  What are the other changes related to this?

ludfjig: Nope it's all gone. I was not a big fan of the original logic but we still merged it in favor of velocity. All logic related to cancellation is now in `HyperlightVm::run`. It's been greatly simplified

ludfjig: This PR no longer changes cancellation logic

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -618,443 +146,191 @@ impl Hypervisor for HypervLinuxDriver {
...
+        ]
+        .contains(&addr)
+        {
+            return Ok(());
         }
-    }
-}
 
-#[cfg(test)]
-mod tests {
```
ludfjig, line 984: Indeed these tests are removed (on purpose). Turns out it does not even run the CODE below... I am planning on re-adding some unit tests in next PR

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -82,23 +78,27 @@ use crate::sandbox::host_funcs::FunctionRegistry;
...
     Unknown(String),
-    /// The operation should be retried
-    /// On Linux this can happen where a call to run the CPU can return EAGAIN
-    /// On Windows the platform could cause a cancelation of the VM run
+    /// The operation should be retried, for example this can happen on Linux where a call to run the CPU can return EAGAIN
+    #[cfg_attr(
+        target_os = "windows",
+        expect(
+            dead_code,
+            reason = "Retry() is never constructed on Windows, but it is still matched on (which dead_code lint ignores)"
```
jsturtevant, line 85: Are we no longer seeing the retries on the windows side?

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -51,61 +51,52 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
 
 /// Determine the reason the vCPU stopped
 /// This is done by checking the DR6 register and the exception id
-/// NOTE: Additional checks are done for the entrypoint, stored hw_breakpoints
-/// and sw_breakpoints to ensure the stop reason is valid with internal state
```
jsturtevant, line 55: Why is this no longer done?

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -51,61 +51,52 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
         // If any of the B0-B3 flags in DR6 register is set, it means a
         // hardware breakpoint triggered the exit
         // Check page 19-4 Vol. 3B of Intel 64 and IA-32
         // Architectures Software Developer's Manual
-        if DR6_HW_BP_FLAGS_MASK & dr6 != 0 && hw_breakpoints.contains(&rip) {
+        if DR6_HW_BP_FLAGS_MASK & dr6 != 0 {
             if rip == entrypoint {
-                return VcpuStopReason::EntryPointBp;
+                vm.remove_hw_breakpoint(entrypoint)?;
+                return Ok(VcpuStopReason::EntryPointBp);
```
jsturtevant, line 77: this not clear why this remove_hw_breakpoint was added

src/hyperlight_host/src/hypervisor/mod.rs
jsturtevant: looks we still handle it in some way

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -618,97 +111,65 @@ impl Hypervisor for HypervLinuxDriver {
...
                 INVALID_GPA_ACCESS_MESSAGE => {
                     let mimo_message = m.to_memory_info().map_err(mshv_ioctls::MshvError::from)?;
                     let gpa = mimo_message.guest_physical_address;
                     let access_info = MemoryRegionFlags::try_from(mimo_message)?;
-                    crate::debug!(
-                        "mshv MMIO invalid GPA access -Details: Address: {} \n {:#?}",
-                        gpa,
-                        &self
-                    );
-                    match get_memory_access_violation(
```
jsturtevant, line 665: Is this check still happening?

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Windows still retries (handled by hyperlight_vm) if wrongly cancelled. But we don't construct a Retry variant in this case because it's a Cancelled variant for a run that wasn't supposed to be cancelled

src/hyperlight_host/src/hypervisor/gdb/arch.rs
ludfjig: We no longer keep track of this state, as it's already being tracked by the vm itself. Tracking this just risks us being out-of-sync with the actual state. It's just simpler this way

src/hyperlight_host/src/hypervisor/gdb/arch.rs
ludfjig: Good question. It used to be removed after the function call. But it's now removed in it instead. https://github.com/hyperlight-dev/hyperlight/blob/485f8e54967f65cfe3e6244de8bd3d70417628cb/src/hyperlight_host/src/hypervisor/gdb/kvm_debug.rs#L121-L138

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Yes here https://github.com/ludfjig/hyperlight/blob/09fdc1c550396181dd6e548bd026db90c3849ef0/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L442-L501

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -107,7 +94,7 @@ impl DebugMemoryAccess {
     ///   This address is shall be translated before calling this function
```
simongdavies, line 94: Not from this PR I know but this does not make sense:

`/// This address is shall be translated before calling this function`

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -172,7 +159,7 @@ impl DebugMemoryAccess {
...
```
simongdavies, line 159: Same here:

`/// This address is shall be translated before calling this function`

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -14,551 +14,242 @@ See the License for the specific language governing permissions and
...
-        let mut partition = VMPartition::new(1)?;
-
-        // get a surrogate process with preallocated memory of size SharedMemory::raw_mem_size()
-        // with guard pages setup
-        let surrogate_process = {
-            let mgr = get_surrogate_process_manager()?;
-            mgr.get_surrogate_process(raw_size, mmap_file_handle)
-        }?;
-
-        partition.map_gpa_range(&mem_regions, &surrogate_process)?;
```
simongdavies, line 75: we seem to have lost the comment about why we need to and why its ok to impl Send

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -14,551 +14,242 @@ See the License for the specific language governing permissions and
...
-        #[cfg(gdb)]
-        let (debug, gdb_conn) = if let Some(gdb_conn) = gdb_conn {
-            let mut debug = HypervDebug::new();
-            debug.add_hw_breakpoint(&proc, entrypoint)?;
-
-            (Some(debug), Some(gdb_conn))
-        } else {
-            (None, None)
+unsafe impl Send for WhpVm {}
+unsafe impl Sync for WhpVm {}
```
simongdavies: Why are we impl Sync for WhpVm?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -73,606 +51,91 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
-            unsafe { vm_fd.set_user_memory_region(kvm_region) }
-        })?;
+static KVM: LazyLock<Result<Kvm>> =
+    LazyLock::new(|| Kvm::new().map_err(|e| new_error!("Failed to open /dev/kvm: {}", e)));
 
+impl KvmVm {
+    /// Create a new instance of a `KvmVm`
+    #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
+    pub(crate) fn new() -> Result<Self> {
+        let hv = KVM
```
simongdavies, line 72: what is the rationale behind the static KVM reference?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -73,606 +51,91 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
-                    self.sandbox_regions
-                        .iter()
-                        .chain(self.mmap_regions.iter().map(|(r, _)| r)),
-                    MemoryRegionFlags::WRITE,
-                ) {
-                    Some(access_violation_exit) => access_violation_exit,
-                    None => HyperlightExit::Mmio(addr),
-                }
-            }
+    fn run_vcpu(&mut self) -> Result<HyperlightExit> {
```
simongdavies, line 106: IS there a reason to remove the debug! output? Also is it intentional that the  trace context has gone?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -42,421 +35,143 @@ pub(crate) mod regs;
...
-        mem_mgr: &mut SandboxMemoryManager<HostSharedMemory>,
-        host_funcs: &Arc<Mutex<FunctionRegistry>>,
-        guest_max_log_level: Option<LevelFilter>,
-        #[cfg(gdb)] dbg_mem_access_fn: Arc<Mutex<SandboxMemoryManager<HostSharedMemory>>>,
-    ) -> Result<()>;
-
-    /// Map a region of host memory into the sandbox.
+/// Trait for single-vCPU VMs. Provides a common interface for basic VM operations.
+/// Abstracts over differences between KVM, MSHV and WHP implementations.
+pub(crate) trait Hypervisor: Send + Sync + Debug {
```
simongdavies: Why is the trait Sync? We don't want hypervisor instances shared across threads? The old hypervisor trait was only send?

simongdavies (commented):
A few comments from a quickish review

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: good catch I will add this back

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: Whoops yes we don't need to I believe

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: Yes I moved the trace context since it's not specific to a particular vm implementation and should live in hyperlight_vm instead. I removed the debug!s in run, because they seem pointless to me since we're returning the same thing we're debug!ing. I think we have an issue for revisiting all logging/tracing, so we might add them back if we think they're useful

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Whoops yes we don't need Sync I believe

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: We had a discussion a **long** time ago about not calling Kvm::new() on each new sandbox. I mentioned that I would add it to (a previous version of) this PR :D

src/hyperlight_host/src/hypervisor/gdb/mod.rs
ludfjig: I think it's just pointing out that it should be a gpa (as opposed to gva). I made an issue relating to this previously https://github.com/hyperlight-dev/hyperlight/issues/1046

src/hyperlight_host/src/hypervisor/gdb/mod.rs
ludfjig: I think it's just pointing out that it should be a gpa (as opposed to gva). I made an issue relating to this previously https://github.com/hyperlight-dev/hyperlight/issues/1046

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -462,40 +462,6 @@ impl Hypervisor for KVMDriver {
...
-        dispatch_func_addr: RawPtr,
-        mem_mgr: &mut SandboxMemoryManager<HostSharedMemory>,
-        host_funcs: &Arc<Mutex<FunctionRegistry>>,
-        #[cfg(gdb)] dbg_mem_access_fn: Arc<Mutex<SandboxMemoryManager<HostSharedMemory>>>,
-    ) -> Result<()> {
-        // Reset general purpose registers, then set RIP and RSP
-        let regs = CommonRegisters {
-            rip: dispatch_func_addr.into(),
-            rsp: self.orig_rsp.absolute()?,
-            ..Default::default()
```
jsturtevant, line 536: noticed here that the flags where not being reset as they were in the other hypervisors?  Is the new behavor correct here?

src/hyperlight_host/src/hypervisor/kvm.rs
simongdavies: Hmmm, I am not sure that these 2 things are equivalent , returning the values does not allow you to see them in a trace or log, i.e. if I turn on logging or tracing at debug level I don't see these values unless I have debug! macros in the code. I think that these were there precisely because they were used to diagnose an issue at some point in the past, I do agree that we should review what/where/level of trace/logs though.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -438,40 +438,18 @@ impl Debug for HypervLinuxDriver {
...
-        if [
-            rgn.guest_region.start,
-            rgn.guest_region.end,
-            rgn.host_region.start,
-            rgn.host_region.end,
-        ]
-        .iter()
-        .any(|x| x % self.page_size != 0)
-        {
-            log_then_return!("region is not page-aligned");
```
jsturtevant, line 514: did we loose this check?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -451,42 +451,7 @@ impl Hypervisor for HypervLinuxDriver {
...
-            let trace_info = self.trace_info_mut();
-            handle_outb(mem_mgr, host_funcs, port, val, &regs, trace_info)?;
-        }
-        #[cfg(not(feature = "mem_profile"))]
-        {
-            handle_outb(mem_mgr, host_funcs, port, val)?;
-        }
-
-        // update rip
-        self.vcpu_fd.set_reg(&[hv_register_assoc {
```
jsturtevant, line 598: This wasn't moved? Also see that KVM didn't do this?

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
```diff
@@ -168,73 +165,33 @@ pub(crate) fn set_up_hypervisor_partition(
...
-            Box::new(hv)
-        }
-
+    HyperlightVm::new(
+        regions,
+        pml4_ptr.absolute()?,
+        entrypoint_ptr.absolute()?,
+        rsp_ptr.absolute()?,
+        config,
         #[cfg(target_os = "windows")]
```
jsturtevant, line 174: should this still be there?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2025 The Hyperlight Authors.
+Copyright 2024 The Hyperlight Authors.
```
jsturtevant: don't update this

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -52,8 +91,124 @@ pub(crate) struct HyperlightVm {
...
+        mem_regions: Vec<MemoryRegion>,
+        _pml4_addr: u64,
+        entrypoint: u64,
+        rsp: u64,
+        #[cfg_attr(not(any(kvm, mshv3)), allow(unused_variables))] config: &SandboxConfiguration,
+        #[cfg(target_os = "windows")] handle: HandleWrapper,
+        #[cfg(target_os = "windows")] raw_size: usize,
+        #[cfg(gdb)] gdb_conn: Option<DebugCommChannel<DebugResponse, DebugMsg>>,
+        #[cfg(crashdump)] rt_cfg: SandboxRuntimeConfig,
+        #[cfg(feature = "mem_profile")] trace_info: MemTraceInfo,
```
jsturtevant, line 113: I am not sure if I am a huge fan of this but there could be some follow up to clean up more

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: It's a reserved flag in x86, and it should always be set. Different hypervisor had different defaults and different requirements on setting it. It should always be set.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: It was moved, see "Implement run_vcpu" commit. Indeed kvm does not do it

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Are you referring that all these are conditionally compiled / feature-gated? I agree it's not ideal

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
ludfjig: The `HandleWrapper`? Yes

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -714,7 +715,7 @@ impl MultiUseSandbox {
     #[cfg(crashdump)]
     #[instrument(err(Debug), skip_all, parent = Span::current())]
     pub fn generate_crashdump(&self) -> Result<()> {
-        crate::hypervisor::crashdump::generate_crashdump(self.vm.as_ref())
+        crate::hypervisor::crashdump::generate_crashdump(self.vm.vm.as_ref() as &dyn Hypervisor)
```
simongdavies: This makes me feel like we got something wrong with the naming here (vm.vm.) seems wrong, maybe we can rename one of those members? or add an accessor method to the struct and make the member private?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -26,9 +26,29 @@ use crate::sandbox::host_funcs::FunctionRegistry;
 use log::LevelFilter;
 use std::sync::{Arc, Mutex};
 
-#[derive(Debug)]
 pub(crate) struct HyperlightVm {
-    pub(crate) vm: Box<dyn Hypervisor>,
+    #[cfg(gdb)]
+    vm: Box<dyn DebuggableVm>,
+    #[cfg(not(gdb))]
+    vm: Box<dyn Hypervisor>,
```
simongdavies, line 78: Related to comment about vm.vm this looks strange that we have member called vm that has a type Hypervisor?

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -263,29 +263,15 @@ mod debug {
...
 
-    sandbox_regions: Vec<MemoryRegion>, // Initially mapped regions when sandbox is created
-    mmap_regions: Vec<(MemoryRegion, u32)>, // Later mapped regions (region, slot number)
-    next_slot: u32,                     // Monotonically increasing slot number
-    freed_slots: Vec<u32>,              // Reusable slots from unmapped regions
-
-    #[cfg(gdb)]
-    debug: Option<KvmDebug>,
+    // KVM as opposed to mshv/whp has no way to get current debug regs, so need to keep a copy here
     #[cfg(gdb)]
```
simongdavies, line 61: IF this is kvm only shouldn't it be dependent on the KVM feature too?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -116,4 +116,236 @@ impl HyperlightVm {
...
+                //    - Will still do a VM entry, but signals will be sent until VM exits
+                let result = self.vm.run_vcpu();
+
+                // End current host trace by closing the current span that captures traces
+                // happening when a guest exits and re-enters.
+                #[cfg(feature = "trace_guest")]
+                tc.end_host_trace();
+
+                // Handle the guest trace data if any
+                #[cfg(feature = "trace_guest")]
```
simongdavies: cant these be combined since the second one is a block?

src/hyperlight_host/src/hypervisor/kvm.rs
jsturtevant: That's what I thought, This is good example of how this helps fix those types of bugs

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
jsturtevant: Oh, I see it now, it is a bit hard to follow the diff.  This is an example of the conditional parameters that makes it hard to read that i mentioned in the other comment.

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -561,51 +553,44 @@ impl Hypervisor for HypervWindowsDriver {
...
-                    exit_context.ExitReason, &self
-                );
-                match self.get_exit_details(exit_context.ExitReason) {
-                    Ok(error) => HyperlightExit::Unknown(error),
-                    Err(e) => HyperlightExit::Unknown(format!("Error getting exit details: {}", e)),
-                }
-            }
+            WHV_RUN_VP_EXIT_REASON(_) => HyperlightExit::Unknown(format!(
+                "Unknown exit reason '{}'",
+                exit_context.ExitReason.0
```
simongdavies, line 284: Why are we directly accessing a member here vs calling a function previously, what changed?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
jsturtevant: I see it now, thanks

src/hyperlight_host/src/hypervisor/kvm.rs
```diff
@@ -579,78 +579,31 @@ impl Hypervisor for KVMDriver {
...
-                libc::EINTR => HyperlightExit::Cancelled(),
-                libc::EAGAIN => HyperlightExit::Retry(),
-                _ => {
-                    crate::debug!("KVM Error -Details: Address: {} \n {:#?}", e, &self);
-                    log_then_return!("Error running VCPU {:?}", e);
-                }
+                // InterruptHandle::kill() sends a signal (SIGRTMIN+offset) to interrupt the vcpu, which causes EINTR
+                libc::EINTR => Ok(HyperlightExit::Cancelled()),
+                libc::EAGAIN => Ok(HyperlightExit::Retry()),
+                _ => Ok(HyperlightExit::Unknown(format!(
```
simongdavies, line 121: Now this has changed to return an OK(unknown) when we get an unexpected error (which looks a bit strange) does this function need to return a Result since it looks like all returns are OK().

Seems like the other 2 implementations have limited cases where they return Err()

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -158,11 +161,9 @@ pub(crate) trait Hypervisor: Debug + Send {
         host_funcs: &Arc<Mutex<FunctionRegistry>>,
     ) -> Result<()>;
 
-    /// Run the vCPU
-    fn run(
-        &mut self,
-        #[cfg(feature = "trace_guest")] tc: &mut crate::sandbox::trace::TraceContext,
-    ) -> Result<HyperlightExit>;
+    /// Runs the vCPU until it exits.
+    /// Note: this function should not emit any traces or spans as it is called after guest span is setup
```
simongdavies, line 108: This comment suggests to me that the guest span is being setup too early.

src/hyperlight_host/src/hypervisor/kvm.rs
simongdavies: I seem to recall we had a bug that was caused by EFLAGs being handled differently (reset or not reset depending on hypervisor) I am guessing this now eradicates that/thos sort of issues (seem to think [this](https://github.com/hyperlight-dev/hyperlight/blob/f4d80bb39989273eb4f347f662be191d9960a996/src/hyperlight_host/src/hypervisor/hyperv_windows.rs#L428) might be it)

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -102,18 +102,42 @@ impl HyperlightVm {
...
+    pub(crate) fn unmap_region(&mut self, region: &MemoryRegion) -> Result<()> {
+        if let Some(pos) = self.mmap_regions.iter().position(|(_, r)| r == region) {
+            let (slot, _) = self.mmap_regions.remove(pos);
+            self.freed_slots.push(slot);
+            self.vm.unmap_memory((slot, region))?;
+        } else {
+            return Err(new_error!("Region not found in mapped regions"));
+        }
+
+        Ok(())
```
simongdavies, line 303:
```suggestion
pub(crate) fn unmap_region(&mut self, region: &MemoryRegion) -> Result<()> {
    let pos = self
        .mmap_regions
        .iter()
        .position(|(_, r)| r == region)
        .ok_or_else(|| new_error!("Region not found in mapped regions"))?;

    let (slot, _) = self.mmap_regions.remove(pos);
    self.freed_slots.push(slot);
    self.vm.unmap_memory((slot, region))?;

    Ok(())
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -52,8 +91,124 @@ pub(crate) struct HyperlightVm {
...
-        Self { vm: inner }
+    /// Create a new HyperlightVm instance (will not run vm until calling `initialise`)
+    #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
+    #[allow(clippy::too_many_arguments)]
+    pub(crate) fn new(
+        mem_regions: Vec<MemoryRegion>,
+        _pml4_addr: u64,
+        entrypoint: u64,
+        rsp: u64,
+        #[cfg_attr(not(any(kvm, mshv3)), allow(unused_variables))] config: &SandboxConfiguration,
```
simongdavies: isn't this the same as `#[cfg_attr(target_os = "windows"), allow(unused_variables))]`

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -52,8 +91,124 @@ pub(crate) struct HyperlightVm {
...
+        });
+
+        #[cfg(target_os = "windows")]
+        let interrupt_handle: Arc<dyn InterruptHandleImpl> = Arc::new(WindowsInterruptHandle {
+            state: AtomicU8::new(0),
+            partition_handle: vm.partition_handle(),
+            dropped: AtomicBool::new(false),
+        });
+
+        #[allow(unused_mut)] // needs to be mutable when gdb is enabled
```
simongdavies: shouldn't this be `#[cfg_attr(not(gdb), allow(unused_mut))]`

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Yep, i plan on renaming the trait to Vm in an upcoming PR

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: yes, but this whole file is already only included if kvm is enabled

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: You're right, KVM does not need to and always returns Ok(...), but since mshv and whp share the same trait and they need to be able to return Err, so the Result is the "lowest common denominator"

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: I can see your point. I preferred making it like this since I think that run_vcpu() should not require any params unrelated to the vm itself. If you feel strongly about it, I am open to fix it in another PR?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Yous is a bit cleaner, will add this!

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: yes good catch, will fix

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: yes that is more accurate, will fix

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Agree. But it was changed in a later commit , it is now `crate::hypervisor::crashdump::generate_crashdump(&self.vm)`

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Fixed!

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: I removed it because I don't see its purpose. It was
```rust
#[inline]
fn get_exit_details(&self, exit_reason: WHV_RUN_VP_EXIT_REASON) -> Result<String> {
    let mut error = String::new();
    error.push_str(&format!(
        "Did not receive a halt from Hypervisor as expected - Received {exit_reason:?}!\n"
    ));
    error.push_str(&format!("Registers: \n{:#?}", self.processor.regs()?));
    Ok(error)
}
```

simongdavies: OK, so we no longer capture registers in the log? Again I think that was there because it was used in the past to try and narrow down where issues were (registers such as RIP are/were useful to try and determine what\where errors were happening) I suspect that this isnt as important now that we have crashdump capability

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: OK lets create an issue? I think its more important that we don't rely on people not making mistakes than we do on having ideal APIs

ludfjig: https://github.com/hyperlight-dev/hyperlight/issues/1076

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -257,24 +257,18 @@ mod debug {
 }
 
 /// A Hypervisor driver for HyperV-on-Windows.
```
dblnz: This comment might benefit from an update Or it changed in a later commit.

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -512,47 +512,39 @@ impl Hypervisor for HypervWindowsDriver {
...
-            }
-            // HvRunVpExitReasonX64Halt
-            WHV_RUN_VP_EXIT_REASON(8i32) => {
-                debug!("HyperV Halt Details :\n {:#?}", &self);
-                HyperlightExit::Halt()
-            }
-            // WHvRunVpExitReasonMemoryAccess
-            WHV_RUN_VP_EXIT_REASON(1i32) => {
+                let rip = exit_context.VpContext.Rip + instruction_length as u64;
+                self.set_registers(&[(
```
dblnz, line 225: Where has this come from, why is it needed?  This wasn't here before, right?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -116,20 +116,41 @@ impl HyperlightVm {
...
+
+        self.run(
             mem_mgr,
             host_funcs,
             #[cfg(gdb)]
             dbg_mem_access_fn,
-        )
+        )?;
+
+        Ok(())
```
dblnz: Nit: Only if you make other modifications to the PR

```suggestion
        self.run(
            mem_mgr,
            host_funcs,
            #[cfg(gdb)]
            dbg_mem_access_fn,
        )
```

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -262,28 +262,11 @@ pub(crate) fn is_hypervisor_present() -> bool {
...
 }
 
-/// A Hypervisor driver for HyperV-on-Linux. This hypervisor is often
-/// called the Microsoft Hypervisor (MSHV)
-pub(crate) struct HypervLinuxDriver {
-    _mshv: Mshv,
-    page_size: usize,
+/// A MSHV implementation of a single-vcpu VM
+#[derive(Debug)]
+pub(crate) struct MshvVm {
```
dblnz, line 55: You changed the struct name to `MshvVm` but line 272 implements the old name

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
dblnz, line 1: Finally, removing all that duplicated code is awesome. Good work!

src/hyperlight_host/src/hypervisor/gdb/arch.rs
dblnz: This is so that the vCPU can continue execution, otherwise it triggers the breakpoint again and again

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -653,81 +469,125 @@ impl Drop for HypervLinuxDriver {
...
-                mem_size
-            );
+#[cfg(gdb)]
+impl DebuggableVm for MshvVm {
+    fn translate_gva(&self, gva: u64) -> Result<u64> {
+        use mshv_bindings::{HV_TRANSLATE_GVA_VALIDATE_READ, HV_TRANSLATE_GVA_VALIDATE_WRITE};
+
+        use crate::HyperlightError;
+
+        let flags = (HV_TRANSLATE_GVA_VALIDATE_READ | HV_TRANSLATE_GVA_VALIDATE_WRITE) as u64;
```
dblnz, line 225: I see you've added flags here, was this necessary?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -653,81 +469,125 @@ impl Drop for HypervLinuxDriver {
...
+
+        use crate::hypervisor::gdb::arch::{BP_EX_ID, DB_EX_ID};
+
+        let access_type_mask = if enabled {
+            HV_INTERCEPT_ACCESS_MASK_EXECUTE
+        } else {
+            HV_INTERCEPT_ACCESS_MASK_NONE
+        };
+
+        for vector in [DB_EX_ID, BP_EX_ID] {
```
dblnz, line 249: I remember doing this once when the `new` function was called, to avoid performance degradation.
The only thing is the intercepts remain installed. But we didn't remove them before, either.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -85,31 +57,17 @@ pub(crate) struct MshvVm {
...
-    /// While registers are set up, they will not have been applied to
-    /// the underlying virtual CPU after this function returns. Call the
-    /// `apply_registers` method to do that, or more likely call
-    /// `initialise` to do it for you.
-    #[allow(clippy::too_many_arguments)]
-    // TODO: refactor this function to take fewer arguments. Add trace_info to rt_cfg
+static MSHV: LazyLock<Result<Mshv>> =
+    LazyLock::new(|| Mshv::new().map_err(|e| new_error!("Failed to open /dev/mshv: {}", e)));
+
+impl MshvVm {
```
dblnz, line 63: The `struct MshvVm` was already defined, but the `impl` block had the wrong name. I think it's an artifact of commits editing

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
jprendes: Was about ot comment the same.

Could they maybe be always present, and become no-op when not needed?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+        };
+
+        for (i, region) in mem_regions.iter().enumerate() {
+            // Safety: slots are unique and region points to valid memory since we created the regions
+            unsafe { vm.map_memory((i as u32, region))? };
+        }
+
+        // Mark initial setup as complete for Windows - subsequent map_memory calls will fail
+        #[cfg(target_os = "windows")]
+        vm.complete_initial_memory_setup();
```
jprendes, line 138: could we have this method in all implementations, and have some of them be a no-op?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+            sig_rt_min_offset: config.get_interrupt_vcpu_sigrtmin_offset(),
+            dropped: AtomicBool::new(false),
+        });
+
+        #[cfg(target_os = "windows")]
+        let interrupt_handle: Arc<dyn InterruptHandleImpl> = Arc::new(WindowsInterruptHandle {
+            state: AtomicU8::new(0),
+            partition_handle: vm.partition_handle(),
+            dropped: AtomicBool::new(false),
+        });
```
jprendes, line 173: nit: Could we have `LinuxInterruptHandle::new(vm, config)` and `WindowsInterruptHandle::new(vm, config)`?
Can we also have `struct GenericInterruptHandle(Arc<dyn InterruptHandleImpl>)` and `GenericInterruptHandle::new(vm, config)`?

jprendes: Maybe in another PR, to avoid churning on this one :-)

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+            #[cfg(crashdump)]
+            rt_cfg,
+        };
+
+        // Send the interrupt handle to the GDB thread if debugging is enabled
+        // This is used to allow the GDB thread to stop the vCPU
+        #[cfg(gdb)]
+        if ret.gdb_conn.is_some() {
+            ret.send_dbg_msg(DebugResponse::InterruptHandle(ret.interrupt_handle.clone()))?;
+        }
```
jprendes, line 206: if you move the `interrupt_handle` creation a bit earlier, then you can do this at the same time that you do ` vm.set_debug(true)?;`

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+        #[cfg(feature = "trace_guest")]
+        let mut tc = crate::sandbox::trace::TraceContext::new();
+
+        let result = loop {
+            // ===== KILL() TIMING POINT 2: Before set_tid() =====
+            // If kill() is called and ran to completion BEFORE this line executes:
+            //    - CANCEL_BIT will be set and we will return an early HyperlightExit::Cancelled()
+            //      without sending any signals/WHV api calls
+            #[cfg(any(kvm, mshv3))]
+            self.interrupt_handle.set_tid();
```
jprendes, line 370: **suggestion:** it would be nice if interupt_handle had a method
```
enum InterruptibleResult<T> {
    Completed(T),
    Interrupted,
    Cancelled,
}
fn run_interruptible<T>(&mut self, func: impl FnOnce() -> T) -> InterruptibleResult<T>
```
and all the complexity of the interrupt handler can be contained there

then you can do
```
let exit_reason = self.interrupt_handle.run_interruptible(|| {
    tc.setup_guest_trace(Span::current().context());
    let result = self.vm.run_vcpu();

    // End current host trace by closing the current span that captures traces
    // happening when a guest exits and re-enters.
    {
        // end tracing stuff
    }
    result
})
```

But I would say not for this PR, it would just add noise

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+                    break Ok(());
+                }
+                Ok(HyperlightExit::IoOut(port, data)) => {
+                    self.handle_io(mem_mgr, host_funcs, port, data)?
+                }
+                Ok(HyperlightExit::MmioRead(addr)) => {
+                    let all_regions = self
+                        .sandbox_regions
+                        .iter()
+                        .chain(self.mmap_regions.iter().map(|(_, r)| r));
```
jprendes:
```suggestion
                        .chain(self.get_mapped_regions());
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+
+                            break Err(new_error!("MMIO READ access address {:#x}", addr));
+                        }
+                    }
+                }
+                Ok(HyperlightExit::MmioWrite(addr)) => {
+                    let all_regions = self
+                        .sandbox_regions
+                        .iter()
+                        .chain(self.mmap_regions.iter().map(|(_, r)| r));
```
jprendes:
```suggestion
                        .chain(self.get_mapped_regions());
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+    ) -> Result<()> {
+        use crate::hypervisor::gdb::DebugMemoryAccess;
+
+        if self.gdb_conn.is_none() {
+            return Err(new_error!("Debugging is not enabled"));
+        }
+
+        let mem_access = DebugMemoryAccess {
+            dbg_mem_access_fn,
+            guest_mmap_regions: self.mmap_regions.iter().map(|(_, r)| r.clone()).collect(),
```
jprendes:
```suggestion
            guest_mmap_regions: self.get_mapped_regions().cloned().collect(),
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+            // Get the filename from the binary path
+            let filename = self.rt_cfg.binary_path.clone().and_then(|path| {
+                Path::new(&path)
+                    .file_name()
+                    .and_then(|name| name.to_os_string().into_string().ok())
+            });
+
+            // Include both initial sandbox regions and dynamically mapped regions
+            let mut regions: Vec<MemoryRegion> = self.sandbox_regions.clone();
+            regions.extend(self.mmap_regions.iter().map(|(_, r)| r).cloned());
```
jprendes:
```suggestion
            regions.extend(self.get_mapped_regions().cloned());
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+    /// The accessed region has the given flags
+    AccessViolation(MemoryRegionFlags),
+    /// The accessed region is a stack guard page
+    StackGuardPageViolation,
+}
+
+/// Determines if a known memory access violation occurred at the given address with the given action type.
+/// Returns Some(reason) if violation reason could be determined, or None if violation occurred but in unmapped region.
+fn get_memory_access_violation<'a>(
+    gpa: usize,
```
jprendes, line 805: **nit:** I'm not a big fan of these acronyms, what is gpa? "Grade Point Average"? "Guest Physical Address"?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+    let region = mem_regions.find(|region| region.guest_region.contains(&gpa));
+
+    if let Some(region) = region {
+        if region.region_type == MemoryRegionType::GuardPage {
+            return Some(MemoryAccess::StackGuardPageViolation);
+        } else if !region.flags.contains(tried) {
+            return Some(MemoryAccess::AccessViolation(region.flags));
+        }
+    }
+    None
```
jprendes:
```suggestion
    let region = mem_regions.find(|region| region.guest_region.contains(&gpa))?;
    if region.region_type == MemoryRegionType::GuardPage {
        return Some(MemoryAccess::StackGuardPageViolation);
    }
    if !region.flags.contains(tried) {
        return Some(MemoryAccess::AccessViolation(region.flags));
    }
    // No violation? is this possible?
    None
```

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -0,0 +1,1075 @@
...
+) -> Option<MemoryAccess> {
+    // find the region containing the given gpa
+    let region = mem_regions.find(|region| region.guest_region.contains(&gpa));
+
+    if let Some(region) = region {
+        if region.region_type == MemoryRegionType::GuardPage {
+            return Some(MemoryAccess::StackGuardPageViolation);
+        } else if !region.flags.contains(tried) {
+            return Some(MemoryAccess::AccessViolation(region.flags));
+        }
```
jprendes: Can we reach this point? is is possible to call this method, match a region, not be a guard page, and neither an access violation?
I assume this matches the previous behavior, if it does, and we are still not sure if this is reachable, consider adding a TODO comment to check if this is reachable

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -741,320 +202,138 @@ impl Hypervisor for HypervLinuxDriver {
...
+        Ok(xsave.buffer.to_vec())
+    }
+}
 
-            let vcpu_regs = self.vcpu_fd.get_regs()?;
-            let sregs = self.vcpu_fd.get_sregs()?;
-            let xsave = self.vcpu_fd.get_xsave()?;
+#[cfg(gdb)]
+impl DebuggableVm for MshvVm {
+    fn translate_gva(&self, gva: u64) -> Result<u64> {
```
jprendes, line 220: **nit:** ahh! again the acronyms! I guess this is "Guest Virtual Address", right? bytes and keystrokes are cheap :-)

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: If we want this, lets add it back to all in hyperlight_vm so it applies to all vms, not just windows. But if it's only used for non-live debugging, I say lets get rid of it. If you need extra info while debugging offline, you can add whatever you want

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: If you look at the "Implement memory[unmapping]" commit, this is already implemented

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: We could but I don't really see the point.  I think this is cleaner. Regardless hopefully we will implement memory mapping on windows soon and we can get rid of this.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: I have experiment with this before because I also like this direction, but I wasn't able to due to some flags needing to be set in a particular order. After this PR is merged I'll check it again and see if it's possible. https://github.com/hyperlight-dev/hyperlight/issues/1078

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: If you look at the "Implement memory[unmapping]" commit, this is already implemented

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Lets investigate this after this PR. Since this is current behavior lets keep it for now

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: It was here before, but in handle_io

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: yeah they were changed in different commits which is not ideal. The final result should be accurate though

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: This is the same as existing code in mshv_debug.rs:

```rust
    fn translate_gva(&self, vcpu_fd: &Self::Vcpu, gva: u64) -> Result<u64> {
        let flags = (HV_TRANSLATE_GVA_VALIDATE_READ | HV_TRANSLATE_GVA_VALIDATE_WRITE) as u64;
        let (addr, _) = vcpu_fd
            .translate_gva(gva, flags)
            .map_err(|_| HyperlightError::TranslateGuestAddress(gva))?;

        Ok(addr)
    }
```

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: yeah sorry about that. The final result should still be correct

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Personally I think gva and gpa are fine :P

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Personally I think gva and gpa are fine :P

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: Fixed!

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Nice this is better

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Fixed by moving everything down instead. I need to contruct the `ret` before I can call `send_dbg_msg`

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: It's unclear whether it can happen or not, at least I myself am not sure, and it might vary by hypervisor. Returning `None` in this case is safe anyway, as it will be treated as a generic access violation.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: https://github.com/hyperlight-dev/hyperlight/issues/1078

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: https://github.com/hyperlight-dev/hyperlight/issues/1078

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: https://github.com/hyperlight-dev/hyperlight/issues/1078

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Yes. Do this think this current implementation looks ok?

src/hyperlight_host/src/hypervisor/kvm.rs
ludfjig: Added back

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: https://github.com/hyperlight-dev/hyperlight/issues/1078

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: Added back!

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: added!

PR #931, opened 2025-10-02
Add poison state to sandbox

### LINKED ISSUE #848: Consider adding some kind of "poisoned sandbox" state to prevent sandbox misuse
Certain scenario can leave a sandbox in a leaked state. For example, calling `kill()` on a sandbox's interrupt handle will leak any heap allocated objects in the guest. The same goes for if the guest panics, or if a host function returns an error. 

We should consider introducing something like a poisoned state, which perhaps should prevent any further use of a sandbox, and require the user to restore the sandbox back into a good state via a snapshot restore, along with maybe an unsafe way to circumvent a guard like this

ludfjig:
This PR adds a poison state to sandbox in order to prevent further operations when the sandbox is left in an inconsistent state that could compromise memory safety, data integrity, or security. The sandbox becomes poisoned when guest functions abort/panic or when host-initiated execution cancellation occurs, leaving behind leaked heap allocations, corrupted data structures, or unreleased resources. For example, interrupting execution while guest is allocating can leave the global allocation lock in an inconsistent state, making future allocations impossible in subsequent runs due to infinite locking/spinning.

Poisoned sandboxes will reject all further operations (guest calls, snapshots, memory mapping) until the inconsistent state is resolved through either restoring to a snapshot or manually (unsafely) clearing the poison state.

Closes #848

Round 2:

I've changed the docs and implementation slightly. `is_poison_error` has been made private. Removed unsafe way of clearing poision. Users of hyperlight are expected on-error to check if the sandbox is poisoned, and deal with it accordingly, for example.

```rust
let result = sandbox.call::<()>("guest_panic", ());
if result.is_err() {
    if sandbox.poisoned() {
        // Restore from snapshot to clear poison
        sandbox.restore(&snapshot)?;
        assert!(!sandbox.poisoned());
        
        // Sandbox is now usable again
        sandbox.call::<String>("Echo", "hello".to_string())?;
    }
}
```

### DIFF
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -56,9 +56,45 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
 ///
 /// Guest functions can be called repeatedly while maintaining state between calls.
 /// The sandbox supports creating snapshots and restoring to previous states.
+///
+/// ## Sandbox Poisoning
+///
+/// The sandbox becomes **poisoned** when the guest is not run to completion, leaving it in
+/// an inconsistent state that could compromise memory safety, data integrity, or security.
+///
+/// ### When Does Poisoning Occur?
+///
+/// Poisoning happens when guest execution is interrupted before normal completion:
+///
+/// - **Guest panics or aborts** - When a guest function panics, crashes, or calls `abort()`,
+///   the normal cleanup and unwinding process is interrupted
+/// - **Invalid memory access** - Attempts to read/write/execute memory outside allowed regions
+/// - **Stack overflow** - Guest exhausts its stack space during execution
+/// - **Heap exhaustion** - Guest runs out of heap memory
+/// - **Host-initiated cancellation** - Calling [`InterruptHandle::kill()`] to forcefully
+///   terminate an in-progress guest function
+///
+/// ### Why This Is Unsafe
+///
+/// When guest execution doesn't complete normally, critical cleanup operations are skipped:
+///
+/// - **Memory leaks** - Heap allocations remain unreachable as the call stack is unwound
+/// - **Corrupted allocator state** - Memory allocator metadata (free lists, heap headers)
+///   left inconsistent
+/// - **Locked resources** - Mutexes or other synchronization primitives remain locked
+/// - **Partial state updates** - Data structures left half-modified (corrupted linked lists,
+///   inconsistent hash tables, etc.)
+///
+/// ### Recovery
+///
+/// Use [`restore()`](Self::restore) with a snapshot taken before poisoning occurred.
+/// This is the **only safe way** to recover - it completely replaces all memory state,
+/// eliminating any inconsistencies. See [`restore()`](Self::restore) for details.
 pub struct MultiUseSandbox {
     /// Unique identifier for this sandbox instance
     id: u64,
+    /// Whether this sandbox is poisoned
+    poisoned: bool,
     // We need to keep a reference to the host functions, even if the compiler marks it as unused. The compiler cannot detect our dynamic usages of the host function in `HyperlightFunction::call`.
     pub(super) _host_funcs: Arc<Mutex<FunctionRegistry>>,
     pub(crate) mem_mgr: SandboxMemoryManager<HostSharedMemory>,
@@ -87,6 +123,7 @@ impl MultiUseSandbox {
     ) -> MultiUseSandbox {
         Self {
             id: SANDBOX_ID_COUNTER.fetch_add(1, Ordering::Relaxed),
+            poisoned: false,
             _host_funcs: host_funcs,
             mem_mgr: mgr,
             vm,
@@ -101,6 +138,11 @@ impl MultiUseSandbox {
     ///
     /// The snapshot is tied to this specific sandbox instance and can only be
     /// restored to the same sandbox it was created from.
+    ///
+    /// ## Poisoned Sandbox
+    ///
+    /// This method will return [`crate::HyperlightError::PoisonedSandbox`] if the sandbox
+    /// is currently poisoned. Snapshots can only be taken from non-poisoned sandboxes.
     ///
     /// # Examples
     ///
@@ -122,6 +164,10 @@ impl MultiUseSandbox {
     /// ```
     #[instrument(err(Debug), skip_all, parent = Span::current())]
     pub fn snapshot(&mut self) -> Result<Snapshot> {
+        if self.poisoned {
+            return Err(crate::HyperlightError::PoisonedSandbox);
+        }
+
         if let Some(snapshot) = &self.snapshot {
             return Ok(snapshot.clone());
         }
@@ -139,6 +185,22 @@ impl MultiUseSandbox {
     /// The snapshot must have been created from this same sandbox instance.
     /// Attempting to restore a snapshot from a different sandbox will return
     /// a [`SnapshotSandboxMismatch`](crate::HyperlightError::SnapshotSandboxMismatch) error.
+    ///
+    /// ## Poison State Recovery
+    ///
+    /// This method automatically clears any poison state when successful. This is safe because:
+    /// - Snapshots can only be taken from non-poisoned sandboxes
+    /// - Restoration completely replaces all memory state, eliminating any inconsistencies
+    ///   caused by incomplete guest execution
+    ///
+    /// ### What Gets Fixed During Restore
+    ///
+    /// When a poisoned sandbox is restored, the memory state is completely reset:
+    /// - **Leaked heap memory** - All allocations from interrupted execution are discarded
+    /// - **Corrupted allocator metadata** - Free lists and heap headers restored to consistent state
+    /// - **Locked mutexes** - All lock state is reset
+    /// - **Partial updates** - Data structures restored to their pre-execution state
+    ///
     ///
     /// # Examples
     ///
@@ -162,6 +224,35 @@ impl MultiUseSandbox {
     /// sandbox.restore(&snapshot)?;
     /// let restored_value: i32 = sandbox.call_guest_function_by_name("GetValue", ())?;
     /// assert_eq!(restored_value, 0); // Back to initial state
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// ## Recovering from Poison
+    ///
+    /// ```no_run
+    /// # use hyperlight_host::{MultiUseSandbox, UninitializedSandbox, GuestBinary, HyperlightError};
+    /// # fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let mut sandbox: MultiUseSandbox = UninitializedSandbox::new(
+    ///     GuestBinary::FilePath("guest.bin".into()),
+    ///     None
+    /// )?.evolve()?;
+    ///
+    /// // Take snapshot before potentially poisoning operation
+    /// let snapshot = sandbox.snapshot()?;
+    ///
+    /// // This might poison the sandbox (guest not run to completion)
+    /// let result = sandbox.call::<()>("guest_panic", ());
+    /// if result.is_err() {
+    ///     if sandbox.poisoned() {
+    ///         // Restore from snapshot to clear poison
+    ///         sandbox.restore(&snapshot)?;
+    ///         assert!(!sandbox.poisoned());
+    ///         
+    ///         // Sandbox is now usable again
+    ///         sandbox.call::<String>("Echo", "hello".to_string())?;
+    ///     }
+    /// }
     /// # Ok(())
     /// # }
     /// ```
@@ -197,12 +288,28 @@ impl MultiUseSandbox {
         // The restored snapshot is now our most current snapshot
         self.snapshot = Some(snapshot.clone());
 
+        // Clear poison state when successfully restoring from snapshot.
+        //
+        // # Safety:
+        // This is safe because:
+        // 1. Snapshots can only be taken from non-poisoned sandboxes (verified at snapshot creation)
+        // 2. Restoration completely replaces all memory state, eliminating:
+        //    - All leaked heap allocations (memory is restored to snapshot state)
+        //    - All corrupted data structures (overwritten with consistent snapshot data)
+        //    - All inconsistent global state (reset to snapshot values)
+        self.poisoned = false;
+
         Ok(())
     }
 
     /// Calls a guest function by name with the specified arguments.
     ///
     /// Changes made to the sandbox during execution are *not* persisted.
+    ///
+    /// ## Poisoned Sandbox
+    ///
+    /// This method will return [`crate::HyperlightError::PoisonedSandbox`] if the sandbox
+    /// is currently poisoned. Use [`restore()`](Self::restore) to recover from a poisoned state.
     ///
     /// # Examples
     ///
@@ -242,6 +349,9 @@ impl MultiUseSandbox {
         func_name: &str,
         args: impl ParameterTuple,
     ) -> Result<Output> {
+        if self.poisoned {
+            return Err(crate::HyperlightError::PoisonedSandbox);
+        }
         let snapshot = self.snapshot()?;
         let res = self.call(func_name, args);
         self.restore(&snapshot)?;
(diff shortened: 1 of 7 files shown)


src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -420,8 +504,12 @@ impl MultiUseSandbox {
         // - the serialized guest function result is zeroed out by us (the host) during deserialization, see `get_guest_function_call_result`
         // - any serialized host function call are zeroed out by us (the host) during deserialization, see `get_host_function_call`
         // - any serialized host function result is zeroed out by the guest during deserialization, see `get_host_return_value`
-        if res.is_err() {
+        if let Err(e) = &res {
             self.mem_mgr.clear_io_buffers();
+
+            // Determine if we should poison the sandbox.
+            // We should poison the sandbox if the error is GuestAborted, ExecutionCancelledByHost,
+            self.poisoned |= Self::is_poison_error(e);
```
dblnz: Is bitwise OR intended here?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -458,6 +546,195 @@ impl MultiUseSandbox {
...
+    ///     // - Validated all data structures are consistent
+    ///     // - Checked all global state is correct
+    ///     unsafe {
+    ///         sandbox.clear_poison();
+    ///     }
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub unsafe fn clear_poison(&mut self) {
```
dblnz: Do we really want to allow clearing the poison? Marking it `unsafe` and thoroughly describing the conditions in the comments does help.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -458,6 +546,195 @@ impl MultiUseSandbox {
...
+            // - Memory allocations that were in progress of being allocated/freed
+            // - I/O operations in inconsistent state (partial writes, half-open connections)
+            // - Critical sections or mutexes that were being acquired/released
+            // - Memory allocator metadata in corrupted state
+            HyperlightError::GuestAborted(_, _) | HyperlightError::ExecutionCanceledByHost() => {
+                true
+            }
+
+            // All other errors do not poison the sandbox
+            HyperlightError::AnyhowError(_)
```
dblnz: Nit: Maybe using `_` to avoid explicitly stating each error type.
Or the intention is to acknowledge the type of error (if it is poisonous or not) when a new variant is introduced?

ludfjig: yes that is the intention

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I think so. It's up to the user at this point

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: yeah but I guess `=` and `|=` are the same here :D

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -512,10 +601,192 @@ impl MultiUseSandbox {
...
+            | HyperlightError::MemoryProtectionFailed(_)
+            | HyperlightError::MemoryRequestTooBig(_, _)
+            | HyperlightError::MetricNotFound(_)
+            | HyperlightError::MmapFailed(_)
+            | HyperlightError::MprotectFailed(_)
+            | HyperlightError::NoHypervisorFound()
+            | HyperlightError::NoMemorySnapshot
+            | HyperlightError::ParameterValueConversionFailure(_, _)
+            | HyperlightError::PEFileProcessingFailure(_)
+            | HyperlightError::PoisonedSandbox
```
jsturtevant: It seems a bit strange to me that IsPoisoned doesn't return true when this is the error.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: Is the anyhow error here allowing for anyone one to return an error via anyhow and circumvent this?

Same for `HyperlightError::Error(_)`?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -512,10 +601,192 @@ impl MultiUseSandbox {
...
+            | HyperlightError::RawPointerLessThanBaseAddress(_, _)
+            | HyperlightError::RefCellBorrowFailed(_)
+            | HyperlightError::RefCellMutBorrowFailed(_)
+            | HyperlightError::ReturnValueConversionFailure(_, _)
+            | HyperlightError::StackOverflow()
+            | HyperlightError::SnapshotSandboxMismatch
```
jsturtevant: Some of these like this one seem like a data corruption. I am a little unclear on now we determine if it puts a function into a "posioned" state.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -56,9 +56,59 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
...
+/// - Corrupted memory allocator metadata (free lists, heap headers)
+///
+/// **Resource Leaks**:
+/// - File descriptors opened but never closed due to interrupted I/O operations
+/// - Network sockets left in half-open states
+/// - Guest-side mutex locks that are never released, causing future deadlocks
+///
+/// **Data Structure Corruption**:
+/// - Hash tables with inconsistent bucket counts vs. actual entries
+/// - Linked lists with broken next/prev pointers
```
jsturtevant: Are these things we have that get corrupted? These seems like general information on corruption but not specific to what we are protecting against.

ludfjig: yes these are just general information warnings

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Only cases where we don't allow the guest run to cleanly finish execution are poisoned. This is only when killing/interrupting a running sandbox, or a rust panic, since we are aborting instead of unwinding. The rest should be fine

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: you're right it's confusing, but it is correct. `PoisonedSandbox` is what we return in case some other error already have caused poison state.

ludfjig: Which reminds we should probably better organize our errors into categories.

jsturtevant: isn't by definition poisoned then?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -512,10 +601,192 @@ impl MultiUseSandbox {
...
+            HyperlightError::AnyhowError(_)
+            | HyperlightError::BoundsCheckFailed(_, _)
+            | HyperlightError::CheckedAddOverflow(_, _)
+            | HyperlightError::CStringConversionError(_)
+            | HyperlightError::Error(_)
+            | HyperlightError::ExecutionAccessViolation(_)
+            | HyperlightError::FailedToGetValueFromParameter()
+            | HyperlightError::FieldIsMissingInGuestLogData(_)
+            | HyperlightError::GuestError(_, _)
+            | HyperlightError::GuestExecutionHungOnHostFunctionCall()
```
jsturtevant: is this one where things could get corrupted?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -56,9 +56,59 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
...
+/// ### Operations That Poison the Sandbox:
+///
+/// **Guest Function Aborts/Panics** ([`HyperlightError::GuestAborted`](crate::HyperlightError::GuestAborted)):
+/// - Guest functions that panic, abort, or crash unexpectedly
+/// - Call stack unwinding interrupted, leaving heap allocations leaked
+///
+/// **Host-Initiated Execution Cancellation** ([`HyperlightError::ExecutionCanceledByHost`](crate::HyperlightError::ExecutionCanceledByHost)):
+/// - Guest execution interrupted by host via interrupt handle
+/// - Operations terminated mid-execution, leaving partial state updates
+/// - Example: `interrupt_handle.kill()` during guest function → sandbox poisoned
```
jsturtevant: Would this mean we need to set poisoned as soon as the 'kill' function is invoked?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -56,9 +56,59 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
...
+/// ### Common Inconsistent States:
+///
+/// **Memory Corruption**:
+/// - Heap allocations that are leaked due to unwound call stacks
+/// - Dangling pointers to freed memory that may be reallocated
+/// - Corrupted memory allocator metadata (free lists, heap headers)
+///
+/// **Resource Leaks**:
+/// - File descriptors opened but never closed due to interrupted I/O operations
+/// - Network sockets left in half-open states
```
jsturtevant: do our sandboxes open network sockets or FDs?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: Would all of these fall under memory corruption?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -56,9 +56,59 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
...
+///
+/// **Data Structure Corruption**:
+/// - Hash tables with inconsistent bucket counts vs. actual entries
+/// - Linked lists with broken next/prev pointers
+/// - Reference-counted objects with incorrect ref counts
+///
+/// **Global State Inconsistency**:
+/// - Static variables left partially updated (e.g., counters, flags, caches)
+/// - Thread-local storage in undefined states
+/// - Interrupt handlers or signal masks in incorrect configurations
```
jsturtevant: Do we have threading support and interrupts on guest side?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -56,9 +56,59 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
...
+/// - Network sockets left in half-open states
+/// - Guest-side mutex locks that are never released, causing future deadlocks
+///
+/// **Data Structure Corruption**:
+/// - Hash tables with inconsistent bucket counts vs. actual entries
+/// - Linked lists with broken next/prev pointers
+/// - Reference-counted objects with incorrect ref counts
+///
+/// **Global State Inconsistency**:
+/// - Static variables left partially updated (e.g., counters, flags, caches)
```
jsturtevant: This also feels like mostly memory related

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: GuestExecutionHungOnHostFunctionCall is never instantiated, we should remove it. Looks like we might have a couple of errors are unused

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: No because it's possible the `kill()` doens't kick the vcpu and therefore doesn't lead to inconsistent state, for example if you call kill() before or after a guest-function-call.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I added this with purpose of showing it's really unsafe. Several things mentioned are not relevant to our guest, but might be to somebody elses

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I added this with purpose of showing it's really unsafe. Several things mentioned are not relevant to our guest, but might be to somebody elses

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I added this with purpose of showing it's really unsafe. Several things mentioned are not relevant to our guest, but might be to somebody elses

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Probably yes

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: No because this error doesn't cause poison. PoisionedSandbox is what we returned if it is poisoned.

jsturtevant: if the sandbox returns this error then `poisoned() == true`?  something feels off about not verifying that if this is returned and the internal state should also be verified to be correct.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: IMO, this list feels overly generic and distracts from the rest of the comment but can see what others think

jsturtevant (commented):
Overall LGTM, a few nits but happy to see what other say

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: the sandbox will never return this, it's only ever returned if we see that the bool is true. But, you're right, I added it here anyway so it won't be confusing in the future.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I trimmed it down

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -512,10 +587,178 @@ impl MultiUseSandbox {
...
+    /// ```
+    pub unsafe fn clear_poison(&mut self) {
+        self.poisoned = false;
+    }
+
+    /// Determines if the given error should poison the sandbox.
+    ///
+    /// Errors that poison the sandbox are those that can leave the sandbox in an inconsistent
+    /// state where memory, resources, or data structures may be corrupted or leaked.
+    fn is_poison_error(err: &HyperlightError) -> bool {
```
simongdavies: Couldn't we do something like this instead:
```
 impl HyperlightError {
    /// Returns whether this error should poison the sandbox.
    ///
    /// Errors that poison the sandbox are those that leave it in an inconsistent
    /// state where memory, resources, or data structures may be corrupted or leaked.
    pub fn is_poison_error(&self) -> bool {
        matches!(
            self,
            HyperlightError::GuestAborted(_, _)
                | HyperlightError::ExecutionCanceledByHost()
                | HyperlightError::PoisonedSandbox
        )
    }
}
```

and then get rid of this function?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: I am all for giving the user control, however I don't see a scenario where a user could resolve this without calling restore or creating a new sandbox, what did you have in mind?

simongdavies (dismissed):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: yes I could move it. But I want to keep the explicit pattern matching because I want to force maintainers who add new errors to think whether the new errors poison a sandbox or not (via compiler error).

ludfjig:
> I am all for giving the user control, however I don't see a scenario where a user could resolve this without calling restore or creating a new sandbox, what did you have in mind?

Not exactly sure. Maybe some usecase for completely state-less guests? Not sure tbh.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: OK, makes sense , I think though it makes even more sense to add it as an impl HyperlightError method in that case

simongdavies:
> > I am all for giving the user control, however I don't see a scenario where a user could resolve this without calling restore or creating a new sandbox, what did you have in mind?
> 
> Not exactly sure. Maybe some usecase for completely state-less guests? Not sure tbh.

No idea why this doesn't show up as a review comment, anyway my instinct is not to add this now and then if someone wants it add it later, seems like its safer that way

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: fixed

jsturtevant (approved):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: removed

Issue #179, opened 2025-01-28
test float_roundtrip fails when using c guest


ludfjig:
...but passes with rust guest. Why?

To replicate run
`GUEST="c" cargo test -p hyperlight-host -- float_roundtrip --ignored`
which fails. But 
`cargo test -p hyperlight-host -- float_roundtrip --ignored` passes

ludfjig:
The issue is that compiling rust with target x86_64-unknown-none does not use any floating point registers. However, compiling clang with target x86_64-unknown-none-elf does. There does not seem to be an easy way change this in clang

jsturtevant:
Something to look into is similar to https://github.com/hyperlight-dev/hyperlight-wasm/pull/166, where the softfloat abi is causing issue between  different between x86_64-unknown-none and cranelifts compiler.  The clang x86_64-unknown-none-elf and x86_64-unknown-none might have a similar mis-match

PR #937, opened 2025-10-06
Adding FFI functions for Hyperlight Guest CAPI

vshailesh:
- Added function for bool, f32, f64, string, vecbytes
- Added tests for bool, f32, f64, string

### DIFF
--- a/src/hyperlight_host/tests/integration_test.rs
+++ b/src/hyperlight_host/tests/integration_test.rs
@@ -21,6 +21,7 @@ use std::time::Duration;
 
 use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
 use hyperlight_common::mem::PAGE_SIZE;
+use hyperlight_host::func::ReturnValue;
 use hyperlight_host::sandbox::SandboxConfiguration;
 use hyperlight_host::{GuestBinary, HyperlightError, MultiUseSandbox, UninitializedSandbox};
 use hyperlight_testing::simplelogger::{LOGGER, SimpleLogger};
@@ -769,7 +770,7 @@ fn log_test_messages(levelfilter: Option<log::LevelFilter>) {
 /// Tests whether host is able to return Bool as return type
 /// or not
 #[test]
-fn test_if_guest_is_able_to_get_return_values_from_host() {
+fn test_if_guest_is_able_to_get_bool_return_values_from_host() {
     let mut sbox1 = new_uninit().unwrap();
 
     sbox1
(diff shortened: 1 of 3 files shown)


vshailesh:
@ludfjig 
I added test for bool and string, and they pass on my local builds.
But tests for f32 and f64 fail, and the values getting returned is some garbage value, maybe because of overflow. I am not sure.

Please review the test cases, and tell me if I have missed something

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -765,3 +765,87 @@ fn log_test_messages(levelfilter: Option<log::LevelFilter>) {
             .unwrap();
     }
 }
+
+/// Tests whether host is able to return Bool as return type
+/// or not
+#[test]
+fn test_if_guest_is_able_to_get_bool_return_values_from_host() {
+    let mut sbox1 = new_uninit().unwrap();
```
ludfjig: Thanks for this. the `new_uninit` will either return a c or rust based sandbox, depending on environment variables. Since this test should only run the c guest, could you create a new function `new_uninit_c` (just like there already exists a `new_uninit_rust`) and use that for all the tests. 

I also believe the reason that the float/double test fails is because of incorrect registers being used, see #179. For now you can mark the float/double tests using #[ignore], and we should be good to merge it, and unignore it when the issues are fixed

ludfjig (changes requested):

src/hyperlight_host/tests/integration_test.rs
vshailesh: sure, i will add that function

vshailesh: added new_uninit_c function to create C simpleguests only

dblnz (commented):
Thanks for your contribution!
It looks good from my point of view.
Let's see what CI says 😄  I've triggered the builds.

ludfjig:
Clippy has a couple of complaints, otherwise looks good to me. For the #[ignore] you added, would you mind putting a small comment about why, and linking the issue #179.

BTW when you want to update branch because it's out of date, I recommend using git rebase main, so the commits will not show up on this branch, instead of using something like git merge which will explicitly put the commits on the branch

vshailesh:
implemented the clippy suggested changes and added comments for ignore attribute

dblnz:
Because of other work merged into main, now there are conflicts with this PR.
Please rebase on the latest main and fix the conflicts.

PR #943, opened 2025-10-07
Add support for mmaped memory in crashdumps and guest debugging

### LINKED ISSUE #747: Investigate make crashdump feature work with mmaped regions
Should we add the `mmap`ed regions also to the crashdump? I think there's some additional logic needed to make that work

_Originally posted by @dblnz in https://github.com/hyperlight-dev/hyperlight/pull/742#discussion_r2243608688_

dblnz:
Fixes #747 and is a blocker for https://github.com/hyperlight-dev/hyperlight-wasm/pull/129.

This modifies the guest debugging by adding a `DebugMemoryAccess` type that handles the memory access for the debugger. It abstracts over the type of memory, shared or mmaped.

Additionally, we now include the mmapped memory regions in the crashdump.


ludfjig:
This looks good to me, but it looks like we're circumventing our SharedMemory abstraction here with unsafe code like from_raw. Would it be possible to avoid this?

dblnz:
That is true, but I think this is a consequence of how the mmapped memory is handled. As a separate set of regions.
What I think I could do is implement something similar to `SharedMemory` abstraction over the mmapped memory that keeps track of accessing and mapping/unmapping.

ludfjig:
> That is true, but I think this is a consequence of how the mmapped memory is handled. As a separate set of regions.
What I think I could do is implement something similar to SharedMemory abstraction over the mmapped memory that keeps track of accessing and mapping/unmapping.

Oh yeah forgot about that... That might be unfeasible

dblnz:
I believe this PR touches on the Sandbox memory model discussion we need to have, specifically about how `mmaped` memory should fit into it. This is tracked in #956 .

jsturtevant (dismissed):

simongdavies:
> > That is true, but I think this is a consequence of how the mmapped memory is handled. As a separate set of regions.
> > What I think I could do is implement something similar to SharedMemory abstraction over the mmapped memory that keeps track of accessing and mapping/unmapping.
> 
> Oh yeah forgot about that... That might be unfeasible

I think we should go forward with this as is and then address the winder memory issues in #956

simongdavies (dismissed):

PR #959, opened 2025-10-16
Fixes a race condition in killing Sandboxes

simongdavies:
Fixes a race condition where a sandbox kill arrives after a sandbox has successfully exited causing the subsequent run to fail.

There is a breaking change in this PR, previously if kill was called on an `InterruptHandle` before or while a guest call was not in progress the next guest call made on the `Sandbox` would be cancelled , now this scenario is a no-op. kill only takes effect if there is a guest call running.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -168,7 +168,9 @@ test-integration guest target=default-target features="":
     {{if os() == "windows" { "$env:" } else { "" } }}GUEST="{{guest}}"{{if os() == "windows" { ";" } else { "" } }} {{ cargo-cmd }} test --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test integration_test execute_on_heap {{ if features =="" {""} else {"--features " + features} }} -- --ignored
     
     @# run the rest of the integration tests
-    {{if os() == "windows" { "$env:" } else { "" } }}GUEST="{{guest}}"{{if os() == "windows" { ";" } else { "" } }} {{ cargo-cmd }} test -p hyperlight-host {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F init-paging," + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test '*'
+    @# skip interrupt_random_kill_stress_test and then run it explicitly so we can see the output more 
+    {{if os() == "windows" { "$env:" } else { "" } }}GUEST="{{guest}}"{{if os() == "windows" { ";" } else { "" } }} {{ cargo-cmd }} test -p hyperlight-host {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F init-paging," + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test '*' --  --skip interrupt_random_kill_stress_test
+    {{if os() == "windows" { "$env:" } else { "" } }}GUEST="{{guest}}"{{if os() == "windows" { ";" } else { "" } }} {{ cargo-cmd }} test -p hyperlight-host {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F init-paging," + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test integration_test interrupt_random_kill_stress_test --  --nocapture --exact
 
 # tests compilation with no default features on different platforms
 test-compilation-no-default-features target=default-target:
--- a/src/hyperlight_host/src/hypervisor/hyperv_windows.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_windows.rs
@@ -17,7 +17,7 @@ limitations under the License.
 use std::fmt;
 use std::fmt::{Debug, Formatter};
 use std::string::String;
-use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 
 use log::LevelFilter;
@@ -327,10 +327,11 @@ impl HypervWindowsDriver {
         };
 
         let interrupt_handle = Arc::new(WindowsInterruptHandle {
-            running: AtomicBool::new(false),
-            cancel_requested: AtomicBool::new(false),
+            running: AtomicU64::new(0),
+            cancel_requested: AtomicU64::new(0),
             #[cfg(gdb)]
             debug_interrupt: AtomicBool::new(false),
+            call_active: AtomicBool::new(false),
             partition_handle,
             dropped: AtomicBool::new(false),
         });
@@ -549,7 +550,8 @@ impl Hypervisor for HypervWindowsDriver {
         &mut self,
         #[cfg(feature = "trace_guest")] tc: &mut crate::sandbox::trace::TraceContext,
     ) -> Result<super::HyperlightExit> {
-        self.interrupt_handle.running.store(true, Ordering::Relaxed);
+        // Get current generation and set running bit
+        let generation = self.interrupt_handle.set_running_bit();
 
         #[cfg(not(gdb))]
         let debug_interrupt = false;
@@ -559,11 +561,10 @@ impl Hypervisor for HypervWindowsDriver {
             .debug_interrupt
             .load(Ordering::Relaxed);
 
-        // Don't run the vcpu if `cancel_requested` is true
+        // Check if cancellation was requested for THIS generation
         let exit_context = if self
             .interrupt_handle
-            .cancel_requested
-            .load(Ordering::Relaxed)
+            .is_cancel_requested_for_generation(generation)
             || debug_interrupt
         {
             WHV_RUN_VP_EXIT_CONTEXT {
@@ -578,12 +579,21 @@ impl Hypervisor for HypervWindowsDriver {
 
             self.processor.run()?
         };
-        self.interrupt_handle
-            .cancel_requested
-            .store(false, Ordering::Relaxed);
-        self.interrupt_handle
-            .running
-            .store(false, Ordering::Relaxed);
+
+        // Clear running bit
+        self.interrupt_handle.clear_running_bit();
+
+        let is_canceled = exit_context.ExitReason == WHV_RUN_VP_EXIT_REASON(8193i32); // WHvRunVpExitReasonCanceled
+
+        // Check if this was a manual cancellation (vs internal Windows cancellation)
+        let cancel_was_requested_manually = self
+            .interrupt_handle
+            .is_cancel_requested_for_generation(generation);
+
+        // Only clear cancel_requested if we're actually processing a cancellation for this generation
+        if is_canceled && cancel_was_requested_manually {
+            self.interrupt_handle.clear_cancel_requested();
+        }
 
         #[cfg(gdb)]
         let debug_interrupt = self
@@ -659,12 +669,32 @@ impl Hypervisor for HypervWindowsDriver {
                     // return a special exit reason so that the gdb thread can handle it
                     // and resume execution
                     HyperlightExit::Debug(VcpuStopReason::Interrupt)
+                } else if !cancel_was_requested_manually {
+                    // This was an internal cancellation
+                    // The virtualization stack can use this function to return the control
+                    // of a virtual processor back to the virtualization stack in case it
+                    // needs to change the state of a VM or to inject an event into the processor
+                    // see https://learn.microsoft.com/en-us/virtualization/api/hypervisor-platform/funcs/whvcancelrunvirtualprocessor#remarks
+                    debug!("Internal cancellation detected, returning Retry error");
+                    HyperlightExit::Retry()
                 } else {
                     HyperlightExit::Cancelled()
                 }
 
                 #[cfg(not(gdb))]
-                HyperlightExit::Cancelled()
+                {
+                    if !cancel_was_requested_manually {
+                        // This was an internal cancellation
+                        // The virtualization stack can use this function to return the control
+                        // of a virtual processor back to the virtualization stack in case it
+                        // needs to change the state of a VM or to inject an event into the processor
+                        // see https://learn.microsoft.com/en-us/virtualization/api/hypervisor-platform/funcs/whvcancelrunvirtualprocessor#remarks
+                        debug!("Internal cancellation detected, returning Retry error");
+                        HyperlightExit::Retry()
+                    } else {
+                        HyperlightExit::Cancelled()
+                    }
+                }
             }
             #[cfg(gdb)]
             WHV_RUN_VP_EXIT_REASON(4098i32) => {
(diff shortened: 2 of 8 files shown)


src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -435,13 +435,51 @@ impl VirtualCPU {
...
+    /// - Only the intended guest function call is affected
+    /// - Multiple rapid kill() calls on the same generation are idempotent
+    ///
+    /// # Blocking Behavior
+    ///
+    /// This function will block while attempting to deliver the signal to the VCPU thread,
+    /// retrying until either:
+    /// - The signal is successfully delivered (VCPU transitions from running to not running)
+    /// - The VCPU stops running for another reason (e.g., call completes normally)
+    /// - No call is active (call_active=false)
```
jsturtevant: Does it block? I thought it would just return false?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -458,69 +496,229 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+    /// This flag prevents kill() from having any effect when called outside of a
+    /// guest function call. This solves the "kill-in-advance" problem where kill()
+    /// could be called before a guest function starts and would incorrectly cancel it.
+    ///
+    /// # Behavior
+    ///
+    /// - Set to true at the start of call_guest_function_by_name_no_reset()
+    /// - Cleared at the end of call_guest_function_by_name_no_reset()
+    /// - kill() only stamps cancel_requested if call_active is true
+    /// - If kill() is called when call_active=false, it returns false and has no effect
```
jsturtevant, line 792: Is this actually the case? I don't see it returning early when call is not active.

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -536,14 +734,28 @@ impl LinuxInterruptHandle {
...
+            // Only stamp if a call is actually active (call_active=true)
+            if stamp_generation
+                && target_generation.is_none()
+                && self.call_active.load(Ordering::Acquire)
+            {
+                self.set_cancel_requested(generation);
+                target_generation = Some(generation);
+            }
+
+            // If not running, we've stamped the generation (if requested), so we're done
```
jsturtevant, line 850: should there be a check that `call_active` is `false` so this exits early in the case that the call ins't active? 

I think this is currently working due to the fact that when `call_active` `running` is also active. I think there is one case where the call might not be active but the vm is running (during initial set up) and i think it might hang in that scenario?

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -663,11 +668,31 @@ impl Hypervisor for HypervWindowsDriver {
...
 
                 #[cfg(not(gdb))]
-                HyperlightExit::Cancelled()
+                {
+                    if !cancel_was_requested_manually {
+                        // This was an internal cancellation
+                        // The virtualization stack can use this function to return the control
+                        // of a virtual processor back to the virtualization stack in case it
+                        // needs to change the state of a VM or to inject an event into the processor
+                        debug!("Internal cancellation detected, returning Retry error");
```
jsturtevant, line 692: see https://learn.microsoft.com/en-us/virtualization/api/hypervisor-platform/funcs/whvcancelrunvirtualprocessor#remarks

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: it actually did this check in `send_signal` but I've put one in the `kill` function now as well

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: Added

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -458,69 +500,261 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+    ///
+    /// Although there's a theoretical race where:
+    /// 1. Thread A checks call_active (false)
+    /// 2. Thread B sets call_active (true) and starts guest call
+    /// 3. Thread A's kill() returns false (no effect)
+    ///
+    /// This is acceptable because the generation tracking provides an additional
+    /// safety layer. Even if a stale kill somehow stamped cancel_requested, the
+    /// generation mismatch would cause it to be ignored.
+    call_active: AtomicBool,
```
ludfjig, line 804: I think this flag is redundant. We will already not send any signals when calling `kill` if the vcpu is not running

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -484,92 +526,303 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+    ///
+    /// # Returns
+    ///
+    /// The new generation number assigned to this guest function call.
+    ///
+    /// # Note
+    ///
+    /// This is only called on Linux (KVM/MSHV). Windows uses a different interrupt mechanism.
+    #[cfg(any(kvm, mshv))]
+    fn increment_call_generation(&self) -> u64;
```
ludfjig: these all method are public right now. I recommend making a new private trait InterruptHandleImpl like so `pub(crate) trait InterruptHandleImpl: InterruptHandle` and store a `dyn` of this trait, but return it as a `dyn InterruptHandle` to the user

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -484,92 +526,303 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+
+    /// Internal method to increment the generation counter.
+    ///
+    /// Returns the new generation value after incrementing.
+    #[cfg(any(kvm, mshv, target_os = "windows"))]
+    fn increment_generation_internal(&self) -> u64;
+}
+
+// Default implementations for common methods
+impl dyn InterruptHandle {
```
ludfjig: I believe you can put default method directly in the trait definition, no need to `impl dyn InterruptHandle`

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -390,6 +390,10 @@ impl MultiUseSandbox {
         return_type: ReturnType,
         args: Vec<ParameterValue>,
     ) -> Result<ReturnValue> {
+        // Mark that a guest function call is now active
+        // (This also increments the generation counter internally)
+        self.vm.interrupt_handle().set_call_active();
```
ludfjig: If you want, you could consider making this a function which takes closure like 
```
start_call(||{
...rest of existing code
})
```

where the function calls set_call_active and clear_call_active for you (and calls your closure in between) in order not to forget about the structure in the future.

Similar to existing `maybe_time_and_emit_host_call`

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: Its short circuits the whole process and returns immediately if there is no call active

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: I added a guard instead which I think is a better solution in this case

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: It will block until one of the conditions listed is true and then return a bool.

Justfile
```diff
@@ -168,7 +168,9 @@ test-integration guest target=default-target features="":
     {{if os() == "windows" { "$env:" } else { "" } }}GUEST="{{guest}}"{{if os() == "windows" { ";" } else { "" } }} {{ cargo-cmd }} test --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test integration_test execute_on_heap {{ if features =="" {""} else {"--features " + features} }} -- --ignored
     
     @# run the rest of the integration tests
-    {{if os() == "windows" { "$env:" } else { "" } }}GUEST="{{guest}}"{{if os() == "windows" { ";" } else { "" } }} {{ cargo-cmd }} test -p hyperlight-host {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F init-paging," + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test '*'
+    @# skip interrupt_random_kill_stress_test and then run it explicitly so we can see the output more
```
ludfjig, line 171: Do we want to keep running this test separately still?

src/hyperlight_host/Cargo.toml
```diff
@@ -67,7 +67,7 @@ windows = { version = "0.62", features = [
     "Win32_System_JobObjects",
     "Win32_System_SystemServices",
 ] }
-windows-sys = { version = "0.61", features = ["Win32"] }
+windows-sys = { version = "0.61", features = ["Win32", "Win32_System", "Win32_System_Performance", "Win32_System_WindowsProgramming", "Win32_Foundation"] }
```
ludfjig: Were these added on purpose? Do we need them?

src/hyperlight_host/Cargo.toml
```diff
@@ -79,8 +79,8 @@ kvm-bindings = { version = "0.14", features = ["fam-wrappers"], optional = true
 kvm-ioctls = { version = "0.24", optional = true }
 mshv-bindings2 = { package="mshv-bindings", version = "=0.2.1", optional = true }
 mshv-ioctls2 = { package="mshv-ioctls",  version = "=0.2.1", optional = true}
-mshv-bindings3 = { package="mshv-bindings", version = "0.6.1", optional = true }
-mshv-ioctls3 = { package="mshv-ioctls",  version = "0.6.1", optional = true}
+mshv-bindings3 = { package="mshv-bindings", version = "0.6", optional = true }
```
ludfjig: Is 0.6.1 -> 0.6 on purpose?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1238,9 +1238,13 @@ mod tests {
                     vec![]
                 };
                 let status = std::process::Command::new("cargo")
-                    .args(["test", "-p", "hyperlight-host"])
+                    .args(["test", "-p", "hyperlight-host", "--lib"])
                     .args(target_args)
-                    .args(["--", "--ignored", test])
+                    .args([
```
ludfjig: what does this change?

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: I still think it's true. If in `send_signal` you move everything to after tye
```
if !running {
    break;
}
```
I believe you'll get the same behavior and you can get rid of the flag?

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -578,12 +579,22 @@ impl Hypervisor for HypervWindowsDriver {
...
-            .store(false, Ordering::Relaxed);
-        self.interrupt_handle
-            .running
-            .store(false, Ordering::Relaxed);
+
+        // Clear running bit
+        self.interrupt_handle.clear_running_bit();
+
+        let exit_reason = exit_context.ExitReason;
+        let is_canceled = exit_reason.0 == 8193; // WHvRunVpExitReasonCanceled
```
ludfjig: Can't you use the constant from the library?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -461,13 +466,50 @@ impl VirtualCPU {
...
+    /// - **No active call**: If called when no guest function call is in progress (call_active=false),
+    ///   this has no effect and returns `false`. This prevents "kill-in-advance" where kill()
+    ///   is called before a guest function starts.
+    ///
+    /// - **During host function**: If the guest call is currently executing a host function
+    ///   (VCPU not running but call_active=true), this stamps cancel_requested. When the
+    ///   host function returns and attempts to re-enter the guest, the cancellation will
+    ///   be detected and the call will abort. Returns `true`.
+    ///
+    /// # Generation Tracking (linux only)
```
jsturtevant:
```suggestion
    /// # Generation Tracking 
```

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -461,13 +466,50 @@ impl VirtualCPU {
 pub trait InterruptHandle: Debug + Send + Sync {
     /// Interrupt the corresponding sandbox from running.
     ///
-    /// - If this is called while the vcpu is running, then it will interrupt the vcpu and return `true`.
-    /// - If this is called while the vcpu is not running, (for example during a host call), the
-    ///   vcpu will not immediately be interrupted, but will prevent the vcpu from running **the next time**
-    ///   it's scheduled, and returns `false`.
+    /// This method attempts to cancel a currently executing guest function call by sending
+    /// a signal to the VCPU thread. It uses generation tracking (linux only) and call_active flag to
```
jsturtevant:
```suggestion
    /// a signal to the VCPU thread. It uses generation tracking and call_active flag to
```

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -596,17 +837,37 @@ impl LinuxInterruptHandle {
...
-        self.send_signal()
+        self.send_signal(false)
     }
-    fn dropped(&self) -> bool {
-        self.dropped.load(Ordering::Relaxed)
+
+    #[cfg(any(kvm, mshv))]
+    fn increment_call_generation(&self) -> u64 {
+        self.increment_generation()
+    }
```
jsturtevant: is this used anywhere?  Windows now has   `increment_generation` can we just use that directly?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -482,94 +524,293 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+    /// used throughout the entire guest call, even if the VCPU is run multiple
+    /// times (due to host function calls, retries, etc.).
+    ///
+    /// # Returns
+    ///
+    /// The new generation number assigned to this guest function call.
+    ///
+    /// # Note
+    ///
+    /// This is only called on Linux (KVM/MSHV). Windows uses a different interrupt mechanism.
```
jsturtevant: There is a `increment_generation` on windows can we just use that?  I not seeing this being used anywere in the code any more?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -482,94 +524,293 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+    /// Returns the dropped atomic bool reference for default implementations.
+    ///
+    /// This is used by default trait methods to access the common dropped field.
+    fn get_dropped(&self) -> &std::sync::atomic::AtomicBool;
+
+    /// Internal method to increment the generation counter.
+    ///
+    /// Returns the new generation value after incrementing.
+    #[cfg(any(kvm, mshv, target_os = "windows"))]
+    fn increment_generation_internal(&self) -> u64;
```
jsturtevant: is the `cfg` needed here?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -482,94 +524,293 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+        self.cancel_requested.store(value, Ordering::Release);
+    }
+
+    // Clear cancel_requested (reset to no cancellation)
+    fn clear_cancel_requested(&self) {
+        self.cancel_requested.store(0, Ordering::Release);
+    }
+
+    // Check if cancel_requested is set for the given generation
+    fn is_cancel_requested_for_generation(&self, generation: u64) -> bool {
```
jsturtevant: This looks to be the same on Windows and linux now.  Could we condense them to be default impls on the Trait?  Several of these are the same (`clear_cancel_requested`, `set_cancel_requested`)

src/hyperlight_host/src/hypervisor/mod.rs
jsturtevant: This does guard around some additional functionality in `call_guest_function_by_name_no_reset` when the guest isn't actually running but being set up to run. Now we have some tests that pass might be interesting experiment to see if we remove this if it simplified?

simongdavies: Well yes perhaps but this just short circuits the whole thing and returns early, I'm a bit loathed to remove it since it doesnt do any harm, it works and you are planning a new implementation anwyay

Justfile
simongdavies: Yes , I did this so that we can see the results in the CI output , otherwise they get lost.

src/hyperlight_host/Cargo.toml
simongdavies: Yes they were , but we might not need them any longer I will check.

src/hyperlight_host/Cargo.toml
simongdavies: Yes, but I probably shouldn't do it here, I just added a suggestion for it to dorus mshv PR , I will remove it from here

src/hyperlight_host/src/mem/shared_mem.rs
simongdavies: I don't recall what this is for, I'll remove it

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: No, I dont think so

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -596,18 +875,43 @@ impl LinuxInterruptHandle {
...
+
+    #[cfg(any(kvm, mshv))]
+    fn get_cancel_requested(&self) -> &AtomicU64 {
+        &self.cancel_requested
+    }
+
+    // #[cfg(any(kvm, mshv))]
+    // fn increment_generation_internal(&self) -> u64 {
+    //     self.increment_generation()
+    // }
```
jsturtevant, line 914: remove this?

jsturtevant (approved):

PR #971, opened 2025-10-23
Remove seccomp

### LINKED ISSUE #970: Remove seccomp features
The implementation to isolate each sandbox execution is not without fault, and it creates more problems than it solves.

The user of the Hyperlight library could potentially choose the best suited constraints for the sandbox based on the scenario they build the application for. 
The executing thread/process isolation is not the goal of this library

dblnz:
Closes #970 by removing all `seccomp` support in `hyperlight`.
This includes tests, documentation and just recipes.

### DIFF
--- a/docs/signal-handlers-development-notes.md
+++ b/docs/signal-handlers-development-notes.md
@@ -1,17 +1,9 @@
 # Signal Handling in Hyperlight
 
-Hyperlight registers custom signal handlers to intercept and manage specific signals, primarily `SIGSYS` and `SIGRTMIN`. Here's an overview of the registration process:
-- **Preserving Old Handlers**: When registering a new signal handler, Hyperlight first retrieves and stores the existing handler using `OnceCell`. This allows Hyperlight to delegate signals to the original handler if necessary.
+Hyperlight registers custom signal handlers to intercept and manage specific signals, primarily `SIGRTMIN`. Here's an overview of the registration process:
 -  **Custom Handlers**:
-  - **`SIGSYS` Handler**: Captures disallowed syscalls enforced by seccomp. If the signal originates from a hyperlight thread, Hyperlight logs the syscall details. Otherwise, it delegates the signal to the previously registered handler. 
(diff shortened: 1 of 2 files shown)


docs/signal-handlers-development-notes.md
```diff
@@ -1,10 +1,9 @@
 # Signal Handling in Hyperlight
 
-Hyperlight registers custom signal handlers to intercept and manage specific signals, primarily `SIGSYS` and `SIGRTMIN`. Here's an overview of the registration process:
+Hyperlight registers custom signal handlers to intercept and manage specific signals, primarily `SIGRTMIN`. Here's an overview of the registration process:
 - **Preserving Old Handlers**: When registering a new signal handler, Hyperlight first retrieves and stores the existing handler using `OnceCell`. This allows Hyperlight to delegate signals to the original handler if necessary.
```
simongdavies: I dont think we do this any longer , I think we only did it for the seccomp signal handler

docs/signal-handlers-development-notes.md
```diff
@@ -1,10 +1,9 @@
...
 -  **Custom Handlers**:
-  - **`SIGSYS` Handler**: Captures disallowed syscalls enforced by seccomp. If the signal originates from a hyperlight thread, Hyperlight logs the syscall details. Otherwise, it delegates the signal to the previously registered handler. 
-  - **`SIGRTMIN` Handler**: Utilized for inter-thread signaling, such as execution cancellation. Similar to SIGSYS, it distinguishes between application and non-hyperlight threads to determine how to handle the signal.
+  - **`SIGRTMIN` Handler**: Utilized for inter-thread signaling, such as execution cancellation. It distinguishes between application and non-hyperlight threads to determine how to handle the signal.
 - **Thread Differentiation**: Hyperlight uses thread-local storage (IS_HYPERLIGHT_THREAD) to identify whether the current thread is a hyperlight thread. This distinction ensures that signals are handled appropriately based on the thread's role.
```
simongdavies: I don't think we do this either for the kill signal because it might be a host thread, might be worth re writing this whole section to describe how the signal handler is used for killing sandbox

simongdavies (approved):
Minor issues with the docs otherwise LGTM.

PR #972, opened 2025-10-23
Crashdump on demand

simongdavies:
This PR adds a function to the crashdump feature to allow a crashdump to be produced by calling a function on a sandbox. 

This can be useful when debugging a process using gdb.

The updated doc has details on how to use this function

### DIFF
--- a/docs/how-to-debug-a-hyperlight-guest.md
+++ b/docs/how-to-debug-a-hyperlight-guest.md
@@ -226,14 +226,14 @@ To selectively disable this feature for a specific sandbox, you can set the `gue
     cfg.set_guest_core_dump(false); // Disable core dump for this sandbox
 ```
 
-### Creating a dump on demand
+## Creating a dump on demand
 
 You can also create a core dump of the current state of the guest on demand by calling the `generate_crashdump` method on the `InitializedMultiUseSandbox` instance. This can be useful for debugging issues in the guest that do not cause crashes (e.g., a guest function that does not return).
 
 This is only available when the `crashdump` feature is enabled and then only if the sandbox
 is also configured to allow core dumps (which is the default behavior).
 
-# Examples
+### Example
 
 Attach to your running process with gdb and call this function:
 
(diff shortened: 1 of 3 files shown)


docs/how-to-debug-a-hyperlight-guest.md
```diff
@@ -227,6 +226,39 @@ To selectively disable this feature for a specific sandbox, you can set the `gue
     cfg.set_guest_core_dump(false); // Disable core dump for this sandbox
 ```
 
+## Creating a dump on demand
+
+You can also create a core dump of the current state of the guest on demand by calling the `generate_crashdump` method on the `InitializedMultiUseSandbox` instance. This can be useful for debugging issues in the guest that do not cause crashes (e.g., a guest function that does not return).
```
dblnz: Nit
```suggestion
You can also create a core dump of the current state of the guest on demand by calling the `generate_crashdump` method on the `MultiUseSandbox` instance. This can be useful for debugging issues in the guest that do not cause crashes (e.g., a guest function that does not return).
```

dblnz (approved):
LGTM!
I don't think there is any security issue with allowing to generate a crashdump of a sandbox from the host, because if the host is compromised, then the access to guest memory is already a given, right?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -475,6 +475,47 @@ impl MultiUseSandbox {
...
+    /// (gdb) print sandbox
+    /// # Call the crashdump function 
+    /// call sandbox.generate_crashdump()
+    /// ```
+    /// The crashdump should be available in crash dump directory (see `HYPERLIGHT_CORE_DUMP_DIR` env var).
+    /// 
+    #[cfg(crashdump)]
+    #[instrument(err(Debug), skip_all, parent = Span::current())]
+
+    pub fn generate_crashdump(&self) -> Result<()> {
```
ludfjig, line 516: A test would be nice to make sure we don't break it in the future

ludfjig (dismissed):
LGTM. Would be nice to have a test that exercises it so we don't break it in the future

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: I think its not likely since this is just a thin layer over the crashdump capability which already has its own tests

PR #977, opened 2025-10-24
Fix guest tracing filter

### LINKED ISSUE #990: Fix guest tracing filtering
Currently, when tracing guests, the spans/events produced are not correctly filtered. Even though the `max_log_level` is correctly received by a guest, only the logs are filtered.

dblnz:
This closes #990.
Currently, when tracing guests, the max log level parameter provided to the guest is not used to filter out traces.

This PR adds the filtering capability for the guest tracing and also modifies some of the trace levels in the guest.
For that to work, a change in the `integration_tests.rs` was necessary to update the number of expected logs in `log_message` test

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
@@ -77,47 +77,64 @@ use crate::sandbox::trace::MemTraceInfo;
 use crate::sandbox::uninitialized::SandboxRuntimeConfig;
 
 /// Get the logging level filter to pass to the guest entrypoint
-fn get_max_log_level_filter() -> LevelFilter {
-    // Check to see if the RUST_LOG environment variable is set
-    // and if so, parse it to get the log_level for hyperlight_guest
-    // if that is not set get the log level for the hyperlight_host
-
+///
+/// The guest entrypoint uses this to determine the maximum log level to enable for the guest.
+/// The `RUST_LOG` environment variable is expected to be in the format of comma-separated
+/// key-value pairs, where the key is a log target (e.g., "hyperlight_guest_bin") and the value is
+/// a log level (e.g., "debug").
+///
+/// NOTE: This prioritizes the log level for the targets containing "hyperlight_guest" string, then
+/// "hyperlight_host", and then general log level. If none of these targets are found, it
+/// defaults to "error".
+fn get_max_log_level_filter(rust_log: String) -> LevelFilter {
     // This is done as the guest will produce logs based on the log level returned here
     // producing those logs is expensive and we don't want to do it if the host is not
     // going to process them
+    let level_str = rust_log
+        .split(',')
+        // Prioritize targets containing "hyperlight_guest"
+        .find_map(|part| {
+            let mut kv = part.splitn(2, '=');
+            match (kv.next(), kv.next()) {
+                (Some(k), Some(v)) if k.trim().contains("hyperlight_guest") => Some(v.trim()),
+                _ => None,
+            }
+        })
+        // Then check for "hyperlight_host"
+        .or_else(|| {
+            rust_log.split(',').find_map(|part| {
+                let mut kv = part.splitn(2, '=');
+                match (kv.next(), kv.next()) {
+                    (Some(k), Some(v)) if k.trim().contains("hyperlight_host") => Some(v.trim()),
+                    _ => None,
+                }
+            })
+        })
+        // Finally, check for general log level
+        .or_else(|| {
+            rust_log.split(',').find_map(|part| {
+                if part.contains("=") {
+                    None
+                } else {
+                    Some(part.trim())
+                }
+            })
+        })
+        .unwrap_or("");
 
-    let val = std::env::var("RUST_LOG").unwrap_or_default();
+    tracing::info!("Determined guest log level: {}", level_str);
 
-    let level = if val.contains("hyperlight_guest") {
-        val.split(',')
-            .find(|s| s.contains("hyperlight_guest"))
-            .unwrap_or("")
-            .split('=')
-            .nth(1)
-            .unwrap_or("")
-    } else if val.contains("hyperlight_host") {
-        val.split(',')
-            .find(|s| s.contains("hyperlight_host"))
-            .unwrap_or("")
-            .split('=')
-            .nth(1)
-            .unwrap_or("")
-    } else {
-        // look for a value string that does not contain "="
-        val.split(',').find(|s| !s.contains("=")).unwrap_or("")
-    };
-
-    tracing::info!("Determined guest log level: {}", level);
-    // Convert the log level string to a LevelFilter
     // If no value is found, default to Error
-    LevelFilter::from_str(level).unwrap_or(LevelFilter::ERROR)
+    LevelFilter::from_str(level_str).unwrap_or(LevelFilter::ERROR)
 }
 
 /// Converts a given [`Option<LevelFilter>`] to a `u64` value to be passed to the guest entrypoint
+/// If the provided filter is `None`, it uses the `RUST_LOG` environment variable to determine the
+/// maximum log level filter for the guest and converts it to a `u64` value.
 fn get_guest_log_filter(guest_max_log_level: Option<LevelFilter>) -> u64 {
     let guest_log_level_filter = match guest_max_log_level {
         Some(level) => level,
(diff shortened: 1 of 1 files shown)


src/hyperlight_guest_tracing/src/subscriber.rs
```diff
@@ -31,22 +31,42 @@ pub(crate) struct GuestSubscriber {
...
+/// Used to check if an event should be recorded based on the maximum log level
+fn convert_level_filter(filter: tracing::log::LevelFilter) -> tracing_core::LevelFilter {
+    match filter {
+        tracing::log::LevelFilter::Off => tracing_core::LevelFilter::OFF,
+        tracing::log::LevelFilter::Error => tracing_core::LevelFilter::ERROR,
+        tracing::log::LevelFilter::Warn => tracing_core::LevelFilter::WARN,
+        tracing::log::LevelFilter::Info => tracing_core::LevelFilter::INFO,
+        tracing::log::LevelFilter::Debug => tracing_core::LevelFilter::DEBUG,
+        tracing::log::LevelFilter::Trace => tracing_core::LevelFilter::TRACE,
+    }
```
ludfjig: Is it possible to tracing_core::LevelFilter everywhere instead of having this? (in entrypoint of hyperlight_guest_bin + whatever the host passes to the guest)?

jsturtevant:
@dblnz would you like to get this into the next release (we have it scheduled for next week (Feb 25th)

dblnz:
> @dblnz would you like to get this into the next release (we have it scheduled for next week (Feb 25th)

Yes, I'm rebasing it to include it in the release

src/hyperlight_guest_tracing/src/subscriber.rs
dblnz: I think it is doable, I just wanted to avoid breaking changes.
But I think we'll have breaking changes this release anyway

dblnz: It turns out it's more complicated that I had anticipated.
First of all, `tracing_core::LevelFIlter` cannot be cast to `u64` as an `entrypoint` argument. So, we still need to do some conversions.
Secondly, the `logger` inside the guest, still needs a `log::LevelFilter`.

The solution I've used is to have a separate `hyperlight_common::log_level::GuestLogFilter` that can convert from/to `u64` and from/to `LevelFilter` (both log/tracing ones).

I think this is a short term solution, because I am going to eliminate the `logger` inside the guest, to only use the `tracing` approach across the guests logs and traces. That will eliminate the need for `log::LevelFilter`.

Let me know your opinion on this.

jsturtevant (dismissed):

src/tests/rust_guests/simpleguest/src/main.rs
```diff
@@ -462,6 +463,8 @@ fn test_rust_malloc(code: i32) -> i32 {
 
 #[guest_function("LogMessage")]
 fn log_message(message: String, level: i32) {
+    let level =
+        LevelFilter::from(GuestLogFilter::try_from(level as u64).expect("Invalid log level"));
     let level = LevelFilter::iter().nth(level as usize).unwrap().to_level();
```
ludfjig: should we remove the second `level`?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -73,6 +75,52 @@ use crate::sandbox::trace::MemTraceInfo;
 #[cfg(crashdump)]
 use crate::sandbox::uninitialized::SandboxRuntimeConfig;
 
+/// Get the logging level filter to pass to the guest entrypoint
```
ludfjig, line 79: not sure whether I agree that this should go in this file but it's not a blocker

src/hyperlight_guest_tracing/Cargo.toml
```diff
@@ -12,7 +12,7 @@ description = """Provides the tracing functionality for the hyperlight guest."""
 [dependencies]
 hyperlight-common = { workspace = true, default-features = false }
 spin = "0.10.0"
-tracing = { version = "0.1.44", default-features = false, features = ["attributes"] }
```
ludfjig, line 15: did you mean to include this added feature `log`? https://docs.rs/tracing/latest/tracing/#emitting-log-records. I thought we wanted to switch to tracing

ludfjig (dismissed):

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
dblnz: I didn't see a reason to have it in a separate file considering it was only used here

src/tests/rust_guests/simpleguest/src/main.rs
dblnz: Oups, good catch

src/hyperlight_guest_tracing/Cargo.toml
dblnz: I'll remove it as I don't think it is necessary here, with the change of `LevelFilter`, but I think this will be required when merging the `logger` with the tracing logic because we still want to emit logs when tracing is not enabled (tracing subscriber is not set).

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -73,6 +75,52 @@ use crate::sandbox::trace::MemTraceInfo;
...
+fn get_max_log_level_filter() -> LevelFilter {
```
ludfjig: I know this is not related to this PR, but this code could be simplified. AI came up with this but I'm sure it can be further tweaked (not blocking, feel free to ignore if you wish)

```rust
fn get_max_log_level_filter() -> LevelFilter {
    let rust_log = std::env::var("RUST_LOG").unwrap_or_default();

    let level = rust_log
        .split(',')
        .find_map(|entry| {
            entry
                .strip_prefix("hyperlight_guest=")
                .or_else(|| entry.strip_prefix("hyperlight_host="))
                .or_else(|| {
                    // Fallback: global log level (no '=' present)
                    if !entry.contains('=') {
                        Some(entry)
                    } else {
                        None
                    }
                })
        })
        .unwrap_or("");

    tracing::info!("Determined guest log level: {}", level);

    LevelFilter::from_str(level).unwrap_or(LevelFilter::ERROR)
}
```

dblnz: This example is not quite the exact behaviour as before.
This matches `hyperlight_guest` and `hyperlight_host` exactly, while before it first checked for something that contains `hyperlight_guest`, afterwards for something that contains `hyperlight_host`. I've changed the logic and maintained the previous behaviour though.

dblnz: I am not sure which is best, because we can have `hyperlight_guest_bin` and missing `hyperlight_guest`, should we exactly check against `hyperlight_guest`?

Maybe we should prioritize `hyperlight_guest` and then if this is missing, take anything containing `hyperlight_guest`? What do you think?

ludfjig: do we currently support different logging levels per-module, or do we have a single level for everything inside the guest? If everything is at the same level, then I think just picking some name like `hyperlight-guest` and documenting it is sufficient. In my opinion, I think I would prefer having users explicilty set the log level for guests, rather than silently inherting hyperlight-host's logging level

I haven't thought much about this, but maybe it makes sense to use a different env var than RUST_LOG for logging inside the guests? Otherwise, say the crate `A` exists both in the guest, and host. How would a user express that they want only the guest's `A` or only the host's `A`?

In any case, I think it would make sense to eventually allow different modules inside the guest to have different logging levels (like RUST_LOG usually allows).

dblnz: We currently have a single level for everything inside the guest.
We currently have no way to specify a trace level for the guest but a different one for the host (for the same module).
I'll make a separate issue to tackle this.
We'd need to change the API for the `entrypoint` for this. Or think of a way to do this per guest call (that would be even better because going forward, taking a snapshot after creation means the trace level is set in stone and cannot be changed afterwards)

dblnz: Created https://github.com/hyperlight-dev/hyperlight/issues/1269

PR #991, opened 2025-10-27
Prevent MSRs from leaking across calls to `MultiuseSandbox::restore`

ludfjig:
The goal is to prevent guest state to persist across snapshot-restores.

- Kvm: directly disallows read/writes to msrs using KVM_X86_SET_MSR_FILTER. A denied access raises #GP in the guest, surfaces as GuestAborted, and poisons the sandbox.
- Hyperv (mshv+whp): The read/writes are allowed (no way to prevent it (intercept exists but doesn't cover all msrs)). Instead `restore` will reset set msrs using a hardcoded list of MSRs that are known to be writeable. The baseline values of these MSRs are captured on vm-creation. A limitation is that any MSR that is guest-writable and not in this list, has the potential to leak.

Future work:

- When scrub-parition hvcall is ready, we will switch to it, which is guaranteed reset all MSRs. (won't exist for kvm though, which is fine). 
- address https://github.com/hyperlight-dev/hyperlight/issues/1685
- batch calls to read/write msrs in vm creation, to improve perf via decreasing number of hv call. The reason it's all done 1 at at time right now it because on hyperv, read/write multiple msrs is either "all or nothing" so not trivial to batch, since it's expected that some should fail (and we need to know exactly which!). Could also be addressed by above item #1685 to perhaps avoid msr probing entirely, or at least make the number of probed msrs fewer

~~Will mark ready for review once KVM releases new version, which should include newly added `KVM_X86_SET_MSR_FILTER` vm ioctl that this PR depends on, see https://github.com/rust-vmm/kvm/pull/359~~


docs/msr.md
```diff
@@ -0,0 +1,304 @@
...
+
+```text
+guest-writable retained state => host-readable and host-writable state
+```
+
+Host-readable state need not be guest-writable. Extra reset entries are safe.
+`EFER`, `APIC_BASE`, `FS_BASE`, and `GS_BASE` belong to the special-register
+state.
+
+The two halves are established differently. Resolution checks the
```
syntactically: I don't follow this paragraph. Is "Resolution" a specific process? If so, what? What are the "two halves"? Why does "the host-wrtiable half" "hold by construction"?  What is mshv doing in here?

docs/msr.md
```diff
@@ -0,0 +1,304 @@
...
+Only `FS_BASE`, `GS_BASE`, and `KERNEL_GS_BASE` are written directly. Every
+other register below is intercepted and stored. Hyper-V stores the value on both
+Intel and AMD hosts.
+
+Interception alone is not the test. Hyper-V also intercepts registers the guest
+can only read, or that return a host-derived value. Those keep no
+guest-controlled state and are not reset. Each row below names the guest state
+that persists.
+
+| MSR (index) | Retained guest state |
```
syntactically: Do we need to regularly cross-check this table against updates to the hypervisor?

docs/msr.md
```diff
@@ -0,0 +1,304 @@
...
+
+## Snapshot validation
+
+Snapshot MSR entries are untrusted. A snapshot records the reset values and the
+capturing sandbox's allow list. `validate_snapshot` enforces two rules against
+the destination VM's reset set:
+
+* The snapshot's allow list must be a subset of the destination's. A
+  destination that allows at least as much accepts the snapshot.
+* Every supplied index must belong to the destination reset set.
```
syntactically: On mshv, are the "supplied index"es in the snapshot "every msr that mshv intercepts" or "just the ones specifically allowed"? If the former, isn't this overly-restrictive? If the latter, isn't this enforced by the previous bullet as well? I assume it's an invariant (independent of snapshotting) that every allowlisted msr is in the reset set.

docs/msr.md
```diff
@@ -0,0 +1,304 @@
...
+clears PASID during XSAVE reset before restoring MSRs. The performance-monitoring
+and last-branch registers remain inaccessible while perfmon is off.
+
+## Snapshot validation
+
+Snapshot MSR entries are untrusted. A snapshot records the reset values and the
+capturing sandbox's allow list. `validate_snapshot` enforces two rules against
+the destination VM's reset set:
+
+* The snapshot's allow list must be a subset of the destination's. A
```
syntactically: Why not reinitialise the allow list from the one in the snapshot (changing the filters if necessary)?

docs/msr.md
```diff
@@ -0,0 +1,304 @@
...
+KVM requires the index in `KVM_GET_MSR_INDEX_LIST` and a successful host read
+and write. MSHV and WHP require a named-register mapping and a successful host
+read.
+
+At most 16 distinct MSRs may be requested. KVM also limits the resulting
+contiguous filter groups to 16.
+
+## KVM
+
+KVM installs a deny filter over the full MSR space. Allowed indices form the
```
syntactically: You probably already checked this but I just wanted to call out the need to be sure that KVM doesn't have any special handling for certain things (I could imagine syscall vectors or something) that bypasses the filter.

docs/msr.md
```diff
@@ -0,0 +1,304 @@
...
+only guest `RDMSR` and `WRMSR` paths through that filter. A denied access exits
+to Hyperlight, injects `#GP`, and poisons the sandbox. The denied write stores
+no state.
+
+The KVM reset set contains the allow list plus `KERNEL_GS_BASE` and `TSC`.
+`KERNEL_GS_BASE` is required because `WRGSBASE` followed by `SWAPGS` changes it
+without `WRMSR`. `TSC` gives restore the same clock semantics on every backend.
+
+KVM does not filter x2APIC indices `0x800..=0x8FF`. Hyperlight keeps the APIC in
+xAPIC mode, where MSR access to that range raises `#GP`. `APIC_BASE` is not an
```
syntactically: How does this interact with the cfg(feature = 'hw-interrupts') support for hardware (vAPIC on x86 kvm I think) interrupt controllers?

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -69,6 +72,26 @@ use crate::sandbox::trace::MemTraceInfo;
...
+#[cfg(gdb)]
+type BoxedVm = Box<dyn DebuggableVm>;
+#[cfg(not(gdb))]
+type BoxedVm = Box<dyn VirtualMachine>;
+
+/// Determines the MTRR reset indices for an MSHV or WHP backend and validates
+/// its allow list. Both hosts lack an MSR filter, so the allow list adds reset
+/// state.
+#[cfg(any(mshv3, target_os = "windows"))]
+fn determine_reset_msrs(
```
syntactically: I'm not totally clear what this function is doing. It's named "determine_reset_msrs", but the thing it returns seems to only have the MTRR registers, not any other MSRs that are in the "reset set"? And it's then doing something with the allow list?

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -199,6 +236,50 @@ impl HyperlightVm {
         Ok(ret)
     }
 
+    /// Determines this VM's MSR reset set: the MSRs reset on restore and
```
syntactically: Why is this code here, rather than in MsrResetState?

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -199,6 +236,50 @@ impl HyperlightVm {
...
+    /// captured in snapshots. `required_mtrrs` is the VCNT-driven MTRR set
+    /// gathered at creation. `allowed` is the validated allow list.
+    fn capture_msr_reset_state(
+        vm: &BoxedVm,
+        required_mtrrs: Vec<u32>,
+        allowed: &[u32],
+    ) -> std::result::Result<MsrResetState, VmError> {
+        let core: Vec<u32> = match get_available_hypervisor() {
```
syntactically: There is a race condition in using `get_available_hypervisor()` twice instead of passing in the same one returned above. For example, on a host with both kvm and mshv available, we could end up in a situation where the code above set up a mshv backend, but someone removed the mshv device so that this code believed it to be kvm.

Also, since this code is hypervisor-specific, maybe it makes sense to put it into the HyperlightVm trait and have e.g. `vm.get_core_msrs()`?

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -199,6 +236,50 @@ impl HyperlightVm {
...
+        allowed.sort_unstable();
+        allowed.dedup();
+        Ok(MsrResetState::new(baseline, allowed))
+    }
+
+    /// Core stateful MSRs a filterless host can read.
+    ///
+    /// MTRRs come from the VCNT-driven required set, so they are excluded.
+    #[cfg(any(mshv3, target_os = "windows"))]
+    fn probe_core_reset_indices(vm: &BoxedVm) -> Vec<u32> {
```
syntactically: The name for this should differentiate these "filterless core" msrs from the "core" above that includes the filtered core msrs.

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -85,18 +108,27 @@ impl HyperlightVm {
         #[cfg(crashdump)] rt_cfg: SandboxRuntimeConfig,
         #[cfg(feature = "mem_profile")] trace_info: MemTraceInfo,
     ) -> std::result::Result<Self, CreateHyperlightVmError> {
-        #[cfg(gdb)]
-        type VmType = Box<dyn DebuggableVm>;
-        #[cfg(not(gdb))]
-        type VmType = Box<dyn VirtualMachine>;
-
-        let vm: VmType = match get_available_hypervisor() {
+        let (vm, required_mtrrs): (BoxedVm, Vec<u32>) = match get_available_hypervisor() {
```
syntactically: Why not have required_mtrrs (and validate_allowed_msrs?) be a method on HyperlightVm so that each hypervisor can implement it independently?

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -385,6 +458,51 @@ pub(crate) trait VirtualMachine: Debug + Send {
...
+/// Returns every guest-visible MTRR a filterless (MSHV/WHP) host must reset.
+#[cfg(all(target_arch = "x86_64", any(mshv3, target_os = "windows")))]
+pub(crate) fn mtrr_reset_indices(
+    vm: &dyn VirtualMachine,
+) -> std::result::Result<Vec<u32>, CreateVmError> {
+    let mtrr_cap = vm
+        .msrs(&[MSR_MTRR_CAP])
+        .map_err(CreateVmError::GetMtrrCap)?[0]
+        .value;
+    let indices = hyperv_mtrr_reset_indices(mtrr_cap)?;
```
syntactically, line 518: Why does this assume hyperv? What about linux builds with both kvm and mshv enabled?

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -667,6 +811,59 @@ impl VirtualMachine for WhpVm {
...
+            .map(|e| {
+                msr_to_whv_register_name(e.index)
+                    .map(|name| (name, Align16(WHV_REGISTER_VALUE { Reg64: e.value })))
+                    .ok_or(RegisterError::MsrsUnsupported)
+            })
+            .collect::<std::result::Result<_, _>>()?;
+        if regs.is_empty() {
+            return Ok(());
+        }
+        self.set_registers(&regs)
```
syntactically, line 837: Why is there a helper for WHvSetVirtualProcessorRegisters, but not WHvGetVirtualProcessorRegisters?

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -180,6 +182,17 @@ pub(super) struct OciSnapshotConfig {
     /// Special registers captured from the paused vCPU, restored
     /// verbatim when resuming the call.
     pub(super) sregs: CommonSpecialRegisters,
+    /// MSR reset state captured from the paused vCPU. `None` uses the
+    /// destination sandbox's reset set.
+    #[cfg(target_arch = "x86_64")]
+    #[serde(default)]
+    pub(super) msrs: Option<Vec<MsrEntry>>,
```
syntactically: Are the indices in this meant to be the same as the allow list? If yes, why combine them? If no, why are there extras that need to be persisted in the snapshot?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -300,6 +302,8 @@ where
         root_pt_gpas: &[u64],
         rsp_gva: u64,
         sregs: CommonSpecialRegisters,
+        #[cfg(target_arch = "x86_64")] msrs: Option<Vec<MsrEntry>>,
```
syntactically: Since there are already 2 things here, and more might arise, I wonder if it makes sense to extract as a separate generic "arch-specific info" structure with all the (de)serialisation code living in an arch-specific directory like the src/arch/<arch>/... we use in hyperlight-common.

Also, why are these `Option`s? At least for `allowed_msrs`. If it's to support loading from old snapshots, surely that should just initialise allowed_msrs to the empty list?

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -180,6 +182,17 @@ pub(super) struct OciSnapshotConfig {
...
+    /// Allow list of the sandbox that captured this snapshot. Present
+    /// exactly when `msrs` is, and checked against the destination allow
+    /// list on load.
+    #[cfg(target_arch = "x86_64")]
+    #[serde(default)]
+    pub(super) allowed_msrs: Option<Vec<u32>>,
```
syntactically: Also other code says that the Option here is meant to be the same as the Option above---so why not combine to `Option<(Vec<MsrEntry>, Vec<u32>)>` (or if needed for serde reasons, `Option<SnapshotMsrState>`?)

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -850,6 +854,15 @@ impl Snapshot {
         // 8. Build entrypoint + sregs back from the config.
         let entrypoint = NextAction::Call(cfg.entrypoint_addr);
 
+        // `msrs` and `allowed_msrs` travel together. A config with one but
```
syntactically: Why not have a single field (that's optional) that holds a dictionary with both?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -92,6 +94,18 @@ pub struct Snapshot {
...
 
+    /// MSR reset state captured from a running vCPU.
+    /// None for a pre-init snapshot or an old on-disk snapshot
+    /// predating MSR capture.
+    #[cfg(target_arch = "x86_64")]
+    msrs: Option<Vec<MsrEntry>>,
+
+    /// The allow list of the sandbox that captured this snapshot,
+    /// sorted and deduplicated. Present exactly when `msrs` is present.
+    /// A restore requires the destination allow list to be a superset.
```
syntactically: See comment above about changing the structure so that it more automatically ensures the "present exactly when `msrs` is present" invariant.

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
```diff
@@ -275,6 +347,21 @@ impl KvmVm {
             Ok(VcpuExit::IoOut(port, data)) => Ok(VmExit::IoOut(port, data.to_vec())),
             Ok(VcpuExit::MmioRead(addr, _)) => Ok(VmExit::MmioRead(addr)),
             Ok(VcpuExit::MmioWrite(addr, _)) => Ok(VmExit::MmioWrite(addr)),
+            // Reentering KVM_RUN completes the failed MSR exit and injects #GP.
```
syntactically: This comment seems out of place?

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
```diff
@@ -292,6 +379,150 @@ impl KvmVm {
...
+            );
+            return Err(CreateVmError::MsrFilterNotSupported);
+        }
+
+        // Every permitted guest write must have restorable host state.
+        for &msr in allowed {
+            self.validate_allowed_msr(msr)?;
+        }
+
+        // Tell KVM to exit to userspace on filtered MSR access.
```
syntactically: Why do we need to exit to userspace rather than just use the filters?

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
```diff
@@ -292,6 +379,150 @@ impl KvmVm {
...
+                let mut bytes = vec![0u8; count.div_ceil(8)];
+                for bit in 0..*count {
+                    bytes[bit / 8] |= 1 << (bit % 8);
+                }
+                bytes
+            })
+            .collect();
+
+        // Default deny requires at least one range.
+        static DENY_BITMAP: [u8; 1] = [0u8];
```
syntactically, line 381: Did you test that this is enough to block the KVM custom MSRs https://docs.kernel.org/virt/kvm/x86/msr.html as well as the normal MSRs?

Also, the KVM docs say that "MSRs that are loaded/stored via dedicated VMCS fields are not filtered as part of VM-Enter/VM-Exit emulation."---is that a problem? Do we need to do anything special to make sure the guest does not have vmx emulation (I assume this means nested virtualisation on vmx) enabled?

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
```diff
@@ -292,6 +379,150 @@ impl KvmVm {
...
+        if !hv.check_extension(Cap::X86UserSpaceMsr) || !hv.check_extension(Cap::X86MsrFilter) {
+            tracing::error!(
+                "KVM does not support KVM_CAP_X86_USER_SPACE_MSR or KVM_CAP_X86_MSR_FILTER."
+            );
+            return Err(CreateVmError::MsrFilterNotSupported);
+        }
+
+        // Every permitted guest write must have restorable host state.
+        for &msr in allowed {
+            self.validate_allowed_msr(msr)?;
```
syntactically: Roughly the same check is also made by the (architecture-independent but only called in the MSHV case, if I'm reading correctly? There's a lot of complexity here that feels like it could be disentangled a little bit by more clearly keeping what can be common, common, and the hv-specific stuff off in hv-specific files) code using `validate_allowed_msrs`; is there any commonality there that should be unified? These are called in different places from different call trees but are doing the same thing (and even doing it roughly the same way, by testing reading/writing each msr?)

src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
```diff
@@ -102,13 +183,13 @@ impl MshvVm {
...
         #[allow(unused_mut)]
-        let mut pr: mshv_create_partition_v2 = Default::default();
+        let mut pr = mshv_create_partition_v2::default();
         // Enable LAPIC for hw-interrupts — required for interrupt delivery
         // via request_virtual_interrupt.
         #[cfg(feature = "hw-interrupts")]
         {
             use mshv_bindings::MSHV_PT_BIT_LAPIC;
-            pr.pt_flags = 1u64 << MSHV_PT_BIT_LAPIC;
+            pr.pt_flags |= 1u64 << MSHV_PT_BIT_LAPIC;
```
syntactically: Is this change preserving any specific other flag?

src/hyperlight_host/src/hypervisor/regs/x86_64/msrs.rs
```diff
@@ -0,0 +1,483 @@
...
+pub(crate) const MSR_VIRT_SPEC_CTRL: u32 = 0xC001_011F;
+
+const HYPERV_VARIABLE_MTRR_COUNT: u8 = 16;
+
+// Every guest-writable retained value must be host-readable and host-writable.
+// EFER, APIC_BASE, FS_BASE, and GS_BASE are part of the special-register state.
+// PRED_CMD (0x49) and FLUSH_CMD (0x10B) are intentionally absent: they are
+// write-only commands with no retained state, so they cannot be reset and
+// therefore cannot be allowed.
+const MSR_TABLE: &[u32] = &[
```
syntactically: Does this need to include the x2APIC accesses for mshv guests which I presume can write APIC_BASE?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -304,6 +306,15 @@ impl MultiUseSandbox {
                         crate::hypervisor::hyperlight_vm::HyperlightVmError::Restore(e.into()),
                     )
                 })?;
+
+            // Restore captured MSR state.
+            #[cfg(target_arch = "x86_64")]
```
syntactically, line 312: It would be nice if most of this x86_64-specific stuff could be factored out into e.g. arch::handle_arch_specific_state.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -2720,6 +2755,1223 @@ mod tests {
...
+                .map(|entry| entry.index)
+                .collect();
+
+            for index in core_reset_indices() {
+                if !reset_indices.contains(&index) {
+                    assert_omitted_msr_does_not_retain(&mut sbox, index);
+                } else if kernel_gs_uses_instruction_side_effect && index == KERNEL_GS_BASE {
+                    // Direct WRMSR is denied. The dedicated SWAPGS test proves
+                    // the instruction-side mutation is restored.
+                } else if (index == 0x10 && !kernel_gs_uses_instruction_side_effect)
```
syntactically: Why is `kernel_gs_uses_instruction_side_effect` relevant to TSC?

Also, for the MSRs which have named constants in regs/x86_64/msrs.rs, it would be nice to use the constants in the tests?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -2720,6 +2755,1223 @@ mod tests {
...
+                    || matches!(index, 0xE7 | 0xE8)
+                {
+                    assert_guest_counter_is_writable_and_restored(&mut sbox, index);
+                } else if let Some(sentinel) = positive_write_sentinel(index) {
```
syntactically: What does "positive_write_sentinel" mean?

syntactically (commented):
This looks functionally quite good! I left a number of review comments inline, which are mostly about improving clarity / things whose function was not obvious to me.

Beyond what I've explicitly called out in the review comments, I think the high level structure of both the code and the documentation could be a little bit clearer. For the code, I felt like it was not always clear what was HV-specific and what wasn't, and there was just generally a lot of jumping around to follow e.g. the initialisation sequences and compare kvm vs mshv for things like allowlist filtering.  For the documentation, there are currently a lot of sections that are all level 2 headings, some of which are related to each other and some of which are not, and it feels like it jumps between different things a few times, making it harder to follow; probably there is some way to use heading structure to organise it a bit more nicely) could be a bit clearer.

docs/msr.md
ludfjig: My idea is that the allow list is a property/policy of a sandbox, not the snapshot, and restoring a snapshot should not change the MSR policy of the sandbox. To use a different compatible policy, you'd need to create a new sandbox with that policy. The allow list in the snapshot is only used to check that the new sandbox provides the permissions the snapshot expects.

docs/msr.md
ludfjig: yes, which is far from ideal, and that's why I'd like to switch to scrub hv call asap

docs/msr.md
ludfjig: I've rewritten msr.md, please take another look

docs/msr.md
ludfjig: The supplied indices are every MSR in the source reset set. Yes it is restrictive but without it I think captured MSR state could be lost. If the source snapshot contains an MSR that cannot be restored on the destination, silently omitting it could cause incorrect behavior if the guest depends on that state

docs/msr.md
ludfjig: yes I checked, everything except 0x800..=0x8ff is correctly denied by the filter filtered

docs/msr.md
ludfjig: Theya reunlreated, hw-interrups doesn't use x2apic. However, you remind me that snapshots are missing this state for when hw-interrupts are enabled. We should add lapic to serialized snapshot in a future pr so hw-interrupt works snapshot works correctly

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
ludfjig: I cleaned this up in latest commit

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
ludfjig: fixed, moved to method `msr_reset_indices`. mshv and hyperv share one impl though

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
ludfjig: fixed!

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
ludfjig: moved to method `msr_reset_indices`. Also technically `get_available_hypervisor()` is cached via OnceLock, but resolved by moving to backend regardless.

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
ludfjig: thanks, renamed `filterless_core_reset_candidates`

src/hyperlight_host/src/hypervisor/regs/x86_64/msrs.rs
ludfjig: No, x2apic requires opt-in partition feature, so any writes there faults. APIC_BASE is restored though

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
ludfjig: deleted

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
ludfjig: thanks, unified it!

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
ludfjig: I totally thought it was required for some reason and tied to the filter, whoopsie. Removed the exit to userspace

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
ludfjig: I confirmed we don't need to do anything, and that they are already denied by the filter. However for completeness I now changed the guests cpuid to also not allow nested

src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
ludfjig: reverted

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
ludfjig: good catch, moved to VirtualMachine

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
ludfjig: true. Added one

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
ludfjig: allowed_msrs is needed for the compatibilty check we have when restoring snapshots, so restoring a snapshot created with broader msr access into a more restricted destinatino is not allowed

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
ludfjig: fixed

src/hyperlight_host/src/sandbox/snapshot/mod.rs
ludfjig: fixed

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I agree, but I'll skip it for this pr if you don't mind

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: whops it's not. Fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: it was just some msrs with specific values that should be accepeted. Renamed

ludfjig (commented):
I think I addressed your comments. I moved the hyperv specific code to be called from respective backend (whp/mshv) but since it's duplicate code the code is still located outside the backends. Would you prefer the duplicate code appraoch inside each backend isntead?

docs/msr.md
syntactically: I thought the point of the allowed set here is that only the MSRs in the allowed set would be persisted across save/restore because the sandbox ought not to be relying on the behaviour of any MSR that it has not declared to need to use?

docs/msr.md
syntactically: That makes sense; I can see the argument for there being a benefit to the host being able to know that snapshots won't use more msrs than it has declared OK.

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
syntactically: > Also technically get_available_hypervisor() is cached via OnceLock, but resolved by moving to backend regardless.

Good point. I do think this is the kind of brittle extra non-local complexity that it is best to avoid without a very good reason, however.

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
syntactically: > that they are already denied by the filter

Did you try executing a vmenter/vmexit in the guest in order to do the store via the vmenter? If so, let's add a test with that code?

> I now changed the guests cpuid to also not allow nested

Did you add a test that this actually disables the vmexit emulation? (e.g. try to execute vmexit on an intel guest and see it fail)?

syntactically:
I didn't have a chance to actually read this yet. A couple of immediate reactions from the comment replies you wrote to me:

- I am still not sure I completely understand the semantics of the allow set; I left an inline reply on that
- I also had a couple of quick follow-up questions about what testing you did for the vmexit/vmenter emulation

> Would you prefer the duplicate code appraoch inside each backend isntead?

Duplicating the code does indeed not seem great. I might go for a mshv_shared module at the same level as kvm/mshv/hvf/whp that has any shared logic for the mshv/whp stuff?  This is getting into the weeds of taste now, so feel free to ignore.

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
```diff
@@ -163,19 +193,34 @@ impl KvmVm {
...
             .get_supported_cpuid(kvm_bindings::KVM_MAX_CPUID_ENTRIES)
             .map_err(|e| CreateVmError::InitializeVm(e.into()))?;
         for entry in kvm_cpuid.as_mut_slice().iter_mut() {
-            if entry.function
-                == CPUID_FUNCTION_PROCESSOR_CAPACITY_PARAMETERS_AND_EXTENDED_FEATURE_IDENTIFICATION
-            {
-                entry.eax &= !0xff;
-                entry.eax |= hyperlight_common::layout::SCRATCH_TOP_GPA.ilog2() + 1;
+            match entry.function {
+                // Hyperlight does not support nested Intel virtualization.
```
jsturtevant: I don't think we support x2APIC either based on 

```
Hyperlight keeps the APIC in xAPIC mode. Guest writes to `APIC_BASE` and the
  x2APIC range `0x800..=0x8FF` are denied. Snapshot restore also rejects an
  `APIC_BASE` value with the x2APIC enable bit set.
```
so we should not advertise it in the cpuid?
`CPUID.1:ECX[21]` from
AMD64 Architecture Programmer’s Manual, Volume 3, Appendix E, E.3.2 Function 1h - Processor and Processor Feature Identifiers

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -177,6 +319,9 @@ impl WhpVm {
             #[cfg(feature = "hw-interrupts")]
             Self::enable_lapic_emulation(p)?;
 
+            // Hyper-V permits MSR intercepts only for unimplemented indices.
+            // Implemented MSR isolation therefore depends on reset.
```
jsturtevant: this comment seems to be hanging out here but I am not sure why

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -159,6 +186,62 @@ impl SandboxConfiguration {
...
     }
 
+    /// Adds MSRs to the sandbox allow list.
+    ///
+    /// Adds MSRs to the state captured by
+    /// [`MultiUseSandbox::snapshot`](crate::MultiUseSandbox::snapshot) and restored
+    /// by [`MultiUseSandbox::restore`](crate::MultiUseSandbox::restore).
+    ///
+    /// If this method is not called, only the platform's standard MSR set is
+    /// captured and restored.
```
jsturtevant: I am not sure I fully understand this statement in relation to resetting MSRs.  I understood that we reset them by default, so will be saving any if this information isn't called?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: We will eventually need this on ARM as well?

syntactically: @jsturtevant In re system registers specifically: I think on kvm and hvf both we use delete+recreate vcpu right now, although I haven't tested the scalability limitations or benchmarked it. If we switch away from doing that, then we will need something similar (though a bit different).

However, I would be unsurprised in general if there was some other kind of architecture-specific state we wanted at some point as well. I also would just like to get all the x86 specific stuff off in its corner not muddying up the core abstractions, to the degree that that's possible.

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
ludfjig: yes and done

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
ludfjig: disabled it

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
ludfjig: removed

src/hyperlight_host/src/sandbox/config.rs
ludfjig: updated hopefully it's better now

jsturtevant (dismissed):

docs/msr.md
```diff
@@ -0,0 +1,254 @@
...
+
+MSHV and WHP read `IA32_MTRRCAP` during VM creation. The reset set includes
+`MTRR_DEF_TYPE`, every variable base and mask pair reported by `VCNT`, and all
+fixed MTRRs. Hyper-V accepts fixed-MTRR writes even when `MTRRCAP.FIX` is
+clear.
+
+VM creation fails if `VCNT` exceeds the supported maximum of 16 or required
+MTRRs cannot be read.
+
+### TSC
```
syntactically, line 181: I'm not totally convinced these are the right TSC semantics (see also the discussion on #1422). This PR is not the right venue to fix it since I understand KVM was doing the same thing anyway, but we should in the future make tsc handling special.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: Moved this to #1694

syntactically (approved):
This looks good now!

docs/msr.md
syntactically: Moved the LAPIC comment to #1695.

PR #992, opened 2025-10-28
DRAFT: Create crashdumps for active sandboxes if process is crashing and creating a crashdump

### LINKED ISSUE #966: Crashdump feature should enable guest dump creation if host process dumps
If a host process crash dump is created and the `crashdump` feature is enabled we currently do not create dumps for the guests, this makes scenarios where the host crashed or hangs difficult to diagnose

simongdavies:
This pull request introduces a new crash handler capability for Hyperlight, enabling automatic sandbox dump generation when the host crashes due to fatal signals (Linux) or exceptions (Windows). The system uses a global, lock-free registry for tracking sandboxes and installs platform-specific crash handlers. The implementation is gated behind a feature flag and is robust against initialization failures and recursive crashes.

Crash handler infrastructure:

* Added a new `crash_handler.rs` module that manages sandbox registration and crash dump generation, using a global lock-free registry (`DashMap`) and lazy initialization with `once_cell`. It ensures safe (for crash context) access to hypervisor pointers and robust initialization.
* Integrated new dependencies `once_cell` and `dashmap` in `Cargo.toml` to support lock-free global state and lazy initialization for the crash handler subsystem.

Linux-specific crash handling:

* Added a new `crash_handler/linux.rs` module that installs signal handlers for fatal signals (e.g., SIGSEGV, SIGABRT) and chains to previous handlers after generating sandbox dumps. It includes robust detection for whether core dumps are enabled on the system and ensures async-signal-safety is intentionally violated only during crash handling.

Safety and error handling:

* The crash handler code is careful to avoid unsafe behavior except during crash contexts, and includes mechanisms to prevent recursive crash handling, detect initialization failures, and fail gracefully when system configuration disables core dumps. 

Closes #966

### DIFF
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -51,6 +51,8 @@ metrics = "0.24.2"
 serde_json = "1.0"
 elfcore = "2.0"
 uuid = { version = "1.18.1", features = ["v4"] }
+once_cell = "1.20"
+dashmap = "6.1"
 
 [target.'cfg(windows)'.dependencies]
 windows = { version = "0.62", features = [
@@ -66,6 +68,8 @@ windows = { version = "0.62", features = [
     "Win32_System_Threading",
     "Win32_System_JobObjects",
     "Win32_System_SystemServices",
+    "Win32_System_Registry",
+    "Win32_System_Kernel",
 ] }
 windows-sys = { version = "0.61", features = ["Win32"] }
 windows-result = "0.4"
--- a/src/hyperlight_host/src/crash_handler.rs
+++ b/src/hyperlight_host/src/crash_handler.rs
@@ -1,0 +1,239 @@
+/*
+Copyright 2025 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Host process crash handler for generating sandbox dumps.
+//!
+//! This module provides crash detection and dump generation for sandboxes
+//! when the host process crashes due to unhandled signals (Linux) or
+//! exceptions (Windows).
+//!
+//! # Architecture
+//!
+//! - **Registry**: Global map of sandbox ID -> (hypervisor raw pointer, dump enabled flag)
+//! - **Linux**: Signal handlers via `sigaction()` for fatal signals
+//! - **Windows**: Vectored exception handler via `AddVectoredExceptionHandler()`
+//! - **Automatic**: Initialized on first sandbox registration
+//! - **Cleanup**: Entries removed on sandbox Drop
+//!
+//! # Usage
+//!
+//! The crash handler is automatically initialized when the first sandbox
+//! is created. No explicit setup is required. When the host process crashes,
+//! dumps are generated for all registered sandboxes that have `guest_core_dump`
+//! enabled in their runtime configuration.
+//!
+//! # Feature Flag
+//!
+//! This entire module requires the `crashdump` feature to be enabled.
+
+use std::sync::Mutex as StdMutex;
+use std::sync::atomic::{AtomicBool, Ordering};
+
+use dashmap::DashMap;
+use once_cell::sync::Lazy;
+
+use crate::hypervisor::Hypervisor;
+use crate::{Result, new_error};
+
+/// Entry in the sandbox registry.
+///
+/// Stores a raw pointer to the hypervisor (unsafe!).
+/// This is safe during crash handling because:
+/// 1. The sandbox owns the hypervisor and won't drop it while registered
+/// 2. During a crash, normal thread-safety doesn't matter
+/// 3. We only access these pointers during crash (process is dying anyway)
+struct SandboxEntry {
+    /// Raw pointer to the hypervisor (UNSAFE - only valid while sandbox is alive)
+    hypervisor_ptr: *const dyn Hypervisor,
+}
+
+// SAFETY: We only access these pointers during crash handling, when the process
+// is dying anyway and normal thread-safety rules don't apply
+unsafe impl Send for SandboxEntry {}
+unsafe impl Sync for SandboxEntry {}
+
+/// Global registry of active sandboxes.
+///
+/// Maps sandbox ID to hypervisor pointer. Uses DashMap for lock-free concurrent access.
+/// Entries are removed when sandboxes are dropped.
+static SANDBOX_REGISTRY: Lazy<DashMap<u64, SandboxEntry>> = Lazy::new(DashMap::new);
+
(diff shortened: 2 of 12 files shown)


src/hyperlight_host/src/crash_handler/linux.rs
```diff
@@ -0,0 +1,301 @@
...
+/// Calls unsafe libc::sigaction. Caller must ensure `prev_handler`
+/// points to valid static storage.
+unsafe fn install_handler(signal: c_int, prev_handler: *mut Option<sigaction>) {
+    // SAFETY: All operations in this block are guarded by unsafe blocks
+    // and are part of the signal handler installation process
+    unsafe {
+        let mut sa: sigaction = std::mem::zeroed();
+        sa.sa_sigaction = crash_signal_handler as usize;
+        sa.sa_flags = SA_SIGINFO | SA_RESTART;
+        libc::sigemptyset(&mut sa.sa_mask);
```
dblnz, line 178: I am not sure how this works, but is setting the signal handler before retrieving the previous the way it works?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -121,8 +121,22 @@ impl MultiUseSandbox {
         dispatch_ptr: RawPtr,
         #[cfg(gdb)] dbg_mem_access_fn: Arc<Mutex<SandboxMemoryManager<HostSharedMemory>>>,
     ) -> MultiUseSandbox {
+        let id = SANDBOX_ID_COUNTER.fetch_add(1, Ordering::Relaxed);
+
+        // Register with crash handler if dumps are enabled
+        #[cfg(feature = "crashdump")]
+        if vm.runtime_config().guest_core_dump
```
dblnz, line 128: Do we want this on by default?
Imagine running a huge number of sandboxes and something hangs, and you want to terminate it and it starts dumping everything.
I know this is not a production use case, but still. Let me know what you think.
Also, I am curious what happens if we're running 100+ sandboxes of big sizes 500MiB 😄

src/hyperlight_host/src/crash_handler/linux.rs
simongdavies: Yes when you set the handler you get back any previous handler which you can then chain to

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: Do you think that we should have a configuration option for this which turns it on or off globally and set it to false by default?

dblnz: I think for now we can leave it as is and introduce a configuration option if necessary.

src/hyperlight_host/Cargo.toml
```diff
@@ -51,6 +51,8 @@ metrics = "0.24.2"
 serde_json = "1.0"
 elfcore = "2.0"
 uuid = { version = "1.18.1", features = ["v4"] }
+once_cell = "1.20"
```
ludfjig, line 54: I think `once_cell` is obsolete, it's already been merged into std

src/hyperlight_host/src/crash_handler.rs
```diff
@@ -0,0 +1,239 @@
...
+//! The crash handler is automatically initialized when the first sandbox
+//! is created. No explicit setup is required. When the host process crashes,
+//! dumps are generated for all registered sandboxes that have `guest_core_dump`
+//! enabled in their runtime configuration.
+//!
+//! # Feature Flag
+//!
+//! This entire module requires the `crashdump` feature to be enabled.
+
+use std::sync::Mutex as StdMutex;
```
ludfjig, line 42: any reason for this alias?

src/hyperlight_host/src/crash_handler.rs
```diff
@@ -0,0 +1,239 @@
...
+            }
+        }
+    }
+
+    // Add entry to registry (lock-free with DashMap!)
+    let hypervisor_ptr = unsafe {
+        std::mem::transmute::<*const dyn Hypervisor, *const dyn Hypervisor>(
+            hypervisor as *const dyn Hypervisor,
+        )
+    };
```
ludfjig, line 150: This transmute is a no-op

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -121,8 +121,22 @@ impl MultiUseSandbox {
...
+            && let Err(e) = crate::crash_handler::register_sandbox(id, vm.as_ref())
+        {
+            tracing::error!(
```
ludfjig, line 131: Any reason this should not be a hard error?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -303,6 +303,9 @@ pub(crate) trait Hypervisor: Debug + Send {
     #[cfg(crashdump)]
     fn crashdump_context(&self) -> Result<Option<crashdump::CrashDumpContext>>;
 
+    #[cfg(crashdump)]
+    fn runtime_config(&self) -> &crate::sandbox::uninitialized::SandboxRuntimeConfig;
```
ludfjig, line 307: I don't think we should add this method to the Vm because it's not conceptually related to vms. It's already part of uninitialized sandbox so I believe you can already grab it from there without going through here

src/hyperlight_host/src/crash_handler/windows.rs
```diff
@@ -0,0 +1,267 @@
...
+    // This check might fail due to permissions, which is fine (fail-open)
+    unsafe {
+        let key_path: Vec<u16> = "SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting\0"
+            .encode_utf16()
+            .collect();
+        let mut hkey = std::mem::zeroed();
+
+        if RegOpenKeyExW(
+            HKEY_LOCAL_MACHINE,
+            PCWSTR(key_path.as_ptr()),
```
ludfjig, line 106: I __think__ there might a macro for this in the windows crate, something like `w!("SOFTWARE\\Microsoft\\Windows\\Windows Error Reporting")`

src/hyperlight_host/src/crash_handler/linux.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        let mut old_sa: sigaction = std::mem::zeroed();
+
+        if libc::sigaction(signal, &sa, &mut old_sa) == 0 {
```
ludfjig, line 182: Is it possible to use the `vmm_sys_util::signal::register_signal_handler` (which we already use somewhere else)? Or is it too limited

src/hyperlight_host/src/crash_handler/windows.rs
```diff
@@ -0,0 +1,267 @@
...
+};
+use windows::Win32::System::Registry::{
+    HKEY_LOCAL_MACHINE, KEY_READ, REG_DWORD, RegCloseKey, RegOpenKeyExW, RegQueryValueExW,
+};
+use windows::core::PCWSTR;
+
+/// Flag to prevent recursive crash handling.
+///
+/// If we crash while handling a crash, don't try to handle it again.
+static IN_CRASH_HANDLER: AtomicBool = AtomicBool::new(false);
```
ludfjig, line 69: This is never reset to false after being set to true. Is this on purpose?

simongdavies:
Closing this as the suggestion #966 is that we implement this feature by processing the host crashdump

PR #994, opened 2025-10-28
Fixes issues from merge and splits trait to expose only public method

simongdavies:
Merge issues with the mshv2 PR messed things up, we  also took the opportunity to refactor the `InterruptHandle` trait so we only exposed public methods

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
@@ -48,7 +48,7 @@ use super::gdb::{
     DebugCommChannel, DebugMemoryAccess, DebugMsg, DebugResponse, GuestDebug, MshvDebug,
     VcpuStopReason,
 };
-use super::{HyperlightExit, Hypervisor, InterruptHandle, LinuxInterruptHandle, VirtualCPU};
+use super::{HyperlightExit, Hypervisor, LinuxInterruptHandle, VirtualCPU};
 #[cfg(gdb)]
 use crate::HyperlightError;
 use crate::hypervisor::get_memory_access_violation;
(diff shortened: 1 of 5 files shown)


jsturtevant (approved):

ludfjig (dismissed):
LGTM, this is more or less what I was thinking too

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -647,11 +647,15 @@ impl Hypervisor for HypervLinuxDriver {
             .tid
             .store(unsafe { libc::pthread_self() as u64 }, Ordering::Release);
         // Note: if `InterruptHandle::kill()` is called while this thread is **here**
+        // Cast to internal trait for access to internal methods
+        let interrupt_handle_internal =
+            self.interrupt_handle.as_ref() as &dyn super::InterruptHandleInternal;
```
simongdavies, line 652: Yes, doesnt compile without.

ludfjig (approved):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -59,17 +59,17 @@ static SANDBOX_ID_COUNTER: AtomicU64 = AtomicU64::new(0);
...
 /// Only one guard can exist per interrupt handle at a time - attempting to create
 /// a second guard will return an error.
-struct CallActiveGuard {
-    interrupt_handle: Arc<dyn InterruptHandle>,
+struct CallActiveGuard<T: crate::hypervisor::InterruptHandleInternal + ?Sized> {
+    interrupt_handle: Arc<T>,
 }
 
-impl CallActiveGuard {
+impl<T: crate::hypervisor::InterruptHandleInternal + ?Sized> CallActiveGuard<T> {
```
dblnz, line 66: Is this change for performance reasons?

simongdavies: No, I think it can be simplified to take `dyn InterruptHandleInternal` rather than be generic

simongdavies: No I think it can be simplified to take/store `dyn InterruptHandleInternal`

PR #997, opened 2025-10-29
Fix possible UB with allocating guest traces on the host

dblnz:
I realized the way the space necessary for the guest traces is allocated on the host, may cause UB.
Here's a fix to address this before the next release.


dblnz:
> This looks highly unsafe, especially `from_raw_parts_mut`. I think we should be able to avoid all unsafe code here (except possibly transmuting the read bytes)

Let's sync offline and find the best solution

dblnz:
I'll reopen with a different implementation

Issue #964, opened 2025-10-21
Address guest tracing improvements


dblnz:
The current version using `opentelemetry` is an incremental step, after which additional `tunables` can be added to improve usability.

Improvements to be considered:
- The guest logging shall use the same logic as tracing for reporting back to host. [Maybe we can have spans/events emitted as logs in the guest](https://github.com/hyperlight-dev/hyperlight/pull/844#discussion_r2444557732).
- Change the tracing underlying storage to a circular buffer for traces/events to avoid context switches. This ensures the guest doesn't do host calls to further degrade performance. Trace/logs loss is an acceptable compromise
- Provide a mechanism on the guest to enable/disable the tracing support, configure the memory used and log level at runtime
- Consider changing to a heap allocated circular buffer, which is runtime configurable

simongdavies:
> Trace/logs loss is an acceptable compromise

I think this is only true if the host can make the decision that they are willing to suffer log/trace loss.

dblnz:
Having the configuration option on the guest side, the guest can say how much space it is acceptable to use for traces/logs buffering.
Making VMExits  to report traces has a performance penalty.

simongdavies:
> Having the configuration option on the guest side, the guest can say how much space it is acceptable to use for traces/logs buffering. Making VMExits to report traces has a performance penalty.

Understood, I am saying that we should not make the decision on if to trade off data loss for perf for a host, they should make that decision themselves.

PR #999, opened 2025-10-31
Serialize guest trace data using flatbuffers

dblnz:
This is part of #964.

## Description
This PR changes the way the guest reports the guest tracing data to the host from simply providing a pointer to the data in the guest to serializing it and providing a pointer to the serialized data in the guest.
The host side can the interpret that pointer as an offset in the shared memory, retrieve a copy of that buffer in the host, deserialize it and then parse the trace events to emit corresponding tracing spans/events on the host.

To do that, the following steps were necessary:
1. Define a flatbuffers schema to represent the guest trace data used on the guest to store the events
2. Implement wrappers on top of the generated flatbuffers Rust code to provide Rust types that the guest/host can use.
These wrappers make use of the flatbuffers code to provide `TryFrom` and `From` implementations for the types used on the guest and host
3. Change the tracing logic to use a single `GuestEvent` type that is an enum with the following variants:
    - `OpenSpan` - is created when a span is opened and stores all the corresponding info for that span including the TSC read on the guest
    - `CloseSpan` - is created when a span is closed and stores the TSC of when that happened
    - `LogEvent` - is created when a log is encountered
4. Update the guest state logic to keep track of the active span and use the new `GuestEvent` defined above. When the guest data vector reaches the capacity, the data is serialized and the pointer and length of that buffer are written in the registers of the CPU right before calling the `out` instruction that yields control to the host
5. Update the host context logic that:
    - retrieves the pointer and length from the vCPU's registers
    - gets exclusive access to the shared memory to retrieve the slice containing the serialized buffer
    - deserializes the buffer
    - parses the events to emit the corresponding trace events on the host (span/log)
    - At the same time, it keeps track of the active span in the guest so that all the new host spans created between this point and the time the guest starts executing again, are set as children of the guest span. This simulates a continuous call trace.

## Special notes
I would advise special consideration to the following areas:
- flatbuffer schema definition and serialization/deserialization
- guest memory access from the host and exclusive memory handling in that case
- guest buffer allocation and serialization of data

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -194,7 +194,8 @@ test-rust-crashdump target=default-target features="":
 # rust test for tracing
 test-rust-tracing target=default-target features="":
     # Run tests for the tracing guest and macro
-    {{ cargo-cmd }} test -p hyperlight-guest-tracing --profile={{ if target == "debug" { "dev" } else { target } }}  {{ target-triple-flag }}
+    {{ cargo-cmd }} test -p hyperlight-guest-tracing -F trace --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }}
+    {{ cargo-cmd }} test -p hyperlight-common -F trace_guest --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }}
 
     # Build the tracing guest to ensure it builds with the tracing feature
     just build-rust-guests {{ target }} trace_guest


src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -15,84 +15,61 @@ limitations under the License.
...
+/// TODO: Change these constants to be configurable at runtime by the guest
+/// Maybe use a weak symbol that the guest can override at link time?
+///
+/// Maximum number of spans that the guest can store
+const MAX_NO_OF_SPANS: usize = 10;
+/// Maximum number of events that the guest can store
+const MAX_NO_OF_EVENTS: usize = 10;
+
+/// Pre-calculated capacity for the events vector
+const EVENTS_VEC_CAPACITY: usize = 2 * MAX_NO_OF_SPANS + MAX_NO_OF_EVENTS;
```
dblnz: This will be changed later when another PR will allow guest configuration of this buffer.

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -143,21 +114,26 @@ impl<
...
             // Pop all remaining spans from the stack
+            let event = GuestEvent::CloseSpan {
+                id,
+                tsc: invariant_tsc::read_tsc(),
+            };
+            let events = &mut self.data.events;
+            // Should never fail because we flush when full
+            events.push(event);
+
+            if events.len() == EVENTS_VEC_CAPACITY {
```
dblnz:
```suggestion
            if events.len() >= EVENTS_VEC_CAPACITY {
```

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -167,11 +143,8 @@ impl<
 
     /// Returns information about the information needed by the host to read the spans.
     pub(crate) fn guest_trace_info(&mut self) -> TraceBatchInfo {
-        TraceBatchInfo {
-            guest_start_tsc: self.guest_start_tsc,
-            spans_ptr: &self.spans as *const _ as u64,
-            events_ptr: &self.events as *const _ as u64,
-        }
+        let serialized_data: Vec<u8> = (&self.data).try_into().unwrap();
```
dblnz: Avoid `unwrap`.
What to do in this case?

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -180,36 +153,31 @@ impl<
...
 
-        let spans = &mut self.spans;
+        let events = &mut self.data.events;
         // Should never fail because we flush when full
-        let _ = spans.push(span);
+        events.push(event);
 
         // In case the spans Vec is full, we need to report them to the host
-        if spans.len() == spans.capacity() {
+        if events.len() == EVENTS_VEC_CAPACITY {
```
dblnz:
```suggestion
if events.len() >= EVENTS_VEC_CAPACITY {
```

src/hyperlight_common/src/flatbuffer_wrappers/guest_trace_data.rs
```diff
@@ -0,0 +1,672 @@
...
+impl TryFrom<FbKeyValue<'_>> for KeyValue {
+    type Error = Error;
+
+    fn try_from(value: FbKeyValue<'_>) -> Result<Self, Self::Error> {
+        let key = value.key().to_string();
+        let value = value.value().to_string();
+
+        Ok(KeyValue { key, value })
+    }
+}
```
ludfjig: nit: if this is is infallible you can implement `From` rather than `TryFrom`

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -223,44 +192,62 @@ impl<
...
+                *id == s_id.into_u64()
+            } else {
+                false
+            }
+        }) {
+            let mut v = Vec::new();
+            values.record(&mut FieldsVisitor { out: &mut v });
+            fields.extend(v);
         }
     }
```
ludfjig, line 230: Would something like this work?

```rust
pub(crate) fn record(&mut self, s_id: &Id, values: &Record<'_>) {
    let target_id = s_id.into_u64();
    
    if let Some(GuestEvent::OpenSpan { fields, .. }) = self
        .data
        .events
        .iter_mut()
        .find(|e| matches!(e, GuestEvent::OpenSpan { id, .. } if *id == target_id))
    {
        let mut v = Vec::new();
        values.record(&mut FieldsVisitor { out: &mut v });
        fields.extend(v);
    }
}
```

ludfjig (dismissed):
LGTM. BTW what happens if guest is build with tracing feature, but host is not?

src/hyperlight_host/src/sandbox/trace/context.rs
```diff
@@ -28,69 +30,80 @@ use tracing_opentelemetry::OpenTelemetrySpanExt;
...
+                    "Failed to parse guest trace data from guest memory: {:?}",
                     e
                 )
             })?;
 
-        let events: Events = unsafe {
-            let raw = events.as_slice() as *const _ as *const Events;
-            raw.read_unaligned()
-        };
+            Ok::<GuestTraceData, HyperlightError>(gtd)
```
jsturtevant: nit: this is possibly contervisial but I prefer full names for variables something like `traceData`

dblnz:
> LGTM. BTW what happens if guest is build with tracing feature, but host is not?

The host crashes with:
```
called `Result::unwrap()` on an `Err` value: AnyhowError(Invalid OutBAction value: 104)
```

Because of https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_host/src/sandbox/outb.rs#L155
stems from https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_common/src/outb.rs#L117

I think it is ok to fail because unless the host expects this from the guest, it shall fail.

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -180,36 +154,31 @@ impl<
...
 
-        let spans = &mut self.spans;
+        let events = &mut self.data.events;
         // Should never fail because we flush when full
-        let _ = spans.push(span);
+        events.push(event);
 
         // In case the spans Vec is full, we need to report them to the host
-        if spans.len() == spans.capacity() {
+        if events.len() >= EVENTS_VEC_CAPACITY {
```
jsturtevant, line 181: is this spans or events? this function is called new_span and now we are doing devents?  Is there a difference between this and the function below?

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -272,12 +259,22 @@ impl<
...
+            tsc: invariant_tsc::read_tsc(),
+        };
+
+        // Should never fail because we flush when full
+        events.push(event);
+
+        if events.len() >= EVENTS_VEC_CAPACITY {
+            self.send_to_host();
         }
+
```
jsturtevant, line 260: Do we have a test case for filling the span somewhere that verifies this assertion?  Could the send_to_host fail in someway and cause the events to not be cleared fully?

jsturtevant (dismissed):

src/hyperlight_guest_tracing/src/state.rs
dblnz: Yeah, this might be confusing, but I called them all `events` and there's a `GuestEvent` that can be `OpenSpan`, `CloseSpan` and `LogEvent` variant.
We used to have separate `Vec`s for spans and log events, now we have a `Vec` of enums that can be either

src/hyperlight_guest_tracing/src/state.rs
dblnz: We don't have tests that exercise this part of the code as it is a `no-std` part that involves `asm` operations.
There's no `fail` way for the `send_to_host` method, as it serializes data, does a VM exit and it cleans the data.

dblnz: I have a plan to add tests/fuzzing, but it's in progress. I know that even though this isn't a `main` functionality for hyperlight, it shall have tests

src/hyperlight_common/src/flatbuffer_wrappers/guest_trace_data.rs
dblnz: Fixed. I changed to implement `From`

src/hyperlight_guest_tracing/src/state.rs
dblnz: Fixed. Much cleaner now! Thanks!

src/hyperlight_host/src/sandbox/trace/context.rs
```diff
@@ -212,101 +225,126 @@ impl TraceContext {
...
+                    if let Some(mut span) = self.guest_spans.remove(&id) {
+                        let end_ts =
+                            self.calculate_guest_time_relative_to_host(trace_batch.start_tsc, tsc)?;
+                        span.end_with_timestamp(end_ts);
+
+                        // The span ids should be closed in order
+                        if let Some(stack_id) = spans_stack.pop()
+                            && stack_id == id
+                        {
+                            tracing::warn!("Guest span with id {} closed out of order", id);
```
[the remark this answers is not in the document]
dblnz: Fixed

src/hyperlight_host/src/sandbox/trace/context.rs
dblnz: Changed

PR #1004, opened 2025-11-03
Allow windows more time to enter guest func for interrupt benchmark

ludfjig:
This benchmark is flaky on windows because sometimes the guest call has not started before issuing the `kill`, causing the benchmark to run forever. For an example where this happened, see [here](https://github.com/hyperlight-dev/hyperlight/actions/runs/19022845389/job/54320982829?pr=1003). The fix is just to sleep for a little longer, hopefully allowing the guest function thread to enter the guest function. This test started being flaky after https://github.com/hyperlight-dev/hyperlight/pull/959 was merged, since we can no longer cancel guest functions in advance.


jsturtevant:
LGTM, I triggered CI to re-run to check for flakes

PR #1007, opened 2025-11-05
Organize benchmarks and introduce 'size' dimension

### LINKED ISSUE #722: Restructure micro benchmarks
I think we should reorientate the benchmarks, we should have a set of tests that we perform on sandboxes of differing sizes , for each sized sandbox we should:
measure the create time 
measure the drop time
have guest calls with a varying number of parameters that does not reset the sandbox
have guest calls with a varying number of parameters that does reset the sandbox.
repeat the above where there are 1-n host function calls made by the guest function
measure the cost of snapshotting  when modifying varying proportions of the sandbox memory
measure the cost of restoring when modifying varying proportions of the sandbox memory.
measure the cost of calling with the largest parameters that sized sandbox can support.

_Originally posted by @simongdavies in https://github.com/hyperlight-dev/hyperlight/pull/683#discussion_r2222801703_

ludfjig:
Many of our benchmarks exhibit different performance characteristics depending on the size of the sandbox. This PR restructures the benchmark suite to run relevant benchmarks across four different heap sizes (default, 8MB, 64MB, 256MB), providing better visibility into how performance scales with memory allocation. This PR increases the number of benchmarks which will increase CI benchmark execution time proportionally.

Also slightly reorganizes the benchmarks into better categories. The multiple consecutive for-loops over sizes might seem weird at first, but it makes sure `cargo bench` runs the same benchmark with all sizes before moving on to the next benchmark. `cargo bench -- --list` now yields the following: 
```
sandboxes/create_uninitialized/default: benchmark
sandboxes/create_uninitialized/small: benchmark
sandboxes/create_uninitialized/medium: benchmark
sandboxes/create_uninitialized/large: benchmark
sandboxes/create_uninitialized_and_drop/default: benchmark
sandboxes/create_uninitialized_and_drop/small: benchmark
sandboxes/create_uninitialized_and_drop/medium: benchmark
sandboxes/create_uninitialized_and_drop/large: benchmark
sandboxes/create_initialized/default: benchmark
sandboxes/create_initialized/small: benchmark
sandboxes/create_initialized/medium: benchmark
sandboxes/create_initialized/large: benchmark
sandboxes/create_initialized_and_drop/default: benchmark
sandboxes/create_initialized_and_drop/small: benchmark
sandboxes/create_initialized_and_drop/medium: benchmark
sandboxes/create_initialized_and_drop/large: benchmark

guest_calls/call/default: benchmark
guest_calls/call/small: benchmark
guest_calls/call/medium: benchmark
guest_calls/call/large: benchmark
guest_calls/call_with_restore/default: benchmark
guest_calls/call_with_restore/small: benchmark
guest_calls/call_with_restore/medium: benchmark
guest_calls/call_with_restore/large: benchmark
guest_calls/call_with_host_function/default: benchmark
guest_calls/call_with_host_function/small: benchmark
guest_calls/call_with_host_function/medium: benchmark
guest_calls/call_with_host_function/large: benchmark
guest_calls/different_thread: benchmark
guest_calls/interrupt_latency: benchmark

snapshots/create/default: benchmark
snapshots/create/small: benchmark
snapshots/create/medium: benchmark
snapshots/create/large: benchmark
snapshots/restore/default: benchmark
snapshots/restore/small: benchmark
snapshots/restore/medium: benchmark
snapshots/restore/large: benchmark

guest_functions_with_large_parameters/guest_call_with_large_parameters: benchmark

function_call_serialization/serialize_function_call: benchmark
function_call_serialization/deserialize_function_call: benchmark

sample_workloads/24K_in_8K_out_c: benchmark
sample_workloads/24K_in_8K_out_rust: benchmark
```

Also adds the `snapshots/create` and `snapshots/restore` benchmarks, which are useful

Closes #722

### DIFF
--- a/src/hyperlight_host/benches/benchmarks.rs
+++ b/src/hyperlight_host/benches/benchmarks.rs
@@ -19,7 +19,9 @@ limitations under the License.
     reason = "This is a benchmark file, so using disallowed macros is fine here."
 )]
 
-use std::sync::Mutex;
+use std::sync::{Arc, Barrier, Mutex};
+use std::thread;
+use std::time::{Duration, Instant};
 
 use criterion::{Criterion, criterion_group, criterion_main};
 use flatbuffers::FlatBufferBuilder;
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/benches/benchmarks.rs
```diff
@@ -30,149 +32,344 @@ use hyperlight_host::GuestBinary;
...
+        group.bench_function(format!("call_with_host_function/{}", size.name()), |b| {
+            bench_guest_call_with_host_function(b, size)
+        });
+    }
+
+    group.bench_function("different_thread".to_string(), |b| {
+        bench_guest_call_different_thread(b, SandboxSize::Default)
+    });
+
+    group.bench_function("interrupt_latency".to_string(), |b| {
```
jsturtevant, line 298: Any reason not to run this with different sizes?  Is there no difference?

ludfjig: Yeah I don't think there's a difference so might as well save some ci cycles

andreiltd (approved):
This looks  great! 

I know this is not part of the changes but there are some tests that could benefit from using `iter_batched` to avoid measuring expensive setup. An example of this is `guest_call_with_large_parameters` that is cloning huge data in measure loop and could be rewritten as:

```rust
    b.iter_batched(
        || (large_vec.clone(), large_string.clone()),
        |(vec, string)| {
            sandbox.call::<()>("LargeParameters", (vec, string)).unwrap()
        },
        criterion::BatchSize::SmallInput,
    );
```

I think this is important because if the time spent on test setup dominates the total measured time (e.g 90%), then only a small fraction of the benchmark reflects the actual code we want to measure. This makes it nearly impossible to detect meaningful performance changes, because any improvements or regressions are drowned out by the setup overhead of cloning so we should pay extra attention if we want to maintain meaningful measuremens -- sorry for the offtopic :-)

ludfjig:
> This looks great!
> 
> I know this is not part of the changes but there are some tests that could benefit from using `iter_batched` to avoid measuring expensive setup. An example of this is `guest_call_with_large_parameters` that is cloning huge data in measure loop and could be rewritten as:
> 
> ```rust
>     b.iter_batched(
>         || (large_vec.clone(), large_string.clone()),
>         |(vec, string)| {
>             sandbox.call::<()>("LargeParameters", (vec, string)).unwrap()
>         },
>         criterion::BatchSize::SmallInput,
>     );
> ```
> 
> I think this is important because if the time spent on test setup dominates the total measured time (e.g 90%), then only a small fraction of the benchmark reflects the actual code we want to measure. This makes it hard to detect meaningful performance changes, because any improvements or regressions are drowned out by the setup overhead of cloning so we should pay extra attention if we want to maintain meaningful measuremens -- sorry for the offtopic :-)

You are totally right! We should fix this!

PR #1013, opened 2025-11-06
Deduplicate mem mgr and host functions from hv drivers

ludfjig:
Remove MemoryManager and FunctionRegistry fields from drivers, passing them as parameters instead. This eliminates duplicate ownership since they are already part of the sandbox itself, simplifies the ownership model and gets rid of a complete category of possible errors.

Easiest reviewed commit by commit

### DIFF
--- a/src/hyperlight_host/src/sandbox/outb.rs
+++ b/src/hyperlight_host/src/sandbox/outb.rs
@@ -25,9 +25,12 @@ use tracing::{Span, instrument};
 use tracing_log::format_trace;
 
 use super::host_funcs::FunctionRegistry;
-use crate::hypervisor::Hypervisor;
+#[cfg(feature = "mem_profile")]
+use crate::hypervisor::regs::CommonRegisters;
 use crate::mem::mgr::SandboxMemoryManager;
 use crate::mem::shared_mem::HostSharedMemory;
+#[cfg(feature = "mem_profile")]
+use crate::sandbox::trace::MemTraceInfo;
 use crate::{HyperlightError, Result, new_error};
 
 #[instrument(err(Debug), skip_all, parent = Span::current(), level="Trace")]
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -491,17 +461,9 @@ impl HyperlightVm {
...
-                                    if !mem_mgr.check_stack_guard()? {
-                                        break Err(HyperlightError::StackOverflow());
-                                    }
-                                }
-                                None => {
-                                    break Err(new_error!("Memory manager not initialized"));
-                                }
-                            }
-
+                            // if !self.mem_mgr.check_stack_guard()? {
```
ludfjig: todo: uncomment this

src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -147,10 +146,10 @@ fn outb_abort(mem_mgr: &mut SandboxMemoryManager<HostSharedMemory>, data: u32) -
 #[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
 pub(crate) fn handle_outb(
     mem_mgr: &mut SandboxMemoryManager<HostSharedMemory>,
-    host_funcs: Arc<Mutex<FunctionRegistry>>,
-    #[cfg(feature = "mem_profile")] _hv: &mut dyn Hypervisor,
+    host_funcs: &Arc<Mutex<FunctionRegistry>>,
     port: u16,
     data: u32,
+    _hv: &mut dyn Hypervisor,
```
jsturtevant: why is this needed?

jsturtevant: oh, i see, should this just be `hv` not `_hv`

ludfjig: Only the registers and TracingInfo is needed, I could pass that instead of the entire dyn Hypervisor

jsturtevant: nope, its only used when `mem_profile` feature is on.

jsturtevant: I think it would be cleaner to pass just the registers/tracing info

PR #1020, opened 2025-11-10
Closes #651 : Modify host_bindgen proc_macro to support selecting specific world

### LINKED ISSUE #651: Support selecting a world for host_bindgen
when using the `host_bindgen!` macro on a wit file that has multiple worlds in it, I end up not getting the expected bindings.  It would be nice to specify the world that the host should care about instead of having to create a separate wit file for the host wit just the world the host should implement.

```
hyperlight_component_macro::host_bindgen!();
```

example wit:

```
interface event {
   handle: func(x: u32) -> u32;
}

world handler {
    export event;
}

world runner {
    import event;
}
```

Currently I don't seem to be able to get bindings for the `handler` world. I am seeing bindings generated for the `Runner` Imports.

vshailesh:
fixes: #651
- adds support to select a specific world from a WIT file if multiple worlds are specified.

host_bindgen and guest_bindgen accepts two type of inputs now
1. old way - currently being used everywhere - directly giving path of the wasm compiled file.
2. path and world_name as input to choose a specific world name if you want that.

### DIFF
--- a/src/hyperlight_component_macro/src/lib.rs
+++ b/src/hyperlight_component_macro/src/lib.rs
@@ -50,6 +50,8 @@ limitations under the License.
 extern crate proc_macro;
 
 use hyperlight_component_util::*;
+use syn::parse::{Parse, ParseStream};
+use syn::{Ident, LitStr, Result, Token};
 
 /// Create host bindings for the wasm component type in the file
 /// passed in (or `$WIT_WORLD`, if nothing is passed in). This will
@@ -66,11 +68,14 @@ use hyperlight_component_util::*;
 #[proc_macro]
 pub fn host_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let _ = env_logger::try_init();
-    let path: Option<syn::LitStr> = syn::parse_macro_input!(input as Option<syn::LitStr>);
-    let path = path
-        .map(|x| x.value().into())
-        .unwrap_or_else(|| std::env::var_os("WIT_WORLD").unwrap());
-    util::read_wit_type_from_file(path, |kebab_name, ct| {
+    let parsed_bindgen_input = syn::parse_macro_input!(input as BindgenInputParams);
+    let path = match parsed_bindgen_input.path {
+        Some(path_env) => path_env.into_os_string(),
+        None => std::env::var_os("WIT_WORLD").unwrap(),
+    };
+    let world_name = parsed_bindgen_input.world_name;
+
+    util::read_wit_type_from_file(path, world_name, |kebab_name, ct| {
         let decls = emit::run_state(false, false, |s| {
             rtypes::emit_toplevel(s, &kebab_name, ct);
             host::emit_toplevel(s, &kebab_name, ct);
@@ -89,11 +94,14 @@ pub fn host_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
 #[proc_macro]
 pub fn guest_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
     let _ = env_logger::try_init();
-    let path: Option<syn::LitStr> = syn::parse_macro_input!(input as Option<syn::LitStr>);
-    let path = path
-        .map(|x| x.value().into())
-        .unwrap_or_else(|| std::env::var_os("WIT_WORLD").unwrap());
-    util::read_wit_type_from_file(path, |kebab_name, ct| {
+    let parsed_bindgen_input = syn::parse_macro_input!(input as BindgenInputParams);
+    let path = match parsed_bindgen_input.path {
+        Some(path_env) => path_env.into_os_string(),
+        None => std::env::var_os("WIT_WORLD").unwrap(),
+    };
+    let world_name = parsed_bindgen_input.world_name;
+
+    util::read_wit_type_from_file(path, world_name, |kebab_name, ct| {
         let decls = emit::run_state(true, false, |s| {
             // Emit type/trait definitions for all instances in the world
             rtypes::emit_toplevel(s, &kebab_name, ct);
@@ -106,4 +114,58 @@ pub fn guest_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream 
         // a debug temporary file be created.
         util::emit_decls(decls).into()
     })
+}
+
+#[derive(Debug)]
+struct BindgenInputParams {
+    world_name: Option<String>,
+    path: Option<std::path::PathBuf>,
+}
+
+impl Parse for BindgenInputParams {
+    fn parse(input: ParseStream) -> Result<Self> {
+        let mut path = None;
+        let mut world_name = None;
+
+        if input.peek(syn::token::Brace) {
+            let content;
+            syn::braced!(content in input);
+
+            // Parse key-value pairs inside the braces
+            while !content.is_empty() {
+                let key: Ident = content.parse()?;
+                content.parse::<Token![:]>()?;
+
+                match key.to_string().as_str() {
+                    "world_name" => {
+                        let value: LitStr = content.parse()?;
+                        world_name = Some(value.value());
+                    }
+                    "path" => {
+                        let value: LitStr = content.parse()?;
+                        path = Some(std::path::PathBuf::from(value.value()));
+                    }
+                    _ => {
+                        return Err(syn::Error::new(
+                            key.span(),
+                            format!(
+                                "unknown parameter '{}'; expected 'path' or 'world_name'",
+                                key
+                            ),
+                        ));
+                    }
+                }
+                // Parse optional comma
+                if content.peek(Token![,]) {
+                    content.parse::<Token![,]>()?;
+                }
+            }
+        } else {
+            let option_path_litstr = input.parse::<Option<syn::LitStr>>()?;
+            if let Some(concrete_path) = option_path_litstr {
+                path = Some(std::path::PathBuf::from(concrete_path.value()));
+            }
+        }
+        Ok(Self { world_name, path })
+    }
 }
--- a/src/hyperlight_component_util/src/component.rs
+++ b/src/hyperlight_component_util/src/component.rs
@@ -82,9 +82,10 @@ fn raw_type_export_type<'p, 'a, 'c>(
 /// export.
 pub fn read_component_single_exported_type<'a>(
     items: impl Iterator<Item = wasmparser::Result<Payload<'a>>>,
+    world_name: Option<String>,
 ) -> Component<'a> {
     let mut ctx = Ctx::new(None, false);
-    let mut last_idx = None;
+    let mut selected_type_idx = None;
     for x in items {
         match x {
             Ok(Version { num, encoding, .. }) => {
@@ -112,8 +113,18 @@ pub fn read_component_single_exported_type<'a>(
                         Err(_) => panic!("invalid export section"),
                         Ok(ce) => {
                             if ce.kind == ComponentExternalKind::Type {
-                                last_idx = Some(ctx.types.len());
                                 ctx.types.push(raw_type_export_type(&ctx, &ce).clone());
+
+                                // picks the world index if world_name is passed in the proc_macro
+                                // else picks the index of last type, exported by core module
+                                if let Some(world) = world_name.as_ref() {
+                                    let name = ce.name.0;
+                                    if name.eq_ignore_ascii_case(world) {
+                                        selected_type_idx = Some(ctx.types.len() - 1);
+                                    }
+                                } else {
+                                    selected_type_idx = Some(ctx.types.len() - 1);
+                                }
                             }
                         }
                     }
@@ -149,11 +160,15 @@ pub fn read_component_single_exported_type<'a>(
             _ => {}
         }
     }
-    match last_idx {
-        None => panic!("no exported type"),
+
+    match selected_type_idx {
         Some(n) => match ctx.types.into_iter().nth(n) {
             Some(Defined::Component(c)) => c,
             _ => panic!("final export is not component"),
+        },
+        None => match &world_name {
+            Some(name) => panic!("world '{}' not found in component", name),
+            None => panic!("no exported type"),
         },
     }
 }
--- a/src/hyperlight_component_util/src/util.rs
+++ b/src/hyperlight_component_util/src/util.rs
@@ -21,6 +21,7 @@ use crate::etypes;
 /// given filename, relative to the cargo manifest directory.
 pub fn read_wit_type_from_file<R, F: FnMut(String, &etypes::Component) -> R>(
     filename: impl AsRef<std::ffi::OsStr>,
+    world_name: Option<String>,
     mut cb: F,
 ) -> R {
     let path = std::path::Path::new(&filename);
@@ -30,7 +31,7 @@ pub fn read_wit_type_from_file<R, F: FnMut(String, &etypes::Component) -> R>(
 
     let bytes = std::fs::read(path).unwrap();
     let i = wasmparser::Parser::new(0).parse_all(&bytes);
-    let ct = crate::component::read_component_single_exported_type(i);
+    let ct = crate::component::read_component_single_exported_type(i, world_name);
 
     // because of the two-level encapsulation scheme, we need to look
     // for the single export of the component type that we just read
@@ -52,8 +53,7 @@ pub fn read_wit_type_from_file<R, F: FnMut(String, &etypes::Component) -> R>(
 
 /// Deal with `$HYPERLIGHT_COMPONENT_MACRO_DEBUG`: if it is present,
 /// save the given token stream (representing the result of
-/// macroexpansion) to the debug file and include that file instead of
-/// directly returning the given token stream.
+/// macroexpansion) to the debug file and then return the token stream
 pub fn emit_decls(decls: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
     if let Ok(dbg_out) = std::env::var("HYPERLIGHT_COMPONENT_MACRO_DEBUG") {
         if let Ok(file) = syn::parse2(decls.clone()) {
--- a/src/hyperlight_host/tests/wit_test.rs
+++ b/src/hyperlight_host/tests/wit_test.rs
@@ -426,3 +426,49 @@ mod wit_test {
         drop(guard);
     }
 }
+
+mod pick_world_bindings {
+    hyperlight_component_macro::host_bindgen!({path: "../tests/rust_guests/witguest/twoworlds.wasm", world_name: "firstworld"});
+}
+mod pick_world_binding_test {
+    use crate::pick_world_bindings::r#twoworlds::r#wit::r#first_import::RecFirstImport;
+
+    impl crate::pick_world_bindings::r#twoworlds::r#wit::r#first_import::RecFirstImport {
+        fn new() -> Self {
+            Self {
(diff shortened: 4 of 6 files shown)


vshailesh:
@ludfjig

src/hyperlight_component_macro/src/lib.rs
```diff
@@ -66,11 +68,19 @@ use hyperlight_component_util::*;
...
-        .unwrap_or_else(|| std::env::var_os("WIT_WORLD").unwrap());
-    util::read_wit_type_from_file(path, |kebab_name, ct| {
+    let parsed_bindgen_input = syn::parse_macro_input!(input as BindgenInputParams);
+    let path = parsed_bindgen_input.path.unwrap_or_else(|| {
+        let wit_world_env = std::env::var_os("WIT_WORLD");
+
+        if let Some(env) = wit_world_env {
+            std::path::PathBuf::from(env)
+        } else {
+            std::path::PathBuf::new()
```
ludfjig: This logic looks slightly different than before: we used to unwrap the `std::env::var_os("WIT_WORLD")`.

src/hyperlight_component_macro/src/lib.rs
```diff
@@ -89,11 +99,19 @@ pub fn host_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
```
ludfjig: Same here regarding unwrapping the env

vshailesh:
- output generated on local testing

```pub mod r#twoworlds {
    pub mod r#wit {
        pub mod r#first_import {
            #[derive(Debug)]
            pub struct RecFirstExport {
                pub r#key: alloc::string::String,
                pub r#value: alloc::string::String,
            }
        }
        pub trait FirstImport {}
        pub trait FirstworldExports<I: FirstworldImports + ::core::marker::Send> {}
        pub trait FirstworldImports {
            type FirstImport: super::super::r#twoworlds::r#wit::FirstImport;
            fn r#first_import(
                &mut self,
            ) -> impl ::core::borrow::BorrowMut<Self::FirstImport>;
        }
        pub trait Firstworld {
            type Exports<
                I: FirstworldImports + ::core::marker::Send,
            >: FirstworldExports<I>;
            fn instantiate<I: FirstworldImports + ::core::marker::Send + 'static>(
                self,
                imports: I,
            ) -> Self::Exports<I>;
        }
    }
}
pub(crate) struct FirstworldResources<I: r#twoworlds::r#wit::FirstworldImports> {
    resource0: (),
    _phantomI: ::core::marker::PhantomData<I>,
}
impl<I: r#twoworlds::r#wit::FirstworldImports> FirstworldResources<I> {
    fn new() -> Self {
        FirstworldResources {
            resource0: (),
            _phantomI: ::core::marker::PhantomData,
        }
    }
}
pub struct FirstworldSandbox<
    T: r#twoworlds::r#wit::FirstworldImports,
    S: ::hyperlight_host::sandbox::Callable,
> {
    pub(crate) sb: S,
    pub(crate) rt: ::std::sync::Arc<::std::sync::Mutex<FirstworldResources<T>>>,
}
pub(crate) fn register_host_functions<
    I: r#twoworlds::r#wit::FirstworldImports + ::std::marker::Send + 'static,
    S: ::hyperlight_host::func::Registerable,
>(sb: &mut S, i: I) -> ::std::sync::Arc<::std::sync::Mutex<FirstworldResources<I>>> {
    let rts = ::std::sync::Arc::new(::std::sync::Mutex::new(FirstworldResources::new()));
    let imports = ::std::sync::Arc::new(::std::sync::Mutex::new(i));
    rts
}
impl<
    I: r#twoworlds::r#wit::FirstworldImports + ::std::marker::Send,
    S: ::hyperlight_host::sandbox::Callable,
> r#twoworlds::r#wit::FirstworldExports<I> for FirstworldSandbox<I, S> {}
impl r#twoworlds::r#wit::Firstworld
for ::hyperlight_host::sandbox::UninitializedSandbox {
    type Exports<I: r#twoworlds::r#wit::FirstworldImports + ::std::marker::Send> = FirstworldSandbox<
        I,
        ::hyperlight_host::sandbox::initialized_multi_use::MultiUseSandbox,
    >;
    fn instantiate<
        I: r#twoworlds::r#wit::FirstworldImports + ::std::marker::Send + 'static,
    >(mut self, i: I) -> Self::Exports<I> {
        let rts = register_host_functions(&mut self, i);
        let sb = self.evolve().unwrap();
        FirstworldSandbox { sb, rt: rts }
    }
}

File: src/hyperlight_component_macro/src/lib.rs (hunk 1 above)
Comment (vshailesh) [reply src/hyperlight_component_macro/src/lib.rs]: unwrap leads to panic at runtime if the value is None, so I used this way, I can do unwrap if that is what we desire.

what if no path was given to macro and no OS env was set?

Comment (ludfjig) [reply src/hyperlight_component_macro/src/lib.rs]: if given no path and no os env, I think it would panic at compile time instead of runtime, which is OK I think

File: src/hyperlight_component_macro/src/lib.rs (hunk 2 above)
Comment (vshailesh) [reply src/hyperlight_component_macro/src/lib.rs]: changed this.

File: src/hyperlight_component_macro/src/lib.rs (hunk 1 above)
Comment (vshailesh) [reply src/hyperlight_component_macro/src/lib.rs]: updated this

File: src/hyperlight_component_util/src/component.rs [hunk 3]
```diff
@@ -112,8 +113,23 @@ pub fn read_component_single_exported_type<'a>(
...
-                                last_idx = Some(ctx.types.len());
                                 ctx.types.push(raw_type_export_type(&ctx, &ce).clone());
+
+                                // picks the world index if world is passed in the proc_macro
+                                // else picks the index of last type, exported by core module
+                                if world_name.is_some() {
+                                    let world = world_name.clone().unwrap();
+                                    match ce.name {
+                                        wasmparser::ComponentExportName(name) => {
+                                            if name.eq_ignore_ascii_case(&world) {
```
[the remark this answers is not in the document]
jsturtevant: I don't think this is correct. I didn't find anything in the spec that states this https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md.

src/hyperlight_component_util/src/component.rs
```diff
@@ -112,8 +113,23 @@ pub fn read_component_single_exported_type<'a>(
...
+                                                picks_one_world_or_last_idx =
+                                                    Some(ctx.types.len() - 1);
```
jsturtevant: Why is this minus one here? Previous it was just `last_idx = Some(ctx.types.len());`

src/hyperlight_component_util/src/component.rs
```diff
@@ -112,8 +113,23 @@ pub fn read_component_single_exported_type<'a>(
...
+                                            }
+                                        }
```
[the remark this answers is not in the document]
jsturtevant: `If let` would be preferable here i think

src/hyperlight_component_util/src/component.rs
vshailesh: The logic is still the same. 
Earlier it was like this 
arr = [a, b, c], 
arr.lenth() = 3, 
last_idx = 3, ---> points outside the array
then append happens -> arr = [a, b, c, d]

but the value at index is not yet inserted, the arr[3] will panic, if you try to use that index in future to extend some code here, and person reading the code is not careful to pick this quirk. Although with current code it would never panic, because we are not performing any operation on the array in between these two operation right now.

now I have just switched the order of this, append element to the array first then get the index of last element.
old_arr = [a, b, c]
appended_arr = [a, b, c, d]
last_idx = appended_arr.size() - 1 ------> this index will always resolve to a real value in the array.

So I just switched the order, I felt it would help in future modifications.

src/hyperlight_component_macro/src/lib.rs
```diff
@@ -107,3 +115,51 @@ pub fn guest_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream
...
+                    "world_name" => {
+                        let value: LitStr = content.parse()?;
+                        world_name = Some(value.value());
+                    }
+                    "path" => {
+                        let value: LitStr = content.parse()?;
+                        path = Some(std::path::PathBuf::from(value.value()));
+                    }
+                    _ => {
+                        return Err(syn::Error::new(key.span(), format!("Unknown key: {}", key)));
```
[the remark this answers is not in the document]
vshailesh: added

src/hyperlight_component_util/src/component.rs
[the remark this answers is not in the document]
vshailesh: thank you @jsturtevant

src/hyperlight_component_util/src/component.rs
vshailesh: I have tested this with original host_bindgen macro as well, it gives correct world, which is always last_world in that case.

src/hyperlight_component_util/src/component.rs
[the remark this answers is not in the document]
vshailesh: I think then this would be better 

``` 
let name = ce.name.0;
if name.eq_ignore_ascii_case(&world) {
    selected_type_idx = Some(ctx.types.len() - 1);
} 
```
because I get compiler warning with `if let` saying -> ``` irrefutable if let pattern
this pattern will always match, so the if let is useless
consider replacing the if let with a let ```

src/hyperlight_component_util/src/component.rs
```diff
@@ -112,8 +113,23 @@ pub fn read_component_single_exported_type<'a>(
...
+                                    }
+                                } else {
+                                    picks_one_world_or_last_idx = Some(ctx.types.len() - 1);
+                                }
```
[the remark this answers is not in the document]
vshailesh: I think I will have to check the name on each iteration, this may not avoidable

src/hyperlight_component_util/src/component.rs
```diff
@@ -149,7 +165,8 @@ pub fn read_component_single_exported_type<'a>(
             _ => {}
         }
     }
-    match last_idx {
+
+    match picks_one_world_or_last_idx {
         None => panic!("no exported type"),
```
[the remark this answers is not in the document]
vshailesh: thanks for suggestion.

src/hyperlight_component_macro/src/lib.rs
```diff
@@ -107,3 +115,51 @@ pub fn guest_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream
...
+                }
+            }
+        } else {
+            let option_path_litstr = input.parse::<Option<syn::LitStr>>()?;
+            if let Some(concrete_path) = option_path_litstr {
+                path = Some(std::path::PathBuf::from(concrete_path.value()));
+            }
+        }
+        Ok(Self { world_name, path })
+    }
```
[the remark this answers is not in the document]
vshailesh: I added this, thank you

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -426,3 +426,26 @@ mod wit_test {
...
+                r#value: String::from("dummyValue"),
+            })
+        }
+    }
+
+    #[test]
+    fn test_first_import_instance() {
+        let first_import = RecFirstImport::new();
+        assert!(first_import.is_some());
+    }
```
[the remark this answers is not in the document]
vshailesh: This is correct, because the struct itself is constructed by the host_bindgen, so if the struct instantiation is not working, it means the host_bindgen macro did not select the world we wanted.

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -426,3 +426,26 @@ mod wit_test {
...
+}
```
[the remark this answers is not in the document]
vshailesh: Point 1. Test added to verify that the second world is not present, when we are selecting firstworld.
Point 2. Not sure how to write tests for this one.
Point 3. This point is covered by the first older testcases.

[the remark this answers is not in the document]
vshailesh: Point 1 -> Changed return type to Self.
Point 2 -> Taken care by resolving point 1.

src/tests/rust_guests/witguest/two_worlds.wit
```diff
@@ -0,0 +1,23 @@
+package twoworlds:wit;
+
+world secondworld {
+  export second-export;
+}
+
+world firstworld {
+    import first-import;
```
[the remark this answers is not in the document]
vshailesh: Fixed indentation here

src/tests/rust_guests/witguest/two_worlds.wit
```diff
@@ -0,0 +1,23 @@
...
+  record rec-second-export {
+    key : string,
+    value: string
+  }
+}
+
+interface first-import {
+    record rec-first-import {
+        key: string,
+        value : string
```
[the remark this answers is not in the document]
vshailesh: Fixed formatting here.

src/hyperlight_component_util/src/component.rs
```diff
@@ -82,9 +82,10 @@ fn raw_type_export_type<'p, 'a, 'c>(
 /// export.
 pub fn read_component_single_exported_type<'a>(
     items: impl Iterator<Item = wasmparser::Result<Payload<'a>>>,
+    world_name: Option<String>,
 ) -> Component<'a> {
     let mut ctx = Ctx::new(None, false);
-    let mut last_idx = None;
+    let mut picks_one_world_or_last_idx = None;
```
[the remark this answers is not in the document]
vshailesh: changed overly verbose variable name

src/hyperlight_component_util/src/component.rs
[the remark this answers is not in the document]
vshailesh: Added borrow by reference for world_name variable.

src/hyperlight_component_util/src/component.rs
jsturtevant: Thanks, this makes more sense!

src/hyperlight_component_macro/src/lib.rs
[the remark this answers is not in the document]
jsturtevant: Looks like might need to push the latest changes?

src/hyperlight_component_macro/src/lib.rs
```diff
@@ -107,3 +115,63 @@ pub fn guest_bindgen(input: proc_macro::TokenStream) -> proc_macro::TokenStream
...
+            if let Some(concrete_path) = option_path_litstr {
+                path = Some(std::path::PathBuf::from(concrete_path.value()));
+            }
+        }
+        if world_name.is_some() && path.is_none() {
+            return Err(syn::Error::new(
+                proc_macro2::Span::call_site(),
+                "If `world_name` is provided in the brace syntax, `path` must also be provided.",
+            ));
+        }
```
jsturtevant: Why this?  In the usage above if the path is none then we get the path from the env, so this check shouldn't be needed?

```
let path = match parsed_bindgen_input.path {
        Some(path_env) => path_env.into_os_string(),
        None => std::env::var_os("WIT_WORLD").unwrap(),
    };
    let world_name = parsed_bindgen_input.world_name;
```

src/hyperlight_component_util/src/util.rs
```diff
@@ -30,7 +31,8 @@ pub fn read_wit_type_from_file<R, F: FnMut(String, &etypes::Component) -> R>(
 
     let bytes = std::fs::read(path).unwrap();
     let i = wasmparser::Parser::new(0).parse_all(&bytes);
-    let ct = crate::component::read_component_single_exported_type(i);
+    // let ct = crate::component::read_component_single_exported_type(i);
```
jsturtevant: nit: remove the comment

src/hyperlight_component_util/src/util.rs
```diff
@@ -62,7 +63,7 @@ pub fn emit_decls(decls: proc_macro2::TokenStream) -> proc_macro2::TokenStream {
             let decls = format!("{}", &decls);
             std::fs::write(&dbg_out, &decls).unwrap();
         }
-        quote::quote! { include!(#dbg_out); }
+        decls
```
jsturtevant: Why this change, I think this would mean that the code file wouldn't be the source which we would want for debugging support.

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -426,3 +426,27 @@ mod wit_test {
         drop(guard);
     }
 }
+
+mod pick_world_bindings {
```
jsturtevant, line 430: let's add a test that picks the second world in this file too.  this is testi gthe last world is selected but below the first world is the last in the file. It should look just like this test just for that other world

```

world secondworld {
  export second-export;
}

world firstworld {
  import first-import;
}
```

src/hyperlight_component_util/src/util.rs
vshailesh: will do

src/hyperlight_component_util/src/util.rs
vshailesh: this is the line that causes race condition, if you call host_bindgen macro multiple times in a file, with different .wasm files, then you will not get the correct generated code back, even though the macro actually generates the correct code.

This is happening because of the following reason
1. host_bindgen() ----------> macro called for the first time --> output gets saved to file.
2. host_bindgen() ----------> macro called for the n(th) time --> it may return the content of the file that was saved by previous run of the macro on some other .wasm file. 

I have tested this multiple times on local pc. 

This only happens when we have the `HYPERLIGHT_COMPONENT_MACRO_DEBUG` os env set, because it is writing to the file system. I think the rust std::fs::write() is either lazy or not synchronous. 

This happens because all the macro runs are manipulating the same file on filesystem.

if we need to access the file, then we should do std::fs::read separately, instead of relying on the return value of this function.

src/hyperlight_host/tests/wit_test.rs
vshailesh: oh, understood. I will add that.

src/hyperlight_component_macro/src/lib.rs
vshailesh: I just added this because copilot suggested it, and I didn't push back.

jsturtevant: We could probably add a test that validates this with the setup you have in the wit file but I would say this looks fine to remove

src/hyperlight_component_util/src/util.rs
jsturtevant: > This only happens when we have the HYPERLIGHT_COMPONENT_MACRO_DEBUG os env set, because it is writing to the file system. I think the rust std::fs::write() is either lazy or not synchronous.

Since this is a debug tool, lets revert this and we can make a note of it and do a follow up to fix this issue if we want.

src/hyperlight_host/tests/wit_test.rs
jsturtevant: I think that is fine. It will be a dev dependency.  it might be useful in some of our other macro tests.  @jprendes might have an opinion here

src/hyperlight_component_util/src/util.rs
jsturtevant: https://github.com/hyperlight-dev/hyperlight/issues/1107

src/hyperlight_host/tests/wit_test.rs
jsturtevant: Thanks, we can follow up with more tests if needed

PR #1021, opened 2025-11-11
Refactor 'run'

ludfjig:
Gives `Hypervisor::run` a default implementation and removes each specialized implementation from kvm/mshv/whp. Adds `run_vcpu` to just run vm for 1 single exit. The rest of the changes are just made to enable the above. Also refactors guest function cancellation slightly (simpler). Certain logic is introduced in this PR that will be removed in my next pr.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
@@ -356,7 +356,7 @@ impl HypervLinuxDriver {
         })?;
 
         let interrupt_handle: Arc<dyn InterruptHandleImpl> = Arc::new(LinuxInterruptHandle {
-            running: AtomicU64::new(0),
+            state: AtomicU64::new(0),
             #[cfg(gdb)]
             debug_interrupt: AtomicBool::new(false),
             #[cfg(all(
--- a/src/hyperlight_host/src/hypervisor/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/mod.rs
@@ -668,13 +668,12 @@ pub(super) struct LinuxInterruptHandle {
     /// Atomic value packing vcpu execution state.
     ///
     /// Bit layout:
-    /// - Bit 63: RUNNING_BIT - set when vcpu is actively running
-    /// - Bit 62: CANCEL_BIT - set when cancellation has been requested
-    /// - Bits 61-0: generation counter - tracks vcpu run iterations to prevent ABA problem
+    /// - Bit 1: RUNNING_BIT - set when vcpu is actively running
+    /// - Bit 0: CANCEL_BIT - set when cancellation has been requested
     ///
     /// CANCEL_BIT persists across vcpu exits/re-entries within a single `HyperlightVm::run()` call
     /// (e.g., during host function calls), but is cleared at the start of each new `HyperlightVm::run()` call.
-    running: AtomicU64,
+    state: AtomicU64,
 
     /// Thread ID where the vcpu is running.
     ///
@@ -699,62 +698,27 @@ pub(super) struct LinuxInterruptHandle {
 
 #[cfg(any(kvm, mshv3))]
 impl LinuxInterruptHandle {
-    const RUNNING_BIT: u64 = 1 << 63;
-    const CANCEL_BIT: u64 = 1 << 62;
-    const MAX_GENERATION: u64 = (1 << 62) - 1;
+    const RUNNING_BIT: u64 = 1 << 1;
+    const CANCEL_BIT: u64 = 1 << 0;
 
-    /// Sets the RUNNING_BIT and increments the generation counter.
-    ///
-    /// # Preserves
-    /// - CANCEL_BIT: The current value of CANCEL_BIT is preserved
-    ///
-    /// # Invariants Maintained
-    /// - Generation increments by 1 (wraps to 0 at MAX_GENERATION)
-    /// - RUNNING_BIT is set
-    /// - CANCEL_BIT remains unchanged
+    /// Get the running and cancel flags atomically.
     ///
     /// # Memory Ordering
-    /// Uses `Release` ordering to ensure that the `tid` store (which uses `Release`)
-    /// is visible to any thread that observes RUNNING_BIT=true via `Acquire` ordering.
-    /// This prevents the interrupt thread from reading a stale `tid` value.
-    #[expect(clippy::expect_used)]
-    fn set_running_and_increment_generation(&self) -> u64 {
-        self.running
-            .fetch_update(Ordering::Release, Ordering::Relaxed, |raw| {
-                let cancel_bit = raw & Self::CANCEL_BIT; // Preserve CANCEL_BIT
-                let generation = raw & Self::MAX_GENERATION;
-                let new_generation = if generation == Self::MAX_GENERATION {
-                    // restart generation from 0
-                    0
-                } else {
-                    generation + 1
-                };
-                // Set RUNNING_BIT, preserve CANCEL_BIT, increment generation
-                Some(Self::RUNNING_BIT | cancel_bit | new_generation)
-            })
-            .expect("Should never fail since we always return Some")
-    }
-
-    /// Get the running and cancel bits, return the previous value.
-    ///
-    /// # Memory Ordering
-    /// Uses `Acquire` ordering to synchronize with the `Release` in `set_running_and_increment_generation()`.
-    /// This ensures that when we observe RUNNING_BIT=true, we also see the correct `tid` value.
-    fn get_running_cancel_and_generation(&self) -> (bool, bool, u64) {
-        let raw = self.running.load(Ordering::Acquire);
-        let running = raw & Self::RUNNING_BIT != 0;
-        let cancel = raw & Self::CANCEL_BIT != 0;
-        let generation = raw & Self::MAX_GENERATION;
-        (running, cancel, generation)
+    /// Uses `Acquire` ordering to synchronize with the `Release` in `set_running()` and `kill()`.
+    /// This ensures that when we observe running=true, we also see the correct `tid` value.
+    fn get_running_and_cancel(&self) -> (bool, bool) {
+        let state = self.state.load(Ordering::Acquire);
+        let running = state & Self::RUNNING_BIT != 0;
+        let cancel = state & Self::CANCEL_BIT != 0;
+        (running, cancel)
     }
 
     fn send_signal(&self) -> bool {
         let signal_number = libc::SIGRTMIN() + self.sig_rt_min_offset as libc::c_int;
         let mut sent_signal = false;
-        let mut target_generation: Option<u64> = None;
 
         loop {
-            let (running, cancel, generation) = self.get_running_cancel_and_generation();
+            let (running, cancel) = self.get_running_and_cancel();
 
             // Check if we should continue sending signals
             // Exit if not running OR if neither cancel nor debug_interrupt is set
@@ -766,13 +730,6 @@ impl LinuxInterruptHandle {
 
             if !should_continue {
                 break;
-            }
-
-            match target_generation {
-                None => target_generation = Some(generation),
-                // prevent ABA problem
-                Some(expected) if expected != generation => break,
-                _ => {}
             }
 
             log::info!("Sending signal to kill vcpu thread...");
@@ -800,25 +757,28 @@ impl InterruptHandleImpl for LinuxInterruptHandle {
     }
 
     fn set_running(&self) {
-        self.set_running_and_increment_generation();
+        // Release ordering to ensure that the tid store (which uses Release)
+        // is visible to any thread that observes running=true via Acquire ordering.
+        // This prevents the interrupt thread from reading a stale tid value.
+        self.state.fetch_or(Self::RUNNING_BIT, Ordering::Release);
     }
 
     fn is_cancelled(&self) -> bool {
         // Acquire ordering to synchronize with the Release in kill()
-        // This ensures we see the CANCEL_BIT set by the interrupt thread
-        self.running.load(Ordering::Acquire) & Self::CANCEL_BIT != 0
+        // This ensures we see the cancel flag set by the interrupt thread
+        self.state.load(Ordering::Acquire) & Self::CANCEL_BIT != 0
     }
 
     fn clear_cancel(&self) {
-        // Relaxed is sufficient here - we're the only thread that clears this bit
-        // at the start of run(), and there's no data race on the clear operation itself
-        self.running.fetch_and(!Self::CANCEL_BIT, Ordering::Relaxed);
+        // Release ordering to ensure that any operations from the previous run()
+        // are visible to other threads. While this is typically called by the vcpu thread
+        // at the start of run(), the VM itself can move between threads across guest calls.
+        self.state.fetch_and(!Self::CANCEL_BIT, Ordering::Release);
     }
 
     fn clear_running(&self) {
-        // Release ordering to ensure all vcpu operations are visible before clearing RUNNING_BIT
-        self.running
-            .fetch_and(!Self::RUNNING_BIT, Ordering::Release);
+        // Release ordering to ensure all vcpu operations are visible before clearing running
+        self.state.fetch_and(!Self::RUNNING_BIT, Ordering::Release);
     }
 
     fn is_debug_interrupted(&self) -> bool {
@@ -838,7 +798,9 @@ impl InterruptHandleImpl for LinuxInterruptHandle {
     }
 
     fn set_dropped(&self) {
-        self.dropped.store(true, Ordering::Relaxed);
+        // Release ordering to ensure all VM cleanup operations are visible
+        // to any thread that checks dropped() via Acquire
+        self.dropped.store(true, Ordering::Release);
     }
 }
 
@@ -847,7 +809,7 @@ impl InterruptHandle for LinuxInterruptHandle {
     fn kill(&self) -> bool {
         // Release ordering ensures that any writes before kill() are visible to the vcpu thread
         // when it checks is_cancelled() with Acquire ordering
-        self.running.fetch_or(Self::CANCEL_BIT, Ordering::Release);
+        self.state.fetch_or(Self::CANCEL_BIT, Ordering::Release);
 
         // Send signals to interrupt the vcpu if it's currently running
         self.send_signal()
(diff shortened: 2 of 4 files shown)


src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+                    // Handle debug event (breakpoints)
+                    let stop_reason = arch::vcpu_stop_reason(rip_gva, _entrypoint, dr6, exception)?;
+
+                    if stop_reason == VcpuStopReason::EntryPointBp {
+                        // TODO in next PR: make sure to remove hw breakpoint here
+                        // In case the hw breakpoint is the entry point, remove it to
+                        // avoid hanging here as gdb does not remove breakpoints it
+                        // has not set.
+                        // Gdb expects the target to be stopped when connected.
+                        // self.remove_hw_breakpoint(vcpu_fd, entrypoint)?;
```
ludfjig, line 279: note this `self.remove_hw_breakpoint` has been moved to another place (handle_debug in each of the 3 drivers), it will be brought back here in next PR.

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -560,273 +544,85 @@ impl Hypervisor for HypervLinuxDriver {
...
-                    let instruction_length = io_message.header.instruction_length() as u64;
-                    crate::debug!("mshv IO Details : \nPort : {}\n{:#?}", port_number, &self);
-                    HyperlightExit::IoOut(
-                        port_number,
-                        rax.to_le_bytes().to_vec(),
-                        rip,
-                        instruction_length,
-                    )
+                    // mshv, unlike kvm, does not automatically increment RIP
+                    self.vcpu_fd.set_reg(&[hv_register_assoc {
```
dblnz, line 580: I don't see where this increment was done before, or is this a fix?

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
```diff
@@ -560,273 +544,85 @@ impl Hypervisor for HypervLinuxDriver {
...
+                UNMAPPED_GPA => {
                     let mimo_message = m.to_memory_info().map_err(mshv_ioctls::MshvError::from)?;
                     let addr = mimo_message.guest_physical_address;
-                    crate::debug!(
-                        "mshv MMIO unmapped GPA -Details: Address: {} \n {:#?}",
-                        addr,
-                        &self
-                    );
-                    HyperlightExit::Mmio(addr)
+                    match MemoryRegionFlags::try_from(mimo_message)? {
```
dblnz, line 593: I like this separation for Read/Write

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -482,158 +460,53 @@ impl Hypervisor for HypervWindowsDriver {
...
-                }
-            }
-            // HvRunVpExitReasonX64Halt
-            WHV_RUN_VP_EXIT_REASON(8i32) => {
-                debug!("HyperV Halt Details :\n {:#?}", &self);
-                HyperlightExit::Halt()
-            }
-            // WHvRunVpExitReasonMemoryAccess
-            WHV_RUN_VP_EXIT_REASON(1i32) => {
+                let rip = exit_context.VpContext.Rip + instruction_length as u64;
```
dblnz, line 494: On windows do we still need to increment RIP?

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
```diff
@@ -482,158 +460,53 @@ impl Hypervisor for HypervWindowsDriver {
...
+            WHvRunVirtualProcessor(
+                self.processor.get_partition_hdl(),
+                0,
+                &mut exit_context as *mut _ as *mut c_void,
+                std::mem::size_of::<WHV_RUN_VP_EXIT_CONTEXT>() as u32,
+            )?;
         }
 
-        #[cfg(gdb)]
-        let debug_interrupt = self
```
dblnz, line 601: Note: I don't know if this has been moved to somewhere else, I am still reviewing, but I'll leave this comment here for now.

I remember this was needed to be able to make the difference between the vCPU stopping because of a cancel request or a debug reason.
This is very much needed, otherwise we might end up letting the user "debug" when a vCPU has been cancelled, and continue execution afterwards.

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+            //    - CANCEL_BIT will be set and we will return an early VmExit::Cancelled()
+            interrupt_handle.set_tid();
+            interrupt_handle.set_running();
+
+            let exit_reason =
+                if interrupt_handle.is_cancelled() || interrupt_handle.is_debug_interrupted() {
+                    Ok(VmExit::Cancelled())
+                } else {
+                    #[cfg(feature = "trace_guest")]
+                    tc.setup_guest_trace(tracing_opentelemetry::OpenTelemetrySpanExt::context(
```
dblnz, line 216: Even though this looks a lot cleaner, the `setup_guest_trace` does a couple of important stuff that are supposed to be run right before entering the `guest`, or at least in the function that calls `run` on the `VcpuFd`.

- It stores time for when the guest is started, meant to calculate the TSC frequency, but this alone is not a hard requirement to run right before entering the guest.
- Most importantly, it stores the current span context as a parent for all subsequent guest spans. With this change, there is an additional span created by the call to `self.run_vcpu` which should be the parent of all the guest spans, but it will not be because the current context (`run`) is set as the parent.

We'll end up with a span hierarchy something like:
```
- run
  - guest_1
    - guest_2
  - guest_3
- run_vcpu
```
instead of
```
- run
  - run_vcpu
    - guest_1
      - guest_2
    - guest_3
```

Of course, this applies when the `level` matches the one for the host spans, in this case `Trace`

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+            // ===== KILL() TIMING POINT 4: Before capturing cancel_requested =====
+            // If kill() is called and ran to completion BEFORE this line executes:
+            //    - CANCEL_BIT will be set
+            //    - Signals may still be sent (RUNNING_BIT=true) but are harmless no-ops
+            //    - kill() will have no effect on this iteration, but CANCEL_BIT will persist
+            //    - If the loop continues (e.g., for a host call), the next iteration will be cancelled
+            //    - Stale signals from before clear_running() may arrive and kick future iterations,
+            //      but will be filtered out by the cancel_requested check below (and retried).
+            let cancel_requested = interrupt_handle.is_cancelled();
+            let debug_interrupted = interrupt_handle.is_debug_interrupted();
```
dblnz, line 252: Found it, it has been moved here 😄

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -314,201 +557,93 @@ pub(crate) trait Hypervisor: Debug + Send {
...
-                    break;
-                }
-                Ok(HyperlightExit::IoOut(port, data, rip, instruction_length)) => {
-                    hv.handle_io(port, data, rip, instruction_length)?
-                }
-                Ok(HyperlightExit::Mmio(addr)) => {
-                    #[cfg(crashdump)]
-                    crashdump::generate_crashdump(hv)?;
+/// A trait for platform-specific interrupt handle implementation details
+pub(crate) trait InterruptHandleImpl: InterruptHandle {
```
dblnz, line 608: Maybe I don't understand it well, but what does this second trait help with?
I see both Linus and Windows variants implement it and `InterruptHandle` at the same time, no default implementation. So it is as if it is a single trait split into two smaller traits, right?

Would this help with next PRs?

src/hyperlight_host/src/hypervisor/mod.rs
dblnz: I've been looking for an alternative to this when I was working on tracing, but getting the context for a function can only be done with that `tracing::Span::current()`, which yields the Span for the current function (the last span created).

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: The trait is split in order to limit what is public

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Oopps, good catch. Do you think it's sufficient to ensure `run_vcpu` does not emit a span (and i guess is part of caller `run`?

dblnz: Yeah, that was a solution I thought about also when implementing, if you choose this solution, just add a comment to emphasize this.

src/hyperlight_host/src/hypervisor/mod.rs
dblnz: Got it. It makes sense

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Fixed!

src/hyperlight_host/src/hypervisor/hyperv_linux.rs
ludfjig: It was done before, but in `handle_io`

src/hyperlight_host/src/hypervisor/hyperv_windows.rs
ludfjig: Yes we are doing it still. It was done before, but in `handle_io`

src/hyperlight_host/src/hypervisor/gdb/arch.rs
```diff
@@ -51,61 +50,50 @@ pub(crate) const DR6_HW_BP_FLAGS_MASK: u64 = 0x0F << DR6_HW_BP_FLAGS_POS;
...
-    single_step: bool,
     rip: u64,
-    dr6: u64,
     entrypoint: u64,
+    dr6: u64,
     exception: u32,
-    hw_breakpoints: &[u64],
-    sw_breakpoints: &HashMap<u64, [u8; SW_BP_SIZE]>,
-) -> VcpuStopReason {
+) -> Result<VcpuStopReason> {
```
simongdavies: Why did the return type change to `Result<VcpuStopReason>`? I don't see anywhere that there is anything other than `Ok(VcpuStopReason::*)` returned?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -97,26 +100,31 @@ cfg_if::cfg_if! {
...
+    /// The vCPU tried to write to the given (unmapped) addr
+    MmioWrite(u64),
     /// The vCPU execution has been cancelled
     Cancelled(),
     /// The vCPU has exited for a reason that is not handled by Hyperlight
     Unknown(String),
-    /// The operation should be retried
-    /// On Linux this can happen where a call to run the CPU can return EAGAIN
-    /// On Windows the platform could cause a cancelation of the VM run
+    /// The operation should be retried, for example this can happen on Linux where a call to run the CPU can return EAGAIN
```
simongdavies, line 116: Dont understand why we need to do this? Cant we exclude this from whatever match is using it on Windows? Also I dont understand why we are not seeing this on Windows since IIRC the platform itself can choose to cancel an execution which we then captured and retried, what happened to this?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+                    tc.setup_guest_trace(tracing_opentelemetry::OpenTelemetrySpanExt::context(
+                        &tracing::Span::current(),
+                    ));
 
-    /// Get InterruptHandle to underlying VM (returns internal trait)
-    fn interrupt_handle(&self) -> Arc<dyn InterruptHandleInternal>;
+                    // ===== KILL() TIMING POINT 3: Before calling run_vcpu() =====
+                    // If kill() is called and ran to completion BEFORE this line executes:
+                    //    - CANCEL_BIT will be set, but it's too late to prevent entering the guest this iteration
+                    //    - Signals will interrupt the guest (RUNNING_BIT=true), causing VmExit::Cancelled()
```
simongdavies, line 223: This isnt quite true since Windows does not use signals?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+                    //    - If the guest completes before any signals arrive, kill() may have no effect
+                    //      - If there are more iterations to do (IO/host func, etc.), the next iteration will be cancelled
```
simongdavies, line 225: What happens if there are not more iterations to do , is it possible to lie about completion or cancel the next use of this Sandbox?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+                        // Is this the right thing to do?
+                        log::debug!("Error handling guest trace: {:?}", e);
+                    }
+                    exit_reason
+                };
+
+            // ===== KILL() TIMING POINT 4: Before capturing cancel_requested =====
+            // If kill() is called and ran to completion BEFORE this line executes:
+            //    - CANCEL_BIT will be set
+            //    - Signals may still be sent (RUNNING_BIT=true) but are harmless no-ops
```
simongdavies, line 246: What about windows?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -159,26 +167,264 @@ pub(crate) trait Hypervisor: Debug + Send {
...
+                            if !mem_mgr.check_stack_guard()? {
+                                break Err(HyperlightError::StackOverflow());
+                            }
+
+                            break Err(new_error!("MMIO WRITE access address {:#x}", addr));
+                        }
+                    }
+                }
+                Ok(VmExit::Cancelled()) => {
+                    // If cancellation was not requested for this specific guest function call,
```
simongdavies, line 368: I don't understand the scope of this problem, is it possible that we try to kill a long running execution and that we end up delivering the signal to the wrong sandbox? If so does this happen in the current implementation? Either way if this does happen we should really think about a way to prevent this from happening as we don't really want one sandboxes behavior to influence the execution of another sandbox.

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,310 +658,113 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
-    ///
-    /// Although there's a theoretical race where:
-    /// 1. Thread A checks call_active (false)
-    /// 2. Thread B sets call_active (true) and starts guest call
-    /// 3. Thread A's kill() returns false (no effect)
-    ///
-    /// This is acceptable because the generation tracking provides an additional
-    /// safety layer. Even if a stale kill somehow stamped cancel_requested, the
-    /// generation mismatch would cause it to be ignored.
-    call_active: AtomicBool,
```
simongdavies, line 781: did you delete this? Seemed like a reasonable optimization to me , it avoided a lot of useless processing

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,310 +658,113 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
+    /// # Preserves
+    /// - CANCEL_BIT: The current value of CANCEL_BIT is preserved
+    ///
+    /// # Invariants Maintained
+    /// - Generation increments by 1 (wraps to 0 at MAX_GENERATION)
+    /// - RUNNING_BIT is set
+    /// - CANCEL_BIT remains unchanged
+    ///
+    /// # Memory Ordering
+    /// Uses `Release` ordering to ensure that the `tid` store (which uses `Relaxed`)
```
simongdavies: This comment is extremely confusing, I don't understand the relationship between store of this running atomic and the tid store or the relevance of the relaxed orderings, I think generally where we are use atomics and ordering we should document why the orderings are the way they are , these things are hard to understand when looking at the code later.

I think it will also make reviewing the code easier.

simongdavies (commented):
Made a few comments, two general comments:

I feel like this PR is doing 2 things and could be split up to make it easier to review:

1. Refactor Run - which is what the title says
2. Changing the way cancellation is implemented 

The second part could (should?) be done in another PR .

I also feel like we  need more information about how cancellation works, this really needs to be well documented , maybe we need a diagram or two somewhere

src/hyperlight_host/src/hypervisor/gdb/arch.rs
ludfjig: Whoops nice catch. Fixed

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: This flag provided no optimization as far as I can tell. We always already exit early if `running` flag is not true.

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Nice catch, typo in the docs. And agree. I'll create a new document for this

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: It does happen in the current implementation and as far as I am aware it's impossible to avoid. Signals are delivered async so there's no guaranteeing what happens in between sending the signal and the signal being delivered. A totally new vm could have started on the same thread or a new guest call could have been made. I think the only thing we can do here is ignore it if signal was not intended for this guest-call or sandbox.

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: I'm not sure I know what you mean. We only construct `VmExit::Retry()` on linux by this mapping `libc::EAGAIN => VmExit::Retry()`. There is no equivalent exit on windows that needs to be retried as far as I know. Regardless, if windows exits with a `VmExit::Cancelled()`, but there is no cancellation requested for the sandbox, then we retry it (but we still don't construct any `VmExit::Retry()` which is why the warning is supressed on windows).

PR #1024, opened 2025-11-17
Simplify cancellation

ludfjig:
Simplifies cancellation logic and unifies logic for all hypervisors, without affecting current behavior. Adds documentation to how everything works

### DIFF
--- a/docs/cancellation.md
+++ b/docs/cancellation.md
@@ -10,9 +10,9 @@ Hyperlight provides a mechanism to forcefully interrupt guest execution through 
 
 ### LinuxInterruptHandle State
 
-The `LinuxInterruptHandle` uses a packed atomic u64 to track execution state:
+The `LinuxInterruptHandle` uses a packed atomic u8 to track execution state:
 
-- **state (AtomicU64)**: Packs three bits:
+- **state (AtomicU8)**: Packs three bits:
   - **Bit 2 (DEBUG_INTERRUPT_BIT)**: Set when debugger interrupt is requested (gdb feature only)
   - **Bit 1 (RUNNING_BIT)**: Set when vCPU is actively running in guest mode
   - **Bit 0 (CANCEL_BIT)**: Set when cancellation has been requested via `kill()`
@@ -255,7 +255,7 @@ While the core cancellation mechanism follows the same conceptual model on Windo
 
 The `WindowsInterruptHandle` uses a simpler structure compared to Linux:
 
-- **state (AtomicU64)**: Packs three bits (RUNNING_BIT, CANCEL_BIT and DEBUG_INTERRUPT_BIT)
+- **state (AtomicU8)**: Packs three bits (RUNNING_BIT, CANCEL_BIT and DEBUG_INTERRUPT_BIT)
 - **partition_handle**: Windows Hyper-V partition handle for the VM
 - **dropped (AtomicBool)**: Set when the corresponding VM has been dropped
 
--- a/src/hyperlight_host/src/hypervisor/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/mod.rs
@@ -63,7 +63,9 @@ pub(crate) mod crashdump;
 
 use std::fmt::Debug;
 use std::str::FromStr;
-use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
+#[cfg(any(kvm, mshv3))]
+use std::sync::atomic::AtomicU64;
+use std::sync::atomic::{AtomicBool, AtomicU8, Ordering};
 use std::sync::{Arc, Mutex};
 #[cfg(any(kvm, mshv3))]
 use std::time::Duration;
@@ -588,7 +590,7 @@ pub(super) struct LinuxInterruptHandle {
     ///
     /// CANCEL_BIT persists across vcpu exits/re-entries within a single `VirtualCPU::run()` call
     /// (e.g., during host function calls), but is cleared at the start of each new `VirtualCPU::run()` call.
-    state: AtomicU64,
+    state: AtomicU8,
 
     /// Thread ID where the vcpu is running.
     ///
@@ -608,10 +610,10 @@ pub(super) struct LinuxInterruptHandle {
 
 #[cfg(any(kvm, mshv3))]
 impl LinuxInterruptHandle {
-    const RUNNING_BIT: u64 = 1 << 1;
-    const CANCEL_BIT: u64 = 1 << 0;
+    const RUNNING_BIT: u8 = 1 << 1;
+    const CANCEL_BIT: u8 = 1 << 0;
     #[cfg(gdb)]
-    const DEBUG_INTERRUPT_BIT: u64 = 1 << 2;
+    const DEBUG_INTERRUPT_BIT: u8 = 1 << 2;
 
     /// Get the running, cancel and debug flags atomically.
     ///
@@ -756,7 +758,7 @@ pub(super) struct WindowsInterruptHandle {
     ///
     /// CANCEL_BIT persists across vcpu exits/re-entries within a single `VirtualCPU::run()` call
     /// (e.g., during host function calls), but is cleared at the start of each new `VirtualCPU::run()` call.
-    state: AtomicU64,
+    state: AtomicU8,
 
     partition_handle: windows::Win32::System::Hypervisor::WHV_PARTITION_HANDLE,
     dropped: AtomicBool,
@@ -764,10 +766,10 @@ pub(super) struct WindowsInterruptHandle {
 
 #[cfg(target_os = "windows")]
 impl WindowsInterruptHandle {
-    const RUNNING_BIT: u64 = 1 << 1;
-    const CANCEL_BIT: u64 = 1 << 0;
+    const RUNNING_BIT: u8 = 1 << 1;
+    const CANCEL_BIT: u8 = 1 << 0;
     #[cfg(gdb)]
-    const DEBUG_INTERRUPT_BIT: u64 = 1 << 2;
+    const DEBUG_INTERRUPT_BIT: u8 = 1 << 2;
 }
 
 #[cfg(target_os = "windows")]
--- a/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_linux.rs
@@ -18,7 +18,7 @@ extern crate mshv_bindings;
 extern crate mshv_ioctls;
 
 use std::fmt::{Debug, Formatter};
-use std::sync::atomic::{AtomicBool, AtomicU64};
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicU64};
 use std::sync::{Arc, Mutex};
 
 use log::{LevelFilter, error};
@@ -375,7 +375,7 @@ impl HypervLinuxDriver {
         })?;
 
         let interrupt_handle: Arc<dyn InterruptHandleImpl> = Arc::new(LinuxInterruptHandle {
-            state: AtomicU64::new(0),
+            state: AtomicU8::new(0),
             #[cfg(all(
                 target_arch = "x86_64",
                 target_vendor = "unknown",
--- a/src/hyperlight_host/src/hypervisor/hyperv_windows.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperv_windows.rs
@@ -17,7 +17,7 @@ limitations under the License.
 use std::fmt;
 use std::fmt::{Debug, Formatter};
 use std::string::String;
-use std::sync::atomic::{AtomicBool, AtomicU64};
+use std::sync::atomic::{AtomicBool, AtomicU8};
 use std::sync::{Arc, Mutex};
 
 use log::LevelFilter;
@@ -324,7 +324,7 @@ impl HypervWindowsDriver {
         };
 
         let interrupt_handle = Arc::new(WindowsInterruptHandle {
-            state: AtomicU64::new(0),
+            state: AtomicU8::new(0),
             partition_handle,
             dropped: AtomicBool::new(false),
         });
--- a/src/hyperlight_host/src/hypervisor/kvm.rs
+++ b/src/hyperlight_host/src/hypervisor/kvm.rs
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 use std::fmt::Debug;
-use std::sync::atomic::{AtomicBool, AtomicU64};
+use std::sync::atomic::{AtomicBool, AtomicU8, AtomicU64};
 use std::sync::{Arc, Mutex};
 
 use kvm_bindings::{kvm_fpu, kvm_regs, kvm_sregs, kvm_userspace_memory_region};
@@ -333,7 +333,7 @@ impl KVMDriver {
         let rsp_gp = GuestPtr::try_from(RawPtr::from(rsp))?;
 
         let interrupt_handle: Arc<dyn InterruptHandleImpl> = Arc::new(LinuxInterruptHandle {
-            state: AtomicU64::new(0),
+            state: AtomicU8::new(0),
             #[cfg(all(
                 target_arch = "x86_64",
                 target_vendor = "unknown",


src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -356,33 +356,78 @@ impl VirtualCPU {
...
+            // If kill() is called and ran to completion BEFORE this line executes:
+            //    - CANCEL_BIT will be set and we will return an early VmExit::Cancelled()
+            interrupt_handle.set_tid();
+            interrupt_handle.set_running();
+
+            let exit_reason = {
+                if interrupt_handle.is_cancelled() || interrupt_handle.is_debug_interrupted() {
+                    Ok(HyperlightExit::Cancelled())
+                } else {
+                    #[cfg(feature = "trace_guest")]
```
simongdavies, line 389: should `interrupt_handle.set_running();` happen here rather than earlier?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -356,33 +356,78 @@ impl VirtualCPU {
...
+            //    - kill() will have no effect on this iteration, but CANCEL_BIT will persist
+            //    - If the loop continues (e.g., for a host call), the next iteration will be cancelled
+            //    - Stale signals from before clear_running() may arrive and kick future iterations,
+            //      but will be filtered out by the cancel_requested check below (and retried).
+            let cancel_requested = interrupt_handle.is_cancelled();
+            let debug_interrupted = interrupt_handle.is_debug_interrupted();
+
+            // ===== KILL() TIMING POINT 5: Before calling clear_running() =====
+            // Same as point 4.
+            interrupt_handle.clear_running();
```
simongdavies: Can this be done inside the else above (on line 390) I think that will close the window a little where signals are sent needlessly?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -425,6 +470,24 @@ impl VirtualCPU {
                     ));
                 }
                 Ok(HyperlightExit::Cancelled()) => {
+                    // If cancellation was not requested for this specific guest function call,
+                    // the vcpu was interrupted by a stale cancellation from a previous call
+                    if !cancel_requested && !debug_interrupted {
```
simongdavies, line 473: It might be nice to have a metric for when cancel_requested is not set and we cancelled the run, it would show how many signals were delivered to the wrong sandbox I think?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -849,42 +663,207 @@ impl LinuxInterruptHandle {
...
+
+#[cfg(target_os = "windows")]
+impl WindowsInterruptHandle {
+    const RUNNING_BIT: u64 = 1 << 1;
+    const CANCEL_BIT: u64 = 1 << 0;
+}
+
+#[cfg(target_os = "windows")]
+impl InterruptHandleImpl for WindowsInterruptHandle {
+    fn set_tid(&self) {
```
simongdavies: Cant we get rid of the set_tid in the trait definition in windows?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -849,42 +663,207 @@ impl LinuxInterruptHandle {
...
+        // No-op on Windows - we don't need to track thread ID
+    }
+
+    fn set_running(&self) {
```
simongdavies, line 777: Arent all these implementations the same on windows and linux?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -849,42 +663,207 @@ impl LinuxInterruptHandle {
...
+        if state & Self::RUNNING_BIT != 0 {
+            unsafe { WHvCancelRunVirtualProcessor(self.partition_handle, 0, 0).is_ok() }
+        } else {
+            false
+        }
     }
 
-    fn get_cancel_requested(&self) -> &AtomicU64 {
-        &self.cancel_requested
+    fn dropped(&self) -> bool {
```
simongdavies, line 857: This is also the same across windows and linux but making it common may be more trouble than its worth.

docs/cancellation.md
```diff
@@ -0,0 +1,439 @@
...
+    Note over VT,IT: Sync #3: clear_running(Release) → send_signal(Acquire)
+    VT-->>IT: synchronizes-with
+    
+    IT->>IT: send_signal() observes
+    Note right of IT: running=false<br/>Stop sending signals
+```
+
+## Interaction with Host Function Calls
+
+When a guest performs a host function call, the vCPU exits and the host function executes with `RUNNING_BIT=false`, preventing signal delivery during host execution. The `CANCEL_BIT` persists across this exit and re-entry, so if `kill()` was called, cancellation will be detected when the guest attempts to resume execution. This ensures cancellation takes effect even if it occurs during a host call, while avoiding signals during non-guest code execution.
```
jsturtevant:
```suggestion
When a guest performs a host function call, the vCPU exits and the host function executes with `RUNNING_BIT=false`, preventing signal delivery during host execution. The `CANCEL_BIT` persists across this exit and re-entry, so if `kill()` was called while executing a host function, cancellation will be detected when the guest attempts to resume execution. This ensures cancellation takes effect even if it occurs during a host call, while avoiding signals during non-guest code execution.
```

docs/cancellation.md
```diff
@@ -0,0 +1,439 @@
...
+
+```
+Timeline:
+t1: kill() sets CANCEL_BIT and sends signal
+t2: Guest completes naturally
+t3: clear_running() clears RUNNING_BIT
+t4: Signal arrives (too late)
+
+Result: If guest completes normally (Halt), returns Ok()
+        If guest exits for I/O, next iteration will be cancelled
```
jsturtevant: this doesn't really fit for this scenario? if it was only exiting for i/o that wouldn't be "completed"?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,322 +575,84 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
-                && self.call_active.load(Ordering::Acquire)
-            {
-                self.set_cancel_requested(generation);
-                target_generation = Some(generation);
-            }
+            // Check if we should continue sending signals
+            // Exit if not running OR if neither cancel nor debug_interrupt is set
+            #[cfg(gdb)]
+            let should_continue =
+                running && (cancel || self.debug_interrupt.load(Ordering::Relaxed));
```
jsturtevant: Why is the debug interrupt load `Relaxed` and the normal one is `Aquire`?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,322 +575,84 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
 
-    /// Thread ID where the VCPU is currently running.
-    ///
-    /// # Invariants
+    /// Thread ID where the vcpu is running.
     ///
-    /// - If VCPU is running: tid contains the thread ID of the executing thread
-    /// - Multiple VMs may share the same tid, but at most one will have running=true
+    /// Note: Multiple VMs may have the same `tid` (same thread runs multiple sandboxes sequentially),
+    /// but at most one VM will have RUNNING_BIT set at any given time.
```
jsturtevant, line 598: How do we know that the cancel was sent to the right vm in this case?  How will the work if the we are running multiple guest running in async manner?

docs/cancellation.md
```diff
@@ -0,0 +1,439 @@
...
+    loop Run Loop
+        Note over VM: === TIMING POINT 2 ===
+        VM->>IH: set_tid()
+        Note right of VM: Store current thread ID
+        VM->>IH: set_running()
+        Note right of VM: Set running=true
+
+        VM->>IH: is_cancelled()
+        
+        alt is_cancelled() == true
```
jsturtevant, line 53: These `alt` seem like they could be block names in the diagrams which would make it more readable: 
image

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: This is a good idea. I'll add this in a future PR #1027

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -356,33 +356,78 @@ impl VirtualCPU {
...
 
-                result
+                    result
+                }
             };
-            #[cfg(not(feature = "trace_guest"))]
-            let result = hv.run();
 
-            match result {
+            // ===== KILL() TIMING POINT 4: Before capturing cancel_requested =====
```
dblnz: I think there's a POINT 3 missing?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,322 +575,84 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
-    /// - All run() calls within the same guest call must share the same generation
-    /// - This ensures kill() affects the entire guest function call atomically
-    ///
-    /// # Invariants
-    ///
-    /// - If VCPU is running: bit 63 is set (neither converse nor inverse holds)
-    /// - If VCPU is running: bits 0-62 match the current guest call's generation
-    running: AtomicU64,
+    /// CANCEL_BIT persists across vcpu exits/re-entries within a single `HyperlightVm::run()` call
+    /// (e.g., during host function calls), but is cleared at the start of each new `HyperlightVm::run()` call.
```
dblnz:
```suggestion
    /// (e.g., during host function calls), but is cleared at the start of each new `VirtualCPU::run()` call.
```

src/hyperlight_host/src/hypervisor/mod.rs
dblnz: I don't think we can make that difference in this implementation, right?
I think that if we really want to make this difference, we can add some kind of identification for the running sandbox.

src/hyperlight_host/src/hypervisor/mod.rs
dblnz: I think that might have been added by me some time ago.
There was no particular reason other than there weren't strong requirements when it came to debugging, to have access ordered in any way, so I used `Relaxed` ordering. Feel free to change

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -849,42 +663,207 @@ impl LinuxInterruptHandle {
...
+    /// Atomic value packing vcpu execution state.
+    ///
+    /// Bit layout:
+    /// - Bit 1: RUNNING_BIT - set when vcpu is actively running
+    /// - Bit 0: CANCEL_BIT - set when cancellation has been requested
+    ///
+    /// `WHvCancelRunVirtualProcessor()` will return Ok even if the vcpu is not running,
+    /// which is why we need the RUNNING_BIT.
+    ///
+    /// CANCEL_BIT persists across vcpu exits/re-entries within a single `HyperlightVm::run()` call
```
dblnz:
```suggestion
    /// CANCEL_BIT persists across vcpu exits/re-entries within a single `VirtualCPU::run()` call
```

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -849,42 +663,207 @@ impl LinuxInterruptHandle {
...
+    /// (e.g., during host function calls), but is cleared at the start of each new `HyperlightVm::run()` call.
```
dblnz:
```suggestion
    /// (e.g., during host function calls), but is cleared at the start of each new `HyperlightVm::run()` call.
```

dblnz (commented):
Some minor comments from me.
Good work with the docs explaining the cancellation feature and splitting the PR 👍

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: No that would be a bug. We must always call `set_running()` before checking `is_cancelled()`, otherwise we could potentially pass the `!is_cancelled()` check, but then `kill()` is ran to completion _without sending any signals_, and we do `hv.run()` to compleltion despite CANCEL being true, resulting in ignoring the `kill()` command. 

By calling `set_running()` first, we guarantee that signals will be sent.

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Good catch, could move it up above these 2 lines to prevent some signals, but I cannot move it inside the else-block.
```
let cancel_requested = interrupt_handle.is_cancelled();
let debug_interrupted = interrupt_handle.is_debug_interrupted();
 ```

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 4 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs]: ok

File: docs/cancellation.md (hunk 7 above)
Comment (ludfjig) [reply docs/cancellation.md]: Fixed entire paragraph

File: docs/cancellation.md (hunk 8 above)
Comment (ludfjig) [reply docs/cancellation.md]: Cleaned up

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 9 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs]: Fixed, that was just an oversight

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 10 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs:598]: We only send signal if RUNNIG_BIT is set. RUNNIG_BIT being set implies `tid` is correct. Any given thread can only be blocked in 1 ioctl. I think that's the only guarantees we need.

File: docs/cancellation.md (hunk 11 above)
Comment (ludfjig) [reply docs/cancellation.md:53]: The block name are located next to the alts. In this case `[is_cancelled()==true] `

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 12 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs]: Fixed

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 13 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs]: Fixed

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 14 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs]: Fixed

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 15 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs]: Not sure what this comment means, but I removed all occurrences of HyperlightVm if that's what you meant

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 6 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs:857]: yes that's what I thought too

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 5 above)
Comment (ludfjig) [reply src/hyperlight_host/src/hypervisor/mod.rs:777]: yes but not easy to combine with current design

File: src/hyperlight_host/src/hypervisor/mod.rs (hunk 10 above)
Comment (jsturtevant) [reply src/hyperlight_host/src/hypervisor/mod.rs:598]: Ok, for now I think this works but we may need to reconsider in scenarios where multiple guests might be doing async host calls

File: docs/cancellation.md (hunk 11 above)
Comment (jsturtevant) [reply docs/cancellation.md:53]: sounds good, its a little confusing but that is indeed what the docs show

File: src/hyperlight_host/src/hypervisor/mod.rs [hunk 16]
```diff
@@ -461,54 +518,44 @@ impl VirtualCPU {
     }
 }
 
-/// A trait for handling interrupts to a sandbox's vcpu (public API)
-pub trait InterruptHandle: Debug + Send + Sync {
+/// A trait for platform-specific interrupt handle implementation details
+pub(crate) trait InterruptHandleImpl: InterruptHandle {
```
jsturtevant, line 526: This says if for platform specific implementation but it looks like `InterruptHandle` trait has to the platform specific implementation details and `InterruptHandleImpl` is the generic stuff (besides `set_tid`).  Plus each are implemetned for each type.  Maybe there is some future clean up we can do here

src/hyperlight_host/src/hypervisor/mod.rs
simongdavies: Could we update the code with comments to make this clear? (since my initial thought when looking at this was "but its not running yet"

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -1346,3 +1346,97 @@ fn interrupt_random_kill_stress_test() {
...
+/// Ensures that `kill()` reliably interrupts a running guest
+///
+/// The test works by:
+/// 1. Guest calls a host function which waits on a barrier, ensuring the guest is "in-progress" and that `kill()` is not called prematurely to be ignored.
+/// 2. Once the guest has passed that host function barrier, the host calls `kill()`. The `kill()` could be delivered at any time after this point, for example while guest is still in the host func, or returning into guest vm.
+/// 3. The guest enters an infinite loop, so `kill()` is the only way to stop it.
+///
+/// This is repeated across multiple threads and iterations to stress test the cancellation mechanism.
+///
+/// **Failure Condition:** If this test hangs, it means `kill()` failed to stop the guest, leaving it spinning forever.
```
jsturtevant, line 1359: Might consider a timeout so this doesn't hang indefinitly

jsturtevant (dismissed):

docs/cancellation.md
```diff
@@ -0,0 +1,292 @@
...
+- Set immediately by the API call
+- Checked at the start of each VM run loop iteration (before entering guest mode)
+- Automatically cleared when it causes a `WHvRunVpExitReasonCanceled` exit
+
+This means if `WHvCancelRunVirtualProcessor()` is called:
+- **While the vCPU is running**: The API signals the hypervisor to exit with `WHvRunVpExitReasonCanceled`
+- **Before VM runs**: The `CancelPending` flag persists and causes an immediate cancellation on the next VM run attempt
+
+Therefore, we only call `WHvCancelRunVirtualProcessor()` after checking that `RUNNING_BIT` is set. This is important because:
+1. If called when not running, the API would still succeed and will unconditionally cancel the next run attempt. This is bad since `kill()` should have no effect if the vCPU is not running
```
jprendes, line 289: Is there a small time window when the guest VM has exited but before RUNNING_BIT has been cleared that could mean we set CancelPending for the next run?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,322 +572,84 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
-    /// - This ensures kill() affects the entire guest function call atomically
-    ///
-    /// # Invariants
-    ///
-    /// - If VCPU is running: bit 63 is set (neither converse nor inverse holds)
-    /// - If VCPU is running: bits 0-62 match the current guest call's generation
-    running: AtomicU64,
+    /// CANCEL_BIT persists across vcpu exits/re-entries within a single `VirtualCPU::run()` call
+    /// (e.g., during host function calls), but is cleared at the start of each new `VirtualCPU::run()` call.
+    state: AtomicU64,
```
jprendes: Could this be AtomicU8?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,322 +572,84 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
-    /// Thread ID where the VCPU is currently running.
-    ///
-    /// # Invariants
+    /// Thread ID where the vcpu is running.
     ///
-    /// - If VCPU is running: tid contains the thread ID of the executing thread
-    /// - Multiple VMs may share the same tid, but at most one will have running=true
+    /// Note: Multiple VMs may have the same `tid` (same thread runs multiple sandboxes sequentially),
+    /// but at most one VM will have RUNNING_BIT set at any given time.
     tid: AtomicU64,
```
jprendes, line 599: Does this need to be Atomic?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -523,322 +572,84 @@ pub trait InterruptHandle: Debug + Send + Sync {
...
-    call_active: AtomicBool,
-
-    /// Debugger interrupt request flag (GDB only).
-    ///
-    /// Set when kill_from_debugger() is called, cleared when VCPU stops running.
-    /// Used to distinguish debugger interrupts from normal kill() interrupts.
+    /// Debugger interrupt flag (gdb feature only).
+    /// Set when `kill_from_debugger()` is called, cleared when vcpu stops running.
     #[cfg(gdb)]
     debug_interrupt: AtomicBool,
```
jprendes: Would it make sense for this to be part of the `state` Atomic?

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -1346,3 +1346,279 @@ fn interrupt_random_kill_stress_test() {
...
+    }
+
+    for handle in handles {
+        handle.join().unwrap();
+    }
+}
+
+// Validates that kill delivery stays accurate when a sandbox hops between threads
+// mid-call and shares a thread ID with another sandbox, ensuring only the intended
+// VM is interrupted while bait sandboxes keep running.
```
jsturtevant, line 1446: bait?

jsturtevant: oh, i get it now :-)

docs/cancellation.md
ludfjig: Yes this is possible. But when the next run immediately cancels, it will be retried since CANCEL_BIT is not set. It is treated similarly to a stale signal on linux

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: I believe so, it needs interior mutability because `set_tid(&self)` need to take `&self` rather that `&mut self`.

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: good idea, added

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: Yes, fixed

Issue #698, opened 2025-07-08
Add tests for handling exceptions in core Hyperlight


syntactically:
Currently, we have support for registering exception handlers, but the only tests of this functionality are in hyperlight-wasm. We should add simpler tests in hyperlight core, and especially a test of allocating a page on demand when first read/written to.

PR #1037, opened 2025-11-22
Update the interrupt handler for 16byte alignment

jsturtevant:
fixes: https://github.com/hyperlight-dev/hyperlight/issues/1047

Adds tests for exception handling but doesn't specifically test the logic for page faulting logic as described in (https://github.com/hyperlight-dev/hyperlight/issues/698)

This pull request updates the exception context saving and restoring logic to properly ensure correct 16-byte stack alignment. It adds a test that demonstrates the failure and then does some additional validation.

### DIFF
--- a/src/hyperlight_guest_bin/src/exceptions/interrupt_entry.rs
+++ b/src/hyperlight_guest_bin/src/exceptions/interrupt_entry.rs
@@ -51,6 +51,8 @@ unsafe extern "C" {
 macro_rules! context_save {
     () => {
         concat!(
+            // Push padding to match Context struct (8 bytes)
+            "    push 0\n",
             // Save general-purpose registers
             "    push rax\n",
             "    push rbx\n",
@@ -67,10 +69,6 @@ macro_rules! context_save {
             "    push r13\n",
             "    push r14\n",
             "    push r15\n",
-            // Save one of the segment registers to get 16-byte alignment for
-            // FXSAVE. TODO: consider packing the segment registers
-            "    mov rax, ds\n",
-            "    push rax\n",
             // Save floating-point/SSE registers
             // TODO: Don't do this unconditionally: get the exn
             //       handlers compiled without sse
@@ -79,7 +77,9 @@ macro_rules! context_save {
             "    sub rsp, 512\n",
             "    mov rax, rsp\n",
             "    fxsave [rax]\n",
-            // Save the rest of the segment registers
+            // Save all segment registers
+            "    mov rax, ds\n",
+            "    push rax\n",
             "    mov rax, es\n",
             "    push rax\n",
             "    mov rax, fs\n",
@@ -93,20 +93,19 @@ macro_rules! context_save {
 macro_rules! context_restore {
     () => {
         concat!(
-            // Restore most segment registers
+            // Restore all segment registers
             "    pop rax\n",
             "    mov gs, rax\n",
             "    pop rax\n",
             "    mov fs, rax\n",
             "    pop rax\n",
             "    mov es, rax\n",
+            "    pop rax\n",
+            "    mov ds, rax\n",
             // Restore floating-point/SSE registers
             "    mov rax, rsp\n",
             "    fxrstor [rax]\n",
             "    add rsp, 512\n",
-            // Restore the last segment register
-            "    pop rax\n",
-            "    mov ds, rax\n",
             // Restore general-purpose registers
             "    pop r15\n",
             "    pop r14\n",
@@ -123,6 +122,8 @@ macro_rules! context_restore {
             "    pop rcx\n",
             "    pop rbx\n",
             "    pop rax\n",
+            // Skip padding (8 bytes)
+            "    add rsp, 8\n",
         )
     };
 }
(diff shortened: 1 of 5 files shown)


src/tests/rust_guests/callbackguest/Cargo.lock
```diff
@@ -1,270 +0,0 @@
-# This file is automatically @generated by Cargo.
```
ludfjig: what happened to this file?

Justfile
```diff
@@ -56,8 +56,8 @@ build-rust-guests target=default-target features="": (witguest-wit)
     cp {{ dummyguest_source }}/{{ target }}/dummyguest* {{ rust_guests_bin_dir }}/{{ target }}/
     cp {{ witguest_source }}/{{ target }}/witguest* {{ rust_guests_bin_dir }}/{{ target }}/
 
-build-and-move-rust-guests: (build-rust-guests "debug") (move-rust-guests "debug") (build-rust-guests "release") (move-rust-guests "release")
-build-and-move-c-guests: (build-c-guests "debug") (move-c-guests "debug") (build-c-guests "release") (move-c-guests "release")
+build-and-move-rust-guests target=default-target: (build-rust-guests target) (move-rust-guests target)
```
simongdavies: I think this means that the guests recipe needs updating too

src/tests/rust_guests/simpleguest/src/main.rs
```diff
@@ -70,6 +82,148 @@ fn set_static(_: &FunctionCall) -> Result<Vec<u8>> {
...
+            "sub rsp, 16",        // Allocate space on stack
+            "movaps [rsp], xmm0", // This will #GP if stack is not 16-byte aligned
+            "add rsp, 16",        // Clean up
+            options(nostack)
+        );
+    }
+
+    // Validate we can read and write to Context registers
+    // First, read the original value from R9 register (index 6)
+    let original_r9 = unsafe { core::ptr::read_volatile(&(*context).gprs[6]) };
```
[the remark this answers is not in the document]
jsturtevant: this would mean the tests would fail.  This isn't right but did catch that the doc is wrong 🤦.  The order of the items pushed onto the stack is not correct in the diagram.

src/tests/rust_guests/simpleguest/src/main.rs
```diff
@@ -70,6 +82,148 @@ fn set_static(_: &FunctionCall) -> Result<Vec<u8>> {
...
+        core::arch::asm!(
+            "sub rsp, 16",        // Allocate space on stack
+            "movaps [rsp], xmm0", // This will #GP if stack is not 16-byte aligned
+            "add rsp, 16",        // Clean up
+            options(nostack)
+        );
+    }
+
+    // Validate we can read and write to Context registers
+    // First, read the original value from R9 register (index 6)
```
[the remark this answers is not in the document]
jsturtevant: this is wrong, the doc was wrong, but working on an update

src/hyperlight_guest_bin/src/exceptions/interrupt_entry.rs
```diff
@@ -178,6 +179,43 @@ macro_rules! generate_exceptions {
...
+// Stack layout after context_save!() (from high to low addresses):
+// ```
+// +------------------+ <-- Higher addresses
+// | SS               | (Pushed by CPU on exception)
+// | RSP              | (Pushed by CPU on exception)
+// | RFLAGS           | (Pushed by CPU on exception)
+// | CS               | (Pushed by CPU on exception)
+// | RIP              | (Pushed by CPU on exception)
+// | Error Code       | (Pushed by CPU or by handler)
+// +------------------+ <-- ExceptionInfo struct starts here
```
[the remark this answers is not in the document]
jsturtevant: fair, updated

src/tests/rust_guests/simpleguest/src/main.rs
```diff
@@ -70,6 +82,148 @@ fn set_static(_: &FunctionCall) -> Result<Vec<u8>> {
...
+        "R9 register should contain original test value"
+    );
+
+    // Modify R9 register directly and write R10 to context
+    unsafe {
+        // Modify R9 register directly using inline assembly
+        core::arch::asm!("mov r9, {0}", in(reg) TEST_R9_MODIFIED_VALUE);
+
+        // Write to R10 via context to verify context modification works
+        core::ptr::write_volatile(&mut (*context).gprs[5], TEST_R10_VALUE); // R10
```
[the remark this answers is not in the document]
jsturtevant: this is wrong, the doc was wrong

Justfile
jsturtevant: added

Justfile
```diff
@@ -40,7 +40,7 @@ build target=default-target:
     {{ cargo-cmd }} build --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }}
 
 # build testing guest binaries
-guests: build-and-move-rust-guests build-and-move-c-guests
+guests target=default-target: (build-and-move-rust-guests target) (build-and-move-c-guests target)
```
ludfjig: benchmark.yml requires both debug and release to be built by `build-guest-binaries`. Maybe it also needs updating?

src/tests/rust_guests/callbackguest/Cargo.lock
jsturtevant: I think you removed the callback guest awhile back, it looks like this file was missing from the removal or it got added back somehow? https://github.com/hyperlight-dev/hyperlight/pull/871/files

jsturtevant: looks like it was me https://github.com/hyperlight-dev/hyperlight/commit/0bd4ac23f8140887b81249f3a1c8f8920d288e81 🤕 , must have mess something up on the rebase

Justfile
jsturtevant: ah, i think i messed this up.  Im just going to drop this commit as i think it worked better prior

jsturtevant: dropped, so there is no change here.

syntactically:
In re: the main change with the stack alignment, you are absolutely correct: the old code was accidentally setting it up so that the stack was aligned at the beginning of the handler function, but the ABI requirement is actually that the stack be aligned _before the call instruction that invokes the function_, which is a different requirement on amd64 since `call` implicitly pushes a return address onto the stack.  Let's merge that part as soon as possible.

The 16 byte alignment requirement for the stack is knowledge that I think is pretty widespread at the moment, but maybe let's add a citation to the ABI spec? I think the authoritative reference here is the x86-64 ELF System V psABI which can be found at https://gitlab.com/x86-psABIs/x86-64-ABI/-/jobs/artifacts/master/raw/x86-64-ABI/abi.pdf, Section 3.2.2.

I am not a huge fan of the new handler registration API---I don't think that what it abstracts is important, and I think that the important stuff that we do need to abstract in the near future is not covered here. I agree that the raw handler array export should go away, but I would like to see it replaced with some more generic and architecture-independent options for e.g. "let me respond to unmapped page accesses in this region", "let me respond to unpermitted accesses in this region", "let me know about any undefined/misaligned/FP faulting instructions", etc.  The main reason that this is not there yet IIRC is that the virtual memory handlers probably need to work with the (types from the) upcoming architecture-independent virtual memory mapping abstractions.  It seems to me like this changeset just means that fixing the exception handler exports is going to require two API breaks instead of one, with the intermediate API requiring downstream changes in hl-wasm an extra time and unlikely to unblock anyone else. Consequently, I am weakly against on that portion of the change.

syntactically (approved):

jsturtevant:
Added the reference and removed the handler API code since we have major changes make that more platform independent.

syntactically (dismissed):

Issue #1032, opened 2025-11-20
Fix guest tracing memory leak when `halt` function is called


dblnz:
### What happened?

After a relative big number of iterations of the fuzzy test, the call to the guest function hangs, never returns.

### What did you expect to happen?

The expectation is for the fuzzy test to never make a guest function call hang or crash.

### Steps to reproduce the behavior

This issue was observed when working on running fuzzy tests that call a recursive guest function that creates spans and log events.
The test calls a guest function that makes up to 255 recursive calls and when the `halt` function is called to signal the completion of the function call, it leaks the serialized data.

### Hyperlight Version

0.11.0

### OS version

<details>

```console
On Linux:
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.3 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo

$ uname -a
Linux laptop 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux

On Windows:
C:\> cmd /c ver
Not tested
````
</details>

### Additional Information

After further debugging, I found there were two issues producing this outcome:
- When the guest calls `halt` function, it wants to report the tracing data to the host, so it calls the `hyperlight-guest-tracing::guest_trace_info` function to serialize the events captured until then.
 This function returns a `Vec<u8>` and after calling the `hlt` instruction, the `hypervisor` never yields back control to the guest to correctly `drop` the allocated memory, which causes a leak.
- The problem that causes the hanging is a deadlock produced by a lock of the trace data because, at some point, the heap allocation fails which in turn generates an exception which calls `outb` function to send the same trace data to the host, which in turn tries to get the lock.

Solution:
- Change the logic to serialize the events as they are created, and only keep a buffer containing a chunk of serialized events. This removes the need to serialize/allocate when we want to send to the host.
[...]

PR #1066, opened 2025-12-02
Fix guest tracing deadlock when exception happens during tracing data serialization

dblnz:
Partially addresses #1032.
This fixes an issue where the guest hangs because of a deadlock produced when an exception occurs in the tracing code and the `GuestState` lock is acquired. When handling the exception, the call to `outb` function, tries to acquire the same lock another time, which leads to a deadlock.

### DIFF
--- a/src/hyperlight_guest_tracing/src/visitor.rs
+++ b/src/hyperlight_guest_tracing/src/visitor.rs
@@ -34,8 +34,7 @@ impl<'a> Visit for FieldsVisitor<'a> {
     /// * `value` - The byte slice value
     fn record_bytes(&mut self, f: &Field, v: &[u8]) {
         let k = String::from(f.name());
-        let mut val = String::new();
-        val.push_str(&alloc::format!("{v:?}"));
+        let val = alloc::format!("{v:?}");
         self.out.push(KeyValue { key: k, value: val });
     }
 
@@ -45,8 +44,7 @@ impl<'a> Visit for FieldsVisitor<'a> {
     /// * `v` - The string value
     fn record_str(&mut self, f: &Field, v: &str) {
         let k = String::from(f.name());
-        let mut val = String::new();
-        val.push_str(v);
+        let val = String::from(v);
         self.out.push(KeyValue { key: k, value: val });
     }
 
@@ -56,8 +54,7 @@ impl<'a> Visit for FieldsVisitor<'a> {
     /// * `v` - The debug value
     fn record_debug(&mut self, f: &Field, v: &dyn Debug) {
         let k = String::from(f.name());
-        let mut val = String::new();
-        val.push_str(&alloc::format!("{v:?}"));
+        let val = alloc::format!("{v:?}");
         self.out.push(KeyValue { key: k, value: val });
     }
 }


src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -72,12 +72,19 @@ mod trace {
...
     pub fn set_start_tsc(guest_start_tsc: u64) {
         if let Some(w) = GUEST_STATE.get()
-            && let Some(state) = w.upgrade()
+            && let Some(state_mutex) = w.upgrade()
         {
-            state.lock().set_start_tsc(guest_start_tsc);
+            // When accessing the GuestState, we try_lock to avoid deadlocks in the guest.
+            // Having a single thread, if we cannot lock it, it means the state is already
+            // locked (e.g. exception while lock is acquired).
+            // In that case, we panic to signal the issue, continuing results incorrect behavior.
```
jsturtevant: its not clear to me if having a single thread is the problem, this doesn't work with a single thread, or if there is another pattern that we should be using when single threaded. I.E if i get the exception during developing my guest, should I be doing something else?

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -72,12 +72,19 @@ mod trace {
...
+            let mut state = state_mutex
+                .try_lock()
+                .expect("Unable to lock GuestState in set_start_tsc");
```
jsturtevant:
```suggestion
                .expect("Unable to lock Guest tracing state in set_start_tsc");
```
or something along those lines so we know it was realted to tracing

dblnz: The `GuestState` is the actual type being locked

src/hyperlight_guest_tracing/src/lib.rs
dblnz: I think I should phrase it better, I wanted to say that "Even though there's a single thread, it doesn't guarantee the resource is unlocked"

src/hyperlight_guest_tracing/src/visitor.rs
```diff
@@ -34,8 +34,7 @@ impl<'a> Visit for FieldsVisitor<'a> {
     /// * `value` - The byte slice value
     fn record_bytes(&mut self, f: &Field, v: &[u8]) {
         let k = String::from(f.name());
-        let mut val = String::new();
-        val.push_str(&alloc::format!("{v:?}"));
+        let val = String::from(&alloc::format!("{v:?}"));
```
ludfjig: Isn't alloc::format! already returning a string?

src/hyperlight_guest_tracing/src/lib.rs
jsturtevant: If it isn't unlocked then something went wrong with our implementation?  This shouldn't actually be hit unless we did something wrong?

dblnz: On normal use, we lock, add traces, then unlock. But if something bad happens (an exception happens), the resource can remain locked and lead to next time creating a deadlock.

Imagine:
1. lock tracing state
2. serialize data
3. allocate memory for something and it panics (no more memory)
4. the exception handler may try to log/trace again, which needs to do the same steps as above.

The fact that the tracing state remained locked is faulty, which we cannot control. We can control the next time it tries to acquire the lock, to panic, so we know why it happened

ludfjig: But the panic should poison the sandbox, right?

dblnz: Yes, this can happen on the same guest call.
While logging/tracing something, an exception occurs, and during the exception, the handler wants to log/trace again, it needs the lock. This prevents those cases.

I could just use `try_lock` and not panic when acquiring fails, but it would silently fail. There might be cases where you expect a log while debugging, which doesn't come because of an exception and you don't know why

src/hyperlight_guest_tracing/src/lib.rs
jsturtevant: maybe `tracing: unable to lock GuestState in set_start_tsc`? I just imagine my self googling this error message not knowing its related to tracing that I enabled

dblnz: You're right 😄

src/hyperlight_guest_tracing/src/lib.rs
jsturtevant: > Imagine:
> 
> 1. lock tracing state
> 2. serialize data
> 3. allocate memory for something and it panics (no more memory)
> 4. the exception handler may try to log/trace again, which needs to do the same steps as above.

super helpful to understand the scenario, might be worth adding this to the comments

>the exception handler may try to log/trace again

Should we avoid logging/tracing in the guest exception handler?

dblnz: > Should we avoid logging/tracing in the guest exception handler?

We can do this, however, I think we could also benefit from some logs/tracing info inside the handlers themselves.

ludfjig (dismissed):
LGTM

src/hyperlight_guest_tracing/src/lib.rs
```diff
@@ -72,12 +77,27 @@ mod trace {
...
+            // - 2. An exception occurs while the span is being created (e.g. not enough memory, etc.)
+            // - 3. The exception handler uses the tracing API to send the trace data to the host
+            // or just create spans/events for logging purposes.
+            // - 4. The tracing API tries to lock the state again, causing a deadlock.
+            // To avoid this, we use try_lock and if we cannot acquire the lock, we panic to signal
+            // the issue.
+            let mut state = state_mutex
+                .try_lock()
+                .expect("guest_tracing: Unable to lock guest tracing state in `set_start_tsc`");
+            state.set_start_tsc(guest_start_tsc);
```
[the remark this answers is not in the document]
dblnz: This is explicit to ensure anyone who modifies sees this

Issue #1065, opened 2025-12-02
Fix guest logging `StackOverflow` issue


dblnz:
### What happened?

When using logging in a hyperlight guest, the guest returns `StackOverflow`.

### What did you expect to happen?

The expected behavior is for hyperlight to correctly receive the logs from the guest through the `OutBAction::Log` VMExit.

### Steps to reproduce the behavior

1. set `RUST_LOG="none,hyperlight_guest=trace,hyperlight_guest_bin=trace"`
2. run a hyperlight guest example
3. The guest returns `StackOverflow`

### Hyperlight Version

0.11.0

### OS version

<details>

```console
On Linux:
$ cat /etc/os-release
PRETTY_NAME="Ubuntu 24.04.3 LTS"
NAME="Ubuntu"
VERSION_ID="24.04"
VERSION="24.04.3 LTS (Noble Numbat)"
VERSION_CODENAME=noble
ID=ubuntu
ID_LIKE=debian
HOME_URL="https://www.ubuntu.com/"
SUPPORT_URL="https://help.ubuntu.com/"
BUG_REPORT_URL="https://bugs.launchpad.net/ubuntu/"
PRIVACY_POLICY_URL="https://www.ubuntu.com/legal/terms-and-policies/privacy-policy"
UBUNTU_CODENAME=noble
LOGO=ubuntu-logo

$ uname -a
Linux laptop 6.6.87.2-microsoft-standard-WSL2 #1 SMP PREEMPT_DYNAMIC Thu Jun  5 18:30:46 UTC 2025 x86_64 GNU/Linux

On Windows:
C:\> cmd /c ver
not tested but should behave the same

</details>


### Additional Information

When tracing is not enabled, the previous way of reporting logs, `outb` with `OutBAction::Log` action, calls [push_shared_output_data](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest/src/guest_handle/io.rs#L92) which is instrumented for logging/tracing. This `instrument` tries to send a new log again, this results in an infinite `loop` that produces the `StackOverflow`.

The reason for this issue is the `log_message` function call that reports the message to the host, more specifically the [hyperlight_guest/src/host_comm.rs#L196](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest/src/guest_handle/host_comm.rs#L196), that uses the `input/output` buffer for the log.

[...]

PR #1067, opened 2025-12-02
Fix StackOverflow produced by guest logging

dblnz:
This is a temporary fix for #1065 so we can use guest logging.
I am working on a better fix that implies using the event batching that the guest tracing uses.


jsturtevant:
Could you add a test to the just file that does:

```
set RUST_LOG="none,hyperlight_guest=trace,hyperlight_guest_bin=trace"
run a hyperlight guest example
```

so we can verify and so when the future fix comes in we know we didn't break it?

dblnz:
> Could you add a test to the just file that does:
> 
> ```
> set RUST_LOG="none,hyperlight_guest=trace,hyperlight_guest_bin=trace"
> run a hyperlight guest example
> ```
> 
> so we can verify and so when the future fix comes in we know we didn't break it?

We need to enable it in the guests also, because we don't have "logging" enabled there. I can do that

PR #1072, opened 2025-12-04
Fix guest call to `halt` not dropping allocated trace data leading to memory leak 

### LINKED ISSUE #1032: Fix guest tracing memory leak when `halt` function is called

dblnz:
## Problem
This fixes #1032.

 When the `trace_guest` feature is enabled, when a function call successfully finishes it calls `halt` which serializes all the stored events until then and returns a buffer with the data, which the `hlt` instruction provides to the host.
 However, the host doesn't continue the guest execution from the same place, so the owned variable on the stack doesn't get dropped, which leads to memory leaks.

## Solution
Avoid allocating on the stack of the `hlt` function and provide a global buffer (which lives during the entire lifetime of the VM).
Serialize each event and append it to a global, pre-allocated buffer, that when it gets filled, it automatically sends the data to the host.
In the case of `halt` call, we now return a `(ptr, len)` to the global buffer, which avoids allocations.

Things done:
- [x] Introduce `Encoder` / `Decoder` traits that define the functionality a used by the `hyperlight_guest_tracing` crate.
- [x] Implement an `EventsBatchEncoder` that batches multiple events, serializing each one individually and appending to a global buffer. The encoder takes a function as an argument that gets called whenever the buffer is filled. This function receives a reference to the buffer to be able to send it to the host. After the call is finished, the buffer gets cleared.
- [x] The encoder uses a heuristic to estimate the necessary size for the Flatbuffers buffer used to serialize each event before appending it to the global buffer (NOTE: The reason we haven't serialized in-place in the global buffer is to avoid reallocations for the global buffer in case the estimation is incorrect and by appending it overflows)
- Tests
  - [x] Added tests and fuzzying for the heuristic to ensure the logic is correct, however, even if the size is smaller than the needed size, Flatbuffers reallocs memory, which has some performance penalty, but still it doesn't fail
  - [x] Added tests for the host logic that parses the events
  - [x] Added fuzzying for end-to-end use-case. Host calls a guest function that recursively creates spans/logs, and the host parses.
  - [x] Added miri checks for `hyperlight-common` that checks the use of flatbuffers (not only tracing features)

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -214,6 +214,12 @@ test-doc target=default-target features="":
 ### LINTING ####
 ################
 
+miri-tests:
+    rustup component add miri --toolchain nightly
+    # For now only run miri tests on hyperlight-common with trace_guest feature
+    # We can add more as needed
+    cargo +nightly miri test -p hyperlight-common -F trace_guest
+
 check:
     {{ cargo-cmd }} check  {{ target-triple-flag }}
     {{ cargo-cmd }} check -p hyperlight-host --features crashdump  {{ target-triple-flag }}
--- a/.github/workflows/dep_rust.yml
+++ b/.github/workflows/dep_rust.yml
@@ -138,6 +138,13 @@ jobs:
         env: 
           TARGET_TRIPLE: ${{ inputs.target_triple }}
 
+      - name: Run Miri tests
+        if: ${{ (runner.os == 'Linux' )}}
+        env:
+          CARGO_TERM_COLOR: always
+          TARGET_TRIPLE: ${{ inputs.target_triple }}
+        run: just miri-tests
+
       - name: Run Rust tests
         env:
           CARGO_TERM_COLOR: always
--- a/src/hyperlight_common/src/flatbuffer_wrappers/util.rs
+++ b/src/hyperlight_common/src/flatbuffer_wrappers/util.rs
@@ -562,7 +562,9 @@ mod tests {
         );
     }
 
+    /// When running under Miri, this test takes too long due to large allocations
     #[test]
+    #[cfg_attr(miri, ignore)]
     fn test_estimate_megabyte_parameters() {
         assert_estimation_accuracy(
             "process_megabyte_data",


src/hyperlight_common/src/flatbuffer_wrappers/guest_trace_data.rs
```diff
@@ -130,274 +131,445 @@ pub enum GuestEvent {
...
+}
+
+impl<T: Fn(&[u8])> EventsEncoder for EventsBatchEncoderGeneric<T> {
+    /// Serialize a single GuestEvent and append it to the internal buffer.
+    /// If the appending of the serialized data exceeds buffer capacity, the
+    /// `report_full` callback is invoked with the current buffer contents,
+    /// and the buffer is cleared for new data.
+    fn encode(&mut self, event: &GuestEvent) {
+        // TODO: Estimate size more accurately
+        let estimated_size = 1024;
```
dblnz: Add heuristic to estimate this.

src/hyperlight_host/src/sandbox/trace/context.rs
```diff
@@ -396,9 +410,7 @@ impl Drop for TraceContext {
 
 #[cfg(test)]
```
dblnz, line 411: Add more tests that check failure scenarios, to verify hyperlight host behavior

src/hyperlight_common/src/flatbuffer_wrappers/guest_trace_data.rs
dblnz: Add fuzzy tests for the heuristic to ensure it is correct

syntactically:
I think I agree that for this specific problem making sure that the trace code does not allocate is the correct approach, since if you were to allocate for the trace buffer you could only tear it down after the vmexit after the host has had a chance to read it.

However, the description also makes me wonder whether it is a problem more generally that we are not tearing down the frame "properly"---one could imagine some e.g. guest lock release ending up in that frame through some optimisation and its loss having effects, perhaps.  Might it also make sense to try to ensure that the frame cleanup happens, perhaps by having the actual dispatch entrypoint be a tiny asm stub that just does a call and halt, so that the rust function(s) involved do properly have their frame torn down at the end before the halt is called?

dblnz:
> I think I agree that for this specific problem making sure that the trace code does not allocate is the correct approach, since if you were to allocate for the trace buffer you could only tear it down after the vmexit after the host has had a chance to read it.
> 
> However, the description also makes me wonder whether it is a problem more generally that we are not tearing down the frame "properly"---one could imagine some e.g. guest lock release ending up in that frame through some optimisation and its loss having effects, perhaps. Might it also make sense to try to ensure that the frame cleanup happens, perhaps by having the actual dispatch entrypoint be a tiny asm stub that just does a call and halt, so that the rust function(s) involved do properly have their frame torn down at the end before the halt is called?

I think your concern is valid, and we should ensure no leaks and stack frames are properly cleaned.
This PR is specific to the tracing data, which triggered this issue faster because of the big amounts of data being allocated.
I'll open an Issue to address this separately

src/hyperlight_common/src/flatbuffer_wrappers/guest_trace_data.rs
dblnz: Added estimation heuristic

src/hyperlight_host/src/sandbox/trace/context.rs
dblnz: Added more tests and additionally miri

Justfile
```diff
@@ -214,6 +214,12 @@ test-doc target=default-target features="":
 ### LINTING ####
 ################
 
+miri-tests:
+    rustup component add miri --toolchain nightly
```
simongdavies, line 218: Curious, why did we add miri for testing for this PR?

Justfile
```diff
@@ -214,6 +214,12 @@ test-doc target=default-target features="":
...
+    # For now only run miri tests on hyperlight-common with trace_guest feature
```
simongdavies, line 219: Do we have an issue to do this?

Justfile
```diff
@@ -342,12 +348,14 @@ bench features="":
 fuzz_memory_limit := "4096"
 
 # Fuzzes the given target
+# Uses *case* for compatibility to determine if the target is a tracing fuzzer or not
 fuzz fuzz-target:
-    cargo +nightly fuzz run {{ fuzz-target }} --release -- -rss_limit_mb={{ fuzz_memory_limit }}
+    case "{{ fuzz-target }}" in *trace*) just fuzz-trace {{ fuzz-target }} ;; *) cargo +nightly fuzz run {{ fuzz-target }} --release -- -rss_limit_mb={{ fuzz_memory_limit }} ;; esac
```
simongdavies, line 353: I don't think this will work on Windows, but I just realised that we don't fuzz on Windows which I think is an issue, I'll open an issue or that

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -138,7 +138,9 @@ pub(crate) extern "C" fn dispatch_function() {
     #[cfg(feature = "trace_guest")]
     {
         let guest_start_tsc = hyperlight_guest_tracing::invariant_tsc::read_tsc();
-        hyperlight_guest_tracing::set_start_tsc(guest_start_tsc);
+        // Reset the trace state for the new guest function call with the new start TSC
+        // This clears any existing spans/events from previous calls ensuring a clean state
+        hyperlight_guest_tracing::new_call(guest_start_tsc);
```
simongdavies, line 143: I am pretty sure that this scenario is totally broken/not possible any longer so this comment might be irrelevant but what happens if a host was to call a guest function? Not sure if we have an issue for this scenario, maybe we should..

src/hyperlight_guest_tracing/src/state.rs
```diff
@@ -33,38 +35,49 @@ pub struct TraceBatchInfo {
...
     /// Next span ID to allocate
     next_id: AtomicU64,
-    /// Trace information that is exchanged with the host
-    data: GuestTraceData,
     /// Stack of active spans
     stack: Vec<u64>,
 }
 
-/// TODO: Change these constants to be configurable at runtime by the guest
+/// TODO: Change these constant to be configurable at runtime by the guest
```
simongdavies, line 46: Do we have an issue for this?

Justfile
dblnz: I wanted to make sure the things I added for tracing don't produce UB.
It was not a hard requirement. Do you think we don't need it?

Justfile
dblnz: I tried running for `hyperlight-host` and got some errors, which weren't related to this. I'll open an issue for this

src/hyperlight_guest_bin/src/guest_function/call.rs
dblnz: I am not sure if I understand your concern.
 This, for example, makes sure that after a guest runs `hlt`, a new guest function call starts with an empty buffer. So we don't end up reporting the same events to the host multiple times

src/hyperlight_guest_tracing/src/state.rs
dblnz: Yes, it is part of an issue where I'll address guest trace buffer configuration

Justfile
simongdavies: No, was just wondering what miri gave us that we needed here

src/hyperlight_guest_bin/src/guest_function/call.rs
simongdavies: Ideally we could allow host->guest-func->host-func->guest-func (i.e. the host re-enters the guest before the original guest call completes, I was wondering if the trace would flow properly in that case (I don't think it would) but I think its a hypothetical anyway because I dont think we ever (re-)enabled the scenario anyway.

Justfile
dblnz: Created issue https://github.com/hyperlight-dev/hyperlight/issues/1083

Issue #319, opened 2025-03-04
Investigate making it possible to run 2 host processes in parallel n Windows


simongdavies:
On Windows we depend on the `HyperlightSurrogate` process to provide a host process for memory that we subsequently map into a VM.

If more than one hyperlight host processes is started on Windows then the second and subsequent ones fail with an error stating that `HyperlightSurrogate.exe` is in use.

I believe that this is caused by hyperlight trying to copy the surrogate process binary to disk when the second or subsequent process starts , we should fix it tom make it possible to run more than one host concurrently

PR #1092, opened 2025-12-09
Remove name from surrogate job obj

ludfjig:
On windows using hyperlight in multiple processes is not supported because CreateJobObjectA will fail due to using identical names. This PR fixes that. See https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-createjobobjecta

The specific use case that inspired this pr was running multiple hyperlight tests at the same time

### DIFF
--- a/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
+++ b/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
@@ -38,7 +38,7 @@ use windows::Win32::System::SystemServices::NUMA_NO_PREFERRED_NODE;
 use windows::Win32::System::Threading::{
     CREATE_SUSPENDED, CreateProcessA, PROCESS_INFORMATION, STARTUPINFOA,
 };
-use windows::core::{PCSTR, s};
+use windows::core::PCSTR;
 
 use super::surrogate_process::SurrogateProcess;
 use super::wrappers::{HandleWrapper, PSTRWrapper};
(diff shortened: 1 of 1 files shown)


simongdavies:
I've not looked inot it but I wonder if the way we [handle the surrogate process binary](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_host/build.rs#L31) will also cause problems?

simongdavies:
Fixes #319

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -285,12 +285,7 @@ pub(crate) fn get_surrogate_process_manager() -> Result<&'static SurrogateProces
 fn create_job_object() -> Result<HandleWrapper> {
     let security_attributes: SECURITY_ATTRIBUTES = Default::default();
 
-    let job_object = unsafe {
-        CreateJobObjectA(
-            Some(&security_attributes),
-            s!("HyperlightSurrogateJobObject"),
-        )?
-    };
+    let job_object = unsafe { CreateJobObjectA(Some(&security_attributes), PCSTR::null())? };
```
jsturtevant, line 288: Is there any reason to want a name?  Only thing I can think of is when debugging in ProcessExplorer

ludfjig: I don't think so

PR #1093, opened 2025-12-09
Unify page table manipulation code between the guest and the host

simongdavies:
Currently, the guest and the host both have code that manipulates architecture-specific page table structures: the guest has a general map operation, and the host has a much more specific routine that builds an identity map.  As we move to more complex virtual memory configurations in the guest, the host will need the ability to build more complex mappings in the guest, so this commit removes the simple implementation in the host, and replaces it with calls to the implementation originally written for the guest (now moved to `hyperlight_common` and factored into an architecture-independent interface and architecture-specific code parts).

### DIFF
--- a/src/hyperlight_host/src/mem/mgr.rs
+++ b/src/hyperlight_host/src/mem/mgr.rs
@@ -90,18 +90,17 @@ impl vmem::TableOps for GuestPageTableBuffer {
     unsafe fn read_entry(&self, addr: (usize, usize)) -> PageTableEntry {
         let b = self.buffer.borrow();
         let byte_offset = addr.0 * PAGE_TABLE_SIZE + addr.1 * 8;
-        unsafe {
-            let ptr = b.as_ptr().add(byte_offset) as *const PageTableEntry;
-            ptr.read_unaligned()
-        }
+        b.get(byte_offset..byte_offset + 8)
+            .and_then(|s| <[u8; 8]>::try_from(s).ok())
+            .map(u64::from_ne_bytes)
+            .unwrap_or(0)
     }
 
     unsafe fn write_entry(&self, addr: (usize, usize), x: PageTableEntry) {
         let mut b = self.buffer.borrow_mut();
         let byte_offset = addr.0 * PAGE_TABLE_SIZE + addr.1 * 8;
-        unsafe {
-            let ptr = b.as_mut_ptr().add(byte_offset) as *mut PageTableEntry;
-            ptr.write_unaligned(x);
+        if let Some(slice) = b.get_mut(byte_offset..byte_offset + 8) {
+            slice.copy_from_slice(&x.to_ne_bytes());
         }
     }
 


src/hyperlight_common/src/arch/amd64/vm.rs
```diff
@@ -0,0 +1,217 @@
...
+        if self.n > 0 {
+            next_vmin &= !lower_bits_mask;
+        }
+        let entry_ptr = Op::entry_addr(
+            self.request.table_base,
+            bits::<HIGH_BIT, LOW_BIT>(next_vmin) << 3,
+        );
+        let len_from_here = self.request.len - (next_vmin - self.request.vmin);
+        let max_len = (1 << LOW_BIT) - (next_vmin & lower_bits_mask);
+        let next_len = core::cmp::min(len_from_here, max_len);
```
jprendes: I'm not too sure what's going on here.
This is somewhat different from the previous implementation, but I'm not too sure what was going on there either, so... 🤷

Maybe some comments would help

src/hyperlight_common/src/arch/amd64/vm.rs
```diff
@@ -0,0 +1,217 @@
...
+/// # Safety
+/// This function modifies page table data structures, and should not be called concurrently
+/// with any other operations that modify the page tables.
+unsafe fn alloc_pte_if_needed<Op: TableOps>(
+    op: &Op,
+    x: MapResponse<Op::TableAddr>,
+) -> MapRequest<Op::TableAddr> {
+    let pte = unsafe { op.read_entry(x.entry_ptr) };
+    let present = pte & 0x1;
+    if present != 0 {
```
jprendes:
```suggestion
    let present = pte & 0x1 != 0;
    if present {
```

or even better

```suggestion
    let present = bits<0,0>(pte) != 0;
    if present {
```

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -0,0 +1,217 @@
...
+            vmin: x.vmin,
+            len: x.len,
+        };
+    }
+
+    let page_addr = unsafe { op.alloc_table() };
+
+    #[allow(clippy::identity_op)]
+    #[allow(clippy::precedence)]
+    let pte = Op::to_phys(page_addr) |
```
jprendes, line 213: why do we do `Op::to_phys(page_addr)`?
the previous code was `page_addr`

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -0,0 +1,217 @@
...
+        op.write_entry(r.entry_ptr, pte);
+    }
+}
+
+// There are no notable architecture-specific safety considerations
+// here, and the general conditions are documented in the
+// architecture-independent re-export in vm.rs
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn map<Op: TableOps>(op: &Op, mapping: Mapping) {
+    modify_ptes::<47, 39, Op>(MapRequest {
```
jprendes, line 279: I know this is code that was just moved around, but it would be nice if there was some comments explaining what's going on here.

src/hyperlight_common/src/arch/amd64/vm.rs
```diff
@@ -0,0 +1,217 @@
...
+/// This function traverses page table data structures, and should not
+/// be called concurrently with any other operations that modify the
+/// page table.
+unsafe fn require_pte_exist<Op: TableOps>(
+    op: &Op,
+    x: MapResponse<Op::TableAddr>,
+) -> Option<MapRequest<Op::TableAddr>> {
+    let pte = unsafe { op.read_entry(x.entry_ptr) };
+    let present = pte & 0x1;
+    if present == 0 {
```
jprendes:
```suggestion
    let present = bits<0,0>(pte) != 0;
    if !present {
```

src/hyperlight_common/src/arch/amd64/vm.rs
```diff
@@ -0,0 +1,217 @@
...
+        return None;
```
jprendes: Looks like this might have changed behavious, is that intentional?
This used to panic, now it retuns None.

src/hyperlight_common/src/vm.rs
```diff
@@ -0,0 +1,131 @@
...
+    /// time as mapping code using the trait.
+    unsafe fn write_entry(&self, addr: Self::TableAddr, x: PageTableEntry);
+
+    /// Convert an abstract physical address to a concrete u64 which
+    /// can be e.g. written into a table
+    fn to_phys(addr: Self::TableAddr) -> PhysAddr;
+
+    /// Convert a concrete u64 which may have been e.g. read from a
+    /// table back into an abstract physical address
+    fn from_phys(addr: PhysAddr) -> Self::TableAddr;
```
jprendes: are these comments correct? or are they swapped?

src/hyperlight_common/Cargo.toml
```diff
@@ -30,6 +30,7 @@ fuzzing = ["dep:arbitrary"]
 trace_guest = []
 mem_profile = []
 std = ["thiserror/std", "log/std", "tracing/std"]
+init-paging = []
```
jprendes, line 33: This feature is not initializaing paging, but providing data structures for managing virtual memory.
I would make the data structures unconditionally available (i.e., no feature needed), as they have no side-effects (no extra dependencies, no runtime side effects, I wouldn't feature gate it here, but would feature gate it in hl-host / hl-guest)
Alternatively, how about calling it "virtual-memory"?

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -38,20 +38,42 @@ pub fn ptov(x: u64) -> *mut u8 {
...
+        addr + offset
+    }
+    unsafe fn read_entry(&self, addr: u64) -> u64 {
+        let ret: u64;
+        unsafe {
+            asm!("mov {}, qword ptr [{}]", out(reg) ret, in(reg) addr);
+        }
+        ret
+    }
+    unsafe fn write_entry(&self, addr: u64, x: u64) {
```
jprendes:
```suggestion
    unsafe fn write_entry(&self, addr: Self::TableAddr, entry: PageTableEntry) {
```

src/hyperlight_common/src/vm.rs
```diff
@@ -0,0 +1,131 @@
...
+    ///
+    /// # Safety
+    /// This writes to the given memory address, and so all the usual
+    /// Rust things about raw pointers apply. This will also be used
+    /// to update guest page tables, so especially in the guest, it is
+    /// important to ensure that the page tables updates do not break
+    /// invariants. The implementor of the trait should ensure that
+    /// nothing else will be reading/writing the address at the same
+    /// time as mapping code using the trait.
+    unsafe fn write_entry(&self, addr: Self::TableAddr, x: PageTableEntry);
```
jprendes:
```suggestion
    unsafe fn write_entry(&self, addr: Self::TableAddr, entry: PageTableEntry);
```

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -65,64 +87,22 @@ struct MapResponse {
...
-        let mut pte: u64 = 0;
-        unsafe {
-            asm!("mov {}, qword ptr [{}]", out(reg) pte, in(reg) r.entry_ptr);
-        }
-        pte
-    })
-    .collect::<alloc::vec::Vec<u64>>();
-    if addrs.len() != 1 {
-        panic!("impossible: 1 page map request resolved to multiple PTEs");
+        vm::map::<GuestMappingOperations>(
```
jprendes:
```suggestion
        vm::map(
```

src/hyperlight_common/src/arch/amd64/vm.rs
```diff
@@ -0,0 +1,217 @@
...
+        vmin: x.vmin,
+        len: x.len,
+    })
+}
+
+// There are no notable architecture-specific safety considerations
+// here, and the general conditions are documented in the
+// architecture-independent re-export in vm.rs
+#[allow(clippy::missing_safety_doc)]
+pub unsafe fn vtop<Op: TableOps>(op: &Op, address: u64) -> Option<u64> {
```
jprendes: I king of liked better the previous name `dbg_print_address_pte`, what is `vtop`?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -40,16 +45,6 @@ use crate::{Result, log_then_return, new_error};
...
-        /// Paging Flags
-        ///
-        /// See the following links explaining paging, also see paging-development-notes.md in docs:
-        ///
-        /// * Very basic description: https://stackoverflow.com/a/26945892
-        /// * More in-depth descriptions: https://wiki.osdev.org/Paging
-        pub(crate) const PAGE_PRESENT: u64 = 1; // Page is Present
-        pub(crate) const PAGE_RW: u64 = 1 << 1; // Page is Read/Write (if not set page is read only so long as the WP bit in CR0 is set to 1 - which it is in Hyperlight)
-        pub(crate) const PAGE_USER: u64 = 1 << 2; // User/Supervisor (if this bit is set then the page is accessible by user mode code)
-        pub(crate) const PAGE_NX: u64 = 1 << 63; // Execute Disable (if this bit is set then data in the page cannot be executed)`
```
jprendes, line 52: I find this clearer than the new approach, can we bring back these constants in the hl-common crate?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -478,16 +457,10 @@ impl SandboxMemoryLayout {
...
-    //
-    // We can calculate the amount of memory needed for the PTs by calculating how much memory is needed for the sandbox configuration in total,
-    // and then add 3 * 4K (for the PML4, PDPT and PD)  to that,
-    // then add 2MB to that (the maximum size of memory required for the PTs themselves is 2MB when we map 1GB of memory in 4K pages),
-    // then divide that by 0x200_000 (as we can map 2MB in each PT).
-    // This will give us the total size of the PTs required for the sandbox to which we can add the size of the PML4, PDPT and PD.
+    // TODO: This over-counts on small sandboxes (because not all 512
+    // PTs may be required), under-counts on sandboxes with more than
+    // 1GiB memory, and would get unreasonably complicated if we
+    // needed to support hugepages.
```
jprendes, line 474: is the old comment not accurate anymore? if it is, can we keep it?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -80,6 +75,62 @@ pub(crate) struct SandboxMemoryManager<S> {
...
+impl vm::TableOps for GuestPageTableBuffer {
+    type TableAddr = (usize, usize);
+    unsafe fn alloc_table(&self) -> (usize, usize) {
+        let mut b = self.buffer.borrow_mut();
+        let page_addr = b.len();
+        b.push([0; PAGE_TABLE_ENTRIES_PER_TABLE]);
+        (page_addr, 0)
+    }
+    fn entry_addr(addr: (usize, usize), offset: u64) -> (usize, usize) {
+        (addr.0, offset as usize >> 3)
```
jprendes: I'm not sure I understand what's going on here.
I thought it would be something like:
```
let resolved_addr = PAGE_TABLE_ENTRIES_PER_TABLE * addr.0 + addr.1 + offset;
(resolved_addr / PAGE_TABLE_ENTRIES_PER_TABLE, resolved_addr % PAGE_TABLE_ENTRIES_PER_TABLE)
```
or in other words
```
let phys = Self::to_phys(addr) + offset;
Self::from_phys(phys)
```
but that's not the case, so I'm definitely missing something.

jprendes:
The following test stalls when I execute it using the changes in this PR
```rust
use hyperlight_host::{GuestBinary, UninitializedSandbox, sandbox::SandboxConfiguration};
use hyperlight_testing::simple_guest_as_string;

#[test]
fn sandboxes_create_initialized_small() {
    cfg.set_heap_size(8 * 1024 * 1024); // 8 MB

    let path = simple_guest_as_string().unwrap();
    let sbox = UninitializedSandbox::new(GuestBinary::FilePath(path), Some(cfg))
        .unwrap()
        .evolve()
        .unwrap();

    drop(sbox);

    println!("done!");
}
```

src/hyperlight_common/Cargo.toml
simongdavies: The feature is here because we need a way of not including the new vm mod if we are compiling for i686(32 bit), I agree the name is not great but it was/is an existing feature that signifies if the guest is responsible for initialising paging, anyway I'll remove it and make the exclude the vm md if we are compiling for 32 bit instead

jprendes: Do we support `i686`??
Also the implementation is gated for `x86_64` in https://github.com/hyperlight-dev/hyperlight/pull/1093/changes#diff-2db0563eeae4262a8989afd517b9d5dcd5641c517a5007010fc63624d8f1c155R17
But it would fail to compile in any non-`x86_64`, as the `arch` mod would then not be defined.

> it was/is an existing feature that signifies if the guest is responsible for initialising paging
Yes, but that's a feature for the hyperlight-host crate, and this one is hyperlight-common. There's no harm for this code to exist in hyperlight-common while the feature is not enabled in hyperlight-host.

simongdavies: I am going to leave this as is for now, when Dan intagrates support for nanvix with COW this will go away

src/hyperlight_common/src/arch/amd64/vm.rs
simongdavies: Updated the implementation

src/hyperlight_common/src/arch/amd64/vm.rs
simongdavies: Updated this

src/hyperlight_common/src/arch/amd64/vmem.rs
simongdavies: I think its because this now covers host and guest addresses which are different types

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -0,0 +1,217 @@
...
+    })
+    .map(|r| unsafe { alloc_pte_if_needed(op, r) })
+    .flat_map(modify_ptes::<38, 30, Op>)
+    .map(|r| unsafe { alloc_pte_if_needed(op, r) })
+    .flat_map(modify_ptes::<29, 21, Op>)
+    .map(|r| unsafe { alloc_pte_if_needed(op, r) })
+    .flat_map(modify_ptes::<20, 12, Op>)
+    .map(|r| unsafe { map_page(op, &mapping, r) })
+    .for_each(drop);
+}
```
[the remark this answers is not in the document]
simongdavies: Done

src/hyperlight_common/src/arch/amd64/vm.rs
simongdavies: its virtuat to physical I updated it to make it cleaer

src/hyperlight_common/src/arch/amd64/vm.rs
simongdavies: The code is only used from virt_to_phys which returns none if an address is not mapped/present, that function doesn't seem to be called anywhere at the moment , I assume its used by a later PR

src/hyperlight_host/src/mem/mgr.rs
simongdavies: Yes this is incorrect, the current code works because its always called with addr.1 == 0, however if that were not true it would be wrong so probably better to update it and add a test just in case something changes

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -80,6 +75,62 @@ pub(crate) struct SandboxMemoryManager<S> {
...
+        b[addr.0][addr.1] = x;
+    }
+    fn to_phys(addr: (usize, usize)) -> PhysAddr {
+        (addr.0 as u64 * PAGE_TABLE_SIZE as u64) + addr.1 as u64
+    }
+    fn from_phys(addr: PhysAddr) -> (usize, usize) {
+        (
+            addr as usize / PAGE_TABLE_SIZE,
+            addr as usize % PAGE_TABLE_SIZE,
+        )
```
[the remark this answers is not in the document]
simongdavies: Fixed

src/hyperlight_common/src/vm.rs
simongdavies: fixed

src/hyperlight_host/src/mem/layout.rs
simongdavies: I think both new and old are true (although we have a hard limit somewhere that prevents a sandbox allocating more than 1GB of memory so that's more something that could happen if we remove or raise than limit. Anyway I'll merge the two together.

simongdavies:
> The following test stalls when I execute it using the changes in this PR
> 
> ```rust
> use hyperlight_host::{GuestBinary, UninitializedSandbox, sandbox::SandboxConfiguration};
> use hyperlight_testing::simple_guest_as_string;
> 
> #[test]
> fn sandboxes_create_initialized_small() {
>     let mut cfg = SandboxConfiguration::default();
>     cfg.set_heap_size(8 * 1024 * 1024); // 8 MB
> 
>     let path = simple_guest_as_string().unwrap();
>     let sbox = UninitializedSandbox::new(GuestBinary::FilePath(path), Some(cfg))
>         .unwrap()
>         .evolve()
>         .unwrap();
> 
>     drop(sbox);
> 
>     println!("done!");
> }
> ```

Thanks, I fixed a bug in the `ModifyPteIterator` which was causing early termination which meant that we were not creating all the PTEs required once a sandbox went beyond 2MB, I also created an issue #1118 for the infinite loop

src/hyperlight_common/src/vm.rs
```diff
@@ -0,0 +1,142 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+
+#[cfg_attr(target_arch = "x86_64", path = "arch/amd64/vm.rs")]
+mod arch;
+
+#[cfg(target_arch = "x86_64")]
+pub use arch::{PAGE_NX, PAGE_PRESENT, PAGE_RW};
```
syntactically: IMO the purpose of the not-inside-arch vm module is to provide an architecture-independent veneer so that other code does not need to worry about the architecture-specific details which are confined to the private module.  I do not like adding architecture-specific exports to the top level module. If it is really necessary in order to allow a test in another file or something (but I am not sure if it is?), perhaps `pub mod arch` would be a way to expose them that at least made it easy to tell which code in fact needed to worry about architecture-dependent things (`use vm::...` vs `use vm::arch::...`).

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -501,17 +494,27 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
     use hyperlight_testing::simple_guest_as_string;
 
     use super::*;
     use crate::GuestBinary;
+    use crate::mem::memory_region::MemoryRegionType;
     use crate::mem::shared_mem::GuestSharedMemory;
     use crate::sandbox::SandboxConfiguration;
     use crate::sandbox::uninitialized::UninitializedSandbox;
 
+    pub(crate) const PML4_OFFSET: usize = 0x0000;
```
syntactically, line 519: Why were these constants reintroduced? A major point of this commit is to _remove_ assumptions throughout the rest of the code about the exact structure of the page tables, which are liable to be broken as the virtual memory setup gets more sophisticated.

If it's just for the tests below, I would envision breaking those up into unit tests for the mapping and vtop code & separately tests of the sandbox creation which could more generically just traverse the address space with vtop after setting it up and make sure it is reasonable.

src/hyperlight_host/src/mem/mgr.rs
syntactically: Semantically, I don't think it makes sense to offset an entry address by another entry address? I would prefer to just panic if the old offset was not zero I think.

src/hyperlight_host/src/mem/layout.rs
syntactically: @simongdavies I am not sure that there is any fundamental limitation on >1GB sandboxes anymore? I think that came from the old page table creation code.

src/hyperlight_common/Cargo.toml
syntactically: I agree with Jorge. If it needs to build on an architecture we should just add another arch mod for that architecture.

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -14,8 +14,63 @@ See the License for the specific language governing permissions and
...
+//! allocating intermediate tables as needed and setting appropriate flags on leaf PTEs.
+
 use crate::vm::{Mapping, MappingKind, TableOps};
 
+/// Paging Flags
+///
+/// See the following links explaining paging:
+///
+/// * Very basic description: https://stackoverflow.com/a/26945892
+/// * More in-depth descriptions: https://wiki.osdev.org/Paging
```
syntactically: Please replace these references with the authoritative ones: the AMD/Intel manuals. The AMD ref is AMD64 Architecture Programmer’s Manual, Volume 2: System Programming, Section 5.3: Long-Mode Page Translation; I don't have the Intel one to hand.

src/hyperlight_common/src/arch/amd64/vm.rs
```diff
@@ -104,12 +195,12 @@ unsafe fn alloc_pte_if_needed<Op: TableOps>(
     #[allow(clippy::identity_op)]
     #[allow(clippy::precedence)]
     let pte = Op::to_phys(page_addr) |
-        1 << 5 | // A   - we don't track accesses at table level
-        0 << 4 | // PCD - leave caching enabled
-        0 << 3 | // PWT - write-back
-        1 << 2 | // U/S - allow user access to everything (for now)
-        1 << 1 | // R/W - we don't use block-level permissions
-        1 << 0; // P   - this entry is present
+        PAGE_ACCESSED_CLEAR | // we don't track accesses at table level
```
syntactically: The comment is wrong for setting the A bit to 0.

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -215,3 +318,420 @@ pub const PAGE_TABLE_SIZE: usize = 4096;
...
+mod tests {
+    use alloc::vec;
+    use alloc::vec::Vec;
+    use core::cell::RefCell;
+
+    use super::*;
+    use crate::vm::{BasicMapping, Mapping, MappingKind, PAGE_TABLE_ENTRIES_PER_TABLE, TableOps};
+
+    /// A mock TableOps implementation for testing that stores page tables in memory
+    struct MockTableOps {
```
syntactically, line 351: Is this materially different from the GuestPageTableBuffer?

src/hyperlight_common/src/arch/amd64/vmem.rs
ludfjig: intel vol3,  chapter 5 paging. note it's called XD on intel

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -80,6 +66,64 @@ pub(crate) struct SandboxMemoryManager<S> {
     pub(crate) abort_buffer: Vec<u8>,
 }
 
+#[cfg(feature = "init-paging")]
+struct GuestPageTableBuffer {
+    buffer: std::cell::RefCell<Vec<[PageTableEntry; PAGE_TABLE_ENTRIES_PER_TABLE]>>,
```
ludfjig: Is the refcell here necessary?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -80,6 +66,64 @@ pub(crate) struct SandboxMemoryManager<S> {
...
+    }
+    fn into_bytes(self) -> Box<[u8]> {
+        let bx = self.buffer.into_inner().into_boxed_slice();
+        let len = bx.len();
+        unsafe {
+            Box::from_raw(std::ptr::slice_from_raw_parts_mut(
+                Box::into_raw(bx) as *mut u8,
+                len * PAGE_TABLE_SIZE,
+            ))
+        }
```
ludfjig: Do we need the unsafe code here? I am not 100% convinced this is not UB. If this is only used to get a reference to the internal buffer, maybe it can be done in different but safe way?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -639,35 +579,55 @@ mod tests {
     }
 
     /// Get expected flags for a memory region type
+    /// Previously we set User RW flag for some regions but new code for building page tables does
```
ludfjig: Personally I think that comments that mention "previous code" tend to get outdated very fast and are not very useful, same for the comment below

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -142,144 +182,37 @@ where
...
-                }
-            }
-
-            // Write the entire PTE buffer to shared memory in a single operation
-            // Convert u64 buffer to bytes for writing to shared memory
-            let pte_bytes = unsafe {
-                std::slice::from_raw_parts(pte_buffer.as_ptr() as *const u8, pte_buffer.len() * 8)
-            };
-            shared_mem.copy_from_slice(pte_bytes, SandboxMemoryLayout::PT_OFFSET)?;
+            shared_mem.copy_from_slice(&buffer.into_bytes(), SandboxMemoryLayout::PML4_OFFSET)?;
```
ludfjig, line 218: This is very clean, big fan

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -142,144 +182,37 @@ where
...
-                let val_to_write: u64 = (SandboxMemoryLayout::PT_GUEST_ADDRESS as u64
-                    + (i * 4096) as u64)
-                    | PAGE_PRESENT
-                    | PAGE_RW;
-                shared_mem.write_u64(offset, val_to_write)?;
+            let buffer = GuestPageTableBuffer::new();
+            for region in regions.iter() {
+                let readable = region.flags.contains(MemoryRegionFlags::READ);
+                let writable = region.flags.contains(MemoryRegionFlags::WRITE)
+                    // Temporary hack: the stack guard page is
```
jsturtevant, line 198: what makes this temporary?  do we know the fix but can't do it to a later changeset? something that needs to be looked into?

syntactically: The host management of the stack guard page is removed entirely in https://github.com/hyperlight-dev/hyperlight/commit/6211e9856f33c3c9bcc9b0a5c797f73d8254b0d1 @jsturtevant

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: Something like `we don't set User RW flag since we do not run code in user mode` would be easier to read anyway.

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -80,6 +66,64 @@ pub(crate) struct SandboxMemoryManager<S> {
     pub(crate) abort_buffer: Vec<u8>,
 }
 
+#[cfg(feature = "init-paging")]
+struct GuestPageTableBuffer {
+    buffer: std::cell::RefCell<Vec<[PageTableEntry; PAGE_TABLE_ENTRIES_PER_TABLE]>>,
+}
+#[cfg(feature = "init-paging")]
+impl vm::TableOps for GuestPageTableBuffer {
```
jsturtevant: It seems we could directly use this in the tests above and would be testing the actual implementation here.  I don't see anything that is hardware bound.  (I believe this is what Lucy mentioned above)

src/hyperlight_host/src/mem/mgr.rs
syntactically: Taking a second look at this, you are totally right---the alignment of the memory is not right, because Vec allocates with an alignment matching that of the slice (which itself matches the contents, which is a u64), but the box type it is being transmuted into has a layout with smaller alignment. Thank you for catching that!!

This could be fixed with Aligned, and I think it would be safe then, since the other guarantees around being valid for [u8] are I think met, since all bit patterns are valid for u8 and a sequence of locations valid for u8 should be valid for [u8].  

[`bytemuck::allocation::cast_slice_box`](https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_slice_box.html) is basically the same function but more generic/with dynamic error checking; perhaps it makes sense to switch to that.

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -65,64 +87,22 @@ struct MapResponse {
 ///   if previously-unmapped ranges are not being mapped, TLB invalidation may need to be performed afterwards.
 #[instrument(skip_all, parent = Span::current(), level= "Trace")]
 pub unsafe fn map_region(phys_base: u64, virt_base: *mut u8, len: u64) {
```
jsturtevant, line 89: we have pretty minimal test coverage on this from an integration perspective, is there any improvements we could make while we are here or is set of tests we have good enough?

src/hyperlight_host/src/mem/mgr.rs
simongdavies: I don't think we should panic , I think we should either leave it as is or we should alter the parameters to make it correct

src/hyperlight_common/src/vm.rs
simongdavies: Yes this is a bit of a pain, the best option would have been to put the test in the common module, but that led to circular references , I'll do the `pub mod arch` change, it may be that as this series of PRs progresses the test will be able to be modified to be put in the common module

src/hyperlight_host/src/mem/mgr.rs
simongdavies: Yes, I wanted some tests just to validate that this PR created the same Pages/entries as we created previously (other than the change to not create all the PD entries), since there was a bug where the memory size went beyond 2MB that was not caught. This test will need to change as we move forward and those constants will disappear

src/hyperlight_common/src/arch/amd64/vmem.rs
simongdavies: Maybe not, will fix it

src/hyperlight_host/src/mem/mgr.rs
simongdavies: I think because the trait defines `read entry` and `write_entry` to take `&self` then this is correct.

src/hyperlight_host/src/mem/mgr.rs
simongdavies: OK, I'll update it , this test is going to get changed in the future anyway, it was more a comment as to why it had changed in this PR.

src/hyperlight_host/src/mem/mgr.rs
simongdavies: yes will update

src/hyperlight_guest_bin/src/paging.rs
simongdavies: I think we need more tests but probably better to do this at the end of these changes since I suspect this function will need to be updated , created #1123 for this

src/hyperlight_common/src/arch/amd64/vmem.rs
simongdavies: I remember the issue now `GuestPageTableBuffer` is in `hyperlight_host` so that would create a circular dependency, so I can move the tests or keep the Mock , for now I will leave as is

src/hyperlight_host/src/mem/mgr.rs
simongdavies: As noted above we cannot do this as it will create a circular dependency

src/hyperlight_host/src/mem/mgr.rs
ludfjig: here's the playground, you can run under miri and indeed it's UB: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=50345655db37b09f5e904b12f2af9937

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -0,0 +1,752 @@
...
+        Some(pte)
+    } else {
+        None
+    }
+}
+
+#[inline(always)]
+/// Utility function to extract an (inclusive on both ends) bit range
+/// from a quadword.
+fn bits<const HIGH_BIT: u8, const LOW_BIT: u8>(x: u64) -> u64 {
```
ludfjig: nit: comment above the attribute

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -0,0 +1,752 @@
...
+/// This function walks the 4-level page table hierarchy (PML4 → PDPT → PD → PT),
+/// allocating intermediate tables as needed via `alloc_pte_if_needed`, and finally
+/// writing the leaf page table entries with the requested permissions via `map_page`.
+///
+/// The iterator chain processes each level:
+/// 1. PML4 (47:39) - allocate PDPT if needed
+/// 2. PDPT (38:30) - allocate PD if needed
+/// 3. PD (29:21) - allocate PT if needed
+/// 4. PT (20:12) - write final PTE with physical address and flags
+pub unsafe fn map<Op: TableOps>(op: &Op, mapping: Mapping) {
```
ludfjig: nit: comment above the attribute

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -0,0 +1,752 @@
...
+        table_base: Op::from_phys(pte & PTE_ADDR_MASK),
+        vmin: x.vmin,
+        len: x.len,
+    })
+}
+
+// There are no notable architecture-specific safety considerations
+// here, and the general conditions are documented in the
+// architecture-independent re-export in vm.rs
+#[allow(clippy::missing_safety_doc)]
```
ludfjig: nit: comment above attribute (or better, put a safety comment)

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -82,7 +80,10 @@ use crate::{Result, new_error};
 ///   of this field is returned by the `stack_size()` method of this struct. in reality,
 ///   the stack might be slightly bigger or smaller than this value since total memory
 ///   size is rounded up to the nearest 4K, and there is a 16-byte stack guard written
-///   to the top of the stack. (see below for more details)
+///   to the top of the stack. (see below for more details
+#[cfg(feature = "init-paging")]
+// The amount of memory that can be mapped per page table
+const AMOUNT_OF_MEMORY_PER_PT: usize = 0x200_000;
```
ludfjig: nit: `//` should probably be `///`. And also comment above the attriubute

src/hyperlight_host/src/mem/layout.rs
ludfjig: what happened to this?

simongdavies: Sorry missed this, we still have this limit defined [here](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_host/src/mem/layout.rs#L244) and still reference it in our docs so the comment in the code is still correct. I think the reason that the page tables were created that way was that we had a somewhat arbitrary limit of 1GB , we should think about what we want that limit to be, if we want it to be configurable etc.

simongdavies: #1124

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: What changed that made this not UB?

simongdavies: Tried [`bytemuck::allocation::cast_slice_box`](https://docs.rs/bytemuck/latest/bytemuck/allocation/fn.cast_slice_box.html) and still got UB (thanks @ludfjig for confirming) Tried Align , ended up using a Vec, similar perf characteristics for `read` and `write` entry and zero allocation, zero copy `into_bytes` with `into_boxed_slice`

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -80,6 +65,75 @@ pub(crate) struct SandboxMemoryManager<S> {
...
+        Self::from_phys(phys)
+    }
+
+    unsafe fn read_entry(&self, addr: (usize, usize)) -> PageTableEntry {
+        let b = self.buffer.borrow();
+        let byte_offset = addr.0 * PAGE_TABLE_SIZE + addr.1 * 8;
+        unsafe {
+            let ptr = b.as_ptr().add(byte_offset) as *const PageTableEntry;
+            ptr.read_unaligned()
+        }
```
ludfjig: I think these unsafe can be avoided by just indexing into a range of the vec<u8> buffer and then convert the slice to u64

ludfjig (approved):
Approved. But we should follow up to deal with the potential silent errors

PR #1095, opened 2025-12-09
Nix updates

simongdavies:
Updates to flake.nix


jsturtevant:
LGTM but I know nothing of NIX, but I also think it is mostly dev configuration.  @dblnz maybe?

jprendes:
> LGTM but I know nothing of NIX, but I also think it is mostly dev configuration. @dblnz maybe?

ditto

dblnz (approved):
I tried it locally, it installs rust 1.90

PR #1098, opened 2025-12-09
Update snapshots

simongdavies:
This PR contains two changes to snapshots:

It moves `mem::shared_mem_snapshot::SharedMemorySNapshot` into `sandbox::snapshot::Snapshot`, because that structure has always contained more than just a snapshot of a shared memory. It also removes the Arc reference counting wrapper from the new structure, and shifts the Arc wrapper into  `sandbox::initialized_multi_use::MultiUseSandbox`, which is where it was actually needed.

Previously, snapshots were compared for equality by checking if they were literally the same object in memory. This replaces that comparison with a deterministic hash of the contents of the sandbox, which can be persisted to and reloaded from disk.

### DIFF
--- a/src/hyperlight_host/src/error.rs
+++ b/src/hyperlight_host/src/error.rs
@@ -32,7 +32,7 @@ use thiserror::Error;
 
 #[cfg(target_os = "windows")]
 use crate::hypervisor::wrappers::HandleWrapper;
-use crate::mem::memory_region::MemoryRegionFlags;
+use crate::mem::memory_region::{MemoryRegion, MemoryRegionFlags};
 use crate::mem::ptr::RawPtr;
 
 /// The error type for Hyperlight operations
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/error.rs
```diff
@@ -322,7 +330,9 @@ impl HyperlightError {
             | HyperlightError::PoisonedSandbox
             | HyperlightError::ExecutionAccessViolation(_)
             | HyperlightError::StackOverflow()
-            | HyperlightError::MemoryAccessViolation(_, _, _) => true,
+            | HyperlightError::MemoryAccessViolation(_, _, _)
+            | HyperlightError::SnapshotSizeMismatch(_, _)
+            | HyperlightError::MemoryRegionSizeMismatch(_, _, _) => true,
```
jprendes, line 335: Are these poisoning? I feel like they aren't.
Usually restoring would clear the poison flag, and in case of this error we shouldn't clear it, but I don't think it should raise the poison flag. Or am I missing something?

simongdavies: I was in two minds, I thought that since something bad must have happened for us to end up with this error it was the safest thing

jprendes: I think it would be posoning if we ever decide to throw one of these in the middle of a restore, but all of this throws during validation before the sandbox memory is modifued.
If you want to err on the side of caution, I'm happy to go with that and a comment to revisit more thoroughly.

PR #1100, opened 2025-12-09
Remove `HyperlightPEB::code_ptr`

simongdavies:
This may have been used when we had in-process mode, but it certainly isn't anymore.


ludfjig (approved):

jprendes (approved):
LGTM

PR #1105, opened 2025-12-10
Move and rename some Vm traits

ludfjig:
Some follow up items from #924. Please review commit by commit

- Moves kvm.rs/hyperv_linux.rs/hyperv_windows.rs into new vm module. 
- Renames Hyperv_linux.rs to mshv.rs. Renames hyperv_windows.rs to whp.rs
- Renames `Hypervisor` trait to `Vm` trait
- Renames `HyperlightExit` to `VmExit`
- Moves some hypervisor specific code out from sandbox module, to hypervisor/vm module

Note that this PR does not modify any code (with exception of renames and moves)

### DIFF
--- a/src/hyperlight_host/src/hypervisor/gdb/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/gdb/mod.rs
@@ -276,7 +276,7 @@ pub(crate) enum DebugResponse {
 }
 
 /// Trait for VMs that support debugging capabilities.
-/// This extends the base Hypervisor trait with GDB-specific functionality.
+/// This extends the base VirtualMachine trait with GDB-specific functionality.
 pub(crate) trait DebuggableVm: VirtualMachine {
     /// Translates a guest virtual address to a guest physical address
     fn translate_gva(&self, gva: u64) -> crate::Result<u64>;
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
@@ -23,12 +23,13 @@ use crate::Result;
 use crate::hypervisor::regs::{CommonFpu, CommonRegisters, CommonSpecialRegisters};
 use crate::mem::memory_region::MemoryRegion;
 
+/// KVM (Kernel-based Virtual Machine) functionality (linux)
 #[cfg(kvm)]
-/// Functionality to manipulate KVM-based virtual machines
 pub(crate) mod kvm;
-/// HyperV-on-linux functionality
+/// MSHV (Microsoft Hypervisor) functionality (linux)
 #[cfg(mshv3)]
 pub(crate) mod mshv;
+/// WHP (Windows Hypervisor Platform) functionality (windows)
 #[cfg(target_os = "windows")]
 pub(crate) mod whp;
 


src/hyperlight_host/src/hypervisor/vm/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+use tracing::{Span, instrument};
+
+use crate::Result;
+use crate::hypervisor::regs::{CommonFpu, CommonRegisters, CommonSpecialRegisters};
+use crate::mem::memory_region::MemoryRegion;
+
+#[cfg(kvm)]
+/// Functionality to manipulate KVM-based virtual machines
+pub(crate) mod kvm;
+/// HyperV-on-linux functionality
```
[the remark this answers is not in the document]
ludfjig: fixed

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+use crate::hypervisor::regs::{CommonFpu, CommonRegisters, CommonSpecialRegisters};
+use crate::mem::memory_region::MemoryRegion;
+
+#[cfg(kvm)]
+/// Functionality to manipulate KVM-based virtual machines
+pub(crate) mod kvm;
+/// HyperV-on-linux functionality
+#[cfg(mshv3)]
+pub(crate) mod mshv;
+#[cfg(target_os = "windows")]
```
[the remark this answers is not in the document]
ludfjig: fixed

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -276,7 +277,7 @@ pub(crate) enum DebugResponse {
 
 /// Trait for VMs that support debugging capabilities.
 /// This extends the base Hypervisor trait with GDB-specific functionality.
```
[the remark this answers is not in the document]
ludfjig: fixed

ludfjig:
> I like the change from `Hypervisor` → `Vm`. I agree that it is more accurate. Same for `VmExit` over `HyperlightExit`.
> 
> Though, iirc, in the past, when we considered changing from `hyperv_*` to `mshv`/`whp`, we opted to keep the previous names.

THanks. I think the renaming was discarded in the past due to making a PR too hard to review

danbugs:
> > I like the change from `Hypervisor` → `Vm`. I agree that it is more accurate. Same for `VmExit` over `HyperlightExit`.
> > Though, iirc, in the past, when we considered changing from `hyperv_*` to `mshv`/`whp`, we opted to keep the previous names.
> 
> THanks. I think the renaming was discarded in the past due to making a PR too hard to review

I think Simon also mentioned the `hyperv_*` names are more descriptive: https://github.com/hyperlight-dev/hyperlight/pull/517#discussion_r2102621326

jsturtevant:
LGTM, I tend to agree that I perfer hyperv_linux and hyperv_windows over mshv and whp.

ludfjig:
I like whp (windows hypervisor platform) because it's the name of the API we use https://learn.microsoft.com/en-us/virtualization/api/hypervisor-platform/hypervisor-platform

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -20,22 +20,14 @@ use crate::Result;
...
 #[cfg(gdb)]
 pub(crate) mod gdb;
 
 /// Abstracts over different hypervisor register representations
 pub(crate) mod regs;
 
-#[cfg(kvm)]
-/// Functionality to manipulate KVM-based virtual machines
-pub(crate) mod kvm;
+pub(crate) mod vm;
```
simongdavies: Can we call this virtual_machine, #1093 introduces a vm module too (we should probably rename that to virtual_memory a well

src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -276,7 +277,7 @@ pub(crate) enum DebugResponse {
...
-pub(crate) trait DebuggableVm: Hypervisor {
+pub(crate) trait DebuggableVm: Vm {
```
simongdavies: Again can we consider calling this `VirtualMachine` and `DebuggableVirtualMachine`? I agree with the rename its just the ambiguity with vm as an abbreviation for VirtualMemory

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -12,6 +14,77 @@ pub(crate) mod mshv;
...
+                }
+            } else if #[cfg(mshv3)] {
+                if mshv::is_hypervisor_present() {
+                    Some(HypervisorType::Mshv)
+                } else {
+                    None
+                }
+            } else if #[cfg(target_os = "windows")] {
+                if whp::is_hypervisor_present() {
+                    Some(HypervisorType::Whp)
```
simongdavies, line 67: I just realised that we have code in unitializedsandbox::new that checks for supported versions of windows , I don't know why that code is there and not wrapped up in the check for whp presence here , IIRC that code is to check that we are running on Windows 11 or Windows 2022 or later because we need an API that is only available on those Os versions

simongdavies:
> > > I like the change from `Hypervisor` → `Vm`. I agree that it is more accurate. Same for `VmExit` over `HyperlightExit`.
> > > Though, iirc, in the past, when we considered changing from `hyperv_*` to `mshv`/`whp`, we opted to keep the previous names.
> > 
> > 
> > THanks. I think the renaming was discarded in the past due to making a PR too hard to review
> 
> I think Simon also mentioned the `hyperv_*` names are more descriptive: [#517 (comment)](https://github.com/hyperlight-dev/hyperlight/pull/517#discussion_r2102621326)

Well I must be mellowing rapidly since I don't really mind that renaming, I'm easy either way (although being totally contradictory to my previous comments the new names do have a certain amount of symmetry to them). Not sure about VM abbreviation though and don't like ::vm::vm much either

src/hyperlight_host/src/hypervisor/mod.rs
jsturtevant: gosh I just reviewed that and didn't notice that was vm as in virtual_memory.  Should #1093 also rename vm->virtual_memory?

simongdavies: I think so I will make that change

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
ludfjig: I think this that is fine as it will return useful error messages. Moving it to is_hypervisor_present would only return true/false

PR #1112, opened 2025-12-12
docs: add HIP 0001 ring buffer proposal for Hyperlight I/O

andreiltd:
[Rendered](https://github.com/andreiltd/hyperlight/blob/tandr/rng-rfc/proposals/0001-rng-buf/README.md)


proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+
+#### Story 2: High-throughput RPC
+
+As a Hyperlight developer, I want to make multiple function calls from guest to host with minimal
+overhead. The ring buffer should let me queue multiple requests, suppress notifications, and process
+responses in batches.
+
+### Risks and Mitigations
+
+**Risk**: Malicious guest corrupting the queue **Mitigation**: Do not expose low level queue API to
```
jsturtevant: I don't know that not exposing it to the guest does anything.  The guest is fully untrusted and so we need to do the mitigation suggested after this, that we serialize all data as known types and do assertions as suggested.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+
+```
+
+When the `INLINE` flag is set, the `addr` is unused. This optimization, inspired by io_uring,
+eliminates memory indirection for common small messages, improving both latency and cache behavior.
+The tradeof is the increased size of descriptor table. Alternatively we could repurpose the `addr`
+and `len` as raw bytes providing 12 bytes of inline storage. We should asses if any of flatbuffer
+schema serialized data can actually fit into small inline data.
+
+**4. Descriptor Chaining - scatter gather list**
```
jsturtevant, line 363: Is it proposed to implement this now or as future option?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+  reached or when the ring is about to fill up
+
+**2. Event based notifications**
+
+In the single threaded application the notification involve VM exit but in multi-thread environment
+where host and guest are running in separate threads we can laverage event-based notifications (for
+example `ioeventfd` for kvm). This is especially useful for streaming scenarios where the guest can
+continue processing while the host consumes data asynchronously.
+
+**3. Inline Descriptors**
```
jsturtevant, line 319: Is this proposed to be implemented now or as possible improvement in the future

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+the driver pre-allocates response buffers, but the device may produce variable-length responses.
+This means that the pre allocated size might not be enough to write a complete response. The
+proposed solution to that is to use truncation protocol. The protocol can be implemented in the
+descriptor layer or in the flatbuffer schema:
+
+1. Driver allocates buffer of estimated size
+2. Device writes up to buffer length
+3. Device sets actual written length in descriptor
+4. If `actual_length > buffer_length`, device sets a `TRUNCATED` flag,
+5. Driver can re-submit with larger buffer if needed
```
jsturtevant, line 384: I can image this would be possible performance issue? Can we provide a metric for when this occurs and a way to improve initial buffer size?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+
+### Snapshotting
+
+Snapshotting requires that the descriptor table has no in-flight guest-to-host requests and any
+attempt to snapshot a sandbox with such pending requests will result in a snapshot failure.
+
+### Difference from spec
```
jsturtevant: from what spec?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+1. Driver allocates buffer of estimated size
+2. Device writes up to buffer length
+3. Device sets actual written length in descriptor
+4. If `actual_length > buffer_length`, device sets a `TRUNCATED` flag,
+5. Driver can re-submit with larger buffer if needed
+
+### Snapshotting
+
+Snapshotting requires that the descriptor table has no in-flight guest-to-host requests and any
+attempt to snapshot a sandbox with such pending requests will result in a snapshot failure.
```
jsturtevant, line 389: Can we ensure this?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+
+This HIP proposes implementing a ring buffer mechanism for Hyperlight I/O, loosely based on the
+virtio packed virtqueue
+[specification](https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-720008).
+The ring buffer will serve as the foundation for host-guest communication, supporting use cases such
+as streams, RPC, and other I/O patterns in both single-threaded and multi-threaded environments.
+
+The design leverages virtio's well-defined publishing/consumption semantics and memory safety
+guarantees while adapting them to Hyperlight's specific needs. Since we control both ends of the
+queue, we can deviate from strict virtio compliance where it makes sense.
```
jsturtevant: could we link to the section below?

proposals/0001-rng-buf/README.md
jsturtevant: Oh from from virtio spec, let's link the mention above to here

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,749 @@
...
+
+    /// Try to receive a message without blocking
+    pub fn try_recv<T>(&mut self) -> Result<T, RingError>
+    where
+        T: FlatbufferSerializable;
+}
+
+```
+
+### Test Plan
```
jsturtevant, line 732: love seeing the comprehensive testing!

jsturtevant, line 732: Maybe worth outlining the implemenation plan a bit?  Will this come in a couple different PRs or behind feature flags, etc?

jsturtevant (dismissed):
This looks great, thanks for such a detailed write up.  Left a few minor comments but otherwise this seems like it would set up the project to work well for the future

proposals/0001-rng-buf/README.md
andreiltd: This will be implemented in the first draft.

proposals/0001-rng-buf/README.md
andreiltd: I would like to give it a try in the first draft, but really if we're gonna keep it or not depends on two factors: 1) can we actually fit any meaningful data into inline slot, e.g. can serialized function call with single argument fit? 2) the performance improvement for such calls is substantial. Both are hard to assess without actual implementation.

proposals/0001-rng-buf/README.md
andreiltd: That's a good point. I would imagine that this only happens when you try to transport a "big" chunk of data for which the base case should be using stream rather then pushing everything at once. That being said the protocol should support full round trip but really the mitigation of this problem should be using streams.

proposals/0001-rng-buf/README.md
andreiltd: Yes, we can query the queue for number of inflight messages during snapshotting and fail if it reports positive number. Because we have to track the number of free slots in the queue getting number of inflight messages is just length of the queue minus free slots.

proposals/0001-rng-buf/README.md
andreiltd: I've added Implementation Plan section.

proposals/0001-rng-buf/README.md
andreiltd: I've changed this to mention `virtio`.

danbugs:
Really great work on this HIP, @andreiltd! Loved seeing the diagrams, extensive testing plan, code, and whatnot.

A few thoughts:

**1. Side-by-side comparison suggestion**

I think the HIP could benefit from a more explicit side-by-side comparison in the "Comparison with current implementation" section. Something like a table showing:
- Current: `push_shared_output_data()`-> `outb` -> VM Exit -> `pop_shared_input_data()` (per call)
- Proposed: `submit_available()` × N -> `notify()` → VM Exit → `poll_used()` × N (batched)

I feel like this would help readers quickly grasp the key difference in the two approaches.

**2. Related work in Nanvix fork**

I've done some similar work in the Nanvix hyperlight fork that might be relevant: https://github.com/hyperlight-dev/hyperlight/commit/76c9e7c83800ec6a74df63c52e94a0eed89a4a54

There, I added a `credits_value` field to HyperlightPEB that the guest can check before performing a VMExit. This way, the guest avoids unnecessary exits when the host signals no resources are available. This is similar to ring buffer notification suppression / checking `num_free()` before submitting, so I feel like Hyperlight-Nanvix could really benefit from this work!

**3. Single-threaded vs multi-threaded phasing**

For the multi-threaded case (guest and host on separate threads), the host would need a way to interrupt the running guest when responses are ready--yes? For that, I've also done related work for that in Nanvix's Hyperlight fork: https://github.com/hyperlight-dev/hyperlight/commit/a897cad907cb6af4d62708649b3759c763af2a87 for KVM. 

My hardware interrupts work enables PIC/APIC interrupt injection which could support the ioeventfd-style notifications mentioned in the HIP. Happy to coordinate on this when the time comes :)

---

The backward compatibility goal (current function call model portable without public API changes) is really cool to see--great for downstream consumers like Nanvix.

Looking forward to seeing this land!

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+
+#### Story 2: High-throughput RPC
+
+As a Hyperlight developer, I want to make multiple function calls from guest to host with minimal
+overhead. The ring buffer should let me queue multiple requests, suppress notifications, and process
+responses in batches.
+
+### Risks and Mitigations
+
+**Risk**: Malicious guest corrupting the queue **Mitigation**: Do not expose low level queue API to
```
syntactically: I agree with @jsturtevant here---the raw state of the queue is fundamentally exposed to the guest. The thing that we need to do on the host for safety is figure out what that means for us: making sure that the buffers are always in guest memory, ensuring that we do not access them in racy ways (even if the guest tries to trick us into having the same buffer both read from and written to at the same time), etc.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+**Risk**: Potential deadlocks in backpressure scenarios **Mitigation**: Provide clear documentation
+of blocking behavior; consider timeout mechanisms; shuttle testing
+
+**Risk**: Memory safety issues with shared buffer management **Mitigation**: Employ a strong type
+system; implement ownership tracking; perform thorough validation and fuzzing
+
+## Design Details
+
+### Packed Virtqueue Overview
+
```
syntactically, line 131: I don't love the producer/consumer terminology here. I think it added to the confusion before about how we should have buffer allocation responsibility be divided.  How about something like "the `driver` is the side that can allocate buffers accessible by both sides (i.e. the guest), and the `device` is the side that reads from/writes to these buffers on request (i.e. the `host`).  The `host` is essentially providing a specialised paravirtualised device to the guest, and so always takes on the `device` role.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+Before diving into the specifics, it will be useful to agree on some terminology. We borrow terms
+specific for virtio spec. Throughout this document, when we refer to the `driver`, we mean the
+`producer` side of the queue (the side that submits elements). When we refer to the `device`, we
+mean the `consumer` side (the side that processes queue elements). In the virtio specification,
+these terms come from the traditional device driver model, but in our case, either the host or guest
+can act as driver or device depending on the direction of communication.
+
+#### Bidirectional Communication
+
```
syntactically, line 147: How about something like:
> Hyperlight needs to support both device-initiated (for host->guest function calls) and driver-initiated (for guest->host function calls) communication. This is generally accomplished by having the guest create during early initialisation and keep populated a set of available buffers into which the host should next write incoming work for the guest.

I am not fundamentally opposed to having split rx/tx queues, like e.g. virtio-net, but unsure if they actually make sense here, at least in the single-threaded case?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+1. Host-to-Guest Queue: Host acts as driver (producer), guest acts as device (consumer), e.g. queue
+   elements represent host to guest function calls,
+2. Guest-to-Host Queue: Guest acts as driver (producer), host acts as device (consumer), e.g. queue
+   elements represent guest to host function calls,
+
+Each queue is independent and implements the same packed virtqueue semantics described below.
+
+#### Memory Layout
+
+The queue structure reside in shared memory and are accessible to both host and guest. The layout
```
syntactically, line 159: Should this be an offset from the mapped region base, or a GPA? I would tend to think GPA, even though in most cases the host will want to ensure that it is in the scratch region and then convert to an offset from there.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+    - Write descriptors: Point to empty buffers where the device should write responses
+3. Device processes request: The device reads data from read-buffers and writes results into
+   write-buffers
+4. Device marks completion: The device updates the descriptor flags to indicate completion
+
+**Step 1:** Driver submits request with read buffer (request) and write buffer (response)
+
+<img width="1324" height="726" alt="submit" src="https://github.com/user-attachments/assets/a3ee2dea-a55b-4d50-8b96-1702617a21f0" />
+
+**Step 2:** Device processes and writes response
```
syntactically, line 197: I assume that in this picture there is implicitly some header inside the buffers passed where some kind of call ID allocated by the caller can be used to link up calls/responses?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+### Dynamic Response Sizing
+
+A slightly annoying consequence of using virtio model is that we have to account for the fact that
+the driver pre-allocates response buffers, but the device may produce variable-length responses.
+This means that the pre allocated size might not be enough to write a complete response. The
+proposed solution to that is to use truncation protocol. The protocol can be implemented in the
+descriptor layer or in the flatbuffer schema:
+
+1. Driver allocates buffer of estimated size
+2. Device writes up to buffer length
```
syntactically, line 381: Is this flag set in the queue flags, or in the header protocol inside of the buffers?  The latter I assume needs to keep track of both a request id and an offset from the beginning of the response to that request for this buffer, for this to work?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,769 @@
...
+3. Device sets actual written length in descriptor
+4. If `actual_length > buffer_length`, device sets a `TRUNCATED` flag,
+5. Driver can re-submit with larger buffer if needed
+
+### Snapshotting
+
+Snapshotting requires that the descriptor table has no in-flight guest-to-host requests and any
```
syntactically, line 388: I guess there is a slightly more complex issue with snapshotting as well: the state of the buffer pool in the driver will need to be considered/updated, because (a) the physical addresses of the inflight "write the next request for me here" buffers will need to be updated to the new scratch region and (b) something will need to be done (perhaps the same update? but slightly unclear, in case the guest is currently using data in those buffers) with the buffers in the allocator pool that are not currently available in the queue.

syntactically (commented):
I mostly really like this!  Left a few minor nits, and a couple of bigger comments clarifying the question of where buffers get allocated.

I didn't review the Rust type design in detail, as I understand that is going to be updated shortly.

syntactically (commented):
Sorry, some kind of glitch seems to have resulted in my comments getting posted twice. I can't delete this duplicate review comment, apparently.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+
+The driver will publish buffers until there is no space left in the descriptor ring, at which point
+it must wait for the device to process some descriptors before it can continue. Both publishing and
+processing wrap around when reaching the end of the descriptor table, with the wrap counter flipping
+to indicate the beginning of a new round through the ring.
+
+This mechanism ensures that no locks are required for synchronization, only memory barriers combined
+with atomic publishing of flags ensure that the other side will never observe a partial update:
+
+- Driver: Write descriptor fields ? memory barrier ? atomic Release-store flags
```
syntactically, line 228: What barrier were you thinking we might need here? I would think that the synchronising release store and acquire load should always be enough.

Separately-ish, there is the issue of how to ensure that the host does not have data race UB, even in the face of a guest that maliciously violates the synchronisation protocol. This is all a bit complicated by the fact that the reads/writes on the other side are outside the Rust abstract machine. We have some notes about this in shared_mem.rs, which were written for the future thinking about this kind of concurrent update situation. There was also a recentish [discussion in unsafe-code-guidelines](https://github.com/rust-lang/unsafe-code-guidelines/issues/152) about this topic, which came to a similar conclusion to our notes: what we really want is a relaxed atomic-per-byte memcpy to get data in/out of the guest (after the acquire atomic load), but unfortunately this operation does not exist in Rust yet.  So, I think that the present approach of having an acquire load and then volatile non-atomic reads/writes is still the "least bad" option, even though it is not exactly correct w.r.t. the memory model---hopefully we can eventually switch to the per-byte atomic memcpy.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+The descriptor format follows the VIRTIO packed virtqueue specification. Each descriptor
+represents a memory buffer in a scatter-gather list.
+
+```rust
+bitflags! {
+    /// Descriptor flags as defined by VIRTIO specification.
+    pub struct DescFlags: u16 {
+        const NEXT     = 1 << 0;   // Buffer continues via next descriptor
+        const WRITE    = 1 << 1;   // Device write-only (otherwise read-only)
+        const INDIRECT = 1 << 2;   // Buffer contains list of descriptors
```
syntactically, line 445: The text mentioned not supporting indirect, at least for now---do we want this flag to be allocated even though we aren't supporting it yet?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+        const ENABLE  = 0x0;  // Always notify
+        const DISABLE = 0x1;  // Never notify (polling mode)
+        const DESC    = 0x2;  // Notify at specific descriptor index
+    }
+}
+
+/// Event suppression structure for controlling notifications.
+/// Both sides can control when they want to be notified.
+#[repr(C)]
+pub struct EventSuppression {
```
syntactically, line 490: Where do these end up?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+/// Submits buffer chains for the device to process and polls for completions.
+pub struct RingProducer<M: MemOps> {
+    mem: M,
+    avail_cursor: RingCursor,      // Next available descriptor position
+    used_cursor: RingCursor,       // Next used descriptor position
+    num_free: usize,               // Free slots in the ring
+    desc_table: DescTable,         // Descriptor table in shared memory
+    id_free: SmallVec<[u16; 256]>, // Stack of free IDs (allows out-of-order completion)
+    id_num: SmallVec<[u16; 256]>,  // Chain length per ID
+    drv_evt_addr: u64,             // Controls when device notifies about used buffers
```
syntactically, line 547: What does _addr mean here? Are these meant to be bitpacked versions of the suppression structure above?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+}
+
+#[repr(C)]
+pub struct Descriptor {
+    /// Physical address of the buffer
+    pub addr: u64,
+    /// Length in bytes (for used: bytes written by device)
+    pub len: u32,
+    /// Buffer ID for correlating completions with submissions
+    pub id: u16,
```
syntactically, line 458: Are these meant to identify an individual buffer, or are they the higher-level semantic protocol thing used to correlate e.g. calls and returns?  If the former, do they gain us anything on just using the address as the buffer ID?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+    desc_table: DescTable,         // Descriptor table in shared memory
+    id_free: SmallVec<[u16; 256]>, // Stack of free IDs (allows out-of-order completion)
+    id_num: SmallVec<[u16; 256]>,  // Chain length per ID
+    drv_evt_addr: u64,             // Controls when device notifies about used buffers
+    dev_evt_addr: u64,             // Reads device event to check notification preference
+}
+
+impl<M: MemOps> RingProducer<M> {
+    /// Submit a buffer chain to the ring.
+    /// Returns the descriptor ID assigned to this chain.
```
syntactically, line 553: Why do we need this as well as the _notify variant?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+    pub fn submit_available(&mut self, chain: &BufferChain) -> Result<u16, RingError>;
+
+    /// Submit with notification check based on device's event suppression settings.
+    pub fn submit_available_with_notify(&mut self, chain: &BufferChain) -> Result<SubmitResult, RingError>;
+
+    /// Poll for a used buffer. Returns WouldBlock if no completions available.
```
syntactically, line 559: We may need to think a little bit about whether there is a way to actually block/return control to the host, and whether that should be integrated into or distinct from the host->guest notification mechanism, which, at least right now, will also exit to the host in the calling thread (but perhaps in the future could be some faster IPI?)

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+/// A buffer returned after being used by the device.
+pub struct UsedBuffer {
+    pub id: u16,   // Descriptor ID
+    pub len: u32,  // Bytes written by device
+}
+```
+
+#### Buffer Chain Builder
+
+Type-state pattern enforces that readable buffers are added before writable buffers, preventing
```
syntactically, line 607: Why do we care about this "readable before writable" invariant?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+    pub written: usize,
+}
+
+/// High-level virtqueue producer for sending requests and receiving responses.
+/// Manages buffer allocation and tracks in-flight requests.
+pub struct VirtqProducer<M: MemOps, N: Notifier, P: BufferProvider> {
+    inner: RingProducer<M>,
+    notifier: N,
+    pool: P,
+    inflight: Vec<Option<ProducerInflight>>,
```
syntactically, line 701: What's this? I don't see the `ProducerInflight` definition here

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,773 @@
...
+```
+
+#### Buffer Pool
+
+An allocator for virtio buffer management. This will operate on guest physical 
+addresses from page allocator.
+
+```rust
+/// Trait for buffer providers.
+pub trait BufferProvider {
```
syntactically, line 505: In case these eventually end up passed around through a lot of parsing code with indeterminate lifetimes, would it be useful for alloc/dealloc here to be more like get/put and keep a refcount for the buffer and return it to the free pool when it's unreferenced?  I have no idea of this is true, just thought it was worth throwing out there.

syntactically (commented):
I have some minor comments on the Rust definitions as well.

proposals/0001-rng-buf/README.md
andreiltd: Agreed! I think all precautions we are currently using stand valid. What I could add here to "mitigation" is:
- validate all addresses point into permitted guest memory regions,
- use volatile/atomic operations for shared memory access,
- never assume buffer contents are stable between reads - which often means volatile-copy out memory first.

proposals/0001-rng-buf/README.md
andreiltd: That's fair, especially that both sides devuce and driver involve producing and consuming descriptors and even more true if we want to maintain that only the guest allocates shared buffers. I will change this.

proposals/0001-rng-buf/README.md
andreiltd: Single queue might be a cleaner solution but I will need to think harder what it means for backpressure. I will do it here. 

Just to not add to the confusion: by split we mean 2 separate queues for each direction and it has nothing to do with virtio terms like split vs packed queue.

In the split queue scenario the backpressure is simple because it is independent in each direction. For single queue on the other hand the backpressure is shared because both directions compete for the same descriptor ring capacity. SO for example, If the guest fills all its available slots with requests, and the host needs to send something to process those requests, this will result in deadlock - so we need to have a mechanism to avoid this.

Maintaining 50/50 split is simple but it's going to be very wasteful for the asymmetric traffic. We could maintain a minimum number of slots that are always available for the device and whenever host mark them used, the driver "immediately" resubmits fresh writable buffers. The explicit pool still introduces a bias -- so should the device be allowed to "request more credits" if it needs increase capacity? This becomes quickly quiet complicated.

proposals/0001-rng-buf/README.md
andreiltd: Yup, GPA is cleaner.

proposals/0001-rng-buf/README.md
andreiltd: So there is an `id` in the descriptor, but that identifies the buffer, not the semantic call. But I don't think there is a need for the payload level id: we can just return the opaque submission token that wraps descriptor id that is used to correlate resp/req.

proposals/0001-rng-buf/README.md
andreiltd: Good point, I would imagine this has to be protocol level as this is not really about the descriptor state but the mutli-part payload state. As you pointed out it will also require an additional context that cannot be expressed with a single flag.

proposals/0001-rng-buf/README.md
andreiltd: Can change this to `RESERVED` maybe?

proposals/0001-rng-buf/README.md
andreiltd: `drv_evt_addr: u64` and `dev_evt_addr: u64` are addresses pointing to the `EventSuppression` structures in shared memory.

proposals/0001-rng-buf/README.md
andreiltd: Yeah, typically these live at the end of the descriptor ring or at fixed offsets in the queue region. I should probably add this to layout diag.

proposals/0001-rng-buf/README.md
andreiltd: Per spec, `id` is for identifying descriptors/chains for out-of-order completion. I think using addresses as ids is problematic because: (a) addresses may be reused after deallocation, (b) chains span multiple buffers and the descriptors will share id.

proposals/0001-rng-buf/README.md
andreiltd: `_with_notify` variant reference the "other" side `EventSuppression` configuration and return if it should be notified while the other omits this step. It's used internally for optimization where for batching you want to read the suppression config once and submit without rechecking.

proposals/0001-rng-buf/README.md
andreiltd: That's a good point. One way of encoding this information would be to make `should_notify` an enum rather than `bool` and then we could express the reason why notification should happen.

proposals/0001-rng-buf/README.md
andreiltd: This is strictly a `virtio` spec requirement: A buffer consists of zero or more device-readable physically-contiguous elements followed by zero or more physically-contiguous device-writable elements (each buffer has at least one element).

proposals/0001-rng-buf/README.md
andreiltd: This tracks buffer allocations for in-flight requests, so they can be freed when the response arrives.

proposals/0001-rng-buf/README.md
andreiltd: Good point, I wanted to use `Bytes::from_owner` specifically for this:

https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner

```rust
/// The owner of mapped buffer, ensuring its lifetime.
#[derive(Debug, Clone)]
pub struct BufferOwner<P: BufferProvider, M: MemOps> {
    pub pool: Arc<P>,
    pub mem: Arc<M>,
    pub alloc: Allocation,
    pub written: usize,
}

impl<P: BufferProvider, M: MemOps> Drop for BufferOwner<P, M> {
    fn drop(&mut self) {
        let _ = self.pool.dealloc(self.alloc);
    }
}

impl<P: BufferProvider, M: MemOps> AsRef<[u8]> for BufferOwner<P, M> {
    fn as_ref(&self) -> &[u8] {
        // MemOps should give us a pointer to stored allocation
        todo!()
    }
}
```

This way you can just use `Bytes` cheap cloning and it will reuse `Bytes` reference counting. My only concern is about safety of moving around a pointer to guest memory region.

proposals/0001-rng-buf/README.md
andreiltd: > I would think that the synchronising release store and acquire load should always be enough.

That was my assumption, yes. I guess the usual "TOCTOU"  issue where the guest modifies memory between host reads remains and we should probably acknowledge this in the proposal?

proposals/0001-rng-buf/README.md
andreiltd: I've rewritten this section.

proposals/0001-rng-buf/README.md
andreiltd: So I went ahead and tried to implement function calls with `virtq` while maintaining guest buffer allocation only.

For mvp I want to replace stack-based function calls with virtqueues, so we need to handle bidirectional RPC:
- Guest -> Host calls: Guest initiates, Host processes, returns result
- Host -> Guest calls: Host initiates, Guest processes, returns result

Let's consider using a separate H2G for Host -> Guest calls with more or less this flow:

- Guest prefills H2G queue with writable request buffers
- Host takes a buffer, writes `FunctionCall` request, marks USED
- Guest sees USED, processes request, sends response back...

The problem is: how does Guest send the response?

1. Response on Same H2G Queue

- H2G Queue State (after prefill): `[A:req0][A:req1][A:req2]...[A:req255]`  <- All AVAIL
- Host takes req0, fills with request, marks USED: `[U:req0][A:req1][A:req2]...[A:req255]`
- Guest sees USED, processes, submits response: `[A:resp0][A:req1][A:req2]...[A:req255]`  <- Response at END

The problem here is that virtq reads AVAIL descriptors in order. To read `resp0`, host must skip past all remaining prefills (`req1..req255`).

We could think about only keeping a single prefill, but even with 1 prefill, batching (e.g., 5 requests at once) requires 5 prefills. Any N from 1 to 256 suffers the same "skip ahead" problem.

2. What about sending the request using H2G prefills but guest will use a G2H queue for sending responses?

- Guest prefills with writable buffers `[A:request_slot]`
- Host writes request, marks USED
- Guest reads request, processes it
- Guest sends response **via G2H** (not H2G)

Response type has to be indicated in message content. Further, it will require having Request/Response correlation id that is outside the virtq. There is also a possible deadlock scneario, similar to what had been discussed earlier:
                                                                                                                                                                                                                                                              
- Host sends H2G request (request_id=42)
- Host blocks waiting for response
- Guest processes request
- G2H is FULL of pending guest -> host requests
- Guest can't send response (G2H blocked)
- DEADLOCK: Host waiting for response, not draining G2H

The design is not very clean because H2G queue has mixed purpose: accommodating for 2 kind of responses.

3. If we absolutely need to maintain the assumption that only guest can allocate buffers for initiated requests we will need 3 queues (possibly 4 for symmetry).

Three queues, each with a single purpose:

- G2H: Guest -> Host calls with response in same chain `[R:request + W:response]`
- H2G_REQ: Host -> Guest requests only (guest prefills `[W:buffer]`, host writes request)
- H2G_RESP: Guest -> Host responses to H2G calls only `[R:response with request_id]`      

No message type disambiguation needed. No deadlock possible H2G_RESP is never blocked by unrelated traffic. Host polls H2G_RESP for responses without draining other queues.

jsturtevant: Do I understand correctly that you experimented with a single Guest side queue and ran into the "skip ahead" problem, tried two queues and ran into a deadlock?  Having 3 queues was essentially the way to avoid this?

With the split queues (as written now) we don't run into these issues?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,805 @@
...
+without incurring a VM exit for each small read/write operation. The ring buffer should allow me to
+batch multiple operations and only exit when buffers are full or when I explicitly flush.
+
+#### Story 2: High-throughput RPC
+
+As a Hyperlight developer, I want to make multiple function calls from guest to host with minimal
+overhead. The ring buffer should let me queue multiple requests, suppress notifications, and process
+responses in batches.
+
+### Risks and Mitigations
```
dblnz, line 98: I don't know if it's a real risk (this thought has crossed my mind while reading until here), but could the guest potentially try a denial of service on a host? Flooding with messages? (By the way, I think this is possible now).
Maybe we need to think about a mechanism against it?

dblnz: After reading further, I think that is already done by restricting the number of available slots in a queue, right?

proposals/0001-rng-buf/README.md
dblnz: This is a nice feature to have.

proposals/0001-rng-buf/README.md
dblnz: How does this look like for the user?
That means that if you have a `busy` guest, that keeps producing messages, the host side needs to somehow coordinate with it and stop/finish processing before creating a snapshot, right?

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,805 @@
...
+- Stress testing under high load
+
+### Implementation Plan
+
+As proposed, the queue will eventually replace the current stack‑based mechanism for function calls.
+The initial implementation will be gated behind a feature flag. The work can be roughly broken down
+into:
+
+- Introduce a low‑level queue implementation that operates on shared memory.
+- Introduce a serialization trait for data that can be safely sent through the queue.
```
dblnz, line 770: Maybe it's worth taking a more abstract approach here in case we decide to change the framework for serialization.

proposals/0001-rng-buf/README.md
```diff
@@ -0,0 +1,805 @@
...
+- Introduce a high‑level API over the queue.
+- Integrate the queue infrastructure with the existing memory model and function‑call API.
```
dblnz, line 772: Before we have the asynchronous APIs the guest function call workflow would be similar to what we have now right? I think there was a mention of this being an `implementation detail`.

proposals/0001-rng-buf/README.md
andreiltd: Pretty much yes. The proposal assumes that for host to guest communication, host acts as a driver but that will require to allocate buffers on the host side for the guest. That is logically the cleanest solution. If we want guest allocation only, the cleanest solution will be to add dedicated queue for h2g responses:

- G2H: Guest -> Host calls with response in same chain [R:request + W:response]
- H2G_REQ: Host -> Guest requests only (guest prefills [W:buffer], host writes request)
- H2G_RESP: Guest -> Host responses to H2G calls only [R:response with request_id]

One extra queue.

proposals/0001-rng-buf/README.md
andreiltd: That is correct, yes. Additionally there needs to be some set of shadow variables that are only accessible by host and cannot be accessed by guest. You can imagine the queue length is one of those.

proposals/0001-rng-buf/README.md
andreiltd: Yes. Although for mvp this will be realized by just reusing FunctionCall API, so maybe I should remove this step from the plan.

proposals/0001-rng-buf/README.md
andreiltd: Yes, for mvp the goal is to replace stack based function calls with queue and maintain backward compatibility. Full backward compatibility is not really possible because you can imagine we need to have some new configuration that replace input/output size etc.

proposals/0001-rng-buf/README.md
jsturtevant: What is the benefit of having the guest controlled queue vs shared (as proposed now)?  Trying to get a grasp of the trade offs

andreiltd: A bit late here, I've missed this comment. I think @syntactically can better explain pros and cons, but how I understand it is that the biggest practical benefit of of guest-controlled allocation is that it keeps the memory model simple - everything lives in guest memory, the host just translates and validates addresses. So there is no need for special regions like we have today: input, output stack and all buffers live in guest memory regardless of direction. That being said, based on the analysis above I think simpler memory model is a tradeoff and the complexity will reappear in H2G communication.

proposals/0001-rng-buf/README.md
dblnz: Ok, this makes sense.

proposals/0001-rng-buf/README.md
jsturtevant: Let's move forward with the HIP, i think the general direction is agreed.  We can mark this a "Uknowns/TBD", flush out implementations and update the HIP later.  

I propose we bring this up at the community meeting on Wednesday

siennathesane:
LGTM, the only questions I had were around serialization performance. As long as anything crossing host-guest boundaries uses standard C ABI shapes this is good with me, I much prefer the ring buffer semantics as they are kinder on performance. I think as a first pass this is good, and the callout with stacked PRs is the right choice. Admittedly I have no idea how they work but I'm interested in finding out!

yoshuawuyts (approved):

PR #1117, opened 2025-12-15
Remove change perissions on /opt/cargo

simongdavies:
This is now done in the base image


simongdavies:
This PR doesn't fix the issue, still get the error:

```
/usr/bin/tar: ../../../../../opt/cargo/bin: Cannot utime: Operation not permitted
/usr/bin/tar: ../../../../../opt/cargo/bin: Cannot change mode to rwxr-xr-x: Operation not permitted
```

PR #1120, opened 2025-12-16
Reset more vcpu state on snapshot::restore

ludfjig:
MSRs will be added in another PR.

- Removed padding fields from fpu register to simplify comparisons. Not sure why I added them to begin with, we don't need C repr on these. It could _possibly_ allow more efficient `into()` implementation for kvm/mshv due to single memcyp, but that seems like premature optimization to me
- Reset vcpu state on snapshot::restore
- Added bunch of tests
- MSRs are not included in this PR. Will add MSRs in future PR

Addresses https://github.com/hyperlight-dev/hyperlight/issues/791 partially

------ After rebase ------
- Added field onto Snapshot which is SpecialRegisters. These are saved on snapshot() and restored when calling restore(). This is since we can no longer just reset them to default since the guest initializes them

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
@@ -46,7 +46,7 @@ use crate::hypervisor::crashdump;
 use crate::hypervisor::gdb::{DebugError, DebugMemoryAccessError};
 #[cfg(gdb)]
 use crate::hypervisor::hyperlight_vm::debug::ProcessDebugRequestError;
-use crate::hypervisor::regs::CommonSpecialRegisters;
+use crate::hypervisor::regs::{CommonDebugRegs, CommonSpecialRegisters};
 #[cfg(not(gdb))]
 use crate::hypervisor::virtual_machine::VirtualMachine;
 #[cfg(kvm)]
@@ -320,6 +320,8 @@ pub enum HyperlightVmError {
     Initialize(#[from] InitializeError),
     #[error("Map region error: {0}")]
     MapRegion(#[from] MapRegionError),
+    #[error("Restore VM (vcpu) error: {0}")]
+    Restore(#[from] RegisterError),
     #[error("Unmap region error: {0}")]
     UnmapRegion(#[from] UnmapRegionError),
     #[error("Update region error: {0}")]
@@ -612,12 +614,11 @@ impl HyperlightVm {
         Ok(sregs.cr3)
     }
 
-    /// Set the current base page table physical address
-    pub(crate) fn set_root_pt(&mut self, addr: u64) -> Result<(), AccessPageTableError> {
-        let mut sregs = self.vm.sregs()?;
-        sregs.cr3 = addr;
-        self.vm.set_sregs(&sregs)?;
-        Ok(())
+    /// Get the special registers that need to be stored in a snapshot.
+    pub(crate) fn get_snapshot_sregs(
+        &mut self,
+    ) -> Result<CommonSpecialRegisters, AccessPageTableError> {
+        Ok(self.vm.sregs()?)
     }
 
     /// Get the current stack top virtual address
@@ -899,6 +900,44 @@ impl HyperlightVm {
         #[cfg(not(feature = "mem_profile"))]
         {
             handle_outb(mem_mgr, host_funcs, port, val)?;
+        }
+
+        Ok(())
+    }
+
+    /// Resets the following vCPU state:
+    /// - General purpose registers
+    /// - Debug registers
+    /// - XSAVE (includes FPU/SSE state with proper FCW and MXCSR defaults)
+    /// - Special registers (restored from snapshot, with CR3 updated to new page table location)
+    // TODO: check if other state needs to be reset
+    pub(crate) fn reset_vcpu(
+        &self,
+        cr3: u64,
+        sregs: &CommonSpecialRegisters,
+    ) -> std::result::Result<(), RegisterError> {
+        self.vm.set_regs(&CommonRegisters {
+            rflags: 1 << 1, // Reserved bit always set
+            ..Default::default()
+        })?;
+        self.vm.set_debug_regs(&CommonDebugRegs::default())?;
+        self.vm.reset_xsave()?;
+
+        #[cfg(feature = "init-paging")]
+        {
+            // Restore the full special registers from snapshot, but update CR3
+            // to point to the new (relocated) page tables
+            let mut sregs = *sregs;
+            sregs.cr3 = cr3;
+            self.vm.set_sregs(&sregs)?;
+        }
+        #[cfg(not(feature = "init-paging"))]
+        {
+            let _ = (cr3, sregs); // suppress unused warnings
+            // TODO: This is probably not correct.
+            // Let's deal with it when we clean up the init-paging feature
+            self.vm
+                .set_sregs(&CommonSpecialRegisters::standard_real_mode_defaults())?;
         }
 
         Ok(())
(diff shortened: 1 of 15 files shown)


src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1080,3 +1111,911 @@ mod debug {
         }
     }
 }
+
+#[cfg(test)]
+#[cfg(feature = "init-paging")]
```
jsturtevant, line 1457: this should be valid for more than just init paging feature?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1080,3 +1111,911 @@ mod debug {
...
+                        .try_into()
+                        .unwrap(),
+                );
+            }
+        }
+
+        xsave
+    }
+
+    fn hyperlight_vm(code: &[u8]) -> Result<HyperlightVm> {
```
jsturtevant: isn't this duplicating alot of the configuration code we do during VM setup in the test?  how will we maintain this with the actual setup? What if there is a difference?

ludfjig: This is calling `HyperlightVm::new` so they should be the same

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: yes will fix before publishing pr

ludfjig: Most of these tests do require 64bit mode. There is an integration test at least

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
jsturtevant: I am concerned about the logic setting up the exclusive memory, stack cookie, etc.  It seems like alot of the same in `set_up_hypervisor_partition` but simplier and having that not be the same might cause difference in configuration that might not be caught.  This seems like a good case of doing integration testing at the host API and guest api layer.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -1298,6 +1300,41 @@ mod tests {
...
+        assert_eq!(dr0_initial, 0, "DR0 should initially be 0");
+
+        // Dirty DR0 by setting it to a known non-zero value
+        const DIRTY_VALUE: u64 = 0xDEAD_BEEF_CAFE_BABE;
+        sandbox.call::<()>("SetDr0", DIRTY_VALUE).unwrap();
+        let dr0_dirty: u64 = sandbox.call("GetDr0", ()).unwrap();
+        assert_eq!(
+            dr0_dirty, DIRTY_VALUE,
+            "DR0 should be dirty after SetDr0 call"
+        );
```
jsturtevant, line 1402: Should check the value it should be set to to ensure it was reset to expected value

jsturtevant: scratch that...

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Yeah it does indeed do a lot of setup... But that's more related to memory, and whether it's different or not shouldn't affect these tests which exercises vcpu state tests (I think)

ludfjig: After rebasing on top of main branch with changes that changed snapshotting, it got even worse. But I still believe it's fine, because memory-specifics is irrelevant to these tests.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -583,6 +588,36 @@ impl HyperlightVm {
...
+    pub(crate) fn reset_vcpu(&self) -> Result<()> {
+        self.vm.set_regs(&CommonRegisters {
+            rflags: 1 << 1, // Reserved bit always set
+            ..Default::default()
+        })?;
+        self.vm.set_debug_regs(&CommonDebugRegs::default())?;
+        // Note: On KVM this ignores MXCSR so it's being set as part of reset_xsave.
+        // See https://github.com/torvalds/linux/blob/d358e5254674b70f34c847715ca509e46eb81e6f/arch/x86/kvm/x86.c#L12554-L12599
+        self.vm.set_fpu(&CommonFpu::default())?;
+        self.vm.reset_xsave()?;
```
jsturtevant, line 924: I believe FPU state xsave overlap, could we just do the reset here and save a few cycles?

src/hyperlight_host/src/hypervisor/regs/debug_regs.rs
jsturtevant, line 1: thought: this would be a smaller pr if this was a sperate PR and would make its easier to review.

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -165,9 +167,21 @@ pub(crate) trait VirtualMachine: Debug + Send {
...
-    #[cfg(crashdump)]
+    /// Get xsave
+    #[allow(dead_code)]
     fn xsave(&self) -> Result<Vec<u8>>;
+    /// Reset xsave to default state
+    fn reset_xsave(&self) -> Result<()>;
+    /// Set xsave - only used for tests
+    #[cfg(test)]
+    #[cfg(feature = "init-paging")]
+    fn set_xsave(&self, xsave: &[u32]) -> Result<()>;
```
jsturtevant: why is it only for the init-paging feature?  shouldn't there be a test for all the features?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1074,3 +1109,1041 @@ mod debug {
...
+            cr3: match get_available_hypervisor() {
+                #[cfg(mshv3)]
+                Some(HypervisorType::Mshv) => _pml4_addr,
+                _ => 0x12345000,
+            },
+            cr4: 0x20, // PAE
+            cr8: 0x5,
+            efer: 0x500, // LME + LMA
+            apic_base: 0xFEE00900,
+            // interrupt_bitmap: [0xFFFFFFFF; 4],
```
jsturtevant: nit: may need to remove this comment? or do something differnet for mshv

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1074,3 +1109,1041 @@ mod debug {
...
+        expected_sregs.fs.g = got_sregs.fs.g;
+        expected_sregs.gs.unusable = got_sregs.gs.unusable;
+        expected_sregs.gs.g = got_sregs.gs.g;
+        expected_sregs.ss.unusable = got_sregs.ss.unusable;
+        expected_sregs.ss.g = got_sregs.ss.g;
+        expected_sregs.tr.unusable = got_sregs.tr.unusable;
+        expected_sregs.tr.g = got_sregs.tr.g;
+        expected_sregs.ldt.unusable = got_sregs.ldt.unusable;
+        expected_sregs.ldt.g = got_sregs.ldt.g;
+        assert_eq!(got_sregs, expected_sregs);
```
jsturtevant: could we move this to a single function (shared with the one below) so the comment stays the same? There is alot of duplication in the function and is really hard to read.  and know the different phases we are in.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1074,3 +1109,1041 @@ mod debug {
...
+        let mut got_fpu = hyperlight_vm.vm.fpu().unwrap();
+        let mut expected_fpu = fpu;
+        // KVM doesn't preserve mxcsr via set_fpu/fpu()
+        #[cfg(kvm)]
+        if available_hv == HypervisorType::Kvm {
+            got_fpu.mxcsr = fpu.mxcsr;
+        }
+        // fpr only uses 80 bits per register. Normalize upper bits for comparison.
+        for i in 0..8 {
+            expected_fpu.fpr[i][10..16].copy_from_slice(&got_fpu.fpr[i][10..16]);
```
jsturtevant, line 2209: this is duplicated in multiple spots in the function, could we condense it?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1074,3 +1109,1041 @@ mod debug {
...
+            None,
+            #[cfg(gdb)]
+            dbg_mem_access_hdl.clone(),
+        )
+        .unwrap();
+        vm
+    }
+
+    #[test]
+    fn reset_vcpu_simple() {
```
jsturtevant, line 2176: In general, this test is really hard to read.  Could we evaluate a way to make it easier to read?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1074,3 +1109,1041 @@ mod debug {
...
+                r10: 0,
+                r11: 0,
+                r12: 0,
+                r13: 0,
+                r14: 0,
+                r15: 0,
+                rip: 0,
+                rflags: 1 << 1, // Reserved bit 1 is always set
+            };
+            assert_eq!(regs, expected_reset);
```
jsturtevant: this looks to be the same as in `simple_reset`: 
  // Verify the fpu was reset to defaults
        assert_eq!(
            hyperlight_vm.vm.regs().unwrap(),
            CommonRegisters {
                rflags: 1 << 1, // Reserved bit 1 is always set
                ..Default::default()
            }
        );

could we add helper funciton like `verify_registers_reset` and re-use this across all the tests?  that would make off the tests more consistent and easier to read.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1074,3 +1109,1041 @@ mod debug {
...
+            #[cfg(kvm)]
+            use crate::hypervisor::regs::MXCSR_DEFAULT;
+
+            #[cfg(kvm)]
+            let available_hv = *get_available_hypervisor().as_ref().unwrap();
+
+            #[rustfmt::skip]
+            const CODE: [u8; 289] = [
+                // xmm0-xmm7: use movd + pshufd to fill with pattern
+                0xb8, 0x11, 0x11, 0x11, 0x11,       // mov eax, 0x11111111
```
jsturtevant: This seems really hard to reason about, Could we use something like https://docs.rs/iced-x86/latest/iced_x86/ or put the files in `test.asm` and compile them at build time with nasm then bring them in with `include_bytes`

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -1296,6 +1298,41 @@ mod tests {
         assert_ne!(sandbox3.id, sandbox_id);
     }
 
+    /// Test that snapshot restore properly resets vCPU debug registers. This test verifies
+    /// that restore() calls reset_vcpu.
+    #[test]
+    fn snapshot_restore_resets_debug_registers() {
```
jsturtevant, line 1382: could we also add a test like this for other registers too? Doesn't need to be all the registers but atleast the common registers kind of like it is done here?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
jsturtevant: I don't love this but I guess it might be fine.  We might be able to add an abstraction that "preps the vm" but I haven't really thought this through fully.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: agree, done

src/hyperlight_host/src/hypervisor/regs/debug_regs.rs
ludfjig: done

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed!

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: tried to make it more readable now

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: should be better now

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
ludfjig: a lot of current code (unrelated to this) depends on `init-paging` feature, and note that it's a default-feature. All hyperlight usages except nanvix must use this feature. For example `GuestPageTableBuffer`, `set_pt_size`, `standard_64bit_defaults` are some things already gated behind this feature. We can consider ungating these, but let's do that in another PR as it would be a substantial effort.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed!

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: This test is just making sure reset_vcpu is called. Doing it with other registers is not really feasible because they would be clobbered before/after the guest call, unlike debug registers. I think this is fine because the other registers are tested in hyperlight_vm.rs's tests

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed

.github/workflows/ValidatePullRequest.yml
```diff
@@ -67,7 +67,7 @@ jobs:
       - docs-pr
       - build-guests
     strategy:
-      fail-fast: true
+      fail-fast: false
```
jsturtevant: is this purposeful?

ludfjig: no it just helps me debugging, will remove this before merging 👍

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -870,6 +877,31 @@ impl HyperlightVm {
         Ok(())
     }
 
+    // Resets the following vCPU state:
+    // - General purpose registers
+    // - Debug registers
+    // - XSAVE (includes FPU/SSE state with proper FCW and MXCSR defaults)
+    // - Special registers (with saved PML4 if feature enabled)
+    // TODO: check if other state needs to be reset
```
dblnz: Nit

```suggestion
    /// Resets the following vCPU state:
    /// - General purpose registers
    /// - Debug registers
    /// - XSAVE (includes FPU/SSE state with proper FCW and MXCSR defaults)
    /// - Special registers (with saved PML4 if feature enabled)
    /// TODO: check if other state needs to be reset
```

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -289,6 +289,9 @@ impl MultiUseSandbox {
             unsafe { self.vm.map_region(region) }.map_err(HyperlightVmError::MapRegion)?;
         }
 
+        // if this fails, then the sandbox will not have been restored so keep it poisoned (if it is)
+        self.vm.reset_vcpu().map_err(HyperlightVmError::Restore)?;
```
dblnz: What happens if the sandbox is not poisoned (good state), and it fails here?
Won't it be partially reset?
I mean, 15 lines above we run the `restore_snapshot` method, which restores the memory, so if this fails, it will remain not poisoned, with the memory reset, but not the registers, right?

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -86,7 +86,7 @@ pub fn is_hypervisor_present() -> bool {
 }
 
 /// The hypervisor types available for the current platform
-#[derive(PartialEq, Eq, Debug)]
+#[derive(PartialEq, Eq, Debug, Copy, Clone)]
```
dblnz, line 91: Unrelated question: Why is the `HypervisorType` defined twice?
Once here, the other in `sandbox/hypervisor.rs`.

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
```diff
@@ -272,14 +273,77 @@ impl VirtualMachine for MshvVm {
...
+            return Err(RegisterError::XsaveSizeMismatch {
+                expected: XSAVE_MIN_SIZE as u32,
+                actual: current_xsave.buffer.len() as u32,
+            });
+        }
+
+        let mut buf = XSave::default(); // default is zeroed 4KB buffer
+
+        // Copy XCOMP_BV (offset 520-527) - preserves feature mask + compacted bit
+        buf.buffer[520..528].copy_from_slice(&current_xsave.buffer[520..528]);
```
jsturtevant, line 314: Do we need to preserve this in kvm as well?

src/hyperlight_host/src/hypervisor/virtual_machine/kvm.rs
```diff
@@ -234,6 +240,56 @@ impl VirtualMachine for KvmVm {
...
     }
+
+    fn reset_xsave(&self) -> std::result::Result<(), RegisterError> {
+        let mut xsave = kvm_xsave::default(); // default is zeroed 4KB buffer with no FAM
+
+        // XSAVE area layout from Intel SDM Vol. 1 Section 13.4.1:
+        // - Bytes 0-1: FCW (x87 FPU Control Word)
+        // - Bytes 24-27: MXCSR
+        // - Bytes 512-519: XSTATE_BV (bitmap of valid state components)
+        xsave.region[0] = FP_CONTROL_WORD_DEFAULT as u32;
```
jsturtevant, line 288: this is writing 32 bits to a 16 bit field.  We only write the 16 bits in mshv.  Is this intended?

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
```diff
@@ -272,14 +273,77 @@ impl VirtualMachine for MshvVm {
...
     fn xsave(&self) -> std::result::Result<Vec<u8>, RegisterError> {
         let xsave = self
             .vcpu_fd
             .get_xsave()
             .map_err(|e| RegisterError::GetXsave(e.into()))?;
         Ok(xsave.buffer.to_vec())
     }
+
+    fn reset_xsave(&self) -> std::result::Result<(), RegisterError> {
+        const XSAVE_MIN_SIZE: usize = 576; // 512 legacy + 64 header
```
jsturtevant: this is the same constant as in whp, consider sharing it?

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -494,6 +494,132 @@ impl VirtualMachine for WhpVm {
...
+
+        // Expect insufficient buffer error; any other error is unexpected
+        if let Err(e) = result
+            && e.code() != windows::Win32::Foundation::WHV_E_INSUFFICIENT_BUFFER
+        {
+            return Err(RegisterError::GetXsaveSize(e.into()));
+        }
+
+        let provided_size = std::mem::size_of_val(xsave) as u32;
+        if buffer_size_needed > provided_size {
```
jsturtevant: in kvm and mshv we do an explicit check for `std::mem::size_of_val(xsave) != KVM_XSAVE_SIZE_BYTES`  and here we don't.  it means we might provide a buffer to overwrite values and only set certain amount. this is just test code but could help with strange bugs where we think we are doing something different.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1381,3 +1413,1039 @@ mod debug {
...
+                assert_eq!(mxcsr, 0x3F80, "MXCSR in XSAVE should be dirty");
+            }
+
+            // Reset vcpu
+            hyperlight_vm.reset_vcpu().unwrap();
+
+            // Check FPU is reset to defaults
+            assert_fpu_reset(hyperlight_vm.vm.as_ref());
+
+            // Verify MXCSR via xsave on KVM
```
jsturtevant: might just leave a comment here, that using fpu to check this doesn't work on kvm and is why we do the extra read via xsave

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1381,3 +1413,1039 @@ mod debug {
...
+            expected_dirty.rip = regs.rip;
+            expected_dirty.rsp = regs.rsp;
+            expected_dirty.rflags = regs.rflags;
+            assert_eq!(regs, expected_dirty);
+
+            // Reset vcpu
+            hyperlight_vm.reset_vcpu().unwrap();
+
+            // Check registers are reset to defaults
+            assert_regs_reset(hyperlight_vm.vm.as_ref());
```
jsturtevant, line 2376: this is much cleaner thanks!

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
ludfjig: fixed

src/hyperlight_host/src/hypervisor/virtual_machine/kvm.rs
ludfjig: Yes, added comment.

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
ludfjig: Nope, no need since different format is used. Updated comment

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
ludfjig: Not sure :D We should remove one if they're the same, but will not do it in this PR

ludfjig: #1202

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Good point. I made it so if it fails, it poisons the sandbox

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: thanks fixed

jsturtevant (dismissed):

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -905,6 +907,32 @@ impl HyperlightVm {
...
+    /// address via `set_root_pt()` after calling this method.
+    // TODO: check if other state needs to be reset
+    pub(crate) fn reset_vcpu(&self) -> std::result::Result<(), RegisterError> {
+        self.vm.set_regs(&CommonRegisters {
+            rflags: 1 << 1, // Reserved bit always set
+            ..Default::default()
+        })?;
+        self.vm.set_debug_regs(&CommonDebugRegs::default())?;
+        self.vm.reset_xsave()?;
+        #[cfg(feature = "init-paging")]
```
syntactically, line 926: I'm not a huge fan of eating an extra hypercall overhead here for little reason. Can we just swap around the order of `set_root_pt()` and `reset_vcpu()` so that the extra call isn't needed?

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
```diff
@@ -283,14 +284,75 @@ impl VirtualMachine for MshvVm {
...
         let xsave = self
             .vcpu_fd
             .get_xsave()
             .map_err(|e| RegisterError::GetXsave(e.into()))?;
         Ok(xsave.buffer.to_vec())
     }
+
+    fn reset_xsave(&self) -> std::result::Result<(), RegisterError> {
+        let current_xsave = self
+            .vcpu_fd
```
syntactically, line 300: I wonder if there is any more efficient way to do this, since it is a lot of hypercalls. It looks like HvCallMapVpStatePage only gives us access to XMM0-6 and not the other FP state that you are resetting, so perhaps it is not possible (it might make sense to use the register state page for some of the other set_ functions, but that's not relevant to this PR specifically).

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -905,6 +907,32 @@ impl HyperlightVm {
...
+        self.vm
+            .set_sregs(&CommonSpecialRegisters::standard_64bit_defaults(0))?;
+        #[cfg(not(feature = "init-paging"))]
+        self.vm
+            .set_sregs(&CommonSpecialRegisters::standard_real_mode_defaults())?;
+
+        Ok(())
+    }
```
syntactically, line 944: How does this interact with the debug code that uses these register purposefully? Perhaps this is part of #1210.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: fixed in 2b0b5f90ed7432ddaca4634f0cb45c3c4a9630e4

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
ludfjig: agree, it's worth thinking about

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: Good catch, it might break something. Let's make it part of #1210

danbugs (approved):

Issue #1118, opened 2025-12-15
Improve guest exception handling to prevent infinite loops


simongdavies:
When the guest exception handler itself fails (due to corrupted stack,  unmapped memory etc.), the guest enters an unrecoverable state— causing either an infinite page fault loop (hang) or a triple fault (crash). This occurs because:

1. The exception handler entry point requires a working stack (for context_save!() pushes)
2. The Rust handler uses format!() which allocates on the heap
3. There's no detection of nested/recursive exceptions

A triple fault is not a terrible outcome as at least the guest terminates but having the infinite page loop is more of a problem, if the host is monitoring execution and kills the guest via the `InterruptHandle` then it should be able to successfully terminate the guest , but if it is not doing this then a badly behaved guest could cause resource (CPU) starvation.

There are a couple of things that we can do to protect against this:

Configure a small permanent fixed stack in the ISS that can be used for a double fault handler.
Detect exception handler re-entrancy and exit immediately with outb data that the host can translate to a `DoubleFault` error.

This is another reason that we should consider not allowing guest code to run in ring 0.

syntactically:
An IST exception stack is added in https://github.com/hyperlight-dev/hyperlight/commit/6211e9856f33c3c9bcc9b0a5c797f73d8254b0d1, since the interrupt handler becomes involved in growing the main stack, and the handler always runs on the top of that stack since there isn't any expected reentrancy. 

Separately, I think that it is a bug that the handler can allocate. When doing the snapshot CoW, it is careful not to allocate before that is done since that would go badly, but it would be preferable not to allocate at all, and I don't believe there is a strong reason to be allocating.

Having the assembly entry stub check for reentrancy is a reasonable idea, but I'm unsure if it is worth it if the exception handler is small and well-behaved (which it probably should be).

PR #1125, opened 2025-12-17
Don't allocate in exception handler

jsturtevant:
Partially addresses: https://github.com/hyperlight-dev/hyperlight/issues/1118

The test run with out this change fails with memory allocation:

```
assertion `left == right` failed: Full error: GuestAborted(8, "panicked at /home/jstur/.rustup/toolchains/1.89-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/alloc/src/alloc.rs:438:13:\nmemory allocation of 182 bytes failed")
  left: 8
 right: 15
```

### DIFF
--- a/src/hyperlight_guest_bin/src/exceptions/handler.rs
+++ b/src/hyperlight_guest_bin/src/exceptions/handler.rs
@@ -14,12 +14,13 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-use alloc::format;
-use core::ffi::c_char;
+use core::fmt::Write;
 
 use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
 use hyperlight_common::outb::Exception;
-use hyperlight_guest::exit::abort_with_code_and_message;
+use hyperlight_guest::exit::write_abort;
+
+use crate::HyperlightAbortWriter;
 
 /// Exception information pushed onto the stack by the CPU during an excpection.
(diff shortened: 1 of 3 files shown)


src/hyperlight_guest_bin/src/exceptions/handler.rs
```diff
@@ -149,10 +141,24 @@ pub(crate) extern "C" fn hl_exception_handler(
...
 
-    unsafe {
-        abort_with_code_and_message(
-            &[ErrorCode::GuestError as u8, exception as u8],
-            msg.as_ptr() as *const c_char,
-        );
+    // begin abort sequence by writing the error code
+    let mut w = HyperlightAbortWriter;
+    write_abort(&[ErrorCode::GuestError as u8, exception as u8]);
+    let write_res = write!(
```
ludfjig, line 147: Are you sure this `write!` doesn't allocate?

jsturtevant: Yes, I believe that is what is state here: https://doc.rust-lang.org/std/fmt/index.html#write

>write! and writeln! are two macros which are used to emit the format string to a specified stream. This is used to prevent intermediate allocations of format strings and instead directly write the output

Our implementation of `write` writes it to outb with no allocations.  

I believe the test would also fail since the heap is exhausted. It fails before this change

PR #1126, opened 2025-12-18
Run co-pilot on 1ES KVM runner

simongdavies:
make /dev/kvm available to co-pilot agent

### DIFF
--- a/.github/workflows/copilot-setup-steps.yml
+++ b/.github/workflows/copilot-setup-steps.yml
@@ -6,8 +6,8 @@ on: workflow_dispatch
 jobs:
   # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
   copilot-setup-steps:
-    # run on either intel or amd kvm runner based on run id last digit is even/odd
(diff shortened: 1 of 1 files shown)


.github/workflows/copilot-setup-steps.yml
```diff
@@ -6,7 +6,8 @@ on: workflow_dispatch
 jobs:
   # The job MUST be called `copilot-setup-steps` or it will not be picked up by Copilot.
   copilot-setup-steps:
-    runs-on: ubuntu-latest
+    # run on either intel or amd kvm runner based on run id last digit is even/odd
+    runs-on: ${{ (endsWith(github.run_id, '0') || endsWith(github.run_id, '2') || endsWith(github.run_id, '4') || endsWith(github.run_id, '6') || endsWith(github.run_id, '8')) && fromJSON('["self-hosted", "Linux", "X64", "1ES.Pool=hld-kvm-amd"]') || fromJSON('["self-hosted", "Linux", "X64", "1ES.Pool=hld-kvm-intel"]') }}
```
jprendes: what are those numbers?

simongdavies: They are the last digit of the run id , its just so that some agents use amd and some use intel mainly not have everything using amd (or intel) and causing delays to other jobs because of depleting one pool (the KVM pools have more quota than others and we have the same number of AMD/Intel runners)

jsturtevant (approved):

PR #1137, opened 2025-12-24
Remove duplicate GVA-to-GPA translation in sw breakpoint functions

### LINKED ISSUE #1046: gdb: potential bug due to double address translation
In `add_sw_breakpoint` and `remove_sw_breakpoint`, the address is translated from GVA to GPA using translate_gva, but then the resulting GPA is passed to read_addrs/write_addrs which internally call translate_gva again, treating the already-translated GPA as a GVA. 

I would expect this to cause incorrect memory access since the physical address gets translated a second time?

IrvingMg:
Fix double GVA-to-GPA translation in `add_sw_breakpoint` and `remove_sw_breakpoint`. These functions were translating addresses before passing them to `read_addrs`/`write_addrs`, which translate internally.

Fixes #1046

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
@@ -1039,41 +1039,37 @@ mod debug {
         // Must be idempotent!
         fn add_sw_breakpoint(
             &mut self,
-            addr: u64,
+            gva: u64,
             mem_access: &DebugMemoryAccess,
         ) -> crate::Result<()> {
-            let addr = self.vm.translate_gva(addr)?;
-
             // Check if breakpoint already exists
-            if self.sw_breakpoints.contains_key(&addr) {
+            if self.sw_breakpoints.contains_key(&gva) {
                 return Ok(());
             }
 
             // Write breakpoint OP code to write to guest memory
             let mut save_data = [0; SW_BP_SIZE];
-            self.read_addrs(addr, &mut save_data[..], mem_access)?;
-            self.write_addrs(addr, &SW_BP, mem_access)?;
+            self.read_addrs(gva, &mut save_data[..], mem_access)?;
+            self.write_addrs(gva, &SW_BP, mem_access)?;
 
             // Save guest memory to restore when breakpoint is removed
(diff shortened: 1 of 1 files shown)


jsturtevant:
Thanks @IrvingMg! this look good on a quick glance.  Might have @dblnz take a look.  Given holidays might be slow.   

Could you sign your commit? https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

IrvingMg:
> Could you sign your commit? https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

Oops, missed that! It’s signed now. Thanks!

ludfjig (changes requested):
Thanks for this, it looks good to me. However I wonder if we can remove these? I'm not sure whether they are related to this or not, but they look suspicious...

https://github.com/IrvingMg/hyperlight/blob/0397689c166da4af27a6b6d102c672107df40c4f/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L650-L653

https://github.com/IrvingMg/hyperlight/blob/0397689c166da4af27a6b6d102c672107df40c4f/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L699-L702

Perhaps @dblnz knows?

dblnz:
> Thanks for this, it looks good to me. However I wonder if we can remove these? I'm not sure whether they are related to this or not, but they look suspicious...
> 
> https://github.com/IrvingMg/hyperlight/blob/0397689c166da4af27a6b6d102c672107df40c4f/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L650-L653
> 
> https://github.com/IrvingMg/hyperlight/blob/0397689c166da4af27a6b6d102c672107df40c4f/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L699-L702
> 
> Perhaps @dblnz knows?

I don't think those can be eliminated. We still need to report an error when the debugger client requests operating on an address that is not valid for the guest.
This PR eliminates the double translation from SW breakpoints, because the virtual address is still translated by the `read_addrs` and `write_addrs` calls.

PR #1173, opened 2026-01-15
FEATURE: add paravirtualized clock support for guest time access

simongdavies:
Hyperlight guests can now read time without expensive VM exits by using a paravirtualized clock shared between host and guest. This enables high-frequency timing operations like benchmarking, rate limiting, and timestamping with minimal overhead.

Paravirtualized clocks work by having the hypervisor populate a shared memory page with clock calibration data. The guest reads this data along with the CPU's TSC to compute the current time entirely in userspace, avoiding the cost of a VM exit.

Reference: https://docs.kernel.org/virt/kvm/x86/msr.html#pvclock

The implementation uses the native mechanism for each hypervisor:

- KVM: pvclock (MSR 0x4b564d01)
- MSHV: Hyper-V Reference TSC page
- WHP: Hyper-V Reference TSC page

Guests have access to:

- Monotonic time: nanoseconds since sandbox creation, guaranteed to never go backwards
- Wall-clock time: UTC nanoseconds since Unix epoch
- Local time: wall-clock adjusted for host timezone captured at sandbox creation

Rust API (hyperlight_guest_bin::time):

- SystemTime/Instant types mirroring std::time
- DateTime type for human-readable date/time formatting
- Weekday/Month enums with name() and short_name() methods

C API (hyperlight_guest_capi):

- POSIX-compatible: clock_gettime, gettimeofday, time
- Broken-down time: gmtime_r, localtime_r, mktime, timegm
- Formatting: strftime with common format specifiers

The feature is gated behind `guest_time` (enabled by default) and documented in docs/guest-time.md.

Note: The timezone offset is a snapshot from sandbox creation and does not update for DST transitions during the sandbox lifetime.

### DIFF
--- a/docs/guest-time.md
+++ b/docs/guest-time.md
@@ -1,0 +1,346 @@
+# Guest Time API
+
+This document describes how to access time from within a Hyperlight guest. Hyperlight provides a paravirtualized clock that allows guests to read time without expensive VM exits.
+
+## Overview
+
+When a sandbox is created, Hyperlight configures a shared clock page between the host and guest. The guest can read time by accessing this shared page and the CPU's Time Stamp Counter (TSC), without requiring any VM exit or host call.
+
+### Supported Hypervisors
+
+- **KVM**: Uses KVM pvclock (MSR `0x4b564d01`)
+- **MSHV**: Uses Hyper-V Reference TSC page
+- **WHP** (Windows): Uses Hyper-V Reference TSC page
+
+### Clock Types
+
+- **Monotonic time**: Time since sandbox creation. Guaranteed to never go backwards. Use for measuring elapsed time.
+- **Wall-clock time**: UTC time since Unix epoch (1970-01-01 00:00:00 UTC). Can be used for timestamps.
+- **Local time**: Wall-clock time adjusted for the host's timezone offset (captured at sandbox creation).
+
+## Feature Flag
+
+The time functionality is controlled by the `guest_time` feature flag, which is enabled by default. To disable:
+
+```toml
+[dependencies]
+hyperlight-guest = { version = "...", default-features = false }
+```
+
+## Rust API
+
+### High-Level API (`hyperlight_guest_bin::time`)
+
+The recommended API for Rust guests mirrors `std::time`:
+
+```rust
+use hyperlight_guest_bin::time::{SystemTime, Instant, UNIX_EPOCH};
+use core::time::Duration;
+
+// Wall-clock time (like std::time::SystemTime)
+let now = SystemTime::now();
+let duration = now.duration_since(UNIX_EPOCH).unwrap();
+let unix_timestamp = duration.as_secs();
+
+// Monotonic time for measuring elapsed time (like std::time::Instant)
+let start = Instant::now();
+// ... do work ...
+let elapsed = start.elapsed();
+
+// Get timezone offset (seconds east of UTC)
+use hyperlight_guest_bin::time::utc_offset_seconds;
+if let Some(offset) = utc_offset_seconds() {
+    // offset is seconds to add to UTC for local time
+    // e.g., +3600 for UTC+1, -18000 for UTC-5
+}
+```
+
+#### `SystemTime`
+
+Represents wall-clock time (UTC). Methods:
+
+- `SystemTime::now()` - Get current wall-clock time
+- `duration_since(earlier)` - Duration between two system times
+- `elapsed()` - Duration since this time was captured
+- `checked_add(duration)` / `checked_sub(duration)` - Arithmetic operations
+
+#### `Instant`
+
+Represents monotonic time for measuring durations. Methods:
+
+- `Instant::now()` - Get current monotonic time
+- `duration_since(earlier)` - Duration between two instants
+- `elapsed()` - Duration since this instant was captured
+- Supports `+`, `-` operators with `Duration`
+- Supports `-` between two `Instant`s to get a `Duration`
+
+#### `DateTime`
+
+For formatting human-readable dates and times:
+
+```rust
+use hyperlight_guest_bin::time::DateTime;
+
+// Get current local time
+let dt = DateTime::now_local();
+
+// Format: "Thursday 15th January 2026 15:34:56"
+let formatted = format!(
+    "{} {} {} {} {:02}:{:02}:{:02}",
+    dt.weekday().name(),      // "Thursday"
+    dt.day_ordinal(),         // "15th"
+    dt.month().name(),        // "January"
+    dt.year(),                // 2026
+    dt.hour(),                // 15
+    dt.minute(),              // 34
+    dt.second()               // 56
+);
+```
+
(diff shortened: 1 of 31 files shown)


src/hyperlight_common/src/time.rs
```diff
@@ -0,0 +1,205 @@
...
+        assert_eq!(ClockType::from(99u64), ClockType::None);
+    }
+
+    #[test]
+    fn test_guest_clock_region_default() {
+        let region = GuestClockRegion::default();
+        assert!(!region.is_available());
+        assert_eq!(region.get_clock_type(), ClockType::None);
+    }
+}
```
ludfjig, line 205: I think most of these can be `const`-time assertions instead of tests

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,645 @@
...
+                options(nostack, nomem, preserves_flags)
+            );
+        }
+        ((hi as u64) << 32) | (lo as u64)
+    }
+    #[cfg(not(target_arch = "x86_64"))]
+    {
+        0 // TSC not available on non-x86_64 architectures
+    }
+}
```
ludfjig, line 83: I think you can replace this with https://doc.rust-lang.org/core/arch/x86/fn._rdtsc.html

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -169,6 +169,23 @@ pub(crate) trait VirtualMachine: Debug + Send {
...
+    /// * `clock_page_gpa` - Guest physical address of the clock page (must be 4KB aligned)
+    ///
+    /// # Returns
+    /// * `Ok(())` if clock was successfully configured
+    /// * `Err` if the hypervisor doesn't support pvclock or configuration failed
+    fn setup_pvclock(&mut self, _clock_page_gpa: u64) -> Result<()> {
+        Err(crate::new_error!(
+            "Paravirtualized clock setup not implemented for this hypervisor",
+        ))
+    }
```
ludfjig, line 187: nit: should this default implementation be removed?

src/hyperlight_common/src/time.rs
```diff
@@ -0,0 +1,205 @@
...
+            clock_type: clock_type as u64,
+            boot_time_ns,
+            utc_offset_seconds,
+            _padding: 0,
+        }
+    }
+
+    /// Returns true if a clock is configured.
+    pub fn is_available(&self) -> bool {
+        self.clock_page_ptr != 0 && self.clock_type != ClockType::None as u64
```
ludfjig, line 158: It's my understand that this is always available, so do we really need `is_available` and `is_clock_available`? Correct me if I am wrong.

ludfjig (commented):
First round of reviews, looks very good to me! Question: maybe we can split out all the time-related math + formatting into a separate crate?

Haven't looked in detail into everythign yet

syntactically:
Do we definitely want to enable this by default? I think a lot of guests will not need high precision time in production, and providing it by default is a significant semantic constraint: e.g. it makes snapshotting observable, makes timing side-channel attacks easier to internalise, etc.

When we first enabled rdtsc for the performance traces, I believe we explicitly had a discussion across all the maintainers about this and said we very much did not want to enable any extra time sources (and especially not wall-clock/referenced to an epoch ones) anywhere.

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,645 @@
...
+        return None; // Must use MSR fallback, not available in guest
+    }
+
+    compiler_fence(Ordering::Acquire);
+
+    // Read clock data
+    let tsc_scale = tsc_page.tsc_scale;
+    let tsc_offset = tsc_page.tsc_offset;
+
+    compiler_fence(Ordering::Acquire);
```
the-eugen, line 153: Did you mean Ordering::Release here? There's also a similar patter in the KVM handler.

A compiler barrier is not enough though, you need an rmb() here and before reading the data on L147. They are paired with the hypervisor wmb() when updating the page. Admittedly it is going to be a noop on amd64, but maybe better to take care of this now if the arm64 port is planned,

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,645 @@
...
+    let tsc_now = rdtsc();
+
+    // Calculate time in 100ns units
+    // Formula: time_100ns = ((tsc * scale) >> 64) + offset
+    let scaled = ((tsc_now as u128 * tsc_scale as u128) >> 64) as i64;
+    let time_100ns = scaled.wrapping_add(tsc_offset);
+
+    if time_100ns < 0 {
+        return None; // Invalid time
+    }
```
the-eugen, line 171: You are making `scaled` a signed integer to catch overflows, however by doing so you also lose precision (it's just 1 bit of precision, but still). So that would detect overflows when the u64 value would not have in fact overflowed. (e.g. i64::MAX + 1).

I know that tsc_offset is signed, but you could let scaled be unsigned and use `u64.overflowing_add_signed` instead to not lose precision and detect over/under-flows. On amd64 it uses just a normal add instruction and saves the overflow flag.

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,645 @@
...
+    // Read clock data
+    let tsc_scale = tsc_page.tsc_scale;
+    let tsc_offset = tsc_page.tsc_offset;
+
+    compiler_fence(Ordering::Acquire);
+
+    // Check sequence again
+    let seq2 = unsafe { core::ptr::read_volatile(&tsc_page.tsc_sequence) };
+    if seq1 != seq2 {
+        return None; // Data changed during read, retry later
```
the-eugen, line 158: I wonder how the caller is supposed to differentiate between "use MSR fallback" and this case here? The sequence mismatch is a transient problem because the hypervisor is modifying the page contents and it is very likely to go away if you retry immediately, and that is still going to be much faster than taking an MSR exit. 
Therefore typically you would loop when trying to read from the TSC page until sequences match because we (the hyper-v hypervisor) do not expect this problem to persist. All client-side implementations i've seen usually just retry immediately maybe having an upper bound on the retry count. Falling back on MSR traps would be much less efficient in P99.9 cases.

the-eugen (commented):
Hi! Some comments from the hyper-v side of things

docs/guest-time.md
```diff
@@ -0,0 +1,346 @@
...
+
+### Supported Hypervisors
+
+- **KVM**: Uses KVM pvclock (MSR `0x4b564d01`)
+- **MSHV**: Uses Hyper-V Reference TSC page
+- **WHP** (Windows): Uses Hyper-V Reference TSC page
+
+### Clock Types
+
+- **Monotonic time**: Time since sandbox creation. Guaranteed to never go backwards. Use for measuring elapsed time.
```
dblnz: How do ensure this?

simongdavies:
Closed in favour of #1422

PR #1174, opened 2026-01-16
Run shared memory tests with miri

ludfjig:
Enables 10 existing shared_memory tests to be run with Miri and adds them to CI. This allows us to exercise some of our unsafe code and increase confidence in correctness and not having UB.

Note:
- For this PR I first intended on adding another rust-native `Vec<u8>` backing for for HostMapping which does not use platform apis (`mmap`, `VirtualAlloc`), but it turns out `mmap` is partially supported by Miri, but not with MAP_SHARED so I did that instead because it is easier. Aside: I don't think we need to use MAP_SHARED  in general for hyperlight, but that not relevant to this PR.
- the change to `base_ptr()` is intentional because we used to cast pointer to integer, which loses pointer provenance. The fix is to use pointer arithmetics instead to preserve provenance.

### DIFF
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -35,13 +35,11 @@ use windows::Win32::System::Memory::{
 #[cfg(target_os = "windows")]
 use windows::core::PCSTR;
 
-#[cfg(target_os = "windows")]
-use crate::HyperlightError::MemoryAllocationFailed;
 use crate::HyperlightError::SnapshotSizeMismatch;
 #[cfg(target_os = "windows")]
-use crate::HyperlightError::{MemoryRequestTooBig, WindowsAPIError};
+use crate::HyperlightError::WindowsAPIError;
 use crate::sandbox::snapshot::Snapshot;
-use crate::{Result, log_then_return, new_error};
+use crate::{HyperlightError, Result, log_then_return, new_error};
 
 /// Makes sure that the given `offset` and `size` are within the bounds of the memory with size `mem_size`.
 macro_rules! bounds_check {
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -646,7 +662,7 @@ pub trait SharedMemory {
     /// not need to be marked as `unsafe` because doing anything with
     /// this pointer itself requires `unsafe`.
     fn base_ptr(&self) -> *mut u8 {
-        self.base_addr() as *mut u8
+        self.region().ptr.wrapping_add(PAGE_SIZE_USIZE)
```
jsturtevant, line 665: was this something miri suggested?

jsturtevant: should we update base_addr with this too since its doing the same thing?

```
    fn base_addr(&self) -> usize {
        self.region().ptr as usize + PAGE_SIZE_USIZE
    }
```

syntactically:
> but not with MAP_SHARED so I did that instead because it is easier. Aside: I don't think we need to use MAP_SHARED in general for hyperlight, but that not relevant to this PR.

I don't know for sure, but from this comment I would tend to assume the reason Miri doesn't support MAP_SHARED is the potential for concurrent modification and/or writes outside the memory model. Even if we use MAP_PRIVATE, those same concerns about potential UB that can't be detected by Miri will remain, because the same mapping is visible in the VM.

src/hyperlight_host/src/mem/shared_mem.rs
andreiltd: I believe that for `usize` casting we should use `addr()` pointer method and to reconstruct pointer from address with provenance we should be using `with_addr`. There is a good write up about strict provenance API here:

https://doc.rust-lang.org/std/ptr/index.html#strict-provenance

PR #1181, opened 2026-01-21
Closes #1076: Code refactor to set trace context in run_vcpu function.

### LINKED ISSUE #1076: Make sure guest spans are setup as late as possible to avoid human error
with #924 merging, guest span is setup before calling run_vcpu. In the future, if any spans are created in run_vcpu method, then the spans will be malformed. We should change the api to make sure this cannot happen

vshailesh:
- Small change to set up the guest trace context function call inside the run_vcpu function instead, to prevent any human error in future code changes.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
@@ -381,13 +381,10 @@ impl HyperlightVm {
             {
                 Ok(VmExit::Cancelled())
             } else {
-                #[cfg(feature = "trace_guest")]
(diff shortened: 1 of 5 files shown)


ludfjig:
Thanks a lot for this! Looks like CI wants you to run a `just fmt-apply`, and a variable seems to be unused unless `tracing` feature is turned on. I would probably just name this parameter `_tc` so clippy doesn't complain about it if it's unused.

vshailesh:
pushed again, completely forgot about clippy checks.

vshailesh:
retracting this PR.

PR #1182, opened 2026-01-22
Add scratch region

syntactically:
This adds the new scratch region, which will eventually become where everything writable is, but is currently not used.

A prerequisite of this is to improve whp support for dynamically mapping file handles into the surrogate process. The same infrastructure could be used to expose (some parts of) the host mapping API on Windows.

### DIFF
--- a/src/hyperlight_common/src/arch/amd64/layout.rs
+++ b/src/hyperlight_common/src/arch/amd64/layout.rs
@@ -14,5 +14,14 @@ See the License for the specific language governing permissions and
 limitations under the License.
  */
 
-// Keep in mind that the minimum upper half GVA is 0xffff_8000_0000_0000
-pub const SNAPSHOT_PT_GVA: usize = 0xffff_ff00_0000_0000;
+/// We have this the top of the page below the top of memory in order
+/// to make working with start/end ptrs in a few places more
+/// convenient (not needing to worry about overflow)
+pub const MAX_GVA: usize = 0xffff_ffff_ffff_efff;
+pub const SNAPSHOT_PT_GVA: usize = 0xffff_8000_0000_0000;
+
+/// We assume 36-bit IPAs for now, since every amd64 processor
+/// supports at least 36 bits.  Almost all of them support at least 40
+/// bits, so we could consider bumping this in the future if we were
+/// ever memory-constrained.
+pub const MAX_GPA: usize = 0x0000_000f_ffff_ffff;
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/kvm.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/kvm.rs
@@ -83,6 +83,23 @@ impl KvmVm {
             .create_vcpu(0)
             .map_err(|e| CreateVmError::CreateVcpuFd(e.into()))?;
 
+        // Set the CPUID leaf for MaxPhysAddr. KVM allows this to
+        // easily be overridden by the hypervisor and defaults it very
+        // low, while mshv passes it through from hardware unless an
+        // intercept is installed.
+        let mut kvm_cpuid = hv
+            .get_supported_cpuid(kvm_bindings::KVM_MAX_CPUID_ENTRIES)
+            .map_err(|e| CreateVmError::InitializeVm(e.into()))?;
+        for entry in kvm_cpuid.as_mut_slice().iter_mut() {
+            if entry.function == 0x8000_0008 {
+                entry.eax &= !0xff;
+                entry.eax |= hyperlight_common::layout::MAX_GPA.ilog2() + 1;
+            }
+        }
+        vcpu_fd
+            .set_cpuid2(&kvm_cpuid)
+            .map_err(|e| CreateVmError::InitializeVm(e.into()))?;
+
         Ok(Self {
             vm_fd,
             vcpu_fd,
--- a/src/hyperlight_host/src/mem/mgr.rs
+++ b/src/hyperlight_host/src/mem/mgr.rs
@@ -45,6 +45,8 @@ pub(crate) const STACK_COOKIE_LEN: usize = 16;
 pub(crate) struct SandboxMemoryManager<S> {
     /// Shared memory for the Sandbox
     pub(crate) shared_mem: S,
+    /// Scratch memory for the Sandbox
+    pub(crate) scratch_mem: S,
     /// The memory layout of the underlying shared memory
     pub(crate) layout: SandboxMemoryLayout,
     /// Pointer to where to load memory from
@@ -145,6 +147,7 @@ where
     pub(crate) fn new(
         layout: SandboxMemoryLayout,
         shared_mem: S,
+        scratch_mem: S,
         load_addr: RawPtr,
         entrypoint_offset: Option<Offset>,
         stack_cookie: [u8; STACK_COOKIE_LEN],
@@ -152,6 +155,7 @@ where
         Self {
             layout,
             shared_mem,
+            scratch_mem,
             load_addr,
             entrypoint_offset,
             mapped_rgns: 0,
@@ -191,12 +195,6 @@ where
             mapped_regions,
         )
     }
-
-    /// This function restores a memory snapshot from a given snapshot.
-    pub(crate) fn restore_snapshot(&mut self, snapshot: &Snapshot) -> Result<()> {
-        self.shared_mem.restore_from_snapshot(snapshot)?;
-        Ok(())
-    }
 }
 
 impl SandboxMemoryManager<ExclusiveSharedMemory> {
@@ -204,6 +202,7 @@ impl SandboxMemoryManager<ExclusiveSharedMemory> {
         let layout = *s.layout();
         let mut shared_mem = ExclusiveSharedMemory::new(s.mem_size())?;
         shared_mem.copy_from_slice(s.memory(), 0)?;
+        let scratch_mem = ExclusiveSharedMemory::new(s.layout().get_scratch_size())?;
         let load_addr: RawPtr = RawPtr::try_from(layout.get_guest_code_address())?;
         let stack_cookie = rand::random::<[u8; STACK_COOKIE_LEN]>();
         let entrypoint_gva = s.preinitialise();
@@ -211,6 +210,7 @@ impl SandboxMemoryManager<ExclusiveSharedMemory> {
         Ok(Self::new(
             layout,
             shared_mem,
+            scratch_mem,
             load_addr,
             entrypoint_offset,
             stack_cookie,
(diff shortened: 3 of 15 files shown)


src/hyperlight_host/src/sandbox/snapshot.rs
```diff
@@ -184,6 +192,23 @@ impl Snapshot {
...
+            // 2. Map the scratch region
+            let scratch_size = cfg.get_scratch_size();
+            let mapping = Mapping {
+                phys_base: hyperlight_common::layout::scratch_base_gpa(scratch_size),
+                virt_base: hyperlight_common::layout::scratch_base_gva(scratch_size),
+                len: scratch_size as u64,
+                kind: MappingKind::BasicMapping(BasicMapping {
+                    readable: true,
+                    writable: true,
+                    executable: true,
```
ludfjig, line 205: should the scratch region be executable? I guess it's fine?

src/hyperlight_host/src/hypervisor/surrogate_process.rs
```diff
@@ -15,60 +15,155 @@ limitations under the License.
...
+use windows::Win32::System::SystemServices::NUMA_NO_PREFERRED_NODE;
 
 use super::surrogate_process_manager::get_surrogate_process_manager;
 use super::wrappers::HandleWrapper;
+use crate::HyperlightError::WindowsAPIError;
+use crate::{Result, log_then_return};
+
+#[derive(Debug)]
+pub(crate) struct HandleMapping {
+    pub(crate) use_count: u64,
```
ludfjig, line 37: Is this `use_count` used for anything or could we remove it?

src/hyperlight_host/src/hypervisor/surrogate_process.rs
```diff
@@ -15,60 +15,155 @@ limitations under the License.
...
+            }
+        }
+    }
+
+    pub(super) fn unmap(&mut self, host_base: usize) {
+        if let Some(entry) = self.mappings.remove(&host_base) {
+            self.unmap_helper(entry.surrogate_base);
+        } else {
+            #[cfg(debug_assertions)]
+            panic!("Attempted to unmap from surrogate a region that was never mapped")
```
ludfjig: I personally am a fan of panicking on user errors like this instead of slapping `Result` everywhere, I think maybe there are other places in our codebase where we can consider starting to panic. I know we are supposed to be "panic-free", but I think we can still improve alot

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -86,11 +86,11 @@ impl GuestView {
         let regions = ctx
             .regions
             .iter()
-            .filter(|r| !r.host_region.is_empty())
+            .filter(|r| !r.guest_region.is_empty())
```
ludfjig, line 89: for my own sake, is there any logic whether we check guest_region or host_region here and in other places?

src/hyperlight_common/src/arch/amd64/layout.rs
```diff
@@ -14,5 +14,14 @@ See the License for the specific language governing permissions and
...
+/// to make working with start/end ptrs in a few places more
+/// convenient (not needing to worry about overflow)
+pub const MAX_GVA: usize = 0xffff_ffff_ffff_efff;
+pub const SNAPSHOT_PT_GVA: usize = 0xffff_8000_0000_0000;
+
+/// We assume 36-bit IPAs for now, since every amd64 processor
+/// supports at least 36 bits.  Almost all of them support at least 40
+/// bits, so we could consider bumping this in the future if we were
+/// ever memory-constrained.
+pub const MAX_GPA: usize = 0x0000_00ff_ffff_ffff;
```
dblnz: Is this a typo or I don't understand it correctly?
Why is the `MAX_GPA` is `0x0000_00ff_ffff_ffff` (40 bit) when the comment above says `36 bit`?

src/hyperlight_host/src/hypervisor/surrogate_process.rs
andreiltd: Shouldn't `unmap` check `use_count` before actually unmapping and only unmap when it equals 1? At least this is what I would expect from this pattern.

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -236,9 +236,11 @@ impl SandboxMemoryManager<ExclusiveSharedMemory> {
         SandboxMemoryManager<GuestSharedMemory>,
     ) {
         let (hshm, gshm) = self.shared_mem.build();
+        let (hscratch, gscratch) = self.scratch_mem.build();
```
andreiltd, line 239: This is probably a silly question but I would like to understand this split better: could you explain why scratch memory gets its own `SharedMemory` host/guest pair as opposed to scratch being just regions in `shared_mem`?

src/hyperlight_host/src/hypervisor/surrogate_process.rs
```diff
@@ -15,60 +15,155 @@ limitations under the License.
...
+            Entry::Occupied(mut oe) => {
+                oe.get_mut().use_count += 1;
+                Ok(oe.get().surrogate_base)
+            }
+            Entry::Vacant(ve) => {
+                // Use MapViewOfFile2 to map memory into the surrogate process, the MapViewOfFile2 API is implemented in as an inline function in a windows header file
+                // (see https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffile2#remarks) so we use the same API it uses in the header file here instead of
+                // MapViewOfFile2 which does not exist in the rust crate (see https://github.com/microsoft/windows-rs/issues/2595)
+                let surrogate_base = unsafe {
+                    MapViewOfFileNuma2(
```
andreiltd, line 76: The docs say the function can fail:

https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-mapviewoffilenuma2#return-value

Should we check we get non null `Value`?

src/hyperlight_host/src/hypervisor/crashdump.rs
```diff
@@ -86,11 +86,11 @@ impl GuestView {
...
             .map(|r| VaRegion {
                 begin: r.guest_region.start as u64,
                 end: r.guest_region.end as u64,
-                offset: r.host_region.start as u64,
+                offset: <_ as Into<usize>>::into(r.host_region.start) as u64,
```
andreiltd, line 93: This looks like a fix straight from rust-analyzer 😄  can this be just `.into()`?

src/hyperlight_host/src/sandbox/snapshot.rs
syntactically: Yes, it should, since e.g. wasm code loaded at runtime will end up here at least briefly. The general ethos here is to move all the page protections into the guest, which can choose to make some regions NX if it wants to.

src/hyperlight_host/src/hypervisor/surrogate_process.rs
syntactically: Yeah, in this case it is a little bit of a cop-out because of the clippy lint, so it silently ignores the error in production, which I don't love. I would like to have another discussion about the use of the anti-panic clippy lint overall / the goal of being mostly panic free. I think that in general there are a large class of internal error things that we currently log or return runtime failure for, but which really are indicative of a bug inside hyperlight (some invariant not being maintained) rather than a mis-use of hyperlight as a library from another application. Being security-focused code, I generally feel that we should check invariants early and often, and indeed take the quickest safe option of "OK just die now" when we find invariant violations, rather than just returning an error but allowing the system to continue to operate in a known-to-be-inconsistent state.

src/hyperlight_host/src/hypervisor/surrogate_process.rs
syntactically: This code was just pulled over from `surrogate_process_manager.rs` where it did the same thing, I think?  So maybe, but I'm not sure this is the correct venue to fix it?

src/hyperlight_host/src/mem/mgr.rs
syntactically: Eventually the original `shared_mem` will all be readonly and hopefully eventually just be a direct mapping of a snapshot file.

src/hyperlight_common/src/arch/amd64/layout.rs
syntactically: Indeed, thanks for catching that.

src/hyperlight_host/src/hypervisor/crashdump.rs
syntactically: Calculating the length of `host_region` is annoying now because on Windows the endpoints aren't just numbers anymore. As far as I know, it is an invariant of the structure that `host_region` and `guest_region` have the same length, and I consider it a bug that they both have a length at all---I would like to just get rid of the endpoint of `host_region` entirely, storing only `host_base`.  I didn't do that in this PR though, so this was a (bit of an ugly short-term) stop-gap. I'll leave this as is and file an issue about removing the endpoint of host_region to make it make more sense perhaps?

src/hyperlight_host/src/hypervisor/surrogate_process.rs
syntactically: Oh, yeah, you are totally correct! Sorry, will fix.

src/hyperlight_host/src/hypervisor/surrogate_process.rs
ludfjig: agree!

syntactically:
@andreiltd @ludfjig Thanks for the reviews---I believe I added fixup! commits for most of your comments? 

There was also quite some churn in the base commit during rebasing due to the error PR merging, but I don't think anything semantically changed.

src/hyperlight_host/src/hypervisor/crashdump.rs
syntactically: I think making this into just `.into()` would require adding an `Into<u64>` implementation for the type, as the annotation is being used to go from the type -> usize -> u64, which I didn't think was worth it for just the one case?

syntactically:
I have squashed the fixup!s in, in preparation for fast-forward merging this.

src/hyperlight_host/src/hypervisor/virtual_machine/kvm.rs
```diff
@@ -83,6 +83,23 @@ impl KvmVm {
...
 
+        // Set the CPUID leaf for MaxPhysAddr. KVM allows this to
+        // easily be overridden by the hypervisor and defaults it very
+        // low, while mshv passes it through from hardware unless an
+        // intercept is installed.
+        let mut kvm_cpuid = hv
+            .get_supported_cpuid(kvm_bindings::KVM_MAX_CPUID_ENTRIES)
+            .map_err(|e| CreateVmError::InitializeVm(e.into()))?;
+        for entry in kvm_cpuid.as_mut_slice().iter_mut() {
+            if entry.function == 0x8000_0008 {
```
dblnz, line 94: Nit: Maybe we could set a `const` instead of the magic number.

dblnz: sorry for the late comment.
It's just a small observation, you can merge it anyway and fix it later, from my point of view

syntactically: I will fix it later so that we don't have to go through the whole round of wait for CI and get approvals again.

PR #1184, opened 2026-01-24
Code refactor to set trace context in run_vcpu function.

### LINKED ISSUE #1076: Make sure guest spans are setup as late as possible to avoid human error

vshailesh:
- code refactor to the trace_guest feature.
- only start trace_guest after entering the run_vcpu function.
- Closes #1076

### DIFF
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
@@ -30,6 +30,8 @@ use mshv_bindings::{
 };
 use mshv_ioctls::{Mshv, VcpuFd, VmFd};
 use tracing::{Span, instrument};
+#[cfg(feature = "trace_guest")]
+use tracing_opentelemetry::OpenTelemetrySpanExt;
 
 #[cfg(gdb)]
 use crate::hypervisor::gdb::{DebugError, DebuggableVm};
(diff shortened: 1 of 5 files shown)


dblnz:
Could you please take care of the clippy warnings. Thanks!

src/hyperlight_host/src/hypervisor/virtual_machine/mshv.rs
```diff
@@ -121,7 +125,12 @@ impl VirtualMachine for MshvVm {
             .map_err(|e| UnmapMemoryError::Hypervisor(e.into()))
     }
 
-    fn run_vcpu(&mut self) -> std::result::Result<VmExit, RunVcpuError> {
+    fn run_vcpu(
+        &mut self,
+        #[cfg(feature = "trace_guest")] tc: &mut SandboxTraceContext,
+    ) -> std::result::Result<VmExit, RunVcpuError> {
+        #[cfg(feature = "trace_guest")]
+        tc.setup_guest_trace(Span::current().context());
```
ludfjig: Thanks a lot for this! Just a minor comment: Would you mind moving this right before the call to vcpu.run() in each driver? And a comment would be great as well that it's important that `setup_guest_trace` is called right before the `run` call.

dblnz:
@ludfjig I see you requested changed, do you have anything more to add?

PR #1194, opened 2026-01-27
Support wasmparser 0.244.0 breaking API changes

jsturtevant:
wasmparser 0.244.0 introduced breaking changes from PR #2376 which bakes async into component model function types. This adds an async_ field to ComponentFuncType and a new Map variant to ComponentDefinedType.

Since Hyperlight does not yet support async component model features, we handle these new API additions by panicking with clear error messages, consistent with how Future and Stream types are already handled.

This unblocks the Dependabot PR #1156 which was failing CI due to these breaking changes.

### DIFF
--- a/src/hyperlight_component_util/src/elaborate.rs
+++ b/src/hyperlight_component_util/src/elaborate.rs
@@ -439,10 +439,16 @@ impl<'p, 'a> Ctx<'p, 'a> {
             ComponentDefinedType::Future(_) | ComponentDefinedType::Stream(_) => {
                 panic!("async not yet supported")
             }
+            ComponentDefinedType::Map(_, _) => {
+                panic!("map type not yet supported")
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/hyperlight_component_util/src/elaborate.rs
```diff
@@ -436,13 +436,18 @@ impl<'p, 'a> Ctx<'p, 'a> {
             ComponentDefinedType::FixedSizeList(vt, size) => {
                 Ok(Value::FixList(Box::new(self.elab_value(vt)?), *size))
             }
-            ComponentDefinedType::Future(_) | ComponentDefinedType::Stream(_) => {
+            ComponentDefinedType::Future(_)
+            | ComponentDefinedType::Stream(_)
+            | ComponentDefinedType::Map(_, _) => {
                 panic!("async not yet supported")
             }
```
ludfjig: should we add "map" to the panic message?

jsturtevant: oh yea a better error message would be good. let me fix that up

jsturtevant: updated

PR #1197, opened 2026-01-28
Avoid stale checked-in Cargo.lock files

ludfjig:
Avoid stale Cargo.lock being checked in. This is annoying because rust-analyzer will rebuild them locally and update Cargo.lock, resulting in locally modified Cargo.lock files, which can interfere with git operations (like git checkout).

Most complications come from the fact that dependabot will update dependencies (like hyperlight-common dependencies) without rebuilding guests resulting in the guest's Cargo.lock being outdated after a dependabot PR.f

I'm not 100% this will work after merging, so might require a future fix. Worst case is dependabot PRs won't get merged for a while, which is not the end of the world

Note: most of the scripting in this PR is made by copilot. 

Prior art:

- Guest check added before in #126 
- Tried to be fixed in #154 and #227 but failed
- Removed again in #240

Note: since this will modify dependabot PRs, commands like `@dependabot rebase` won't work. Would need to do `@dependabot recreate` isntead, since dependabot won't rebase if somebody else touched the PR. This might make it a little more cumbersome to merge dependabot PRs...

### DIFF
--- a/.github/workflows/dependabot-update-guest-locks.yml
+++ b/.github/workflows/dependabot-update-guest-locks.yml
@@ -1,0 +1,175 @@
+# This workflow automatically updates the Cargo.lock files in guest crates when
+# Dependabot updates dependencies. Without this, Dependabot PRs only update the
+# root Cargo.lock, leaving the guest crate Cargo.lock files stale.
+#
+# See: https://docs.github.com/en/code-security/tutorials/secure-your-dependencies/automating-dependabot-with-github-actions
+
+name: Update Guest Cargo.lock for Dependabot PRs
+
+on:
+  pull_request:
+    branches: [main]
+    paths:
+      - 'Cargo.toml'
+      - 'Cargo.lock'
+      - 'src/hyperlight_*/Cargo.toml'
+
+permissions:
+  contents: read       # Required for actions/checkout to clone the repo
+  pull-requests: read  # Required for dependabot/fetch-metadata to read PR info
+
+env:
+  CARGO_TERM_COLOR: always
+
+jobs:
+  update-guest-locks:
+    # Only run for Dependabot PRs - check the PR author, not the actor
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    runs-on: [self-hosted, Linux, X64, "1ES.Pool=hld-kvm-amd"]
+    timeout-minutes: 15
+    steps:
+      # Fetch metadata about the Dependabot PR
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v2
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      # Only proceed for cargo ecosystem updates
+      - name: Check if cargo update
+        id: check-ecosystem
+        run: |
+          if [ "${{ steps.metadata.outputs.package-ecosystem }}" = "cargo" ]; then
+            echo "is_cargo=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "is_cargo=false" >> "$GITHUB_OUTPUT"
+            echo "Skipping non-cargo dependency update"
+          fi
+
+      # Get GitHub App token for pushing commits back to the PR
+      # Uses the same app as auto-merge-dependabot.yml
+      - name: Get GitHub App token
+        if: steps.check-ecosystem.outputs.is_cargo == 'true'
+        uses: actions/create-github-app-token@v2
+        id: get-app-token
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/dependabot-update-guest-locks.yml
```diff
@@ -0,0 +1,128 @@
...
+
+      - name: Fix cargo home permissions
+        if: steps.check-ecosystem.outputs.is_cargo == 'true'
+        run: |
+          sudo chown -R $(id -u):$(id -g) /opt/cargo || true
+
+      - name: Update simpleguest Cargo.lock
+        if: steps.check-ecosystem.outputs.is_cargo == 'true'
+        working-directory: src/tests/rust_guests/simpleguest
+        run: cargo update
```
syntactically: Is `cargo update` more aggressive than we want here? IIUC `cargo update` would update all deps to the latest compatible version, whereas I wonder if `cargo update simpleguest` to do only the updates that are necessary to make the lockfiles consistent would be useful to make the PRs a bit more granular and isolate the effect of a dependency with a failing update?

syntactically:
@ludfjig Absolutely no pressure if you have higher-priority things, but is there anything we should do to progress this? It's only a tiny papercut, but I am constantly running into the out-of-date guest lock files because I now use Nix-vendored dependencies, which makes them a bit annoying (have to get a cargo executable not configured for vendoring, fix the lock file, wait for the dev environment to rebuild, etc).

.github/workflows/dependabot-update-guest-locks.yml
ludfjig: Perhaps building the guest would do the minimal necessary updates. I'm not sure `cargo update simpleguest` would work

ludfjig: Actually `cargo fetch` should probably do it! (without the `--locked`)

.github/workflows/dependabot-update-guest-locks.yml
```diff
@@ -0,0 +1,175 @@
...
+          private-key: ${{ secrets.DEPENDABOT_APP_KEY }}
+          permission-contents: write
+
+      - name: Checkout PR branch
+        if: steps.check-ecosystem.outputs.is_cargo == 'true'
+        uses: actions/checkout@v6
+        with:
+          token: ${{ steps.get-app-token.outputs.token }}
+          ref: ${{ github.head_ref }}
+          fetch-depth: 0
```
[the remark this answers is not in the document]
ludfjig: fixed

.github/workflows/dependabot-update-guest-locks.yml
```diff
@@ -0,0 +1,175 @@
...
+
+      # Commits created via the Git Data API are automatically signed/verified
+      # by GitHub when authenticated as a GitHub App and no custom author or
+      # committer info is provided.
+      #
+      # References:
+      # - Signature verification for bots:
+      #   https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification#signature-verification-for-bots
+      # - How to Use Commit Signing with GitHub Apps:
+      #   https://github.com/orgs/community/discussions/50055
```
jsturtevant, line 105: looks like this process is documented officially to https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-your-git-database?versionId=free-pro-team%40latest&productId=authentication&restPage=managing-commit-signature-verification%2Cabout-commit-signature-verification&apiVersion=2022-11-28#overview

.github/workflows/dep_code_checks.yml
```diff
@@ -64,7 +64,11 @@ jobs:
             src/tests/rust_guests/witguest -> target
 
       - name: Ensure up-to-date Cargo.lock
-        run: cargo fetch --locked
+        run: |
+          cargo fetch --locked
+          cargo fetch --manifest-path src/tests/rust_guests/simpleguest/Cargo.toml --locked
+          cargo fetch --manifest-path src/tests/rust_guests/dummyguest/Cargo.toml --locked
+          cargo fetch --manifest-path src/tests/rust_guests/witguest/Cargo.toml --locked
```
jsturtevant, line 71: Nit: I think we could do this as `just check cargo.lock` and have it all in one place instead of copy paste across all the files

ludfjig:
> It's kind of unfortunate that this isn't handled automatically.
> 
> It seems like
> 
> https://github.com/hyperlight-dev/hyperlight/blob/8c991d10566d1e5e9a52418bc02ebe53982ad587/.github/dependabot.yml#L34
> 
> should be doing this for us?

I filed this a long time ago because I felt like it is a bug https://github.com/dependabot/dependabot-core/issues/11500

PR #1203, opened 2026-01-31
Consolidate duplicate HypervisorType enum

### LINKED ISSUE #1202: Consolidate duplicate `enum HypervisorType`
We seem to have 2 of these for no reason, we should get rid of one.

Jgprog117:
## Summary

Resolves #1202 by consolidating duplicate  enum definitions.

## Changes

- Removed duplicate enum definition from 
- Added import to use the existing definition from  module
- Both enums were identical, ensuring no functional changes

## Impact

This eliminates code duplication and establishes a single source of truth for the  enumeration, improving maintainability.

Fixes #1202

### DIFF
--- a/src/hyperlight_host/src/sandbox/hypervisor.rs
+++ b/src/hyperlight_host/src/sandbox/hypervisor.rs
@@ -21,6 +21,7 @@ use std::sync::OnceLock;
 use crate::hypervisor::virtual_machine::kvm;
 #[cfg(mshv3)]
 use crate::hypervisor::virtual_machine::mshv;
+use crate::hypervisor::virtual_machine::HypervisorType;
 
 static AVAILABLE_HYPERVISOR: OnceLock<Option<HypervisorType>> = OnceLock::new();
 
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/src/sandbox/hypervisor.rs
dblnz, line 1: It looks like there are also a couple other things that are defined in both files.

I've had a look, and it looks like this file is not used at all, it may be an artifact of some conflict.
Can you remove it please?

If it turns out to not be used at all, could you modify the PR title and/or commit message to reflect that?

ludfjig: agree with this! Thanks a lot for you contribution!

jsturtevant: Thanks for the contribution @Jgprog117, since this doesn't seem to be used i just went ahead and removed it https://github.com/hyperlight-dev/hyperlight/pull/1248

Issue #1074, opened 2025-12-04
Check stack frames are correctly cleared when `hlt` is called


dblnz:
We should ensure that the call to `hlt` instruction is only made after everything has been torn down so we do not leak anything.

An idea of how to implement such a fix can be found at [PR comment](https://github.com/hyperlight-dev/hyperlight/pull/1072#issuecomment-3609786020).
Which basically says we should use minimal `assembly` that only calls a rust function and then runs `hlt`, which ensures anything has been torn down before calling `hlt`

PR #1229, opened 2026-02-11
Enable CoW

syntactically:
There are still a number of things that need to be done before the performance improvements from CoW will actually be apparent (e.g. making it possible to mmap the snapshot mapping readonly directly from the page cache, rather than copy it into a SharedMemory).


src/hyperlight_host/src/error.rs
```diff
@@ -160,6 +160,11 @@ pub enum HyperlightError {
     #[error("Memory requested {0} exceeds maximum size allowed {1}")]
     MemoryRequestTooBig(usize, usize),
 
+    /// The memory request is too small to contain everything that is
+    /// required
+    #[error("Memory requested {0} exceeds maximum size allowed {1}")]
```
ludfjig: nit: this attribute has an incorrect message

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -398,29 +367,47 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
-            .write::<u64>(snapshot_pt_base_gpa_offset, snapshot_pt_base_gpa)?;
+        // Copy the page tables into the scratch region
+        let snapshot_pt_end = self.shared_mem.mem_size();
+        let snapshot_pt_start = snapshot_pt_end - self.layout.get_pt_size();
+        self.shared_mem.with_exclusivity(|snap| {
+            self.scratch_mem.with_exclusivity(|scratch| {
+                let bytes = &snap.as_slice()[snapshot_pt_start..snapshot_pt_end];
+                scratch.copy_from_slice(bytes, self.layout.get_pt_base_scratch_offset())
+            })
+        })???;
```
ludfjig, line 410: nit: This looks a bit funky but it's probably correct

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -631,6 +639,16 @@ impl HyperlightVm {
...
+    /// Get the current entrypoint action
+    pub(crate) fn get_entrypoint(&mut self) -> NextAction {
+        self.entrypoint
+    }
+
+    /// Set the current entrypoint action
+    pub(crate) fn set_entrypoint(&mut self, entrypoint: NextAction) {
+        self.entrypoint = entrypoint
+    }
+
```
ludfjig: nit: this could take &self

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -29,31 +29,27 @@ use hyperlight_guest::prim_alloc::alloc_phys_pages;
...
-            snapshot_pt_base_gpa: unsafe {
-                hyperlight_guest::layout::snapshot_pt_gpa_base_gva().read_volatile()
-            },
-            snapshot_pt_base_gva: hyperlight_common::layout::SNAPSHOT_PT_GVA_MIN as u64,
             scratch_base_gpa: hyperlight_guest::layout::scratch_base_gpa(),
             scratch_base_gva: hyperlight_guest::layout::scratch_base_gva(),
         }
     }
-    fn phys_to_virt(&self, addr: u64) -> *mut u8 {
+    fn fallible_phys_to_virt(&self, addr: u64) -> Option<*mut u8> {
```
andreiltd:
```suggestion
    fn try_phys_to_virt(&self, addr: u64) -> Option<*mut u8> {
```

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -374,16 +399,25 @@ impl SandboxMemoryLayout {
...
+        self.pt_size = Some(size);
+        Ok(())
     }
 
-    /// Sets the size of the memory region used for page tables
+    /// Get the size of the memory region used for page tables
     #[instrument(skip_all, parent = Span::current(), level= "Trace")]
-    pub(crate) fn set_pt_size(&mut self, size: usize) {
-        self.pt_size = Some(size);
+    pub(crate) fn get_pt_size(&mut self) -> usize {
```
simongdavies: think this can just be `get_pt_size(&self)`

src/hyperlight_host/src/mem/mgr.rs
syntactically: Curious what you meant here? Happy to take suggestions for a more clear way to do this. The same nested with_exclusivity pattern is used in snapshot.rs and I don't presently see a cleaner way to do it.

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -115,34 +108,3 @@ fn internal_dispatch_function() {
...
-    #[cfg(feature = "trace_guest")]
-    {
-        let guest_start_tsc = hyperlight_guest_tracing::invariant_tsc::read_tsc();
-        // Reset the trace state for the new guest function call with the new start TSC
-        // This clears any existing spans/events from previous calls ensuring a clean state
-        hyperlight_guest_tracing::new_call(guest_start_tsc);
-    }
-
-    internal_dispatch_function();
-    halt();
```
dblnz, line 147: I am not sure if this is true, but having eliminated the use of `halt()` here, can we remove the function also?

I'm referring to https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest/src/exit.rs#L29.
And, if that is the case, I think this fixes https://github.com/hyperlight-dev/hyperlight/issues/1074 , right?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -755,13 +783,13 @@ impl HyperlightVm {
             match exit_reason {
                 #[cfg(gdb)]
                 Ok(VmExit::Debug { dr6, exception }) => {
+                    let initialise = match self.entrypoint {
+                        NextAction::Initialise(initialise) => initialise,
+                        _ => 0,
```
dblnz, line 788: Nice, this gave me an idea: I think this can be used to add an option `stop_on_entry` for the sandbox under debug and stop on each call of a function (dispatch_function). So that it would solve the issue we have now that we only allow debugging of a newly created sandbox (we rely on the call to initialise to stop).

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -115,34 +108,3 @@ fn internal_dispatch_function() {
...
-    crate::paging::flush_tlb();
-
-    // Read the current TSC to report it to the host with the spans/events
-    // This helps calculating the timestamps relative to the guest call
-    #[cfg(feature = "trace_guest")]
-    {
-        let guest_start_tsc = hyperlight_guest_tracing::invariant_tsc::read_tsc();
-        // Reset the trace state for the new guest function call with the new start TSC
-        // This clears any existing spans/events from previous calls ensuring a clean state
-        hyperlight_guest_tracing::new_call(guest_start_tsc);
```
jsturtevant, line 143: It looks like this was removed from the new implementation.  Does this mean tracing might not work as expected going forward?

src/hyperlight_guest_bin/src/arch/amd64/exception/handle.rs
```diff
@@ -57,6 +60,107 @@ pub type ExceptionHandler = fn(
...
+    }
+}
+
+fn handle_cow_pagefault(_phys: PhysAddr, virt: VirtAddr, perms: CowMapping) {
+    unsafe {
+        let new_page = hyperlight_guest::prim_alloc::alloc_phys_pages(1);
+        let target_virt = virt as *mut u8;
+        core::ptr::copy(
+            target_virt,
+            crate::paging::phys_to_virt(new_page).unwrap(),
```
jsturtevant: This could cause a triple fault which might make it really hard to diagnose?  Should we maybe abort with a clear error message here?

src/hyperlight_guest_bin/src/paging.rs
```diff
@@ -137,45 +123,79 @@ impl vmem::TableOps for GuestMappingOperations {
...
+    ///         Section 5: Page Translation and Protection
+    ///             §5.5: Translation-Lookaside Buffer
+    ///                 §5.5.3: TLB Management
+    #[inline(always)]
+    pub fn first_valid_same_ctx() {
+        unsafe {
+            core::arch::asm!("
+                mov rax, cr0
+                mov cr0, rax
+            ", out("rax") _, out("rcx") _);
```
jsturtevant:
```suggestion
            ", out("rax") _);
```

rcx isn't used here

src/hyperlight_guest_bin/src/guest_function/call.rs
syntactically: Yeah, you might be right there, good catch! @dblnz  do you know if we can just move this into internal_dispatch_function, which is the first thing called with a normal ABI after the little entry stub?

src/hyperlight_guest_bin/src/arch/amd64/exception/handle.rs
syntactically: I'm not sure I see how this would cause a triple fault? If you are referring to the unwrap, then I think it will just abort?   Or is the concern about something a bit second order here like a format call that allocates in the unwrap code path or something? If the latter & that is there then happy to change it.

Apart from that, I don't think we should need to worry about reading from target_virt (since by this point we have already confirmed that the original mapping of this page was a CoW mapping, so just reading from it should be fine) or writing to the new page in the physmap (since all the physmap entries are RW).

src/hyperlight_guest_bin/src/paging.rs
syntactically: Good catch, thanks, that is a holdover from a different implementation, will fix.

src/hyperlight_guest_bin/src/arch/amd64/exception/handle.rs
jsturtevant: I was referring to the unwrap.  Previously, when working on the handler I saw a panic combined with an alloc and it was hard to diagnose the initial error since it triggered the exception recursively till it stack overflowed.  I think it would be easy enough to do a `.or_else()` and abort with a specific error message here?

syntactically: OK, I didn't think there would necessarily be an alloc here, but perhaps there is.  I don't think it's a huge deal, because this case should be impossible, but happy to make the change until we get around to proving that it is!

jsturtevant: There may not be at this point. I think being defensive here might save us some heartache down the road if something acciendelty gets added later

src/hyperlight_guest_bin/src/paging.rs
syntactically: Done in b6f0ca98.

src/hyperlight_guest_bin/src/arch/amd64/exception/handle.rs
syntactically: OK, made the change in 5e239b42.

src/hyperlight_guest_bin/src/guest_function/call.rs
syntactically: Under the assumption we can, I did that change in 01c7666c.

src/hyperlight_guest_bin/src/guest_function/call.rs
syntactically: Good point! Fixed (inter alia) in 01c7666c.

src/hyperlight_host/src/mem/layout.rs
syntactically: Fixed in 889e891e.

src/hyperlight_guest_bin/src/paging.rs
syntactically: Oh, that's an obviously more idiomatic name---don't know why I didn't think of it. Thanks! Changed in 1a32096a.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
syntactically: Thanks! fixed in 63099b88.

src/hyperlight_host/src/error.rs
syntactically: Thanks! Fixed in 3d17099.

syntactically:
@ludfjig @dblnz @simongdavies @jsturtevant any further thoughts on this?

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -231,6 +234,13 @@ pub(crate) extern "C" fn generic_init(peb_address: u64, seed: u64, ops: u64, max
     unsafe {
         hyperlight_main();
     }
+
+    // Ensure that any tracing output from the initialisation phase is
+    // flushed to the host, if necessary.
+    #[cfg(all(feature = "trace_guest", target_arch = "x86_64"))]
+    hyperlight_guest_tracing::flush();
```
dblnz, line 241: Perfect, this fixes #1242

src/hyperlight_guest_bin/src/guest_function/call.rs
dblnz: I think it should be alright

src/hyperlight_host/src/mem/mgr.rs
ludfjig: Nevermind, I don't mind it

Issue #1028, opened 2025-11-19
Use `tracing` crate instead of `log` to create logs


dblnz:
We are currently using both the `tracing` and `log` crate to emit logs in `Hyperlight`. We should only use `tracing` that extends `log`'s functionality and adds more features.

Furthermore, some examples would be nice to demonstrate the use of the hyperlight crate as a dependency and show how the logs can be retrieved regardless of which crate (log or tracing) the example uses.

dblnz:
https://github.com/hyperlight-dev/hyperlight/pull/1338 is also part of this issue

PR #1234, opened 2026-02-13
Refactor hyperlight_component_util to use tracing crate

vshailesh:
- Inital changes towards addressing #1028

### DIFF
--- a/src/hyperlight_host/src/lib.rs
+++ b/src/hyperlight_host/src/lib.rs
@@ -109,33 +109,33 @@ macro_rules! log_then_return {
             None => std::format!($msg),
         };
         let __err = $crate::HyperlightError::Error(__err_msg);
-        log::error!("{}", __err);
+        tracing::error!("{}", __err);
         return Err(__err);
     }};
     ($err:expr $(,)?) => {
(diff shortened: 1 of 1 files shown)


vshailesh:
miri test failed here, does it not produce any logs? is there a way to know what is causing this failure?

syntactically:
I didn't have a chance to review the change or the motivation for it yet. Just commenting on the CI issue.

> miri test failed here, does it not produce any logs? is there a way to know what is causing this failure?

It looks like the CI runner died. I triggered a re-run.

src/hyperlight_component_util/Cargo.toml
```diff
@@ -21,4 +21,5 @@ proc-macro2 = { version = "1.0.106" }
 syn = { version = "2.0.114" }
 itertools = { version = "0.14.0" }
 prettyplease = { version = "0.2.37" }
-log = { version = "0.4" }
\ No newline at end of file
+log = { version = "0.4" }
```
dblnz: Is `log` still neded?

vshailesh: no, we can actually remove it. I will do that

PR #1241, opened 2026-02-17
Save an unconditional clone of guest parameters for registered guest functions

ludfjig:
vs main:
image

Only affects registered guest functions (not c api, nor when someone implements their own `guest_dispatch_function`).

GuestFunctionDefinition was made generic to be able to contain both GuestFunc and CGuestFunc (for use with c-api), also removes some unsafe code and adds some type-safety

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Prerelease] - Unreleased
 
+### Changed
+* **Breaking:** `GuestFunctionDefinition::new` now takes a typed function pointer instead of `usize` by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/1241
+
 ## [v0.12.0] - 2025-12-09
 
 ### Fixed
--- a/src/hyperlight_guest_bin/src/host_comm.rs
+++ b/src/hyperlight_guest_bin/src/host_comm.rs
@@ -80,12 +80,12 @@ pub fn read_n_bytes_from_user_memory(num: u64) -> Result<Vec<u8>> {
 ///
 /// This function requires memory to be setup to be used. In particular, the
 /// existence of the input and output memory regions.
-pub fn print_output_with_host_print(function_call: &FunctionCall) -> Result<Vec<u8>> {
+pub fn print_output_with_host_print(function_call: FunctionCall) -> Result<Vec<u8>> {
     let handle = unsafe { GUEST_HANDLE };
-    if let ParameterValue::String(message) = function_call.parameters.clone().unwrap()[0].clone() {
+    if let ParameterValue::String(message) = function_call.parameters.unwrap().remove(0) {
         let res = handle.call_host_function::<i32>(
             "HostPrint",
-            Some(Vec::from(&[ParameterValue::String(message.to_string())])),
+            Some(Vec::from(&[ParameterValue::String(message)])),
             ReturnType::Int,
         )?;
 
--- a/src/hyperlight_guest_capi/src/dispatch.rs
+++ b/src/hyperlight_guest_capi/src/dispatch.rs
@@ -18,7 +18,6 @@ use alloc::boxed::Box;
 use alloc::slice;
 use alloc::vec::Vec;
 use core::ffi::{CStr, c_char};
-use core::mem;
 
 use hyperlight_common::flatbuffer_wrappers::function_call::FunctionCall;
 use hyperlight_common::flatbuffer_wrappers::function_types::{ParameterType, ReturnType};
@@ -29,7 +28,8 @@ use hyperlight_guest_bin::guest_function::register::GuestFunctionRegister;
 use hyperlight_guest_bin::host_comm::call_host_function_without_returning_result;
 
 use crate::types::{FfiFunctionCall, FfiVec};
-static mut REGISTERED_C_GUEST_FUNCTIONS: GuestFunctionRegister = GuestFunctionRegister::new();
+static mut REGISTERED_C_GUEST_FUNCTIONS: GuestFunctionRegister<CGuestFunc> =
+    GuestFunctionRegister::new();
 
 type CGuestFunc = extern "C" fn(&FfiFunctionCall) -> Box<FfiVec>;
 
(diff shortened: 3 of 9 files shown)


dblnz (approved):
This is fine from my point of view. I've got burnt by this slight difference between the `guest_dispatch_function` definition and the other function definitions.

syntactically:
The perf improvement looks great, and I don't think that the amount of breaking that you mention there is necessarily a barrier. If we wanted to be very proper I suppose we would introduce a new `GuestFunctionDefinitionV2` structure and mark the original deprecated so that a misuse threw an error---but I don't know if it is worth it?

Two slight comments:
- Is there a reason why we have the function pointer on the GuestFunctionDefinition be a usize, rather than an appropriate fn() type, which might help catch any other misuses of this?
- The other direction that we could go here is perhaps to see if it is possible to get rid of the requirement of having an owned `FunctionCall` struct in order to build the parameter tuple to hand off to the registered guest function. I would be curious if you have any thoughts on doing that. Having this be owned the way it is seems a little questionable, because it means that the copy that is currently happening in `try_pop_shared_input_data_into` is impossible to soundly eliminate. Of course, the lifetime constraints get a little bit more hairy with the changes to buffer lifetime and support for async code in #1112, but it's still not totally obvious to me that making this owned would not preclude some nice ability to make direct references into the I/O buffers during the lifetime of the call. @andreiltd would you care to chime in on this at all?

syntactically:
> If we wanted to be very proper I suppose we would introduce a new GuestFunctionDefinitionV2 structure and mark the original deprecated so that a misuse threw an error---but I don't know if it is worth it?

I know @jsturtevant has strong opinions on the permissibility of breaking changes that can cause silent failure in downstreams, so paging him on this as well.

ludfjig:
Maybe we can rename `guest_dispatch_function` randomly between every version, and provide a macro that defines it, which would force users to use our macro :D We can then make changes in the future without breaking people (hopefully).

Regarding the owned `FunctionCall` indeed it does make things less flexible in the future, especially if we'd want to borrow directly from raw buffer with appropriate lifetimes. Maybe we can provide a method that users should call in order to get the parameters, to allow even more flexibility

jprendes:
This LGTM.
I agree with Lucy's concerns about borrowing data from the buffer directly.
But borrowing from the buffer would imply a complete redesign of the parameters / return value, and will certainly introduce a breaking chage regardless of what we do here.
For example, a guest function that would take a `String` would not be able to borrow from the buffer, it would have to take a &str, same with `Vec<u8>`. And those are very common patterns that we have encouraged people to use.
Until then this gives us a significant performance win, so it gets a thumbs up from me :-)

jprendes:
> Maybe we can rename `guest_dispatch_function` randomly between every version, and provide a macro that defines it, which would force users to use our macro :D We can then make changes in the future without breaking people (hopefully).
> 
> Regarding the owned `FunctionCall` indeed it does make things less flexible in the future, especially if we'd want to borrow directly from raw buffer with appropriate lifetimes. Maybe we can provide a method that users should call in order to get the parameters, to allow even more flexibility

Or we can make the type private, and make people use the `guest_function` macro :-D (joking not joking)

jsturtevant:
> > If we wanted to be very proper I suppose we would introduce a new GuestFunctionDefinitionV2 structure and mark the original deprecated so that a misuse threw an error---but I don't know if it is worth it?
> 
> I know @jsturtevant has strong opinions on the permissibility of breaking changes that can cause silent failure in downstreams, so paging him on this as well.

This seems like a pretty small and reasonable thing to do to avoid any downstream users having any issues with this.   I believe building a habit of recognizing and coming up with patterns to avoid unexpected downstream issues is a good thing.

ludfjig:
Updated @syntactically @jprendes @dblnz @jsturtevant

src/hyperlight_guest_capi/src/dispatch.rs
```diff
@@ -40,7 +40,7 @@ unsafe extern "C" {
 }
 
 #[unsafe(no_mangle)]
-pub fn guest_dispatch_function(function_call: FunctionCall) -> Result<Vec<u8>> {
+pub fn guest_dispatch_function_v2(function_call: FunctionCall) -> Result<Vec<u8>> {
```
jsturtevant: does this still need to be public? could we just use the marco going forward?

src/hyperlight_guest_capi/src/dispatch.rs
```diff
@@ -55,10 +55,7 @@ pub fn guest_dispatch_function(function_call: FunctionCall) -> Result<Vec<u8>> {
         registered_func.verify_parameters(&function_call_parameter_types)?;
 
         let ffi_func_call = FfiFunctionCall::from_function_call(function_call)?;
-
-        let guest_func =
-            unsafe { mem::transmute::<usize, CGuestFunc>(registered_func.function_pointer) };
-        let function_result = guest_func(&ffi_func_call);
+        let function_result = (registered_func.function_pointer)(&ffi_func_call);
```
jsturtevant, line 58: Is this borrow right here?

ludfjig: good observation, but yes it is correct. The guest c-api has it's own function signature that needs to be `hl_Vec *c_guest_dispatch_function(const hl_FunctionCall *function_call)`

src/hyperlight_guest_capi/src/dispatch.rs
ludfjig: oh yeah let's switch to the macro here too! Fixed

CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 
 ## [Prerelease] - Unreleased
 
+### Changed
+* **Breaking:** `GuestFunctionDefinition::new` now takes a typed function pointer instead of `usize` by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/1241
```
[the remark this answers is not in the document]
ludfjig: out of date desc

src/hyperlight_guest_bin/src/host_comm.rs
```diff
@@ -80,9 +80,9 @@ pub fn read_n_bytes_from_user_memory(num: u64) -> Result<Vec<u8>> {
 ///
 /// This function requires memory to be setup to be used. In particular, the
 /// existence of the input and output memory regions.
-pub fn print_output_with_host_print(function_call: &FunctionCall) -> Result<Vec<u8>> {
+pub fn print_output_with_host_print(function_call: FunctionCall) -> Result<Vec<u8>> {
     let handle = unsafe { GUEST_HANDLE };
-    if let ParameterValue::String(message) = function_call.parameters.clone().unwrap()[0].clone() {
+    if let ParameterValue::String(message) = function_call.parameters.unwrap().remove(0) {
```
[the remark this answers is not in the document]
ludfjig: old behavior

jsturtevant (approved):

PR #1252, opened 2026-02-23
Explicitly error out on host-guest version mismatch

### LINKED ISSUE #845: What are the backwards compatibility guarantees for guest binaries?
If I have a guest binary compiled with an older hyperlight_guest, will it run in a newer hyperlight_host?
In particular, I see there have been a few changes in [outb.rs](https://github.com/hyperlight-dev/hyperlight/blame/main/src/hyperlight_common/src/outb.rs) in the last few months.

ludfjig:
Hyperlight currently provides no backwards compatibility guarantees for guests. We should therefore error out explicitly if someone loads an old guest with new host, as they ABI might no longer match. 

This PR enforces that hyperlight-host version must match the guest's hyperlight-guest-bin version.

Question: If the elf section is not found, we still load it. This is good for running guests that don't use hyperlight-guest-bin. However, should we enforce that the section must exist? Does anyone run guests without depending on hyperlight-guest-bin? In addition, stripping the binary will remove the new elf section, which is also something to consider

Closes #845

### DIFF
--- a/src/hyperlight_common/src/version_note.rs
+++ b/src/hyperlight_common/src/version_note.rs
@@ -63,7 +63,7 @@ pub const fn padded_desc_size(desc_len_with_nul: usize) -> usize {
 /// `str.len() + 1` (the null-terminated length).
 ///
 /// The constructor enforces these constraints with compile-time assertions.
-#[repr(C)]
+#[repr(C, align(8))]
 pub struct ElfNote<const NAME_SZ: usize, const DESC_SZ: usize> {
     namesz: u32,
     descsz: u32,


ludfjig:
> This looks good, but do we want to match against the patch version also?
> 
> I think it would be better to ensure that the guest's major/minor are the same, but the patch version is greater or equal the host's version.

I thought about this, and then decided to be as conservative as possible, because this would limit some things we can update in patch versions. Maybe this is a good start, and then we can consider later whether we should relax it?

jsturtevant:
> I thought about this, and then decided to be as conservative as possible, because this would limit some things we can update in patch versions. Maybe this is a good start, and then we can consider later whether we should relax it?

Keeping it restrictive to start then relaxing later is a good option.  Its harder to go the other way.

src/hyperlight_host/src/mem/exe.rs
```diff
@@ -103,3 +112,110 @@ impl ExeInfo {
...
+            info.guest_bin_version().unwrap(),
+            env!("CARGO_PKG_VERSION"),
+            "patched version should differ from host version"
+        );
+    }
+
+    /// Load an unpatched simpleguest through `Snapshot::from_env` and verify
+    /// that it succeeds when the embedded version matches the host version.
+    #[test]
+    fn from_env_accepts_matching_version() {
```
jsturtevant, line 208: do we have an `accepts_guest_not_built_with_higihterlight guest`?

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -119,6 +119,19 @@ pub static mut GUEST_HANDLE: GuestHandle = GuestHandle::new();
 pub(crate) static mut REGISTERED_GUEST_FUNCTIONS: GuestFunctionRegister =
     GuestFunctionRegister::new();
 
+// Embed the hyperlight-guest-bin crate version in a dedicated ELF section so
+// the host can verify ABI compatibility at load time.  The section name is
+// defined by [`hyperlight_common::hyperlight_guest_bin_version_section!`].
+core::arch::global_asm!(concat!(
```
syntactically: We do a similar thing in hyperlight-wasm's build.rs, which uses `.note_hyperlight_metadata` and does it in Rust via `#[link_section]`---is there a reason to do it differently here/there?

I think SHT_PROGBITS is not the right section type for this. I went to look at what we are doing in hlwasm, and it looks like it is automagically getting SHT_NOTE, probably because the section name starts with `.note`.  I dion't think that that's strictly correct either---it looks like in the System V gABI there is an explicit structure that SHT_NOTE sections are meant to follow: https://www.sco.com/developers/gabi/2003-12-17/ch5.pheader.html#note_section. However, the intent of the note section does seem like the best fit for this information, and the way we are using it seems not totally out of line with e.g. `.note.gnu.build-id` and `.note.package`, so I feel like perhaps we should just go with name = .note.whatever, sh_type = SHT_NOTE, and put the little note header structure in.

One difference here is that we may not actually need our sections to be `SHF_ALLOC` / covered by any segment, since we are checking them at load time, and we probably will do our own format for conveying the same information in snapshots.  However, the waste of memory is very small, so if it isn't easy to avoid that (it might be easy with some assembler directives? not sure), it might not be worth doing anything about.

We could also use a custom section type from the SHT_LOUSER to SHT_HIUSER range, but I think using SHT_NOTE makes more sense.

If we do this just with the link_section attribute, we should check where that is getting turned into an SHT_NOTE section type---it looks like there are incompatibilities between ld.bfd and ld.lld if that's what's being relied on (https://reviews.llvm.org/D118490)

src/hyperlight_host/src/mem/exe.rs
ludfjig: added

src/hyperlight_guest_bin/src/lib.rs
ludfjig: pushed an update!

src/hyperlight_common/src/version_note.rs
```diff
@@ -0,0 +1,84 @@
...
+
+// SAFETY: ElfNote contains only plain data (`u32` and `[u8; N]`).
+// Required because ElfNote is used in a `static` (for `#[link_section]`),
+// and `static` values must be `Sync`.
+unsafe impl<const N: usize, const D: usize> Sync for ElfNote<N, D> {}
+
+impl<const NAME_SZ: usize, const DESC_SZ: usize> ElfNote<NAME_SZ, DESC_SZ> {
+    /// Create a new ELF note from a name string, descriptor string, and type.
+    ///
+    /// `NAME_SZ` and `DESC_SZ` must equal `name.len() + 1` and `desc.len() + 1`
```
syntactically: Nit: can you add a little to this comment about the consequences if they do not? It looks like there is an ABI invariant that is broken (although I don't know if anybody relies on that invariant). Probably should be either unsafe or checked in the constructor to be sure?

docs/how-to-build-a-hyperlight-guest-binary.md
```diff
@@ -30,3 +30,15 @@ latest release page that contain: the `hyperlight_guest.h` header and the
 C API library.
 The `hyperlight_guest.h` header contains the corresponding APIs to register
 guest functions and call host functions from within the guest.
+
+## Version compatibility
+
+Guest binaries built with `hyperlight-guest-bin` automatically embed the crate
+version in an ELF note section (`.note.hyperlight.version`). When the host
```
syntactically: Very minor nit---I think `.note.hyperlight-version` is a little more consistent with other section names I've seen, e.g. `.note.GNU-stack`, `.note.build-id`, `.note.ABI-tag`, etc.

src/hyperlight_common/src/version_note.rs
```diff
@@ -0,0 +1,84 @@
...
+/// The owner name used in the ELF note header for hyperlight version metadata.
+pub const HYPERLIGHT_NOTE_NAME: &str = "Hyperlight";
+
+/// The note type value used in the ELF note header for hyperlight version metadata.
+pub const HYPERLIGHT_NOTE_TYPE: u32 = 1;
+
+/// A byte array with 4-byte alignment, used for ELF note name/descriptor
+/// fields. The compiler inserts trailing padding automatically so that the
+/// next field starts at a 4-byte boundary.
+#[repr(C, align(4))]
```
syntactically: I think the alignment is meant to be 8 byte here for us? In the [System V ABI](https://www.sco.com/developers/gabi/latest/ch5.pheader.html#note_section):
> In 64-bit objects (files with e_ident[EI_CLASS] equal to ELFCLASS64), each entry is an array of 8-byte words in the format of the target processor

src/hyperlight_common/src/version_note.rs
ludfjig: this is asserted at compile time now

docs/how-to-build-a-hyperlight-guest-binary.md
ludfjig: fixed

src/hyperlight_common/src/version_note.rs
ludfjig: fixed. Removed the alignment wrapper since align8 would move `name` to incorrect offset

syntactically: > since align8 would move name to incorrect offset

Didn't have a chance to look at the new code yet, will shortly. This comment makes sense but is also a bit odd. The ABI seems a little underspecified here, since it doesn't mention the sizes of namesz and sizesz. Can you check what the layout/alignment being used in other tools is to make sure we match up?

ludfjig: yeah it is indeed a bit odd. I just meant to highligh that `name` must start at offset 12 which a struct like Align8 would prevent. And yes, everyone seems to use u32 for these, even on 64-bit targets

syntactically: I mostly wanted to double-check that others were also using the same 12 as start byte and the same approach to the alignment.

ludfjig: https://github.com/torvalds/linux/blob/master/include/uapi/linux/elf.h#L573-L577 using https://github.com/torvalds/linux/blob/master/include/uapi/linux/elf.h#L22

dblnz (approved):

Issue #1191, opened 2026-01-26
Make it easier to create guests


ludfjig:
It's currently pretty hard as can be seen here https://github.com/hyperlight-dev/hyperlight/issues/1175

danbugs:
Discussed w/ the team. It would be great to create some sort of tool for this–e.g., add a scaffolding command to cargo-hyperlight.

jsturtevant:
this is resolved by https://github.com/hyperlight-dev/cargo-hyperlight/pull/36 and readme will be updated in #1359

Issue #1175, opened 2026-01-16
duplicate lang item `panic_impl'


sarocu:
### What happened?

Compiling a simple guest example fails with:
```
error[E0152]: found duplicate lang item `panic_impl`
   --> /home/sarocu/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/hyperlight-guest-bin-0.12.0/src/lib.rs:139:1
    |
139 | / fn panic(info: &core::panic::PanicInfo) -> ! {
140 | |     _panic_handler(info)
141 | | }
    | |_^
    |
    = note: the lang item is first defined in crate `std` (which `tracing` depends on)
    = note: first definition in `std` loaded from /home/sarocu/.rustup/toolchains/1.89-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-unknown-linux-gnu/lib/libstd-099ea48a26be4ce8.rlib
    = note: second definition in the local crate (`hyperlight_guest_bin`)
```

Dependencies:
```
[dependencies]
hyperlight-common = "0.12.0"
hyperlight-guest = "0.12.0"
hyperlight-guest-bin = "0.12.0"
hyperlight-guest-tracing = "0.12.0"
```

Rust:
```
cargo --version
cargo 1.89.0 (c24e10642 2025-06-23)
```

### What did you expect to happen?

Successful build

### Steps to reproduce the behavior

1. Install the shared dependencies in a new project
2. Copy the example snippet for a guest hyperlight binary in the README
3. `cargo build`

### Hyperlight Version

0.12.0

### OS version

```
             /////////////                
         /////////////////////            -------------
      ///////*767////////////////         OS: Pop!_OS 24.04 LTS x86_64
    //////7676767676*//////////////       Kernel: 6.17.9-76061709-generic
   /////76767//7676767//////////////      Uptime: 3 hours, 32 mins
  /////767676///*76767///////////////     Packages: 2082 (dpkg), 13 (flatpak)
 ///////767676///76767.///7676*///////    Shell: bash 5.2.21
/////////767676//76767///767676////////   Resolution: 3440x1440
//////////76767676767////76767/////////   DE: COSMIC
///////////76767676//////7676//////////   Theme: adw-gtk3 [GTK3]
////////////,7676,///////767///////////   Icons: Cosmic [GTK3]
/////////////*7676///////76////////////   Terminal: cosmic-term
[...]

PR #1254, opened 2026-02-24
docs: improve guest creation documentation

jsturtevant:
Add a few more directions to readme and add some trouble shooting.

addresses #1191

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3205,9 +3205,9 @@ dependencies = [
 
 [[package]]
 name = "serial_test"
-version = "3.3.1"
+version = "3.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0d0b343e184fc3b7bb44dff0705fffcf4b3756ba6aff420dddd8b24ca145e555"
+checksum = "911bd979bf1070a3f3aa7b691a3b3e9968f339ceeec89e08c280a8a22207a32f"
 dependencies = [
  "futures-executor",
  "futures-util",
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -69,10 +69,30 @@ fn main() -> hyperlight_host::Result<()> {
...
+> **Important:** The `hyperlight-common` crate must have `default-features = false` to avoid pulling in
+> the standard library, which conflicts with the `no_std` requirement for guests.
+
+Then, create `src/main.rs`:
+
 ```rust
 #![no_std]
 #![no_main]
 extern crate alloc;
+extern crate hyperlight_guest_bin;
```
dblnz, line 95: Is this really needed? The `simpleguest` doesn't do this, it does it for `hyperlight_guest` though 🤔

ludfjig: I think it's necessary when the crate doesn't use any other hyperlight_guest_bin features

jsturtevant: yea i think that is what we found out here: https://github.com/hyperlight-dev/hyperlight/issues/1175#issuecomment-3786471028

jsturtevant:
>LGTM. What do you think about adding a minimal (weekly?) test that tries to compile a minimal guest like the one I posted in https://github.com/hyperlight-dev/hyperlight/issues/1175 using cargo-hyperlight?

Yea, maybe using the code from the readme would be a good thing since it isn't currently tested as is

dblnz:
> LGTM. What do you think about adding a minimal (weekly?) test that tries to compile a minimal guest like the one I posted in #1175 using cargo-hyperlight? Maybe we can also link to it from the docs to show what a minimal guest looks like. I think it could be useful, and by making it weekly, we don't slow down existing stuff like `just guests`

Wouldn't the `cargo-hyperlight` repository be a better match for this?

jsturtevant:
> Wouldn't the `cargo-hyperlight` repository be a better match for this?

we have one there: https://github.com/hyperlight-dev/cargo-hyperlight/tree/main/examples so maybe this isn't needed. but would be nice to have a test that validates our sample works.

PR #1260, opened 2026-02-25
Fix tracing cow changes

### LINKED ISSUE #1242: Fix guest tracing related to CoW changes
With the CoW changes, there are two things that are not working any more:
- `hyperlight_main` tracing - because tracing logic that relied on the fact that before each `hlt` instruction made by the `halt` function, the traces were sent to the host.
This does not hold true any more because the `hlt` instruction is written in assembly after the call to `generic_init`.
- `dispatch_function` - in the future PRs for CoW, most likely, `dispatch_function` will have the same approach.

We need to consider some changes to accommodate the new approach.

Furthermore, the `Info` trace for the `dispatch_function` is not shown because the tracing state is reset right after the call to the function comes from the guest. This produces warnings because the trace event closing the span is sent to the host, but no open span event is sent.

dblnz:
## Description

This PR fixes #1242.
The guest tracing data received by the host assumes identity mapped addresses, the address needs to be translated to correctly decode the flatbuffers data.

This can be reviewed commit by commit.
The first commit tackles this issue, the other two improve on the spans and usability.

### DIFF
--- a/src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
+++ b/src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
@@ -41,7 +41,7 @@ pub enum FunctionCallType {
 }
 
 /// `Functioncall` represents a call to a function in the guest or host.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct FunctionCall {
     /// The function name
     pub function_name: String,
--- a/src/hyperlight_guest_bin/src/guest_function/call.rs
+++ b/src/hyperlight_guest_bin/src/guest_function/call.rs
@@ -91,9 +91,6 @@ pub(crate) fn internal_dispatch_function() {
         .try_pop_shared_input_data_into::<FunctionCall>()
         .expect("Function call deserialization failed");
 
-    #[cfg(debug_assertions)]
-    tracing::trace!("{:?}", function_call);
-
     let res = call_guest_function(function_call);
 
     match res {
--- a/src/hyperlight_host/tests/integration_test.rs
+++ b/src/hyperlight_host/tests/integration_test.rs
@@ -744,14 +744,10 @@ fn log_message() {
     //  - internal_dispatch_function does a log::trace! in debug mode
     //  - logs from trace level tracing spans created as logs because of the tracing `log` feature
     //    - 4 from evolve call (hyperlight_main + halt)
-    //    - 8 from guest call (internal_dispatch_function + others)
+    //    - 7 from guest call
     // and are multiplied because we make 6 calls to `log_test_messages`
     // NOTE: These numbers need to be updated if log messages or spans are added/removed
-    let num_fixed_trace_log = if cfg!(debug_assertions) {
-        (1 + 12) * 6
-    } else {
-        12 * 6
-    };
+    let num_fixed_trace_log = 12 * 6;
 
     let tests = vec![
         (LevelFilter::Trace, 5 + num_fixed_trace_log),


src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -117,8 +119,20 @@ pub(crate) fn internal_dispatch_function() {
...
+        // NOTE: This is necessary to avoid closing the span twice. Flush closes all the open
+        // spans, when preparing to close a guest function call context.
+        // It is not mandatory, though, but avoids a warning on the host that alerts a spans
+        // that has not been opened but is being closed.
+        _entered.exit();
+
+        // Ensure that any tracing output during the call is flushed to
+        // the host, if necessary.
+        hyperlight_guest_tracing::flush();
+    }
```
ludfjig, line 128: Could this be factored out to not be duplicated here and in lib.rs?

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -235,10 +241,22 @@ pub(crate) extern "C" fn generic_init(
...
     }
 
-    // Ensure that any tracing output from the initialisation phase is
-    // flushed to the host, if necessary.
+    // All this tracing logic shall be done right before the call to `hlt` which is done after this
+    // function returns
     #[cfg(all(feature = "trace_guest", target_arch = "x86_64"))]
-    hyperlight_guest_tracing::flush();
+    {
+        // This span captures the internal dispatch function only, without tracing internals.
```
ludfjig: I think this comment shoud say generic_init as opposed to internal dispatch function maybe?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -752,12 +752,26 @@ impl HyperlightVm {
...
-                    // Handle the guest trace data if any
-                    let regs = self.vm.regs().map_err(RunVmError::GetRegs)?;
-                    if let Err(e) = tc.handle_trace(&regs, mem_mgr) {
-                        // If no trace data is available, we just log a message and continue
-                        // Is this the right thing to do?
-                        log::debug!("Error handling guest trace: {:?}", e);
+                    match result.as_ref() {
+                        // On the Halt exit we don't receive any trace information
+                        Ok(VmExit::Halt()) => {}
+                        _ => {
```
ludfjig: nit: could be simplified slightly to something like 

```rust
if !matches!(result.as_ref(), Ok(VmExit::Halt())) {
    let regs = self.vm.regs().map_err(RunVmError::GetRegs)?;
    if let Ok(cr3) = self.vm.sregs().map(|s| s.cr3) {
        tc.handle_trace(&regs, mem_mgr, cr3).unwrap_or_else(|e| {
            tracing::error!("Cannot handle trace data: {}", e);
        });
    } else {
        tracing::error!("Failed to get sregs for trace handling");
    }
}
```

src/hyperlight_host/src/sandbox/trace/context.rs
```diff
@@ -28,73 +28,50 @@ use tracing::span::{EnteredSpan, Span};
...
+impl EventsBatch {
+    /// Extract a batch of guest trace events from guest memory.
+    ///
+    /// The guest passes the trace data pointer as a Guest Virtual Address (GVA)
+    /// in register r9. With Copy-on-Write enabled, this GVA may not be
+    /// identity-mapped to its physical address, so we walk the guest page
+    /// tables to translate GVA → GPA before reading the data.
+    ///
+    /// # Arguments
+    /// * `regs` - The guest registers (r8 = magic, r9 = GVA pointer, r10 = length)
```
ludfjig, line 49: I think this comment is using outdated regsiters

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -411,6 +411,102 @@ impl SandboxMemoryManager<HostSharedMemory> {
 
         Ok(())
     }
+
+    /// Read guest memory at a Guest Virtual Address (GVA) by walking the
+    /// page tables to translate GVA → GPA, then reading from the correct
+    /// backing memory (shared_mem or scratch_mem).
+    ///
+    /// This is necessary because with Copy-on-Write (CoW) the guest's
+    /// virtual pages are be backed by physical pages in the scratch
```
ludfjig: typo "are be"

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -411,6 +411,102 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
+
+                if mappings.is_empty() {
+                    return Err(new_error!(
+                        "No page table mappings found for GVA {:#x} (len {})",
+                        gva,
+                        len,
+                    ));
+                }
+
+                let mut result = Vec::with_capacity(len);
```
jsturtevant, line 459: Is this length controlled by the guest? coming from context.rs:

```
let trace_data_len = regs.r14 as usize;
        let buf = mem_mgr.read_guest_memory_by_gva(trace_data_gva, trace_data_len, root_pt)?;
```

jsturtevant: nice this was cuaght in https://github.com/hyperlight-dev/hyperlight/pull/1260#discussion_r2854770009

ludfjig: very good point, we need to have some max size we allow for sure

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -752,12 +752,26 @@ impl HyperlightVm {
...
+                            // Handle the guest trace data if any
+                            let regs = self.vm.regs().map_err(RunVmError::GetRegs)?;
```
jsturtevant: since we know the `outb action (   in("r12") OutBAction::TraceBatch as u64)`, could we skip doing this for every instance with something like: 

```
     
if regs.r12 == OutBAction::TraceBatch then
 do all the tracing stuff here
```

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -411,6 +411,102 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
+
+        let scratch_size = self.scratch_mem.mem_size();
+
+        self.shared_mem.with_exclusivity(|snap| {
+            self.scratch_mem.with_exclusivity(|scratch| {
+                let pt_buf = SharedMemoryPageTableBuffer::new(snap, scratch, scratch_size, root_pt);
+
+                // Walk page tables to get all mappings that cover the GVA range
+                let mappings: Vec<_> = unsafe {
+                    hyperlight_common::vmem::virt_to_phys(&pt_buf, gva, len as u64)
```
jsturtevant, line 446: This lengt isn't necessarily aligned to a page I think? 

This isn't in your code, but I think this might expose an issue where if this length goes past one page, it could truncate the result and not give you the right mappings.  We are missing a unit test that covers that scenario in vmem. 

You could do ` len as u64 + page_offset` but we should probably fix it in vmem so that we don't accidently mis use this function again

src/hyperlight_guest_bin/src/guest_function/call.rs
dblnz: not really, we need `flush` to be right before the call to `hlt` so that no other spans/logs are added.

src/hyperlight_host/src/mem/mgr.rs
dblnz: yeah, I think I'll limit it at 4K for now, but in the future we'd be able to shrink it because we would use the virtio queues and we won't care about this. We could even not batch them at all, the queue would do that for us

src/hyperlight_host/src/mem/mgr.rs
dblnz: I'll add some tests for this

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -752,12 +752,26 @@ impl HyperlightVm {
...
+                            // If something goes wrong with parsing the trace data, we log the error and
+                            // continue execution instead of returning an error since this is not critical
+                            // to correct execution of the guest
+                            match self.vm.sregs().map(|s| s.cr3) {
+                                Ok(cr3) => {
+                                    tc.handle_trace(&regs, mem_mgr, cr3).unwrap_or_else(|e| {
```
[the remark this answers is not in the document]
dblnz: I am not sure if I need to do this.
@syntactically, I've seen it's not done on the host, should I change this?

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -920,6 +923,52 @@ mod tests {
         assert_eq!(mapping.phys_base, 0x1000);
     }
 
+    #[test]
+    fn test_virt_to_phys_unaligned_virt_and_across_pages_len() {
+        let ops = MockTableOps::new();
+        let mapping = Mapping {
+            phys_base: 0x1000,
+            virt_base: 0x1000,
+            len: 2 * PAGE_SIZE as u64, // 2 page + 512 bytes
```
dblnz:
```suggestion
            len: 2 * PAGE_SIZE as u64, // 2 page
```

dblnz:
> BTW why were the registers changed?

The registers changed because I wanted to ensure the registers used couldn't be overwritten when returning from a function (currently, we set the registers in preparation for a `hlt` to execute, and the `hlt` instruction is written in assembly after the `call generic_init` or `call internal_dispatch_function` finishes).
By reading some more info on [calling convention](https://wiki.osdev.org/Calling_Conventions), it looks like `r12,r13,r14` are preserved registers.

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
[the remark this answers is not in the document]
syntactically: Actually, I think it is a bug that it's not done on the host---we're not using any of those bits right now, but we might in the future. I would do this early on in `HyperlightVm::get_root_pt()` actually, so that architecture-specific stuff like "the mapping from the sysreg to the linear gpa of the root page table base" is dealt with as early as possible.

src/hyperlight_common/src/arch/amd64/vmem.rs
```diff
@@ -539,7 +539,10 @@ pub unsafe fn virt_to_phys<'a, Op: TableReadOps + 'a>(
 ) -> impl Iterator<Item = Mapping> + 'a {
     // Undo sign-extension, and mask off any sub-page bits
     let vmin = (address & ((1u64 << VA_BITS) - 1)) & !(PAGE_SIZE as u64 - 1);
-    let vmax = core::cmp::min(vmin + len, 1u64 << VA_BITS);
+    let addr_canonical = address & ((1u64 << VA_BITS) - 1);
```
syntactically: Good catch! This is a minor nit, but the name `addr_canonical` is slightly confusing here because this is actually iirc a non-canonical form---the canonical form is when the high bits agree with the msb.

src/hyperlight_guest/src/exit.rs
```diff
@@ -103,9 +103,9 @@ pub(crate) unsafe fn out32(port: u16, val: u32) {
                 asm!("out dx, eax",
                     in("dx") port,
                     in("eax") val,
-                    in("r8") OutBAction::TraceBatch as u64,
```
syntactically, line 106: Is there any specific motivation for switching these registers from before?

src/hyperlight_common/src/arch/amd64/vmem.rs
dblnz: yeah, I'll rename. Thanks!

src/hyperlight_guest/src/exit.rs
dblnz: Yeah, I checked the [calling convention](https://wiki.osdev.org/Calling_Conventions) and wanted to ensure the registers are preserved when `hlt` is executed.
```
core;:arch::global_asm!("
    .
    .
    call {generic_init}\n
    hlt\n
", ..);
```

syntactically:
At a high level, I think it adds some complexity in the long term to do everything in terms of VAs here, even if it requires fewer changes in the guest immediately to not need to deal with making sure the buffer allocation is contiguous (think like how big kernels do DMA-pinned memory). The specific way that the host code accesses the trace buffers looks to me like it will be fairly slow as well. However, perhaps these are not major concerns if we are going to switch to using the same ring buffer transport as all other asyncio soon. (Side comment: we should definitely still be batching these trace records when we do that---descriptor capacity is limited and these records are tiny individually afaiui.)

> > BTW why were the registers changed?
> 
> The registers changed because I wanted to ensure the registers used couldn't be overwritten when returning from a function (currently, we set the registers in preparation for a `hlt` to execute, and the `hlt` instruction is written in assembly after the `call generic_init` or `call internal_dispatch_function` finishes). By reading some more info on [calling convention](https://wiki.osdev.org/Calling_Conventions), it looks like `r12,r13,r14` are preserved registers.

- Where are we setting the registers and then returning to hlt? Both of the call sites in fa7873dafaa094a8968dbbc42c0b56ef8b82fce6 look like they directly do the outb.
- If we are actually doing that somewhere, this reading of "preserved" is backwards---if you look at [the abi](https://gitlab.com/x86-psABIs/x86-64-ABI/-/jobs/artifacts/master/raw/x86-64-ABI/abi.pdf?job=build) it uses the much more illuminating name "callee-save": i.e. if some code calls another function, it should expect by the time of the return that those registers have been restored to whatever they were before, and can do whatever it wants with them. So, not only does using these registers not prevent them being overwritten during the prologue for `generic_init`, it actually makes it an ABI violation for generic_init _not_ to do that.

dblnz:
> At a high level, I think it adds some complexity in the long term to do everything in terms of VAs here, even if it requires fewer changes in the guest immediately to not need to deal with making sure the buffer allocation is contiguous (think like how big kernels do DMA-pinned memory). The specific way that the host code accesses the trace buffers looks to me like it will be fairly slow as well. However, perhaps these are not major concerns if we are going to switch to using the same ring buffer transport as all other asyncio soon. (Side comment: we should definitely still be batching these trace records when we do that---descriptor capacity is limited and these records are tiny individually afaiui.)

I chose to use VAs because it seemed simpler to access the guest trace buffers that way. My plan is to definitely use the asyncio features once we have them and get rid of the logic handling the buffer info exchanges between the guest and host.

> Where are we setting the registers and then returning to hlt? Both of the call sites in [fa7873d](https://github.com/hyperlight-dev/hyperlight/commit/fa7873dafaa094a8968dbbc42c0b56ef8b82fce6) look like they directly do the outb.

That's correct, we used to also set those registers before `hlt` to avoid executing an `out` instruction before it, be we don't do that any more.

> If we are actually doing that somewhere, this reading of "preserved" is backwards---if you look at [the abi](https://gitlab.com/x86-psABIs/x86-64-ABI/-/jobs/artifacts/master/raw/x86-64-ABI/abi.pdf?job=build) it uses the much more illuminating name "callee-save": i.e. if some code calls another function, it should expect by the time of the return that those registers have been restored to whatever they were before, and can do whatever it wants with them. So, not only does using these registers not prevent them being overwritten during the prologue for `generic_init`, it actually makes it an ABI violation for generic_init _not_ to do that.

Thanks for clearing this up for me, I got it backwards

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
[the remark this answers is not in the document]
dblnz: Fixed!

src/hyperlight_common/src/arch/amd64/vmem.rs
dblnz: Fixed!

src/hyperlight_guest/src/exit.rs
dblnz: Reverted to `r8,r9,r10`. Thanks for your input

dblnz:
@jsturtevant @ludfjig do you have something else to add?

src/hyperlight_common/src/flatbuffer_wrappers/function_call.rs
```diff
@@ -41,7 +41,7 @@ pub enum FunctionCallType {
 }
 
 /// `Functioncall` represents a call to a function in the guest or host.
-#[derive(Clone)]
+#[derive(Clone, Debug)]
```
ludfjig: did you need this for anything particular?

dblnz: Yeah, I added a `trace` event to the `dispatch function` in the guest.

```
 tracing::trace!("{:?}", function_call);
```

dblnz: It is useful for the traces

ludfjig: ok just curious. Could potentially contain user secrets or similar so might want to be careful, just a random thought

dblnz: Yeah, that's right, I'll remove it.

ludfjig (approved):

dblnz:
Needed to modify the `log_message` tests also when I removed the `trace` log

jsturtevant (approved):

jsturtevant (approved):

PR #1262, opened 2026-02-25
Validate guest-writable fields in try_pop_buffer_into before allocation

ludfjig:
`try_pop_buffer_into` reads two values from guest-writable shared memory

- **Size prefix**: A malicious guest can set this to near u32::MAX, causing the host to attempt a ~4 GB allocation via vec![0; fb_buffer_size] (DoS).
- **Back-pointer**: A malicious guest can point this outside the buffer, causing fill(0, ...) to zero unrelated host-managed data within shared memory (e.g., PEB fields, input buffer) and leaving the buffer stack in an inconsistent state.
- **Arithmetic overflow/underflow**: A back-pointer just below the stack pointer causes underflow in stack_pointer_rel - last_element_offset_rel - 8. A size prefix of 0xFFFF_FFFD causes +4 to wrap around in debug (panic) or release (silent wrap to a small value).

Changes:

- Validate last_element_offset_rel (back-pointer) is within [[8, stack_pointer_rel) before use
- Validate fb_buffer_size does not exceed the actual element slot size before allocation
- Use checked_add(4) on the size prefix instead of bare + 4, returning a descriptive error on u32 overflow
- Add unit tests covering corrupt size prefix, corrupt back-pointer (too small / too large), off-by-one, back-pointer near stack pointer (underflow), and u32 size prefix overflow
- Add integration tests with guest functions that write corrupt output buffers and halt

### DIFF
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -1121,17 +1121,46 @@ impl HostSharedMemory {
         let last_element_offset_rel: usize =
             self.read::<u64>(last_element_offset_abs - 8)? as usize;
 
+        // Validate element offset (guest-writable): must be in [8, stack_pointer_rel - 16]
+        // to leave room for the 8-byte back-pointer plus at least 8 bytes of element data
+        // (the minimum for a size-prefixed flatbuffer: 4-byte prefix + 4-byte root offset).
+        if last_element_offset_rel > stack_pointer_rel.saturating_sub(16)
+            || last_element_offset_rel < 8
+        {
+            return Err(new_error!(
+                "Corrupt buffer back-pointer: element offset {} is outside valid range [8, {}].",
+                last_element_offset_rel,
+                stack_pointer_rel.saturating_sub(16),
+            ));
+        }
+
         // make it absolute
         let last_element_offset_abs = last_element_offset_rel + buffer_start_offset;
 
+        // Max bytes the element can span (excluding the 8-byte back-pointer).
+        let max_element_size = stack_pointer_rel - last_element_offset_rel - 8;
+
         // Get the size of the flatbuffer buffer from memory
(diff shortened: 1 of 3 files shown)


src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1121,18 +1121,46 @@ impl HostSharedMemory {
         let last_element_offset_rel: usize =
             self.read::<u64>(last_element_offset_abs - 8)? as usize;
 
+        // Validate element offset (guest-writable): must be in [8, stack_pointer_rel - 8]
+        // to leave room for at least the 8-byte back-pointer.
+        if last_element_offset_rel > stack_pointer_rel.saturating_sub(16)
```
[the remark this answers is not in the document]
ludfjig: 1 byte elements are not valid flatbuffer encodings

andreiltd (approved):

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1631,6 +1660,192 @@ mod tests {
...
+            data.extend_from_slice(&(payload.len() as u32).to_le_bytes());
+            data.extend_from_slice(payload);
+
+            hshm.push_buffer(0, mem_size, &data).unwrap();
+            let result: RawBytes = hshm.try_pop_buffer_into(0, mem_size).unwrap();
+            assert_eq!(result.0, data);
+        }
+
+        #[test]
+        fn malicious_flatbuffer_size_prefix() {
```
syntactically, line 1702: Not suggesting this for this PR since I'm not sure if we have the infrastructure, but wanted to make a drive-by comment that it would be nice to see if we could use proptests (even better: coverage-guided proptests) more widely as supplemental testing for this type of invariant.

ludfjig: I think that is a great idea!

PR #1263, opened 2026-02-25
fix: use checked arithmetic in bounds_check! to prevent overflow bypass

ludfjig:
The bounds check silently wraps on usize overflow in release builds, bypassing the check entirely. Not exploitable today since callers validate offsets before reaching the macro, but any future caller that doesn't could trigger it.

### DIFF
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -48,7 +48,7 @@ use crate::{HyperlightError, Result, log_then_return, new_error};
 /// Makes sure that the given `offset` and `size` are within the bounds of the memory with size `mem_size`.
 macro_rules! bounds_check {
     ($offset:expr, $size:expr, $mem_size:expr) => {
-        if $offset + $size > $mem_size {
+        if $offset.checked_add($size).is_none_or(|end| end > $mem_size) {
             return Err(new_error!(
                 "Cannot read value from offset {} with size {} in memory of size {}",
                 $offset,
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1220,6 +1220,34 @@ mod tests {
         assert!(hshm.fill(0, mem_size, 1).is_err());
     }
 
+    /// Verify that `bounds_check!` rejects offset + size combinations that
+    /// would overflow `usize`.
+    #[test]
+    fn bounds_check_overflow() {
+        let mem_size: usize = 4096;
+        let mut eshm = ExclusiveSharedMemory::new(mem_size).unwrap();
```
jsturtevant, line 1228: nit: I prefer exclusive_shared_memory for the variable name, makes it easier to read later, but I know thats not for everyone

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -48,7 +48,7 @@ use crate::{HyperlightError, Result, log_then_return, new_error};
 /// Makes sure that the given `offset` and `size` are within the bounds of the memory with size `mem_size`.
 macro_rules! bounds_check {
     ($offset:expr, $size:expr, $mem_size:expr) => {
-        if $offset + $size > $mem_size {
+        if $offset.checked_add($size).is_none_or(|end| end > $mem_size) {
             return Err(new_error!(
```
jsturtevant, line 52: Should we update this to be a typed error? inline with your other work?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1220,6 +1220,34 @@ mod tests {
...
+    /// Verify that `bounds_check!` rejects offset + size combinations that
+    /// would overflow `usize`.
+    #[test]
+    fn bounds_check_overflow() {
+        let mem_size: usize = 4096;
+        let mut eshm = ExclusiveSharedMemory::new(mem_size).unwrap();
+
+        // ExclusiveSharedMemory methods
+        assert!(eshm.read_i32(usize::MAX).is_err());
+        assert!(eshm.write_i32(usize::MAX, 0).is_err());
```
jsturtevant, line 1232: Could we check if this error is of a specific type or message. to ensure we are getting the right message?

dblnz (approved):

src/hyperlight_host/src/mem/shared_mem.rs
ludfjig: Yes I will do that when converting this module later

src/hyperlight_host/src/mem/shared_mem.rs
ludfjig: This is a good idea, but will do this once I add better error typing for this module

Issue #1200, opened 2026-01-28
Guest Copy on Write


jsturtevant:
Tracking issues Guest Copy on Write.  

## Brief design overview

When running on a Type 1 hypervisor, servicing a Stage 2 translation
page fault is relatively quite expensive, since it requires quite a
lot of context switches.  To help alleviate this, Hyperlight will use a
design in which the guest is aware of readonly snapshot from
which it is being run, and manages its own copy-on-write.

Because of this, there are two very fundamental regions of the guest
physical address space, which are always populated: one, at the very
bottom of memory, is a (hypervisor-enforced) readonly mapping of the
base snapshot from which this guest is being evolved. Another, at the
top of memory, is simply a large bag of blank pages: scratch memory
into which this VM can write.

See more in the proposal at https://github.com/hyperlight-dev/hyperlight/commit/3d5940080a75e3e7924957431db79e688afb4a66

TODO:

POC on older branch: https://github.com/hyperlight-dev/hyperlight/compare/main...lm/guest-cow
old PR that tracked things: https://github.com/hyperlight-dev/hyperlight/pull/1091

- [x] #1182
- [x] #1178
- [x] #1093 
- [x] #1095
- [x] #1096
- [x] #1097
- [x] #1098
- [x] #1099
- [x] #1100
- [x] #1127
- [x] #1198 
- [x] #1205
- [ ] Remove old host managed state
- [ ] Enable Snap shots and turn it on
- [ ] Docs

This is particularly improtant for Hyperlight-wasm.  There are steps to leverage this there:

- [ ] update to take these changes
- [ ] more to tbd

danbugs:
Feature implemented by #1229 . Waiting on #1260 (#1242 ) to close. Will open a follow-up issue for future related work.

PR #1264, opened 2026-02-26
Add crashdump example and include snapshot/scratch in core dumps

jsturtevant:
Core dumps generated by Hyperlight were missing the snapshot and scratch memory regions, making post-mortem debugging with GDB incomplete — register state was present but the guest's code, stack, heap, and page tables were absent. This adds the snapshot and scratch regions to the ELF core dump alongside any dynamically mapped regions so that GDB can show full backtraces, disassemble at the crash site, and inspect guest memory. A new runnable crashdump example demonstrates automatic dumps (VM-level faults), on-demand dumps (guest-caught exceptions), and per-sandbox opt-out, with GDB-based integration tests that validate register and memory content in the generated ELF files. The debugging docs are also updated with practical GDB commands for inspecting crash dumps.

### DIFF
--- a/src/hyperlight_host/src/mem/mgr.rs
+++ b/src/hyperlight_host/src/mem/mgr.rs
@@ -512,14 +512,7 @@ impl SandboxMemoryManager<HostSharedMemory> {
     /// Build the list of guest memory regions for a crash dump.
     ///
     /// With `init-paging` enabled, walks the guest page tables to discover
-    /// GVA→GPA mappings and translates them to host-backed regions. Also
-    /// includes any dynamic mmap regions that may not be covered by the
-    /// page table walk, and a sentinel page at the stack top for clean
-    /// GDB backtraces.
-    ///
-    /// Without `init-paging`, GVA == GPA (identity mapped), so the snapshot
-    /// and scratch regions are returned directly at their known addresses
-    /// alongside any dynamic mmap regions.
+    /// GVA→GPA mappings and translates them to host-backed regions.
     #[cfg(all(feature = "crashdump", feature = "init-paging"))]
     pub(crate) fn get_guest_memory_regions(
         &mut self,


src/hyperlight_host/examples/crashdump/main.rs
```diff
@@ -0,0 +1,599 @@
+/*
+Copyright 2025  The Hyperlight Authors.
```
dblnz, line 2: Unrelated question, I remember from some trainings I've taken some very long time ago, that if we add a new file we should have the year reflect that. Also, if any modifications to the file are made in a different year, then a new header with the new year should be applied.

Is that true here also? I am curious how we should approach these scenarios.

src/hyperlight_host/examples/crashdump/main.rs
```diff
@@ -0,0 +1,599 @@
...
+fn main() -> hyperlight_host::Result<()> {
+    env_logger::init();
+
+    let guest_path =
+        hyperlight_testing::simple_guest_as_string().expect("Cannot find simpleguest binary");
+
+    println!("=== Hyperlight Crash Dump Example ===\n");
+
+    // -----------------------------------------------------------------------
+    // Part 1: Automatic crash dump (VM-level fault bypasses guest handler)
```
dblnz: nit: Maybe `guest caused crash dump` or something like that is more suggestive than `automatic`?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1167,9 +1167,24 @@ impl HyperlightVm {
...
             };
 
-            // Include dynamically mapped regions
-            // TODO: include the snapshot and scratch regions
-            let regions: Vec<MemoryRegion> = self.get_mapped_regions().cloned().collect();
+            // Include the snapshot, scratch, and dynamically mapped regions
+            let mut regions: Vec<MemoryRegion> = Vec::new();
+
+            // Snapshot region: contains guest code, read-only data, page tables, etc.
+            if let Some(snapshot) = &self.snapshot_memory {
```
dblnz: I don't know for sure about this, I'll try and run it locally, see how it looks

src/hyperlight_host/examples/crashdump/main.rs
devigned: The way I've done this in the past is to have a linter in the CI that would verify the header on files in the PR. [Example](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/8867b3b8b9a31c0d2937a99c0b887ad0f8bb8a29/hack/verify-boilerplate.sh) inherited from long ago.

The date doesn't need to be updated. It can stay the creation date for the file. Some folks will opt for a date range $creation_year - $most_recent_update_year. For example, check out https://github.com/kubernetes/kubernetes/blob/07a1af766fd54f1f495a854ddf3e5227241fb961/pkg/api/node/util.go in the K/K repo.

syntactically:
@jsturtevant I believe you were working on another version of this that explicitly walks through the whole guest virtual address space?

jsturtevant:
> @jsturtevant I believe you were working on another version of this that explicitly walks through the whole guest virtual address space?

Yes, been working with @dblnz on getting it working. Should have something today

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
jsturtevant: udpated

src/hyperlight_host/examples/crashdump/main.rs
jsturtevant: updated to the latest date

src/hyperlight_host/src/sandbox/snapshot.rs
```diff
@@ -181,9 +181,11 @@ pub(crate) fn access_gpa<'a>(
     gpa: u64,
 ) -> Option<(&'a ExclusiveSharedMemory, usize)> {
     let scratch_base = scratch_base_gpa(scratch_size);
-    if gpa >= scratch_base {
+    if gpa >= scratch_base && gpa < scratch_base + scratch_size as u64 {
         Some((scratch, (gpa - scratch_base) as usize))
-    } else if gpa >= SandboxMemoryLayout::BASE_ADDRESS as u64 {
+    } else if gpa >= SandboxMemoryLayout::BASE_ADDRESS as u64
+        && gpa < SandboxMemoryLayout::BASE_ADDRESS as u64 + snap.mem_size() as u64
+    {
```
dblnz, line 188: I remember this was a bug fix for some regions that are not in the snapshot and not in the scratch region either, but in between, but the previous logic returned the snapshot memory

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -767,8 +767,24 @@ impl MultiUseSandbox {
     ///
     #[cfg(crashdump)]
     #[instrument(err(Debug), skip_all, parent = Span::current())]
-    pub fn generate_crashdump(&self) -> Result<()> {
-        crate::hypervisor::crashdump::generate_crashdump(&self.vm)
+    pub fn generate_crashdump(&mut self) -> Result<()> {
```
dblnz, line 770: The thing I don't really like about this fix is that it changes this mutability, but it is necessary for the `with_exclusivity` pattern for the memory accesses, so I don't see any other way

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -412,6 +504,173 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
+
+                        regions.push(CrashDumpRegion {
+                            guest_region: virt_base..virt_end,
+                            host_region: host_base..host_base + host_len,
+                            flags,
+                            region_type,
+                        });
+                    } else {
+                        // GPA not in snapshot/scratch — check dynamic mmap regions
+                        resolve_from_mmap_regions(
```
dblnz, line 577: I am not 100%, but from what I see, this function creates a region that exactly matches the mapping in size.
I am wondering, wouldn't the inclusion of dynamic mmap regions below take care of this by including the whole mmap region?

Note: I remember that when I was working on this, the dumps wouldn't correctly load without this function call, but I'm just saying, maybe I'm missing something.

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -412,6 +504,173 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
+                    region_type: rgn.region_type,
+                });
+            }
+        }
+
+        // Add a zero-filled sentinel page at the stack top so GDB can
+        // read a null return address and terminate the backtrace cleanly
+        // Not strictly necessary but avoids "Cannot access memory at address \u2026"
+        // in the backtrace in gdb
+        static SENTINEL_PAGE: [u8; 4096] = [0u8; 4096];
```
dblnz: Not sure about this, maybe @syntactically has a better idea how to avoid having this static.

syntactically: I don't know this off the top of my head, but looking around quickly, it seems that there's a general consensus that it ought to be enough to
- (a) use `.cfi_undefined rip` at the beginning of the entry function, so that a debugger using dwarf unwind tables knows there's no way to find a previous rip; and
- (b) set the frame pointer to 0 so that a debugger walking via frame pointers knows that there's no base frame
 
Does adding those to the assembly stubs for `entrypoint` and `dispatch_function` make any difference?

jsturtevant: I can try this. I did a bit of research and found that it was related to the way we set the stack up generally setup since we were in a different env and we might not be doing this step properly. 

As this isn't critical, I can also drop this commit and follow up with more research.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: I think it might make sense? In the future, if there are multiple threads/users of the sandbox, it makes sense to want them all to have given up their &self access to the sandbox in order to be able to take a crash dump?

I guess maybe the one worry here is about the automatic generation of crashdumps when a call goes wrong?

src/hyperlight_host/examples/crashdump/main.rs
```diff
@@ -0,0 +1,598 @@
...
+
+/// Demonstrates an **automatic** crash dump triggered by a VM-level fault
+/// that bypasses the guest exception handler entirely.
+///
+/// The guest has an IDT (Interrupt Descriptor Table) that catches most CPU
+/// exceptions (page faults, undefined instructions, etc.) and reports them
+/// back to the host as `GuestAborted` errors.
+///
+/// However, some faults are intercepted by the **hypervisor** before the
+/// CPU delivers them to the guest. For example, when the guest writes to
```
syntactically: Minor nit, but this example is not the hypervisor "intercepting a fault that was originally destined for the guest": the fault was always targeted at the hypervisor, because it was the hypervisor's own page tables that caused it.  There are intercepts that allow us to redirect faults to the hv (on amd64), but we don't use a tonne of them.

src/hyperlight_host/examples/crashdump/main.rs
```diff
@@ -0,0 +1,598 @@
...
+
+    let mut sandbox: MultiUseSandbox = uninitialized_sandbox.evolve()?;
+
+    // A normal call succeeds.
+    let message = "Hello from crashdump example!\n".to_string();
+    sandbox.call::<i32>("PrintOutput", message)?;
+
+    // This call triggers a ud2 instruction in the guest. The guest's IDT
+    // catches the #UD exception and reports it back to the host as a
+    // GuestAborted error via I/O. This does NOT trigger an automatic crash
```
syntactically, line 219: Why do we have this distinction? We should be clear about exactly which exceptions result in dumps and which do not, and whether there is any relationship between that and which faults poison the sandbox.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
dblnz: If I understand correctly to what you're referring, for automatic generation, I think the `call` API already takes `mut self`, so it'll be a similar pattern

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: Thanks for the pointer, this does work and found docs in https://gitlab.com/x86-psABIs/x86-64-ABI update incoming

jsturtevant: addressed in a84321274717f53da8676746c4c75c6f9a425560, when I auto squash I can change that commit message

src/hyperlight_host/src/sandbox/snapshot.rs
jsturtevant: this makes it explicit that it will look for a GPA in both were as previously it was also looking at values that where past the end of either. Here `snap` is https://github.com/hyperlight-dev/hyperlight/blob/7741d51b7aeadcca193864ab168ac46f727d98b8/src/hyperlight_host/src/mem/layout.rs#L20-L35 and the `scratch` is https://github.com/hyperlight-dev/hyperlight/blob/7741d51b7aeadcca193864ab168ac46f727d98b8/src/hyperlight_host/src/mem/layout.rs#L48-L61

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: this makes sure the dump preserves the GVA (virt_base, virt_end) instead of the physical address which is what the other loop does. The below mmap regions includes the GVA locations just for full dump but likely won't be used in gdb

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -248,6 +248,50 @@ pub(crate) struct MemoryRegion_<K: MemoryRegionKind> {
 
 pub(crate) type MemoryRegion = MemoryRegion_<HostGuestMemoryRegion>;
 
+/// A [`MemoryRegionKind`] for crash dump regions that always uses raw
+/// `usize` host addresses.  The crash dump path only reads host memory
+/// through raw pointers, so it never needs the file-mapping metadata
+/// stored in [`HostRegionBase`] on Windows.
+#[cfg(feature = "crashdump")]
```
ludfjig: The new types in memory_region.rs and helpers in mgr.rs use #[cfg(feature = "crashdump")] while most of the rest of the crate the #[cfg(crashdump)] alias. Maybe we can remove the alias since it doens't serve a purpose (I seem to remember it was only usable in debug profile before, but that no longer seems to be the case), OR switch to using the alias everywhere

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -412,6 +504,157 @@ impl SandboxMemoryManager<HostSharedMemory> {
         Ok(())
     }
 
+    /// Build the list of guest memory regions for a crash dump.
+    ///
+    /// With `init-paging` enabled, walks the guest page tables to discover
+    /// GVA→GPA mappings and translates them to host-backed regions. Also
+    /// includes any dynamic mmap regions that may not be covered by the
+    /// page table walk, and a sentinel page at the stack top for clean
```
ludfjig: outdated doc since the sentinel page was removed i believe

src/hyperlight_host/examples/crashdump/main.rs
```diff
@@ -0,0 +1,589 @@
...
+
+    /// Verify that GDB can read the mapped memory region from the core dump
+    /// and that it contains the sentinel string we wrote before the crash.
+    #[test]
+    #[serial]
+    fn test_crashdump_gdb_memory() {
+        if !gdb_is_available() {
+            eprintln!("Skipping test: {GDB_COMMAND} not found on PATH");
+            return;
+        }
```
ludfjig: Was this needed in CI? Do we not want the test to fail here?

ludfjig (dismissed):
this is great, i like the added tests!

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -44,6 +44,11 @@ pub(crate) struct SandboxRuntimeConfig {
     pub(crate) debug_info: Option<super::config::DebugInfo>,
     #[cfg(crashdump)]
     pub(crate) guest_core_dump: bool,
+    /// The original entry point address of the loaded guest binary
+    /// (load_addr + ELF entry offset). Used for AT_ENTRY in core dumps
+    /// so GDB can compute the correct load offset for PIE binaries.
+    #[cfg(crashdump)]
+    pub(crate) entry_point: u64,
 }
```
syntactically, line 55: This is not new to this PR, so feel free to file an issue and defer, but I wanted to see if you had thoughts in this context first. This addition brings to my attention the question: what is the expected lifecycle of SandboxRuntimeConfig? Some parts, like `debug_info` and `guest_core_dump`, seem like they are config of a specific sandbox instance (regardless of what guest/snapshot is being run inside of that instance), while things like `entry_point` and `guest_core_dump` seem like they are related fundamentally to the snapshot that's underlying the sandbox, and would need to be reset on restore to have sensible behaviour for a single sandbox that urns multiple different guests at different times.

src/hyperlight_host/src/sandbox/snapshot.rs
```diff
@@ -181,9 +181,11 @@ pub(crate) fn access_gpa<'a>(
     gpa: u64,
 ) -> Option<(&'a ExclusiveSharedMemory, usize)> {
     let scratch_base = scratch_base_gpa(scratch_size);
-    if gpa >= scratch_base {
+    if gpa >= scratch_base && gpa < scratch_base + scratch_size as u64 {
```
syntactically, line 184: I think this change makes sense, since it's good for the contract here to be that the returned offset exists in the memory (and an error would happen fast if not).

However, I'm curious if this change was motivated by actually finding a case where it would make a difference or not---it feels like a bug if there is a pointer to above the scratch region (only 1 unmapped page exists there), so if that was the cause of the change I'd be curious where the access was coming from.

src/hyperlight_host/src/sandbox/snapshot.rs
syntactically: Yes, in the case of the snapshot code, which was formerly the only user of this code, this was being handled in the function `guest_page`, and the entire call to `access_gpa` was avoided.  Now that there are more users of this VA enumeration, and crashdump/gdb are likely to continue to need it, we should see if the interface can be factored out a bit better (made to not rely on exclusive memory access, but to take advantage of it where available; and made to unify scratch/snapshot handling and mapped region handling).

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -20,15 +20,107 @@ use hyperlight_common::flatbuffer_wrappers::function_call::{
...
+            if *readable {
+                flags |= MemoryRegionFlags::READ;
+            }
+            if *writable {
+                flags |= MemoryRegionFlags::WRITE;
+            }
+            if *executable {
+                flags |= MemoryRegionFlags::EXECUTE;
+            }
+            (flags, MemoryRegionType::Code)
```
syntactically: This `MemoryRegionType` doesn't seem right, but then again I'm not sure if we're actually using the `MemoryRegionTypes` for anything, and they were already a bit inconsistent/nonsensical...

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -412,6 +504,152 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
+                            &mut regions, mapping, virt_base, virt_end, mmap_regions,
+                        );
+                    }
+                }
+
+                Ok(regions)
+            })
+        })???;
+
+        // Include dynamic mmap regions at their GVA addresses as a fallback
```
syntactically: This doesn't seem quite right. The mmap region `guest_region`s are GPAs not GVAs, and the rest of the crashdump is (much like a snapshot) handled entirely in GVA space. Code could well be doing something important at some GVA that is equal to a GPA where something entirely different is present, and this would overwrite the former with the contents of the latter as far as I understand it.  Regions that are mapped into some guest physical address space, but not mapped from there to anywhere in VA space, are basically inaccessible to any code (unless it bothers to map them in first), since we don't have a complete physmap anyway...

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -20,15 +20,107 @@ use hyperlight_common::flatbuffer_wrappers::function_call::{
...
+    let phys_start = mapping.phys_base as usize;
+    let phys_end = (mapping.phys_base + mapping.len) as usize;
+
+    for rgn in mmap_regions {
+        if phys_start >= rgn.guest_region.start && phys_end <= rgn.guest_region.end {
+            let offset = phys_start - rgn.guest_region.start;
+            let host_base = HostGuestMemoryRegion::to_addr(rgn.host_region.start) + offset;
+            let host_end = host_base + mapping.len as usize;
+
+            regions.push(CrashDumpRegion {
```
syntactically, line 119: Should the coalesce logic above run here as well?

src/hyperlight_host/src/mem/memory_region.rs
jsturtevant: I've used #[cfg(crashdump)] everywhere for now

src/hyperlight_host/examples/crashdump/main.rs
jsturtevant: we should have gdb in CI, removed this.

src/hyperlight_host/examples/crashdump/main.rs
jsturtevant: I am not 100% sure here, @dblnz or @ludfjig do you have any ideas or history on why some cause an automatic crashdump and other do not?  I will take a look at where this distinction is happening too.

src/hyperlight_host/src/sandbox/snapshot.rs
jsturtevant: I ran a Copilot code review and this was flag and it seemed like an appropriate fix.

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: is snapshot type more appropriate?

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: We've discussed this offline and I had a miss understanding here. I've removed this because for now we are only collecting the GVA space for the dump otherwise we will have potential overlaps. I.E the GPA 0x1000000 would overwrite the GVA 0x1000000 if I did it this way.

src/hyperlight_host/examples/crashdump/main.rs
ludfjig: I'm not sure

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: i've added this

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: i've changed this to snapshot.  I don't think it necessarily being used in this case

src/hyperlight_host/src/sandbox/uninitialized.rs
jsturtevant: https://github.com/hyperlight-dev/hyperlight/issues/1279

src/hyperlight_host/examples/crashdump/main.rs
jsturtevant: I've updated teh comment and removed some of this

src/hyperlight_host/examples/crashdump/main.rs
jsturtevant: It is coming from here: https://github.com/hyperlight-dev/hyperlight/blob/64aa448dd5b2a80eb4c2f07d1c9274350033c2b6/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L950-L961

and 

https://github.com/hyperlight-dev/hyperlight/blob/64aa448dd5b2a80eb4c2f07d1c9274350033c2b6/src/hyperlight_host/src/hypervisor/hyperlight_vm.rs#L870-L893

I can create an issue for this if we believe this is wrong

jsturtevant: #1280

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -1234,14 +1255,24 @@ impl HyperlightVm {
...
+                    "entry_point was never set in SandboxRuntimeConfig; AT_ENTRY will be 0"
+                );
+                0
+            });
+            let mmap_regions: Vec<MemoryRegion> = self.get_mapped_regions().cloned().collect();
+            let root_pt = self.get_root_pt()?;
+
+            let regions = mem_mgr
+                .get_guest_memory_regions(root_pt, &mmap_regions)
+                .map_err(|e| CrashDumpError::AccessPageTable(Box::new(e)))?;
```
dblnz, line 1275: This looks a lot cleaner now

dblnz:
This PR is part of #1200 work

PR #1270, opened 2026-03-03
feat: add GuestCounter API behind nanvix-unstable feature flag

danbugs:
## Summary

- Add `GuestCounter`, a host-authoritative shared `u64` counter at a fixed offset in scratch memory, gated behind `nanvix-unstable`
- Add `guest_counter_gva()` helper in hyperlight-guest so the guest can locate the counter without hardcoding addresses
- Add `nanvix-unstable` feature to hyperlight-guest (forwards to hyperlight-common)

## Motivation

Nanvix uses a credits-based flow-control mechanism for guest-host synchronization. The host increments/decrements a counter; the guest reads it with a volatile load. `GuestCounter` provides a type-safe, thread-safe way to do this via a known scratch-memory offset that both sides can compute independently.

## Design

**Host-authoritative**: the host owns the counter value in a `Mutex<GuestCounterInner>` and is the only writer. `increment()` and `decrement()` update the cached value and issue a single `write_volatile` to shared memory. `value()` returns the cached value — the host never reads back from guest memory, preventing a confused-deputy attack where a malicious guest could manipulate the counter.

**Fixed offset**: the counter lives at `SCRATCH_TOP_GUEST_COUNTER_OFFSET = 0x1008` bytes from the top of scratch memory. This places it in page `0xffffe000`.

## Changes

- `hyperlight_common/src/layout.rs` — add `SCRATCH_TOP_GUEST_COUNTER_OFFSET` constant (0x1008) and `MAX_GVA`
- `hyperlight_guest/Cargo.toml` — add `nanvix-unstable` feature
- `hyperlight_guest/src/layout.rs` — add `guest_counter_gva()` helper
- `hyperlight_host/src/sandbox/uninitialized.rs` — add `GuestCounter` struct with `increment()`, `decrement()`, `value()` methods; add `UninitializedSandbox::guest_counter()` constructor
- `hyperlight_host/src/lib.rs` — re-export `GuestCounter`

### DIFF
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -20,9 +20,7 @@ use std::io::Error;
 use std::mem::{align_of, size_of};
 #[cfg(target_os = "linux")]
 use std::ptr::null_mut;
-#[cfg(feature = "nanvix-unstable")]
-use std::sync::atomic::AtomicBool;
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{Arc, RwLock};
 
 use hyperlight_common::mem::PAGE_SIZE_USIZE;
 use tracing::{Span, instrument};
@@ -138,17 +136,7 @@ impl Drop for HostMapping {
 /// and taking snapshots.
 #[derive(Debug)]
 pub struct ExclusiveSharedMemory {
-    pub(crate) region: Arc<HostMapping>,
-    /// Populated by [`build()`](Self::build) with a [`HostSharedMemory`]
-    /// view of this region. Code that needs host-style volatile access
-    /// before `build()` (e.g. `GuestCounter`) can clone this `Arc` and
-    /// will see `Some` once `build()` completes.
-    pub(crate) deferred_hshm: Arc<Mutex<Option<HostSharedMemory>>>,
-    /// Set to `true` once a [`GuestCounter`] has been handed out via
-    /// [`UninitializedSandbox::guest_counter()`]. Prevents creating
-    /// multiple counters that would have divergent cached values.
-    #[cfg(feature = "nanvix-unstable")]
-    pub(crate) counter_taken: AtomicBool,
+    region: Arc<HostMapping>,
 }
 unsafe impl Send for ExclusiveSharedMemory {}
 
@@ -436,9 +424,6 @@ impl ExclusiveSharedMemory {
                 ptr: addr as *mut u8,
                 size: total_size,
             }),
-            deferred_hshm: Arc::new(Mutex::new(None)),
-            #[cfg(feature = "nanvix-unstable")]
-            counter_taken: AtomicBool::new(false),
         })
     }
 
@@ -557,9 +542,6 @@ impl ExclusiveSharedMemory {
                 size: total_size,
                 handle,
             }),
-            deferred_hshm: Arc::new(Mutex::new(None)),
-            #[cfg(feature = "nanvix-unstable")]
-            counter_taken: AtomicBool::new(false),
         })
     }
 
@@ -670,15 +652,6 @@ impl ExclusiveSharedMemory {
             region: self.region.clone(),
             lock: lock.clone(),
         };
-        // Publish the HostSharedMemory so any pre-existing GuestCounter
-        // can begin issuing volatile writes via the proper protocol.
-        // The mutex is only locked here and during GuestCounter
-        // operations; since build() consumes self, there is no
-        // concurrent access that could poison it.
-        #[allow(clippy::unwrap_used)]
-        {
-            *self.deferred_hshm.lock().unwrap() = Some(hshm.clone());
-        }
         (
             hshm,
             GuestSharedMemory {
@@ -688,23 +661,22 @@ impl ExclusiveSharedMemory {
         )
     }
 
-    /// Populate the deferred `HostSharedMemory` slot without consuming
-    /// `self`. Used in tests where `evolve()` / full `build()` is not
-    /// available.
-    #[cfg(all(test, feature = "nanvix-unstable"))]
-    pub(crate) fn simulate_build(&self) {
-        let lock = Arc::new(RwLock::new(()));
-        let hshm = HostSharedMemory {
-            region: self.region.clone(),
-            lock,
-        };
-        *self.deferred_hshm.lock().unwrap() = Some(hshm);
-    }
-
     /// Gets the file handle of the shared memory region for this Sandbox
     #[cfg(target_os = "windows")]
     pub fn get_mmap_file_handle(&self) -> HANDLE {
         self.region.handle
+    }
+
+    /// Create a [`HostSharedMemory`] view of this region without
+    /// consuming `self`. Used in tests where the full `build()` /
+    /// `evolve()` pipeline is not available.
+    #[cfg(all(test, feature = "nanvix-unstable"))]
+    pub(crate) fn as_host_shared_memory(&self) -> HostSharedMemory {
+        let lock = Arc::new(RwLock::new(()));
+        HostSharedMemory {
+            region: self.region.clone(),
+            lock,
+        }
     }
 }
 
@@ -888,9 +860,6 @@ impl SharedMemory for GuestSharedMemory {
             .map_err(|e| new_error!("Error locking at {}:{}: {}", file!(), line!(), e))?;
         let mut excl = ExclusiveSharedMemory {
             region: self.region.clone(),
-            deferred_hshm: Arc::new(Mutex::new(None)),
-            #[cfg(feature = "nanvix-unstable")]
-            counter_taken: AtomicBool::new(false),
         };
         let ret = f(&mut excl);
         drop(excl);
@@ -1256,9 +1225,6 @@ impl SharedMemory for HostSharedMemory {
             .map_err(|e| new_error!("Error locking at {}:{}: {}", file!(), line!(), e))?;
         let mut excl = ExclusiveSharedMemory {
             region: self.region.clone(),
-            deferred_hshm: Arc::new(Mutex::new(None)),
-            #[cfg(feature = "nanvix-unstable")]
-            counter_taken: AtomicBool::new(false),
         };
         let ret = f(&mut excl);
         drop(excl);
--- a/src/hyperlight_host/src/sandbox/uninitialized.rs
+++ b/src/hyperlight_host/src/sandbox/uninitialized.rs
@@ -33,6 +33,8 @@ use crate::mem::memory_region::{DEFAULT_GUEST_BLOB_MEM_FLAGS, MemoryRegionFlags}
 use crate::mem::mgr::SandboxMemoryManager;
 use crate::mem::shared_mem::ExclusiveSharedMemory;
 #[cfg(feature = "nanvix-unstable")]
+use crate::mem::shared_mem::HostSharedMemory;
+#[cfg(feature = "nanvix-unstable")]
 use crate::mem::shared_mem::SharedMemory;
 use crate::sandbox::SandboxConfiguration;
 use crate::{MultiUseSandbox, Result, new_error};
@@ -70,11 +72,9 @@ pub(crate) struct SandboxRuntimeConfig {
 /// `decrement()` take `&self` rather than `&mut self`.
 ///
 /// The counter holds an `Arc<Mutex<Option<HostSharedMemory>>>` that is
-/// shared with [`ExclusiveSharedMemory`]. The `Option` is `None` until
-/// [`build()`](ExclusiveSharedMemory::build) populates it, at which point
-/// the counter can issue volatile writes via the proper protocol. This
-/// avoids adding lock overhead to `ExclusiveSharedMemory` during the
-/// exclusive setup phase.
+/// shared with [`UninitializedSandbox`]. The `Option` is `None` until
+/// [`evolve()`](UninitializedSandbox::evolve) populates it, at which point
+/// the counter can issue volatile writes via the proper protocol.
 ///
 /// Only one `GuestCounter` may be created per sandbox; a second call to
 /// [`UninitializedSandbox::guest_counter()`] returns an error.
@@ -85,7 +85,7 @@ pub struct GuestCounter {
 
 #[cfg(feature = "nanvix-unstable")]
 struct GuestCounterInner {
-    deferred_hshm: Arc<Mutex<Option<crate::mem::shared_mem::HostSharedMemory>>>,
+    deferred_hshm: Arc<Mutex<Option<HostSharedMemory>>>,
     offset: usize,
     value: u64,
 }
@@ -111,11 +111,12 @@ impl GuestCounter {
                 })?
                 .clone()
         };
-        inner.value = inner
+        let new_value = inner
             .value
             .checked_add(1)
             .ok_or_else(|| new_error!("GuestCounter overflow"))?;
-        shm.write::<u64>(inner.offset, inner.value)?;
+        shm.write::<u64>(inner.offset, new_value)?;
+        inner.value = new_value;
         Ok(())
     }
 
@@ -131,11 +132,12 @@ impl GuestCounter {
                 })?
                 .clone()
         };
-        inner.value = inner
+        let new_value = inner
             .value
             .checked_sub(1)
             .ok_or_else(|| new_error!("GuestCounter underflow"))?;
-        shm.write::<u64>(inner.offset, inner.value)?;
+        shm.write::<u64>(inner.offset, new_value)?;
+        inner.value = new_value;
         Ok(())
     }
 
@@ -170,6 +172,17 @@ pub struct UninitializedSandbox {
     // This is needed to convey the stack pointer between the snapshot
     // and the HyperlightVm creation
     pub(crate) stack_top_gva: u64,
+    /// Populated by [`evolve()`](Self::evolve) with a [`HostSharedMemory`]
+    /// view of scratch memory. Code that needs host-style volatile access
+    /// before `evolve()` (e.g. `GuestCounter`) can clone this `Arc` and
+    /// will see `Some` once `evolve()` completes.
+    #[cfg(feature = "nanvix-unstable")]
+    pub(crate) deferred_hshm: Arc<Mutex<Option<HostSharedMemory>>>,
+    /// Set to `true` once a [`GuestCounter`] has been handed out via
+    /// [`guest_counter()`](Self::guest_counter). Prevents creating
+    /// multiple counters that would have divergent cached values.
+    #[cfg(feature = "nanvix-unstable")]
+    counter_taken: std::sync::atomic::AtomicBool,
 }
 
 impl Debug for UninitializedSandbox {
@@ -267,10 +280,9 @@ impl UninitializedSandbox {
     /// the top of scratch memory, so both host and guest can locate it
     /// without an explicit GPA parameter.
     ///
-    /// The returned counter holds an `Arc` clone of the scratch memory's
+    /// The returned counter holds an `Arc` clone of the sandbox's
     /// `deferred_hshm`, so it will automatically gain access to the
-    /// [`HostSharedMemory`] once [`build()`](ExclusiveSharedMemory::build)
-    /// is called during [`evolve()`](Self::evolve).
+    /// [`HostSharedMemory`] once [`evolve()`](Self::evolve) completes.
     ///
     /// This method can only be called once; a second call returns an error
     /// because multiple counters would have divergent cached values.
@@ -280,12 +292,7 @@ impl UninitializedSandbox {
 
         use hyperlight_common::layout::SCRATCH_TOP_GUEST_COUNTER_OFFSET;
 
-        if self
-            .mgr
-            .scratch_mem
-            .counter_taken
-            .swap(true, Ordering::Relaxed)
-        {
+        if self.counter_taken.swap(true, Ordering::Relaxed) {
             return Err(new_error!(
                 "GuestCounter has already been created for this sandbox"
             ));
@@ -301,7 +308,7 @@ impl UninitializedSandbox {
         }
 
         let offset = scratch_size - SCRATCH_TOP_GUEST_COUNTER_OFFSET as usize;
-        let deferred_hshm = self.mgr.scratch_mem.deferred_hshm.clone();
+        let deferred_hshm = self.deferred_hshm.clone();
 
         Ok(GuestCounter {
             inner: Mutex::new(GuestCounterInner {
@@ -370,6 +377,10 @@ impl UninitializedSandbox {
             rt_cfg,
             load_info: snapshot.load_info(),
             stack_top_gva: snapshot.stack_top_gva(),
+            #[cfg(feature = "nanvix-unstable")]
+            deferred_hshm: Arc::new(Mutex::new(None)),
+            #[cfg(feature = "nanvix-unstable")]
+            counter_taken: std::sync::atomic::AtomicBool::new(false),
         };
 
         // If we were passed a writer for host print register it otherwise use the default.
@@ -448,6 +459,18 @@ impl UninitializedSandbox {
         print_func: impl Into<HostFunction<i32, (String,)>>,
     ) -> Result<()> {
         self.register("HostPrint", print_func)
+    }
+
+    /// Populate the deferred `HostSharedMemory` slot without running
+    /// the full `evolve()` pipeline. Used in tests where guest boot
+    /// is not available.
+    #[cfg(all(test, feature = "nanvix-unstable"))]
+    fn simulate_build(&self) {
+        let hshm = self.mgr.scratch_mem.as_host_shared_memory();
+        #[allow(clippy::unwrap_used)]
+        {
+            *self.deferred_hshm.lock().unwrap() = Some(hshm);
+        }
     }
 }
 // Check to see if the current version of Windows is supported
@@ -1500,7 +1523,7 @@ mod tests {
         fn increment_decrement() {
             let mut sandbox = make_sandbox();
             let counter = sandbox.guest_counter().unwrap();
-            sandbox.mgr.scratch_mem.simulate_build();
+            sandbox.simulate_build();
 
             counter.increment().unwrap();
             assert_eq!(counter.value().unwrap(), 1);
@@ -1514,7 +1537,7 @@ mod tests {
         fn underflow_returns_error() {
             let mut sandbox = make_sandbox();
             let counter = sandbox.guest_counter().unwrap();
-            sandbox.mgr.scratch_mem.simulate_build();
+            sandbox.simulate_build();
 
             assert_eq!(counter.value().unwrap(), 0);
             let result = counter.decrement();
--- a/src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
+++ b/src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
@@ -38,6 +38,21 @@ use crate::{MultiUseSandbox, Result, UninitializedSandbox};
 #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
 pub(super) fn evolve_impl_multi_use(u_sbox: UninitializedSandbox) -> Result<MultiUseSandbox> {
     let (mut hshm, gshm) = u_sbox.mgr.build()?;
+
+    // Publish the HostSharedMemory for scratch so any pre-existing
+    // GuestCounter can begin issuing volatile writes.
+    #[cfg(feature = "nanvix-unstable")]
+    {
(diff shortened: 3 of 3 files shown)


syntactically:
I wouldn't like this to be part of the public API. Can we go into a little more detail about how this is being used right now, and whether there are alternative ways to get the same result?

jsturtevant:
>This provides a narrow, type-safe accessor for downstream consumers (e.g., [Nanvix](https://github.com/nanvix/nanvix)) that need to write data into the sandbox's shared memory region before initialization 

What region is it writing too?

>I wouldn't like this to be part of the public API. Can we go into a little more detail about how this is being used right now, and whether there are alternative ways to get the same result?

If we constrained the usage to something like `write_shared_mem(&mut self, src: &[u8], offset: usize)` where we could do checks and things would the be helpful?

ludfjig:
The main worry I have is public usage of this API which is probably somethign we don't want. I'd also want to know what it's useful more specifically, but I'd be open to add this under a new feature flag with maybe a scary name since that would discourage general use

jsturtevant:
> The main worry I have is public usage of this API which is probably somethign we don't want. I'd also want to know what it's useful more specifically, but I'd be open to add this under a new feature flag with maybe a scary name since that would discourage general use

hmm, it seems the requirement is some system need some extra metadata during boot sequence, I wonder for those systems not using paging (or maybe even with paging), if they could use the sratch region or a subset of it?

dblnz:
I have the same concerns other people mentioned around here. Exposing an implementation detail binds us to keeping it that way.
Maybe we can find another way to do this that doesn't imply knowing the memory layout to be able to interact with the API.

danbugs:
Addressed the concerns about exposing `shared_mem_mut()` as a public API. The latest push replaces it with:

```rust
pub fn guest_memory_ptr(&mut self, gpa: u64) -> Result<*mut u8>
```

This does bounds-checked GPA-to-host-pointer translation — the caller provides a guest physical address and gets back a host pointer into the mapped sandbox memory. The API no longer exposes `SharedMemory` internals or the memory manager.

Nanvix uses this for its credits-based flow control: the host and guest agree on a fixed GPA (`CREDITS_GPA`, last 8 bytes of the kernel pool) where a shared credit counter lives. The host increments/decrements the counter via volatile writes through this pointer, and the guest reads it directly from memory.

This is narrower than the previous `shared_mem_mut()` approach: it validates that the GPA falls within the sandbox's mapped region and returns an error otherwise. It doesn't expose the base pointer or total size, so callers can't accidentally access memory outside their intended range.

andreiltd (dismissed):

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +220,48 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+    /// operations that use volatile writes so the guest can observe
+    /// changes.
+    ///
+    /// # Safety contract
+    ///
+    /// The caller must ensure the `GuestSemaphore` does not outlive the
+    /// sandbox's underlying shared memory mapping (which stays alive
+    /// through `evolve()` into `MultiUseSandbox`).
+    #[cfg(feature = "nanvix-unstable")]
+    pub fn guest_semaphore(&mut self, gpa: usize) -> Result<GuestSemaphore> {
```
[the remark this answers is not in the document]
danbugs: Fixed. `guest_semaphore()` is now `pub unsafe fn`. The `# Safety` doc section documents that the caller must ensure the returned `GuestSemaphore` does not outlive the sandbox's shared memory mapping.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +220,48 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+            return Err(new_error!(
+                "GPA {:#x} is below the sandbox base address ({:#x})",
+                gpa,
+                base
+            ));
+        }
+
+        let offset = gpa - base;
+
+        if offset >= mem_size {
```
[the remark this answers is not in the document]
danbugs: Fixed. The function now uses `offset.checked_add(core::mem::size_of::<u64>())` to compute `end`, then checks `end > mem_size`. The `checked_add` also guards against overflow.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +220,48 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+            return Err(new_error!(
+                "GPA {:#x} (offset {:#x}) is outside the sandbox memory region (size {:#x})",
+                gpa,
+                offset,
+                mem_size
+            ));
+        }
+
+        let ptr = unsafe { self.mgr.shared_mem.base_ptr().add(offset) as *mut u64 };
```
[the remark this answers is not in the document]
danbugs: Fixed. Added `gpa % core::mem::align_of::<u64>() != 0` check that returns an error with a descriptive message. There's also a dedicated `misaligned_gpa` test.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +220,48 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+        Ok(GuestSemaphore { ptr, value: 0 })
```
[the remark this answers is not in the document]
danbugs: Fixed. The constructor now reads the current value with `core::ptr::read_volatile(ptr)` before returning the `GuestSemaphore`.

src/hyperlight_host/src/sandbox/uninitialized.rs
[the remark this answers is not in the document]
danbugs: Fixed. Added a guest_semaphore_tests module with tests for: valid GPA, out-of-range GPA, below-base/misaligned GPA, explicitly misaligned GPA, increment/decrement with value checks, and underflow-from-zero. All gated behind `#[cfg(feature = "nanvix-unstable")]`.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -29,9 +29,13 @@ use crate::func::host_functions::{HostFunction, register_host_function};
 use crate::func::{ParameterTuple, SupportedReturnType};
 #[cfg(feature = "build-metadata")]
 use crate::log_build_details;
+#[cfg(feature = "nanvix-unstable")]
+use crate::mem::layout::SandboxMemoryLayout;
 use crate::mem::memory_region::{DEFAULT_GUEST_BLOB_MEM_FLAGS, MemoryRegionFlags};
 use crate::mem::mgr::SandboxMemoryManager;
 use crate::mem::shared_mem::ExclusiveSharedMemory;
+#[cfg(feature = "nanvix-unstable")]
+use crate::mem::shared_mem::SharedMemory;
```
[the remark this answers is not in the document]
danbugs: Not an issue. The `SharedMemory` import is the trait that provides `base_ptr()` and `mem_size()`—both are called inside `guest_semaphore()`.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -54,6 +58,56 @@ pub(crate) struct SandboxRuntimeConfig {
...
+impl GuestSemaphore {
+    /// Increments the counter by one and writes it to guest memory with
+    /// a volatile store.
+    pub fn increment(&mut self) -> Result<()> {
+        self.value = self
+            .value
+            .checked_add(1)
+            .ok_or_else(|| new_error!("GuestSemaphore overflow"))?;
+        // SAFETY: `ptr` was validated as aligned and in-bounds at construction time.
+        unsafe { core::ptr::write_volatile(self.ptr, self.value) };
```
jsturtevant: Is the `Mutex<GuestSemaphore>` described above the guard against reading this from the `value` then writing directly to memory?

What happens here if `value` in self isn't the value that is actually in the memory slot?  I.e. if there are two uses of this guestsemaphore are incrementing on it at the same time? Maybe nanvix doesn't do this? or Is the mutex the gaurd in nanvix?

syntactically: Yes, it doesn't seem like this actually implements a semaphore, since there is nothing stopping the guest mutating the ptr independently of the host. Maybe I misunderstood and the Nanvix use case does not actually require a semaphore as a synchronisation primitive? If so, perhaps a better name could be found. 

Also, agreed on the general point that if there's things that need extra synchronisation/etc to be safe, I would rather see that in the implementation inside here, rather than commented saying to synchronise elsewhere, if at all possible.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +223,54 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+    /// The GPA must fall within the sandbox's allocated memory region and
+    /// be 8-byte aligned.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the returned `GuestSemaphore` does not
+    /// outlive the sandbox's underlying shared memory mapping (which
+    /// stays alive through `evolve()` into `MultiUseSandbox`).
+    #[cfg(feature = "nanvix-unstable")]
+    pub unsafe fn guest_semaphore(&mut self, gpa: usize) -> Result<GuestSemaphore> {
```
syntactically: Since we only need one of these per sandbox, can we get rid of the configurability here? Maybe in arch/i86/layout.rs in hyperlight-common we could statically determine an address at e.g. the top of scratch memory or the bottom of physical memory or something?

danbugs: Removed the configurability and made it a set offset off the top of scratch.

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Renamed to `GuestCounter` instead of `GuestSemaphore` and internalized the expected `Mutex`.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -54,6 +56,75 @@ pub(crate) struct SandboxRuntimeConfig {
     pub(crate) entry_point: Option<u64>,
 }
 
+/// A host-authoritative shared counter backed by a u64 in guest scratch memory.
```
syntactically: Minor nit: I don't think that "backed by" is an appropriate thing here, because that isn't the actual backing store of the value (the value field of the struct is, as far as I can tell?). Maybe something like "exposed to the guest via a u64 ..."?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -54,6 +56,75 @@ pub(crate) struct SandboxRuntimeConfig {
...
+///
+/// Created via [`UninitializedSandbox::guest_counter()`]. The host owns
+/// the counter value and is the only writer: [`increment()`](Self::increment)
+/// and [`decrement()`](Self::decrement) update the cached value and issue a
+/// single `write_volatile` to shared memory. [`value()`](Self::value) returns
+/// the cached value — the host never reads back from guest memory, preventing
+/// a confused-deputy attack where a malicious guest could manipulate the counter.
```
syntactically: I don't quite see the confused-deputy dimension to this?  Is this more directly related to what nanvix is actually using this counter value for?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -54,6 +56,75 @@ pub(crate) struct SandboxRuntimeConfig {
...
+/// Thread safety is provided by an internal `Mutex`, so `increment()` and
+/// `decrement()` take `&self` rather than `&mut self`.
+#[cfg(feature = "nanvix-unstable")]
+pub struct GuestCounter {
+    inner: Mutex<GuestCounterInner>,
+}
+
+#[cfg(feature = "nanvix-unstable")]
+struct GuestCounterInner {
+    ptr: *mut u64,
```
syntactically: Is there a reason to go through raw memory operations here, instead of holding onto the `SharedMemory` structure and using its read and write methods? Presently the code in the shared memory file is a nice single source of truth for the nasty details around preventing host side UB due to guest side races and so forth.

src/hyperlight_common/src/layout.rs
```diff
@@ -35,6 +35,15 @@ pub const SCRATCH_TOP_ALLOCATOR_OFFSET: u64 = 0x10;
 pub const SCRATCH_TOP_SNAPSHOT_PT_GPA_BASE_OFFSET: u64 = 0x18;
 pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x20;
 
+/// Offset from the top of scratch memory for a shared host-guest u64 counter.
+///
+/// This is placed at 0x1008 (rather than the next sequential 0x28) so that the
+/// counter falls in scratch page 0xffffe000 instead of the very last page
+/// 0xfffff000, which on i686 guests would require frame 0xfffff — exceeding the
+/// maximum representable frame number.
```
syntactically, line 47: Maybe MAX_GPA on i686 should just be changed to point to the top of actually representable memory? I believe the current value in there was a guess---if you haven't changed it, I didn't check the i686 docs, just truncated to 32 bits the value from amd64.

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +240,40 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+    ///
+    /// The caller must ensure the returned `GuestCounter` does not
+    /// outlive the sandbox's underlying shared memory mapping (which
+    /// stays alive through `evolve()` into `MultiUseSandbox`).
+    #[cfg(feature = "nanvix-unstable")]
+    pub unsafe fn guest_counter(&mut self) -> Result<GuestCounter> {
+        use hyperlight_common::layout::SCRATCH_TOP_GUEST_COUNTER_OFFSET;
+
+        let scratch_size = self.mgr.scratch_mem.mem_size();
+        if (SCRATCH_TOP_GUEST_COUNTER_OFFSET as usize) + core::mem::size_of::<u64>() > scratch_size
```
syntactically: I'm not sure this is the right check? If the intent here is just to make sure that the guest counter exists (e.g. if the memory is one 4k page or less), then I would expect you to not need the sizeof u64 addend here? That part is definitely fine as long as the offset from the top is >= sizeof(u64) which is statically clear.  There also might be a min_scratch_size() invariant that means this can never fire?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +240,40 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+    /// the top of scratch, so both host and guest can locate it without
+    /// an explicit GPA parameter.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure the returned `GuestCounter` does not
+    /// outlive the sandbox's underlying shared memory mapping (which
+    /// stays alive through `evolve()` into `MultiUseSandbox`).
+    #[cfg(feature = "nanvix-unstable")]
+    pub unsafe fn guest_counter(&mut self) -> Result<GuestCounter> {
```
jsturtevant: I think this is probably minor given the current usage, but this allows you to get two instances of GuestCounter, and the mutext is local to the guestcounter so potentially could have two instances overwriting themselves (looks like same type of thing as before) consider:

```
let c1 = unsafe { sandbox.guest_counter()? };
let c2 = unsafe { sandbox.guest_counter()? };

// c1 reads cached value 0, increments to 1, writes 1 to memory
c1.increment()?;
assert_eq!(c1.value()?, 1);

// c2 still has cached value 0, increments to 1, writes 1 to memory
c2.increment()?;
assert_eq!(c2.value()?, 1); // ← should be 2, but c2 doesn't know about c1's write
```

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -54,6 +56,75 @@ pub(crate) struct SandboxRuntimeConfig {
...
+///
+/// Thread safety is provided by an internal `Mutex`, so `increment()` and
+/// `decrement()` take `&self` rather than `&mut self`.
+#[cfg(feature = "nanvix-unstable")]
+pub struct GuestCounter {
+    inner: Mutex<GuestCounterInner>,
+}
+
+#[cfg(feature = "nanvix-unstable")]
+struct GuestCounterInner {
```
jsturtevant, line 87: should this have a lifetime so the inner pointer doesn't live past the sandbox?

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Fixed — changed to "exposed to the guest via a u64 in guest scratch memory."

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Fair point — removed. Now reads: "the host never reads back from guest memory, so a malicious guest cannot influence the host's view of the counter."

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: GuestCounter is created on UninitializedSandbox (which holds ExclusiveSharedMemory) but must survive across evolve() into MultiUseSandbox (which holds HostSharedMemory). There's no single SharedMemory reference we can hold across that lifecycle transition — ExclusiveSharedMemory is consumed by evolve() -> build(). The raw write_volatile is the same operation HostSharedMemory::write::<u64>() ultimately performs. I've added an "Implementation note" to the type docs explaining this trade-off.

src/hyperlight_common/src/layout.rs
danbugs: It already points to the top of representable 32-bit memory: MAX_GVA = 0xffff_ffff and MAX_GPA = 0xffff_ffff. This specific offset was just picked because of how Nanvix's memory booking works (couldn't have anything at the top); plus, I thought it'd be nice to leave 0x28 and near offsets open for our own future usage so we don't have to skip offsets depending on feature use.

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Documented as a safety requirement on the constructor: "Only one GuestCounter is created per sandbox. Multiple instances are memory-safe but their cached values will diverge, leading to incorrect writes." In Nanvix the counter is created once and wrapped in Arc — there's never a second call.

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: The pointer remains valid because evolve() internally clones the Arc<HostMapping> into both HostSharedMemory and GuestSharedMemory, keeping the mmap alive for the entire sandbox lifetime. A lifetime parameter would complicate usage across evolve() since the sandbox type changes. The unsafe constructor documents the invariant: "The returned GuestCounter does not outlive the sandbox's underlying shared memory mapping."

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Agreed — simplified to just SCRATCH_TOP_GUEST_COUNTER_OFFSET as usize > scratch_size. The offset is 0x1008 which already statically includes room for the u64.

src/hyperlight_common/src/layout.rs
syntactically: > Nanvix's memory booking works (couldn't have anything at the top)

The comment should reflect that then please.

src/hyperlight_host/src/sandbox/uninitialized.rs
syntactically: I think this should actually make sure they are synchronised, or return failure instead.  Safety requirements being documented is not a replacement for making things safe when it could be done relatively easily.

src/hyperlight_common/src/layout.rs
syntactically: The comment should explain that instead of saying that the frame is unrepresentable then.

src/hyperlight_host/src/sandbox/uninitialized.rs
syntactically: I really don't like the duplication here---it's too easy to miss in the future, especially for a fairly rarely-used feature.

Could you just hold on to the sandbox object and use that directly, or, to slightly reduce the footprint on the sandbox object / make it more flexible to add more in the future, have short-lived guest counter references that come off of a specific sandbox? Something like:

```rust
struct GuestCounter<'a, T> {
  host_value: u32,
  shm: &'a T,
  offset: usize,
}
impl<'a, T> GuestCounter<'a, T> {
  fn increment(&mut self) {
    // ...
  }
  fn decrement(&mut self) {
    // ...
  }
}
impl UninitialisedSandbox {
  fn<'a> guest_counter(&'a mut self) -> GuestCounter<'a, ExclusiveSharedMemory> {
    GuestCounter {
      host_value: self.counter_value,
      shm: &'a self.mem_mgr.scratch_mem,
      offset: //..
  }
}
impl InitialisedMultiUseSandbox {
  fn<'a> guest_counter(&'a mut self) -> GuestCounter<'a, HostSharedMemory> {
    GuestCounter {
      host_value: self.counter_value,
      shm: &'a self.mem_mgr.scratch_mem,
      offset: //..
  }
}
```

src/hyperlight_host/src/sandbox/uninitialized.rs
syntactically: Again, I don't think that documentation is a replacement for safety. There are multiple ways we could address this. For example, in rough order of preference:
- Scoped GuestCounter re-derived from the Sandbox as necessary, as I suggested in the other comment
- Sandbox keeps a copy of the GuestCounterInner structure and revokes the pointer when the sandbox is dropped
- Rework SHM code to know about GuestCounter and keep memory alive as long as necessary

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -136,7 +138,17 @@ impl Drop for HostMapping {
 /// and taking snapshots.
 #[derive(Debug)]
 pub struct ExclusiveSharedMemory {
-    region: Arc<HostMapping>,
+    pub(crate) region: Arc<HostMapping>,
```
syntactically: I think you said these don't need to be pub(crate) anymore, because `simulate_build` is in this module.

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -136,7 +138,17 @@ impl Drop for HostMapping {
...
+    /// Populated by [`build()`](Self::build) with a [`HostSharedMemory`]
+    /// view of this region. Code that needs host-style volatile access
+    /// before `build()` (e.g. `GuestCounter`) can clone this `Arc` and
+    /// will see `Some` once `build()` completes.
+    pub(crate) deferred_hshm: Arc<Mutex<Option<HostSharedMemory>>>,
+    /// Set to `true` once a [`GuestCounter`] has been handed out via
```
syntactically: I'm not totally convinced that this should be here as opposed to being in `UninitializedSandbox`.

danbugs:
Resolved a lot of the comments in this PR discussing w/ @jsturtevant and @syntactically privately. Here's a summary of the changes made from our discussions:

**(1) refactor: use deferred HostSharedMemory for GuestCounter**
- Replaced the raw-pointer +  approach with Arc<Mutex<Option<HostSharedMemory>>> shared between ExclusiveSharedMemory and GuestCounter
- ExclusiveSharedMemory holds the Arc<Mutex<Option<...>>>, which starts as None
- build() creates the RwLock fresh and populates the Option with Some(hshm)
- GuestCounter clones the Arc at construction time and polls for Some on each increment()/decrement()
- No unsafe code — constructor is now safe (pub fn instead of pub unsafe fn)
- Counter operations return a clear error before build() completes
- HostSharedMemory::write() remains the single source of truth for host-side shared memory access (volatile + RwLock-guarded)

**(2) refactor: enforce single GuestCounter, harden encapsulation**
- Added counter_taken: AtomicBool to ExclusiveSharedMemory; guest_counter() atomically swaps it and returns an error on a second call (addresses the &mut self not preventing multiple counters)
- Initialized counter value to 0 directly instead of reading from the offset (scratch memory is always zero at that point)
- Reverted HostSharedMemory fields back to private; added #[cfg(all(test, feature = "nanvix-unstable"))]
- simulate_build() on ExclusiveSharedMemory so tests don't leak shm internals
- Added only_one_counter_allowed test

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -136,7 +138,17 @@ impl Drop for HostMapping {
 /// and taking snapshots.
 #[derive(Debug)]
 pub struct ExclusiveSharedMemory {
-    region: Arc<HostMapping>,
+    pub(crate) region: Arc<HostMapping>,
+    /// Populated by [`build()`](Self::build) with a [`HostSharedMemory`]
+    /// view of this region. Code that needs host-style volatile access
+    /// before `build()` (e.g. `GuestCounter`) can clone this `Arc` and
+    /// will see `Some` once `build()` completes.
+    pub(crate) deferred_hshm: Arc<Mutex<Option<HostSharedMemory>>>,
```
jsturtevant: should this also be being the nanvix-unstable flag?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -54,6 +56,96 @@ pub(crate) struct SandboxRuntimeConfig {
...
+                .ok_or_else(|| {
+                    new_error!("GuestCounter cannot be used before shared memory is built")
+                })?
+                .clone()
+        };
+        inner.value = inner
+            .value
+            .checked_add(1)
+            .ok_or_else(|| new_error!("GuestCounter overflow"))?;
+        shm.write::<u64>(inner.offset, inner.value)?;
```
jsturtevant: Should switch this around to avoid a cache write and failure later on the write to memory:

  shm.write::<u64>(inner.offset, new_value)?;
        inner.value = new_value;

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -169,6 +261,57 @@ impl<'a> From<GuestBinary<'a>> for GuestEnvironment<'a, '_> {
...
+        }
+
+        let offset = scratch_size - SCRATCH_TOP_GUEST_COUNTER_OFFSET as usize;
+        let deferred_hshm = self.mgr.scratch_mem.deferred_hshm.clone();
+
+        Ok(GuestCounter {
+            inner: Mutex::new(GuestCounterInner {
+                deferred_hshm,
+                offset,
+                value: 0,
```
jsturtevant, line 317: Do we need to zero the actual memory?  Or maybe not because its scratch and it will always be zero'd?

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Moved it to behind the flag too in uninit sbox.

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Switched ✅

src/hyperlight_host/src/sandbox/uninitialized.rs
danbugs: Yeah, no need. Scratch is already zero-ed. I was reading the value before but removed that as part of one of rounds of review from syntactically's feedback.

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Moved to uninit.

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: No longer pub(crate).

ludfjig:
I would have preferred if the commits got cleaned up (no commits for refactors + fmt) before this got merged, since it was not squashed and they land directly on main @danbugs 1 or 2 commits would probably have been sufficient for this PR

syntactically:
Strong agree with @ludfjig here---the commits that actually go into main should be well-organised, clean, and tell a useful story. There shouldn't be any "address comments" or "refactor" commits that rework the stuff from the same PR in response to review feedback.  There should not be any commits in main at which the tree is broken/containing code that code review found was unacceptable.  For this particular PR, it probably only needed one (fairly small on its own) commit.

Issue #1268, opened 2026-03-02
Snapshot-First Sandbox Lifecycle


simongdavies:
I've moved this content to [hackmd](https://hackmd.io/@8sYwbcKoR36MNDyyFvtecQ/HJnxgWyqWe) . I've moved Dorus and Lucys feedback (thanks) over there.

dblnz:
This is awesome! I appreciate the effort to write this.
I think the new proposed API looks a lot cleaner.

Below are some comments that came to mind while reading:
- R2.7 - This assumes that the same external file is always used when loading a Snapshot, is this the intended behaviour? R1.8 mentions the OCI manifests, containing the companion blobs, but what about other use cases?
R3.4 makes this clearer. By having the host application specifically map the external file, we can fail at that point in case something doesn't match.
- R1.9 - I am not sure that at the moment the memory profiling works, with the new cow changes, we need to check this out also.
- R4.4 - We need to measure this, I don't remember seeing a measurement for KVM
- R5.5 - This means that the user will have to redo snapshots from scratch (with all the layers) if we change the format version. Can we possibly make this more user-friendly to maybe somehow convert?
- R6 - Do I understand this correctly?
Let's say we have a base snapshot we want to expand the heap for, and take a snapshot to save that change.
The base defines the heap as, _X_, and, when loading, the host expands it to _Y_ and takes a snapshot.
As mentioned, the hash of the base only covers the original contents and is not invalidated, but the new snapshot's hash covers the expanded heap pages also, right?
- R10.5 - I think that it would also be helpful for one of the CLI tools to print what registered host function and guest functions are available for a snapshot (something like what `wasm-tools` does with the WIT imports/exports)
- One more thing we need to think about is how to handle debugging a snapshot (I already linked an issue with this one).
Now we rely on an `entrypoint` breakpoint that makes the guest stop for the user to attach with the debugger and start the debugging session. This will have to change once we start working towards this proposal
- Q1-Q2 - I'm for the shorter extensions (`.hls` and `.hll`)
- Q3 - I don't have any insight here
- Q5 - I think having a deprecation period is nicer for our users

syntactically:
In the ongoing discussion about the role of HIPs, I'll say that some of this (more the API design/implementation plan than the requirements) might make sense as one (at least in the "current" process).

I have a few comments on the requirements:

[...]

PR #1271, opened 2026-03-03
feat: use i686 layout for nanvix-unstable guests and make snapshot RWX

danbugs:
## Summary

When `nanvix-unstable` is enabled, selects the i686 hyperlight-common memory layout on x86_64 hosts and makes the snapshot memory region writable. Also propagates the `nanvix-unstable` feature from `hyperlight-host` to `hyperlight-common` so both crates share the same cfg gates.

## Motivation

Nanvix runs as a 32-bit (i386) guest without CoW guest paging. Two things need to change for this to work:

1. **i686 layout**: The amd64 layout defines `MAX_GPA`/`MAX_GVA` for a 36-bit/48-bit address space and includes `SNAPSHOT_PT_GVA_*` constants for the CoW page table region. A 32-bit guest needs 32-bit address limits (`0xFFFF_FFFF`) and has no snapshot page tables. With `nanvix-unstable`, the i686 layout module is used even on x86_64 hosts.

2. **Snapshot RWX**: Without CoW, the guest writes directly to snapshot pages — `.data`, `.bss`, kernel page pool, and even GDT descriptors (the CPU sets the "Accessed" bit during segment loads). If the snapshot isn't writable, these writes cause EPT violations that KVM retries forever.

## Changes

- `hyperlight_host/Cargo.toml`: `nanvix-unstable = ["hyperlight-common/nanvix-unstable"]` — propagates the feature
- `hyperlight_common/src/layout.rs`: Three-way `cfg_attr` selects i686 layout for native x86 OR x86_64 with `nanvix-unstable`; gates `SNAPSHOT_PT_GVA_*` exports behind `all(target_arch = "x86_64", not(nanvix-unstable))`
- `hyperlight_common/src/arch/i686/layout.rs`: Removes stale `SNAPSHOT_PT_GVA_*` constants and fixes `MAX_GVA` to `0xFFFF_FFFF`; updates `min_scratch_size` signature to match the amd64 layout
- `hyperlight_host/src/mem/shared_mem.rs`: Snapshot region is `READ | EXECUTE` by default (CoW handles writes), `READ | WRITE | EXECUTE` with `nanvix-unstable`
- `hyperlight_host/src/sandbox/snapshot.rs`: Gates the snapshot page-table GVA filter behind `not(nanvix-unstable)`

### DIFF
--- a/src/hyperlight_common/src/arch/i686/layout.rs
+++ b/src/hyperlight_common/src/arch/i686/layout.rs
@@ -17,11 +17,9 @@ limitations under the License.
 // This file is just dummy definitions at the moment, in order to
 // allow compiling the guest for real mode boot scenarios.
 
-pub const MAX_GVA: usize = 0xffff_efff;
-pub const SNAPSHOT_PT_GVA_MIN: usize = 0xef00_0000;
-pub const SNAPSHOT_PT_GVA_MAX: usize = 0xefff_efff;
+pub const MAX_GVA: usize = 0xffff_ffff;
 pub const MAX_GPA: usize = 0xffff_ffff;
 
-pub fn min_scratch_size() -> usize {
-    1 * crate::vmem::PAGE_SIZE
+pub fn min_scratch_size(_input_data_size: usize, _output_data_size: usize) -> usize {
+    crate::vmem::PAGE_SIZE
 }
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -14,11 +14,20 @@ See the License for the specific language governing permissions and
 limitations under the License.
  */
 
-#[cfg_attr(target_arch = "x86_64", path = "arch/amd64/layout.rs")]
 #[cfg_attr(target_arch = "x86", path = "arch/i686/layout.rs")]
+#[cfg_attr(
+    all(target_arch = "x86_64", not(feature = "nanvix-unstable")),
+    path = "arch/amd64/layout.rs"
+)]
+#[cfg_attr(
+    all(target_arch = "x86_64", feature = "nanvix-unstable"),
+    path = "arch/i686/layout.rs"
+)]
 mod arch;
 
-pub use arch::{MAX_GPA, MAX_GVA, SNAPSHOT_PT_GVA_MAX, SNAPSHOT_PT_GVA_MIN};
+pub use arch::{MAX_GPA, MAX_GVA};
+#[cfg(all(target_arch = "x86_64", not(feature = "nanvix-unstable")))]
+pub use arch::{SNAPSHOT_PT_GVA_MAX, SNAPSHOT_PT_GVA_MIN};
 
 // offsets down from the top of scratch memory for various things
 pub const SCRATCH_TOP_SIZE_OFFSET: u64 = 0x08;
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -137,7 +137,7 @@ mshv3 = ["dep:mshv-bindings", "dep:mshv-ioctls"]
 gdb = ["dep:gdbstub", "dep:gdbstub_arch"]
 fuzzing = ["hyperlight-common/fuzzing"]
 build-metadata = ["dep:built"]
-nanvix-unstable = []
+nanvix-unstable = ["hyperlight-common/nanvix-unstable"]
 
 [[bench]]
 name = "benchmarks"
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -679,7 +679,22 @@ impl GuestSharedMemory {
             MemoryRegionType::Scratch => {
                 MemoryRegionFlags::READ | MemoryRegionFlags::WRITE | MemoryRegionFlags::EXECUTE
             }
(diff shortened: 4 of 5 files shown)


src/hyperlight_host/src/hypervisor/regs/special_regs.rs
```diff
@@ -106,34 +106,55 @@ impl CommonSpecialRegisters {
...
                 limit: 0xFFFF,
-                type_: 11,
+                type_: 2, // LDT descriptor
                 present: 1,
-                s: 0,
+                s: 0, // system segment
                 ..Default::default()
             },
+            // CR0.ET (bit 4) is hardwired to 1 on modern x86 CPUs.
+            // MSHV rejects the vCPU state if ET is not set.
```
dblnz: How hasn't mshv failed so far? was this scenario not tested before?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
```diff
@@ -374,6 +376,12 @@ impl HyperlightVm {
         vm.set_sregs(&CommonSpecialRegisters::standard_real_mode_defaults())
             .map_err(VmError::Register)?;
 
+        // Initialize XSAVE state with proper FPU defaults (FCW, MXCSR).
+        // MSHV (unlike KVM) does not provide sane defaults for the XSAVE
+        // area on a freshly created vCPU, and will reject the VP state
+        // on the first run if it is left uninitialized.
+        vm.reset_xsave().map_err(VmError::Register)?;
```
dblnz: I'm curious how this worked before

syntactically: It looks like at issue here is the difference between bringing up the vcpu in real vs long mode. Does the xsave reset call need to be there in both, or should it just be in real mode? I know we had tests of the long mode mshv stuff passing before.

src/hyperlight_host/src/hypervisor/regs/special_regs.rs
syntactically: I don't believe there are many tests for the 16/32-bit guests.

syntactically:
I'm not clear on the need for a configurable region base for scratch. You can just set MAX_GPA/MAX_GVA based on target architecture + feature flags? (see src/hyperlight_common/src/arch/i686/layout.rs).  Maybe the selection based on target_arch of which layout module to use is broken in the host for the i686 case where it's the guest arch that counts; maybe that needs to change to something like `#[cfg_attr(any(target_arch = "x86", all(target_arch = "x86-64", not(feature = "init-paging")), path = "arch/i686/layout.rs")]`.  (Or, we could add cargo cfg aliases for guest arch, or something?). I would really prefer to avoid re-adding more configuration knobs.

I'm not clear on how the idea of making snapshot pages writable is going to work with the snapshot-first lifecycle (#1268). This seems like it will break a lot of assumptions and make a _lot_ of code more complex/rife with opportunities for vulnerabilities. If you really are unable to run on amd64 natively, would teaching hyperlight about i686 paging structures be a slightly smaller break of the conceptual model?

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
jsturtevant: I think the real vs long is the issue here. When I did some investigation in mxcsr on kvm awhile back it did work fine in long mode.

jsturtevant:
>I'm not clear on how the idea of making snapshot pages writable is going to work with the snapshot-first lifecycle (https://github.com/hyperlight-dev/hyperlight/issues/1268).

This seems like something we could flush out in that discussion? With CoW changes we've broken some downstream consumers, so it makes sense give an escape hatch (maybe behind a feature flag) while we sort through what we want to do.  As far as I understand, this isn't a breaking change and doesn't introduce issues now? or does it?  To be clear I think we do need to reconcile this requirement and enabling this wouldn't be a long term solution

src/hyperlight_host/src/hypervisor/hyperlight_vm.rs
ludfjig: I agree that if we don't need this for some targets it would be nice to not waste cycles here

danbugs:
Addressed the config option concern. Latest push removes `set_scratch_base_gpa()` from `SandboxConfiguration` entirely. Instead, the i686 memory layout is now auto-selected at compile time:

- When `init-paging` feature is **enabled** (standard hyperlight guests), the existing amd64 layout is used unchanged.
- When `init-paging` feature is **dsaibled** (Nanvix), an i686-compatible layout is selected in `hyperlight-common` that places the scratch region at a lower GPA (below 4GB), compatible with 32-bit page tables.

The `init-paging` feature is propagated from `hyperlight-host` to `hyperlight-common` via Cargo feature unification, so no runtime configuration is needed.

Regarding the xsave commit that was on this branch — it's been dropped from the latest push since it's orthogonal to the scratch GPA changes. The writable snapshot commit remains, gated behind `init-paging`. Nanvix needs writable snapshots because its guest page tables and kernel state live in the snapshot EPT region and must be writable during execution.

Note that Nanvix doesn't use Hyperlight's snapshotting mechanism. Instead, it has a kernel call that, being guest-aware, can VMExit at any time to snapshot. Nanvix does this because it doesn't execute like our purpose-built guest binaries (e.g., it doesn't use guest function calls). Instead, it just calls `evolve` once.

PR #1285, opened 2026-03-06
feat: enable hyperlight-host compilation for aarch64

ludfjig:
Add conditional compilation gates and aarch64 stub modules so that cargo build --target aarch64-unknown-linux-gnu -p hyperlight-host succeeds. No aarch64 implementation is added. All aarch64 code paths are stubs that panic at runtime.

**important**: There will be some code duplication addition for aarch64 initially, espeically regarding HyperlightVm and KvmVm and MshvVm. It will make sense to factor some of this out (trait, etc) when we figure out more what the code looks like with aarch64 support added, but I don't think it's worth doing before adding the actual implementation.

I recommend reviewing commits locally using `--color-moved` flag to easier identify moved code vs changed code

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
@@ -121,49 +121,6 @@ pub(super) fn get_guest_log_filter(guest_max_log_level: Option<LevelFilter>) -> 
     GuestLogFilter::from(guest_log_level_filter).into()
 }
 
-/// Represents a Hyperlight Virtual Machine instance.
-///
-/// This struct manages the lifecycle of the VM, including:
-/// - The underlying hypervisor implementation (e.g., KVM, MSHV, WHP).
-/// - Memory management, including initial sandbox regions and dynamic mappings.
-/// - The vCPU execution loop and handling of VM exits (I/O, MMIO, interrupts).
-pub(crate) struct HyperlightVm {
-    #[cfg(gdb)]
-    pub(super) vm: Box<dyn DebuggableVm>,
-    #[cfg(not(gdb))]
-    pub(super) vm: Box<dyn VirtualMachine>,
-    pub(super) page_size: usize,
-    pub(super) entrypoint: NextAction, // only present if this vm has not yet been initialised
-    pub(super) rsp_gva: u64,
-    pub(super) interrupt_handle: Arc<dyn InterruptHandleImpl>,
-
-    pub(super) next_slot: u32, // Monotonically increasing slot number
-    pub(super) freed_slots: Vec<u32>, // Reusable slots from unmapped regions
-
-    pub(super) snapshot_slot: u32,
-    // The current snapshot region, used to keep it alive as long as
-    // it is used & when unmapping
-    pub(super) snapshot_memory: Option<GuestSharedMemory>,
-    pub(super) scratch_slot: u32, // The slot number used for the scratch region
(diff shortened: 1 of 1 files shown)


docs/aarch64-architecture-plan.md
```diff
@@ -0,0 +1,258 @@
...
+| **`stdarch`** / `core::arch` (rust-lang) | Separate arch crate modules | `x86_64/`, `aarch64/` directories. Each arch is self-contained. |
+| **`rustix`** (bytecodealliance) | Backend abstraction layer | `backend/` directory with `linux_raw/`, `libc/` backends, each internally split by arch. |
+
+### Recommended Pattern for Hyperlight
+
+Hyperlight should use a combination of **Firecracker's re-export pattern** and **Cloud Hypervisor's backend structure**:
+
+1. **`cfg_attr` path switching** for arch-specific module files (already used in `hyperlight-common` and `hyperlight-guest`). Extend to aarch64.
+2. **`cfg(target_arch)` module gating with wildcard re-exports** for register types in `hyperlight-host` (like `kvm-bindings`).
+3. **`cfg(target_arch)` on entire impl blocks and functions** rather than on individual struct fields — avoids `#[cfg]` spaghetti inside struct definitions.
```
dblnz: Although I don't really like having `impl` blocks for the same type in separate files (it's difficult to follow), I agree it's better than having `#[cfg]` inside struct definitions.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -0,0 +1,730 @@
...
+                                addr,
+                                access_type: MemoryRegionFlags::READ,
+                                region_flags,
+                            });
+                        }
+                        None => {
+                            break Err(RunVmError::MmioReadUnmapped(addr));
+                        }
+                    }
+                }
```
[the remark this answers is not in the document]
ludfjig: existing bug, will fix in follow up

docs/aarch64-architecture-plan.md
ludfjig: I removed the planf rom the PR

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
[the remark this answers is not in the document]
ludfjig: fixed in #1295

ludfjig:
> LGTM!.
> 
> The only outstanding comment I have is related to a build job for aarch64 to make sure at least `hyperlight-common` compiles. But this can be added in one of the later PRs.
> 
> So, it's your choice

Will add later, don't want to block other PRs on this

src/hyperlight_common/src/layout.rs
```diff
@@ -23,10 +23,14 @@ limitations under the License.
...
 )]
+#[cfg_attr(target_arch = "aarch64", path = "arch/aarch64/layout.rs")]
 mod arch;
 
 pub use arch::{MAX_GPA, MAX_GVA};
-#[cfg(all(target_arch = "x86_64", not(feature = "nanvix-unstable")))]
+#[cfg(any(
+    all(target_arch = "x86_64", not(feature = "nanvix-unstable")),
+    target_arch = "aarch64"
+))]
```
jsturtevant, line 33: nit: would be simpler to just include this for anything not nanvix?

ludfjig: I prefer this I think

jsturtevant (approved):

PR #1294, opened 2026-03-09
ci: add nanvix-unstable compile checks to CI and Justfile

danbugs:
## Summary

Adds compile-time CI coverage for the `nanvix-unstable` feature flag introduced in #1291 to catch regressions on the non-CoW code path.

Only compile checks are added — not test runs — because the standard Hyperlight test guests require CoW paging, which is disabled under `nanvix-unstable`.

## Changes

- `Justfile` `check`: verify `nanvix-unstable` and `nanvix-unstable,executable_heap` compile on the host
- `Justfile` `check-i686`: verify `hyperlight-common` builds for i686 with `nanvix-unstable` enabled

The `check` recipe is already invoked by both `test-like-ci` and `code-checks-like-ci`, so these checks run automatically in CI without needing a new workflow step.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -261,6 +261,7 @@ test-rust-tracing target=default-target features="":
 check-i686 target=default-target:
     cargo check -p hyperlight-common --target i686-unknown-linux-gnu --profile={{ if target == "debug" { "dev" } else { target } }}
     cargo check -p hyperlight-guest --target i686-unknown-linux-gnu --profile={{ if target == "debug" { "dev" } else { target } }}
+    cargo check -p hyperlight-common --target i686-unknown-linux-gnu --features nanvix-unstable --profile={{ if target == "debug" { "dev" } else { target } }}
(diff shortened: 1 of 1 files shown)


ludfjig (approved):

jsturtevant:
should we update `clippy-exhaustive` as well?

danbugs:
> should we update `clippy-exhaustive` as well?

clippy-exhaustive already tests w/ nanvix-unstable. clippy-exhaustive auto-discovers features to test w/ using `cargo metadata`. Verified it does pick it up locally rn:
```
❯ cargo metadata --format-version 1 --no-deps | jq -r '.packages[] | select(.name == "hyperlight-host") | .features | keys[]'
build-metadata
crashdump
default
executable_heap
function_call_metrics
fuzzing
gdb
kvm
mem_profile
mshv3
nanvix-unstable
print_debug
trace_guest
```

PR #1296, opened 2026-03-10
Map file cow windows

simongdavies:
This pull request introduces an implementation of `map_file_cow` on Windows and adds Windows specific and additional tests for file mapping.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
@@ -67,53 +67,22 @@ pub(crate) fn is_hypervisor_present() -> bool {
     }
 }
 
-/// Tracks a host-side file mapping created by [`MultiUseSandbox::map_file_cow`],
-/// providing RAII cleanup of the underlying Windows handles.
-///
-/// When dropped, releases both the `MapViewOfFile` view and the
-/// `CreateFileMappingW` handle, preventing resource leaks.
-struct OwnedFileMapping {
-    /// Host-side base pointer returned by `MapViewOfFile`.
-    view_base: *mut core::ffi::c_void,
-    /// File mapping handle returned by `CreateFileMappingW`.
-    mapping_handle: HandleWrapper,
-}
-
-impl std::fmt::Debug for OwnedFileMapping {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("OwnedFileMapping")
-            .field("view_base", &self.view_base)
-            .field("mapping_handle", &self.mapping_handle)
-            .finish()
-    }
-}
-
-impl Drop for OwnedFileMapping {
-    fn drop(&mut self) {
-        // Unmap the view first, then close the file mapping handle.
-        unsafe {
-            if let Err(e) = UnmapViewOfFile(MEMORY_MAPPED_VIEW_ADDRESS {
-                Value: self.view_base,
-            }) {
-                tracing::error!("Failed to unmap file view at {:?}: {:?}", self.view_base, e);
-            }
-            if let Err(e) = CloseHandle(self.mapping_handle.into()) {
-                tracing::error!(
-                    "Failed to close file mapping handle {:?}: {:?}",
-                    self.mapping_handle,
-                    e
-                );
-            }
+/// Helper: release a host-side file mapping view and its handle.
+/// Called from both `unmap_memory` and `WhpVm::drop`.
+fn release_file_mapping(view_base: *mut c_void, mapping_handle: HandleWrapper) {
+    unsafe {
+        if let Err(e) = UnmapViewOfFile(MEMORY_MAPPED_VIEW_ADDRESS { Value: view_base }) {
+            tracing::error!("Failed to unmap file view at {:?}: {:?}", view_base, e);
+        }
+        if let Err(e) = CloseHandle(mapping_handle.into()) {
+            tracing::error!(
+                "Failed to close file mapping handle {:?}: {:?}",
+                mapping_handle,
+                e
+            );
         }
     }
 }
-
-// SAFETY: `view_base` is a pointer to a Windows memory-mapped view created by
-// `MapViewOfFile`. Both `UnmapViewOfFile` and `CloseHandle` are thread-safe
-// Win32 APIs that operate on kernel objects, so they can safely be called
-// from any thread. The `HandleWrapper` is already `Send + Sync`.
-unsafe impl Send for OwnedFileMapping {}
-unsafe impl Sync for OwnedFileMapping {}
 
 /// A Windows Hypervisor Platform implementation of a single-vcpu VM
 #[derive(Debug)]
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -498,4 +499,127 @@ mod tests {
...
+        let surrogate_address = process
+            .map(
+                HandleWrapper::from(mem.get_mmap_file_handle()),
+                mem.raw_ptr() as usize,
+                mem.raw_mem_size(),
+                &crate::mem::memory_region::SurrogateMapping::ReadOnlyFile,
+            )
+            .unwrap();
+
+        let buffer = vec![0u8; SIZE];
```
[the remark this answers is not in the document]
simongdavies: The test is correct as written — it's validating that guard pages exist on both sides of the mapped region in the surrogate process. Copilot is confusing host-side shared memory size with surrogate-side virtual allocation size.

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -498,4 +499,127 @@ mod tests {
...
+            );
+            assert!(
+                success.is_ok(),
+                "First page should be readable with ReadOnlyFile (no guard page)"
+            );
+
+            // read the middle page — should also succeed
+            let success = windows::Win32::System::Diagnostics::Debug::ReadProcessMemory(
+                process_handle,
+                surrogate_address.wrapping_add(SIZE),
```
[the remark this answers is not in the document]
simongdavies: Same as above

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -243,7 +272,7 @@ pub struct MemoryRegion_<K: MemoryRegionKind> {
     /// memory access flags for the given region
     pub flags: MemoryRegionFlags,
     /// the type of memory region
-    pub region_type: MemoryRegionType,
+    pub(crate) region_type: MemoryRegionType,
```
jsturtevant, line 285: This makes this type not usable outside the crate?  which negates these changes I just made: https://github.com/hyperlight-dev/hyperlight/pull/1293/changes for HL-wasm's load_module_by_mapping.  

we use `map_region` currnetly.  Is there a different way we are suppose to map these files in hl-wasm?

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -413,3 +442,161 @@ impl From<&MemoryRegion> for kvm_bindings::kvm_userspace_memory_region {
...
+            assert_ne!(hash_of(&sandbox), hash_of(&readonly));
+        }
+
+        #[test]
+        fn host_region_base_same_surrogate_mapping_equal_and_same_hash() {
+            let a = make_host_region_base(SurrogateMapping::SandboxMemory);
+            let b = make_host_region_base(SurrogateMapping::SandboxMemory);
+            assert_eq!(a, b);
+            assert_eq!(hash_of(&a), hash_of(&b));
+        }
```
ludfjig: These tests seem unnecessary

src/hyperlight_host/src/mem/memory_region.rs
simongdavies: OK I will clean this up, I only did it because I was adding a new `MemoryRegionType` member which technically is breaking and I didn't think there was anyone using it

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -144,6 +144,7 @@ impl VirtualMachine for WhpVm {
                 region.host_region.start.from_handle,
                 region.host_region.start.handle_base,
                 region.host_region.start.handle_size,
+                &region.host_region.start.surrogate_mapping,
```
ludfjig: I **think** we can get rid of the entire SurrogateMapping from this PR and instead just pass region.flags here. I think this would make this PR a lot smaller

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -99,6 +156,11 @@ pub struct MultiUseSandbox {
     /// If the current state of the sandbox has been captured in a snapshot,
     /// that snapshot is stored here.
     snapshot: Option<Arc<Snapshot>>,
+    /// Tracks host-side file mappings created by [`Self::map_file_cow`] for
+    /// RAII cleanup. Declared **after** `vm` so that Rust's field drop order
+    /// releases hypervisor mappings before closing host-side file handles.
+    #[cfg(target_os = "windows")]
+    file_mappings: Vec<OwnedFileMapping>,
```
ludfjig: Are you certain this order/comment is necessary? Even if OwnedFileMapping is dropped firest, I suspect the surrgoate process would keep it alive until SuroogateProcess::drop calls UnmapViewOfFile2

ludfjig (changes requested):

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
ludfjig: if contains WRITE -> PAGE_READWRITE and guardpages , otherwise PAGE_READONLY and no guardapges

ludfjig:
Also, could memmap2 crate be useful here instead of our own OwnedFileMapping? I'm not sure if it has accessors we need for the raw windows calls we need, but worth investigating

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: I'm also wondering if this is preferable to just doing the unmap in WhpVm (in drop and unmap_region), based on the MemoryRegionType?

src/hyperlight_host/src/mem/memory_region.rs
simongdavies: Done

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
simongdavies: This is gone and based on MemoryRegionType now

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: Good idea, moved to `WhpVm`.

simongdavies:
> Also, could memmap2 crate be useful here instead of our own OwnedFileMapping? I'm not sure if it has accessors we need for the raw windows calls we need, but worth investigating

This seems like a bigger change outside the scope of this PR , we already use the Windows APIs for Windows Memory allocation, we would also probably want to  use this crate on Linux for consistency

ludfjig:
> > Also, could memmap2 crate be useful here instead of our own OwnedFileMapping? I'm not sure if it has accessors we need for the raw windows calls we need, but worth investigating
> 
> This seems like a bigger change outside the scope of this PR , we already use the Windows APIs for Windows Memory allocation, we would also probably want to use this crate on Linux for consistency

agree sounds reasonble

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -552,17 +560,106 @@ impl MultiUseSandbox {
...
+            {
+                // Clean up host-side resources on failure
+                unsafe {
+                    let _ = UnmapViewOfFile(view);
+                    let _ = CloseHandle(mapping_handle);
+                }
+                return Err(err);
+            }
+
+            self.mem_mgr.mapped_rgns += 1;
```
ludfjig, line 649: is this used for anything or can we get rid of it?

simongdavies: I don't know , the same code is in the Linux `map_regions` which is called from Linux impl of `map_file_cow` , I wonder if its a hangover from before the COW work  , created issue for this.

Issue #1190, opened 2026-01-26
Generate coverage reports


dblnz:
It would be nice to get some test coverage reports to know where we are in terms of testing.

dblnz:
We now have [this ](https://github.com/hyperlight-dev/hyperlight/actions/workflows/Coverage.yml) report that shows the test coverage for kvm.
The next step is to extend this to cover mshv and windows.
The logical steps would be:
- run tests on kvm, mshv and windows as separate jobs
- store the results
- Separate job to merge all the coverage data and generate a report.
I'm not sure yet, but we might need to save other artifacts beside the coverage data.

PR #1301, opened 2026-03-11
feat: add coverage report generation with cargo-llvm-cov

dblnz:
Add coverage infrastructure for the Hyperlight project:

Partially addresses #1190

- Justfile: add coverage, coverage-html, coverage-lcov, and coverage-ci recipes using cargo-llvm-cov for LLVM source-based code coverage
- CI: add Coverage.yml weekly workflow (Monday 06:00 UTC, manual trigger) running on kvm/amd with self-built guest binaries
- docs: add how-to-run-coverage.md with local and CI usage instructions

Guest/no_std crates are excluded from coverage because they define #[panic_handler] and cannot be compiled for the host target under coverage instrumentation. Coverage targets host-side crates only.

This PR:
- only runs this on linux with KVM

### DIFF
--- a/.github/workflows/Coverage.yml
+++ b/.github/workflows/Coverage.yml
@@ -1,0 +1,108 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Weekly Coverage
+
+on:
+  pull_request:
+    paths:
+      - .github/workflows/Coverage.yml
+  schedule:
+    # Runs every Monday at 06:00 UTC
+    - cron: '0 6 * * 1'
+  workflow_dispatch: # Allow manual trigger
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  coverage:
+    timeout-minutes: 90
+    strategy:
+      fail-fast: false
+      matrix:
+        hypervisor: [kvm]
+        cpu: [amd]
+    runs-on: ${{ fromJson(
+        format('["self-hosted", "Linux", "X64", "1ES.Pool=hld-{0}-{1}"]',
+          matrix.hypervisor,
+          matrix.cpu)) }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: hyperlight-dev/ci-setup-workflow@v1.9.0
+        with:
+          rust-toolchain: "1.89"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Fix cargo home permissions
+        run: |
+          sudo chown -R $(id -u):$(id -g) /opt/cargo || true
+
+      - name: Rust cache
+        uses: Swatinem/rust-cache@v2
+        with:
+          shared-key: "${{ runner.os }}-debug"
+          cache-on-failure: "true"
+
+      - name: Build guest binaries
+        run: just guests
+
+      - name: Install nightly toolchain
+        run: |
+          rustup toolchain install nightly
+          rustup component add llvm-tools --toolchain nightly
+
+      - name: Generate coverage report
+        run: just coverage-ci ${{ matrix.hypervisor }}
+
+      - name: Coverage summary
+        run: |
+          echo '## Code Coverage Report' >> $GITHUB_STEP_SUMMARY
+          echo '' >> $GITHUB_STEP_SUMMARY
+          if [ -f target/coverage/summary.txt ]; then
+            echo '```' >> $GITHUB_STEP_SUMMARY
+            cat target/coverage/summary.txt >> $GITHUB_STEP_SUMMARY
+            echo '```' >> $GITHUB_STEP_SUMMARY
+          else
+            echo 'Coverage report was not generated.' >> $GITHUB_STEP_SUMMARY
+          fi
+          echo '' >> $GITHUB_STEP_SUMMARY
+          echo '> For a detailed per-file breakdown, download the **HTML coverage report** from the Artifacts section below.' >> $GITHUB_STEP_SUMMARY
+
+      - name: Upload HTML coverage report
+        uses: actions/upload-artifact@v7
+        with:
+          name: coverage-html-${{ matrix.hypervisor }}-${{ matrix.cpu }}
+          path: target/coverage/html/
+          if-no-files-found: error
+
+      - name: Upload LCOV coverage report
+        uses: actions/upload-artifact@v7
+        with:
+          name: coverage-lcov-${{ matrix.hypervisor }}-${{ matrix.cpu }}
+          path: target/coverage/lcov.info
+          if-no-files-found: error
+
+  notify-failure:
+    runs-on: ubuntu-latest
+    needs: [coverage]
+    if: always() && needs.coverage.result == 'failure'
+    permissions:
+      issues: write
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+
+      - name: Notify Coverage Failure
+        run: ./dev/notify-ci-failure.sh --title="Weekly Coverage Failure - ${{ github.run_number }}" --labels="area/ci-periodics,area/testing,lifecycle/needs-review"
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/Justfile
+++ b/Justfile
@@ -459,6 +459,105 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
 build-trace-fuzzers:
     cargo +nightly fuzz build fuzz_guest_trace --features trace
 
+####################
+### COVERAGE #######
+####################
+
+# install cargo-llvm-cov if not already installed and ensure nightly toolchain + llvm-tools are available
+ensure-cargo-llvm-cov:
+    command -v cargo-llvm-cov >/dev/null 2>&1 || cargo install cargo-llvm-cov --locked
+    rustup toolchain install nightly 2>/dev/null
+    rustup component add llvm-tools --toolchain nightly 2>/dev/null
+
+# host-side packages to collect coverage for (guest/no_std crates are excluded because they
+# define #[panic_handler] and cannot be compiled for the host target under coverage instrumentation)
+coverage-packages := "-p hyperlight-common -p hyperlight-host -p hyperlight-testing -p hyperlight-component-util -p hyperlight-component-macro"
+
+# run all tests and examples with coverage instrumentation, collecting profdata without
+# generating a report. Mirrors test-like-ci + run-examples-like-ci to exercise all code paths
+# across all feature combinations. Uses nightly for branch coverage.
+#
+# Uses the show-env approach so that cargo produces separate binaries per feature combination.
+# This avoids "mismatched data" warnings that occur when `cargo llvm-cov --no-report` recompiles
+# a crate with different features, overwriting the previous binary and orphaning its profraw data.
+#
+# (run `just guests` first to build guest binaries)
+coverage-run hypervisor="kvm": ensure-cargo-llvm-cov
+    #!/usr/bin/env bash
+    set -euo pipefail
+
+    # Set up coverage instrumentation environment variables (RUSTFLAGS, LLVM_PROFILE_FILE, etc.)
+    # and clean previous artifacts. All subsequent cargo commands inherit instrumentation.
+    source <(cargo +nightly llvm-cov show-env --export-prefix --branch)
+    cargo +nightly llvm-cov clean --workspace
+
+    # tests with default features (all drivers; skip stress tests — too slow under instrumentation)
+    cargo +nightly test {{ coverage-packages }} --tests -- --skip stress_test
+
+    # tests with single driver + build-metadata
+    cargo +nightly test {{ coverage-packages }} --no-default-features --features build-metadata,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- --skip stress_test
+
+    # isolated tests (require running separately due to global state)
+    cargo +nightly test -p hyperlight-host --lib -- sandbox::uninitialized::tests::test_log_trace --exact --ignored
+    cargo +nightly test -p hyperlight-host --lib -- sandbox::outb::tests::test_log_outb_log --exact --ignored
+    cargo +nightly test -p hyperlight-host --test integration_test -- log_message --exact --ignored
+    cargo +nightly test -p hyperlight-host --no-default-features -F function_call_metrics,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --lib -- metrics::tests::test_metrics_are_emitted --exact
+
+    # integration test with executable_heap feature
+    cargo +nightly test {{ coverage-packages }} --no-default-features -F executable_heap,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --test integration_test -- execute_on_heap
+
+    # crashdump tests + example
+    cargo +nightly test {{ coverage-packages }} --no-default-features --features crashdump,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- test_crashdump
+    cargo +nightly run --no-default-features --features crashdump,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --example crashdump
+
+    # tracing feature tests (host-side only; hyperlight-guest-tracing is no_std)
+    cargo +nightly test -p hyperlight-common --no-default-features --features trace_guest --tests -- --skip stress_test
+    cargo +nightly test -p hyperlight-host --no-default-features --features trace_guest,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- --skip stress_test
+
+    # examples: metrics, logging, tracing
+    cargo +nightly run --example metrics
+    cargo +nightly run --no-default-features -F function_call_metrics,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --example metrics
+    cargo +nightly run --example logging
+    cargo +nightly run --example tracing
+    cargo +nightly run --no-default-features -F function_call_metrics,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --example tracing
+    cargo +nightly test --no-default-features -F gdb,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --example guest-debugging
+
+# generate a text coverage summary to stdout
+# for this to work you need to run `coverage-run hypervisor` beforehand
+coverage hypervisor="kvm":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    source <(cargo +nightly llvm-cov show-env --export-prefix --branch)
+    cargo +nightly llvm-cov report
+
+# generate an HTML coverage report to target/coverage/html/
+# for this to work you need to run `coverage-run hypervisor` beforehand
+coverage-html hypervisor="kvm":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    source <(cargo +nightly llvm-cov show-env --export-prefix --branch)
+    cargo +nightly llvm-cov report --html --output-dir target/coverage/html
+
+# generate LCOV coverage output to target/coverage/lcov.info
+# for this to work you need to run `coverage-run hypervisor` beforehand
+coverage-lcov hypervisor="kvm":
+    #!/usr/bin/env bash
+    set -euo pipefail
+    source <(cargo +nightly llvm-cov show-env --export-prefix --branch)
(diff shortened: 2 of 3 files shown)


docs/how-to-run-coverage.md
```diff
@@ -0,0 +1,83 @@
...
+| `just coverage-lcov` | `target/coverage/lcov.info` | LCOV format for tooling |
+| `just coverage-ci <hypervisor>` | All of the above | CI recipe: HTML + LCOV + text summary |
+
+## CI Integration
+
+Coverage runs automatically on a **weekly schedule** (every Monday at 06:00 UTC) via the `Coverage.yml` workflow. It can also be triggered manually from the Actions tab using `workflow_dispatch`. The workflow runs on a single configuration (kvm/amd) to keep resource usage reasonable. It:
+
+1. Builds guest binaries (`just guests`)
+2. Installs `cargo-llvm-cov`
+3. Runs `just coverage-ci kvm` — this mirrors `test-like-ci` by running multiple test phases with different feature combinations and merging the results into a single coverage report
```
jsturtevant: should we cover other hypervisors?

Justfile
```diff
@@ -446,6 +446,57 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    cargo llvm-cov clean --workspace
+
+    @# Phase 1: default features (all drivers)
+    cargo llvm-cov {{ coverage-packages }} --no-report
+
+    @# Phase 2: single driver + build-metadata (matches test-like-ci single-driver pass)
+    cargo llvm-cov {{ coverage-packages }} --no-default-features --features build-metadata,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --no-report
+
+    @# Phase 3: crashdump feature tests
+    cargo llvm-cov {{ coverage-packages }} --no-default-features --features crashdump,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --no-report -- test_crashdump
```
jsturtevant: some of the examples have tests too, like `--example crashdump` should those be included?  Do examples count as coverage?

jsturtevant (commented):
Its nice to see a report. Do you have thoughts on how do we ensure we are looking at this report and not just another artifact?  Do we post an issue on run? post in release?  Review at community meetings? 

I ran it locally and I am seeing 70.42%, which quite high!  Its nice to see this. I am not to worried about minor changes but would like to see it trending even or even improving over time.

Justfile
ludfjig: I would say probably not

Justfile
```diff
@@ -449,6 +446,57 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    cargo llvm-cov {{ coverage-packages }}
+
+# generate an HTML coverage report to target/coverage/html/ (run `just guests` first to build guest binaries)
+coverage-html: ensure-cargo-llvm-cov
+    cargo llvm-cov {{ coverage-packages }} --html --output-dir target/coverage/html
+
+# generate LCOV coverage output to target/coverage/lcov.info (run `just guests` first to build guest binaries)
+coverage-lcov: ensure-cargo-llvm-cov
+    mkdir -p target/coverage
+    cargo llvm-cov {{ coverage-packages }} --lcov --output-path target/coverage/lcov.info
```
ludfjig: Does running these 3 in sequence re-run the tests 3 times? It would be nice if we could run once, and then you can choose what format to generate the report after (maybe separated into 2 different just-recipes)

.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,94 @@
...
+        with:
+          name: coverage-html-${{ matrix.hypervisor }}-${{ matrix.cpu }}
+          path: target/coverage/html/
+
+      - name: Upload LCOV coverage report
+        if: always()
+        uses: actions/upload-artifact@v7
+        with:
+          name: coverage-lcov-${{ matrix.hypervisor }}-${{ matrix.cpu }}
+          path: target/coverage/lcov.info
```
ludfjig, line 92: I think there's an option to fail the workflow if the file is not found which could be useful. Should we also using existing framework/script to create an automatic issue if this workflow fails (like we do in another workflow)?

.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,94 @@
...
+          shared-key: "${{ runner.os }}-debug"
+          cache-on-failure: "true"
+
+      - name: Build guest binaries
+        run: just guests
+
+      - name: Install cargo-llvm-cov
+        run: |
+          cargo install cargo-llvm-cov
+          rustup component add llvm-tools
```
ludfjig: technically this shouldn't be needed since ensure-tools will be called below I think. Do we still want it? (I am indifferent). Also, have you considred https://github.com/taiki-e/cargo-llvm-cov#on-github-actions ?

Justfile
```diff
@@ -449,6 +446,57 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    command -v cargo-llvm-cov >/dev/null 2>&1 || cargo install cargo-llvm-cov
+    rustup component add llvm-tools 2>/dev/null || true
+
+# host-side packages to collect coverage for (guest/no_std crates are excluded because they
+# define #[panic_handler] and cannot be compiled for the host target under coverage instrumentation)
+coverage-packages := "-p hyperlight-common -p hyperlight-host -p hyperlight-testing -p hyperlight-component-util -p hyperlight-component-macro"
+
+# generate a text coverage summary to stdout (run `just guests` first to build guest binaries)
+coverage: ensure-cargo-llvm-cov
+    cargo llvm-cov {{ coverage-packages }}
```
ludfjig: What does llvm-cov do by default? Run all unit and integration tests?

dblnz:
> This is awesome, I also tried it locally and works great. I only have one question: is this the recommended way to generate coverage reports (using cargo-llvm-cov), or are there alternative ways? I'm only asking because I am not familiar with this area.

From the research I've done, there are two other alternatives to `llvm-cov`: `tarpaulin` and `grcov`.
Tarpaulin uses `ptrace` by default (only on Linux), can be configured to use LLVM, but it has some limitations (coverage data not returned when non-zero exit, some areas of thread unsafety).
`Grcov` could also be a good alternative for us, it supports multiple aggregation sources, but it's slightly more complicated to set up. Given that I don't think we need that much flexibility at the moment, I haven't used this one.
I chose `llvm-cov` due to its accuracy (uses the llvm backend), support across OS (we could enable windows also) and its relatively simple setup.

Justfile
dblnz: It runs the tests (cargo test)

.github/workflows/Coverage.yml
dblnz: Haven't considered the gh action, good suggestion.
But I think I'll remove this and only rely on the `ensure-` recipe

Justfile
dblnz: I am not sure.

As they exercise various areas of our code, I think we should run everything we have (tests + examples) and aggregate them in a single report so that we have a broad view of the degree of coverage we have.

docs/how-to-run-coverage.md
dblnz: We could do this. I'll see if I include it in this PR or in a follow-up one, but I think it only needs some additional runners in the workflow to do that. I'll check it out

dblnz:
> Its nice to see a report. Do you have thoughts on how do we ensure we are looking at this report and not just another artifact? Do we post an issue on run? post in release? Review at community meetings?
> 
> I ran it locally and I am seeing 70.42%, which quite high! Its nice to see this. I am not to worried about minor changes but would like to see it trending even or even improving over time.

About this, I think we should definitely have this in mind when we do changes. Although, I wouldn't condition any PR on it as it takes a lot to run (same time the PR validations take).
Considering we are still making fundamental changes in the codebase, this metric could change a lot.

So maybe start with having this info as part of the release artifacts is a good start? Then we could see if we need to discuss it more often.

One nice to have thing I was thinking about is to try and correlate the coverage output with the changes made in a PR so that we could have a check saying we are only testing `x` percent of the newly introduced code. I'll hack at this once I get some more time

Justfile
dblnz: Yeah, there's an option to select the type of report at the end of the run. Let me change this to use that

dblnz:
You can check out how the summary looks like at: https://github.com/hyperlight-dev/hyperlight/actions/runs/22959438794?pr=1301

dblnz:
I am also looking at enabling the branch coverage report (it needs nightly).

jsturtevant:
>So maybe start with having this info as part of the release artifacts is a good start? Then we could see if we need to discuss it more often.

Another thought is if it drops by %5 to create an issue (we have some machinery that does this as part of other ci workflows).  that would inform us that we've been going down.

Otherwise this looks great and is a good first start.  Feel free to move forward with this and do follow ups.

dblnz:
> > So maybe start with having this info as part of the release artifacts is a good start? Then we could see if we need to discuss it more often.
> 
> Another thought is if it drops by %5 to create an issue (we have some machinery that does this as part of other ci workflows). that would inform us that we've been going down.
> 
> Otherwise this looks great and is a good first start. Feel free to move forward with this and do follow ups.

Now that I'm back, I'll move this forward by addressing all the comments and suggestions.
Thanks for having a look at this!

docs/how-to-run-coverage.md
dblnz: I'll leave the other hypervisors for later

Justfile
dblnz: I added the examples to be run as tests

Justfile
dblnz: Fixed

.github/workflows/Coverage.yml
dblnz: Removed it from here

.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,111 @@
...
+      - name: Install nightly toolchain
+        run: |
+          rustup toolchain install nightly
+          rustup component add llvm-tools --toolchain nightly
+
+      - name: Generate coverage report
+        run: just coverage-ci ${{ matrix.hypervisor }}
+
+      - name: Coverage summary
+        if: always()
```
ludfjig: why do some of these have `if: always()`?

dblnz: Oh, my bad, leftover things. I'll remove

Justfile
```diff
@@ -459,6 +459,84 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    mkdir -p target/coverage
+    cargo +nightly llvm-cov report --lcov --output-path target/coverage/lcov.info
+
+# generate all coverage reports for CI: HTML + LCOV + text summary.
+# (run `just guests` first to build guest binaries)
+coverage-ci hypervisor="kvm": (coverage-run hypervisor)
+    mkdir -p target/coverage
+    cargo +nightly llvm-cov report --html --output-dir target/coverage/html
+    cargo +nightly llvm-cov report --lcov --output-path target/coverage/lcov.info
+    cargo +nightly llvm-cov report | tee target/coverage/summary.txt
```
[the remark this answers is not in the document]
dblnz: This is not needed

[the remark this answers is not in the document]
dblnz: This is not needed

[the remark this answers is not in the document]
dblnz: This is not needed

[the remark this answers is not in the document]
dblnz: This is not needed

Justfile
```diff
@@ -459,6 +459,84 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
 
+####################
+### COVERAGE #######
+####################
+
+# install cargo-llvm-cov if not already installed and ensure nightly toolchain + llvm-tools are available
+ensure-cargo-llvm-cov:
+    command -v cargo-llvm-cov >/dev/null 2>&1 || cargo install cargo-llvm-cov
+    rustup toolchain install nightly 2>/dev/null || true
+    rustup component add llvm-tools --toolchain nightly 2>/dev/null || true
```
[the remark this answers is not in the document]
dblnz: Removed

Justfile
```diff
@@ -459,6 +459,84 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
 build-trace-fuzzers:
     cargo +nightly fuzz build fuzz_guest_trace --features trace
 
+####################
+### COVERAGE #######
+####################
+
+# install cargo-llvm-cov if not already installed and ensure nightly toolchain + llvm-tools are available
+ensure-cargo-llvm-cov:
+    command -v cargo-llvm-cov >/dev/null 2>&1 || cargo install cargo-llvm-cov
```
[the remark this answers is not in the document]
dblnz: Fixed

Justfile
```diff
@@ -459,6 +459,84 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+coverage hypervisor="kvm":
+    cargo +nightly llvm-cov report
+
+# generate an HTML coverage report to target/coverage/html/
+# for this to work you need to run `coverage-run hypervisor` before hand
+coverage-html hypervisor="kvm":
+    cargo +nightly llvm-cov report --html --output-dir target/coverage/html
+
+# generate LCOV coverage output to target/coverage/lcov.info
+# for this to work you need to run `coverage-run hypervisor` before hand
```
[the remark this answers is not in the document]
dblnz: Fixed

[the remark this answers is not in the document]
dblnz: Fixed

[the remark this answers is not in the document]
dblnz: Fixed

docs/how-to-run-coverage.md
```diff
@@ -0,0 +1,104 @@
...
+
+## Available Recipes
+
+| Recipe | Output | Description |
+|---|---|---|
+| `just coverage-run` | profiling data | Runs tests with coverage instrumentation (must be run first) |
+| `just coverage` | stdout | Text summary of line coverage |
+| `just coverage-html` | `target/coverage/html/` | HTML report for browsing |
+| `just coverage-lcov` | `target/coverage/lcov.info` | LCOV format for tooling |
+| `just coverage-ci <hypervisor>` | All of the above | CI recipe: runs tests + generates HTML + LCOV + text summary |
```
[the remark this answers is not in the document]
dblnz: Doesn't start with `| |`. Also, I see it rendering correctly

.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,111 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Weekly Coverage
+
+on:
+  pull_request:
+      paths:
+        - .github/workflows/Coverage.yml
```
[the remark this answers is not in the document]
dblnz: Fixed

Justfile
```diff
@@ -459,6 +459,105 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    cargo +nightly test -p hyperlight-host --lib -- sandbox::uninitialized::tests::test_log_trace --exact --ignored
+    cargo +nightly test -p hyperlight-host --lib -- sandbox::outb::tests::test_log_outb_log --exact --ignored
+    cargo +nightly test -p hyperlight-host --test integration_test -- log_message --exact --ignored
+    cargo +nightly test -p hyperlight-host --no-default-features -F function_call_metrics,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --lib -- metrics::tests::test_metrics_are_emitted --exact
+
+    # integration test with executable_heap feature
+    cargo +nightly test {{ coverage-packages }} --no-default-features -F executable_heap,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --test integration_test -- execute_on_heap
+
+    # crashdump tests + example
+    cargo +nightly test {{ coverage-packages }} --no-default-features --features crashdump,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- test_crashdump
```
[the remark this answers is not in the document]
dblnz: This is from the cargo docs:
```
--tests

    Test all targets that have the test = true manifest flag set. By default this includes the library and binaries built as unittests, and integration tests. Be aware that this will also build any required dependencies, so the lib target may be built twice (once as a unittest, and once as a dependency for binaries, integration tests, etc.). Targets may be enabled or disabled by setting the test flag in the manifest settings for the target.
```
So the command with `--tests` will run everything (which includes the unit-tests)

Justfile
```diff
@@ -459,6 +459,105 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    # Set up coverage instrumentation environment variables (RUSTFLAGS, LLVM_PROFILE_FILE, etc.)
+    # and clean previous artifacts. All subsequent cargo commands inherit instrumentation.
+    source <(cargo +nightly llvm-cov show-env --export-prefix --branch)
+    cargo +nightly llvm-cov clean --workspace
+
+    # tests with default features (all drivers; skip stress tests — too slow under instrumentation)
+    cargo +nightly test {{ coverage-packages }} --tests -- --skip stress_test
+
+    # tests with single driver + build-metadata
+    cargo +nightly test {{ coverage-packages }} --no-default-features --features build-metadata,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- --skip stress_test
```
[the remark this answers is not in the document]
dblnz: This is from the cargo docs:
```
--tests

    Test all targets that have the test = true manifest flag set. By default this includes the library and binaries built as unittests, and integration tests. Be aware that this will also build any required dependencies, so the lib target may be built twice (once as a unittest, and once as a dependency for binaries, integration tests, etc.). Targets may be enabled or disabled by setting the test flag in the manifest settings for the target.
```
So the command with `--tests` will run everything (which includes the unit-tests)

Justfile
```diff
@@ -459,6 +459,105 @@ fuzz-trace-timed max_time fuzz-target="fuzz_guest_trace":
...
+    # integration test with executable_heap feature
+    cargo +nightly test {{ coverage-packages }} --no-default-features -F executable_heap,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --test integration_test -- execute_on_heap
+
+    # crashdump tests + example
+    cargo +nightly test {{ coverage-packages }} --no-default-features --features crashdump,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- test_crashdump
+    cargo +nightly run --no-default-features --features crashdump,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --example crashdump
+
+    # tracing feature tests (host-side only; hyperlight-guest-tracing is no_std)
+    cargo +nightly test -p hyperlight-common --no-default-features --features trace_guest --tests -- --skip stress_test
+    cargo +nightly test -p hyperlight-host --no-default-features --features trace_guest,{{ if hypervisor == "mshv3" { "mshv3" } else { "kvm" } }} --tests -- --skip stress_test
```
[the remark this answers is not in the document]
dblnz: This is from the cargo docs:
```
--tests

    Test all targets that have the test = true manifest flag set. By default this includes the library and binaries built as unittests, and integration tests. Be aware that this will also build any required dependencies, so the lib target may be built twice (once as a unittest, and once as a dependency for binaries, integration tests, etc.). Targets may be enabled or disabled by setting the test flag in the manifest settings for the target.
```
So the command with `--tests` will run everything (which includes the unit-tests)

docs/how-to-run-coverage.md
```diff
@@ -0,0 +1,111 @@
+# How to Run Coverage
+
+This guide explains how to generate code coverage reports for Hyperlight.
+
+## Prerequisites
+
+- A working Rust toolchain
+- Rust nightly toolchain (required for branch coverage; installed automatically by the `just` recipes)
+- [cargo-llvm-cov](https://github.com/taiki-e/cargo-llvm-cov) (installed automatically by the `just` recipes)
+- Guest binaries must be built first: `just guests`
```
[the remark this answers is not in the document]
dblnz: Coverage is runnable on windows also, but I only configured it to run on kvm linux in CI.
Specifying here, that the workflow only runs on linux, might get desynchronized once we modify the workflow.

.github/workflows/Coverage.yml
```diff
@@ -0,0 +1,108 @@
...
+        hypervisor: [kvm]
+        cpu: [amd]
+    runs-on: ${{ fromJson(
+        format('["self-hosted", "Linux", "X64", "1ES.Pool=hld-{0}-{1}"]',
+          matrix.hypervisor,
+          matrix.cpu)) }}
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: hyperlight-dev/ci-setup-workflow@v1.8.0
```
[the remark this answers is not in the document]
dblnz: fixed

dblnz:
@jsturtevant @ludfjig do you have anything else to add to this PR?

dblnz:
> Nice. What are your plans on windows support?

I wanted to see how it works for a couple of weeks, then I was thinking maybe I can run the windows coverage also and merge together with linux so we have a complete report

Issue #1261, opened 2026-02-25
Guest Copy on Write Followups


jsturtevant:
We have a few follow ups form Cow work in #1200 

- [x] Docs
- [ ] HL wasm updates

PR #1305, opened 2026-03-12
CoW Developer docs

jsturtevant:
Part of #1261

Docs updated from @syntactically's original proposal https://github.com/hyperlight-dev/hyperlight/commit/3d5940080a75e3e7924957431db79e688afb4a66

### DIFF
--- a/docs/paging-development-notes.md
+++ b/docs/paging-development-notes.md
@@ -91,7 +91,7 @@ on amd64 as well.
 ## Top-of-scratch metadata layout
 
 The top of the scratch region contains structured metadata at fixed
-offsets such as the scratch size, allocator state and where the excpetions starts.
+offsets such as the scratch size, allocator state and where the exceptions starts.
 These offsets are defined as `SCRATCH_TOP_*` constants in
 [`hyperlight_common::layout`](../src/hyperlight_common/src/layout.rs), which has detailed comments on each
 field.


docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
 which it is being run, and manages its own copy-on-write.
 
 Because of this, there are two very fundamental regions of the guest
-physical address space, which are always populated: one, at the very
-bottom of memory, is a (hypervisor-enforced) readonly mapping of the
-base snapshot from which this guest is being evolved. Another, at the
-top of memory, is simply a large bag of blank pages: scratch memory
-into which this VM can write.
+physical address space, which are always populated: one, near the
+bottom of memory (starting at GPA `0x1000`), is a
```
syntactically: The specific base address is sort of a nasty implementation detail that might go away at some point anyway. Morally, it doesn't matter where it is and "at the bottom" is really just a general location. I would rather not have this in the high-level docs because it feels like noise to me.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+(hypervisor-enforced) readonly mapping of the base snapshot from which
+this guest is being evolved. Another, at the top of memory, is simply
+a large bag of blank pages: scratch memory into which this VM can
+write.
+
+```
+  Guest Physical Address Space (GPA)
+
+  +-------------------------------+ MAX_GPA
```
syntactically: I think this is duplicating a diagram from src/hyperlight_host/src/mem/layout.rs. Maybe we can merge them. I'm ambivalent about whether the merged diagram ends up here or there.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+  |    Exn Stack, Bookkeeping     |
+  |    (scratch size, allocator   |
```
syntactically: Rather than explicitly listing metadata/bookkeeping items, I would just in text reference the layout.rs files (primarily the hyperlight_common one) that have fairly detailed comments on each item along with its offsets.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+  |     state, reserved PT slot)  |
+  +-------------------------------+
+  |      Free Scratch Memory      |
+  +-------------------------------+
+  |         Output Data           |
```
syntactically: Similarly, the io data being here right now is sort of a stupid hack because we don't have the better phys alloc / buffer management from #1112. I would like to avoid having that in this diagram if possible and we can talk about the fact that that is the stopgap in the section below which I see was not fully written in the original commit (sorry).

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+
+    Snapshot page at GPA 0x5000 is untouched.
+```
+
+The page table
 entries to do this will likely need to be copied themselves, and so a
 ready supply of already-mapped scratch pages to use for replacement
-page tables is needed. Currently, the guest accomplishes this by
-keeping an identity mapping of the entire scratch memory around.
+page tables is set up by the Host. The guest keeps a mapping of the entire scratch
```
syntactically: I don't fully understand what this change is trying to say. I prefer the original wording I think.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+physical memory into virtual memory at a fixed offset
+(`scratch_base_gva - scratch_base_gpa`), so that any scratch physical
```
syntactically: This isn't really an offset (rather an address), and the number calculated in this expression doesn't seem right.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+address can be accessed by adding this offset.
 
 The host and the guest need to agree on the location of this mapping,
 so that (a) the host can create it when first setting up a blank guest
 and (b) the host can ignore it when taking a snapshot (see below).
 
-Currently, the host always creates the scratch map at the top of
-virtual memory.  In the future, we may add support for a guest to
```
syntactically: I'm not sure what this rephrasing is trying to accomplish? I think I prefer the original.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+The host creates the scratch map at the top of virtual memory
+(`MAX_GVA - scratch_size + 1`) and at the top of physical memory
+(`MAX_GPA - scratch_size + 1`). In the future, we may add support for a guest to
```
syntactically: The _scratch map_ is just the _mapping into VA space of the scratch region_, so it is not correct to say that it is "create[d]...at the top of physical memory"---it is created nowhere in _physical_ memory, because it exists only in virtual memory.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
 
-Do we actually need to have a physmap type mapping of the entire
-snapshot memory? We only really use it when copying from it in which
-case we ought to have the VA that we need to copy from already. There
-is one major exception to this, which is the page tables
-themselves. The page tables themselves must be mapped at some VA so
-that we can copy them.
+The snapshot page tables must be mapped at some virtual address so
+that the guest can read and copy them during CoW operations.
+Today, the host simply
```
syntactically: This is kind of an implementation detail of amd64 being annoying. I think in the near future we will get rid of this on aarch64 and with some minor hv changes we can get rid of it in some cases on amd64. I would rather keep this section explaining that this mapping of the pt region from the snapshot is necessary, and then add another (sub)section explaining that this is not the approach taken on amd64 at the moment. (Thank you for rewriting the section from the original question/points that were more temporary notes that I had in here).

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+
+These offsets are defined as `SCRATCH_TOP_*` constants in
+`hyperlight_common::layout`.
 
 ## The physical page allocator
 
 The host needs to be able to reset the state of the physical page
-allocator when resuming from a snapshot. Currently, we use a simple
-bump allocator as a physical page allocator, with no support for free,
+allocator when resuming from a snapshot. We use a simple bump
```
syntactically: Please retain the "currently" language to imply that this decision is not set in stone (it isn't).

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+allocator as a physical page allocator, with no support for free,
 since pages not in use will automatically be omitted from a snapshot.
-Therefore, the allocator state is nothing but a single `u64` that
-tracks the address of the first free page. This `u64` will always be
-located at the top of scratch physical memory.
+The allocator state is a single `u64` tracking the address of the
+first free page, located at offset `0x10` from the top of scratch
+(see layout above). The guest advances it atomically via `lock xadd`.
```
syntactically: The specific instruction used here isn't really relevant, and is an amd64-specific implementation detail. Mentioning that it's done via atomic makes sense to me though!

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
 
 ## The guest exception stack
 
 Similarly, the guest needs a stack that is always writable, in order
-to be able to take exceptions to it.  The remainder of the top page of
-the scratch memory is used for this.
+to be able to take exceptions to it. The exception stack begins at
+offset `0x20` from the top of the scratch region (below the metadata
```
syntactically: Again, I would avoid duplicating the actual offset here (too easy to get out of date) and just say below the metadata.

docs/paging-development-notes.md
```diff
@@ -2,66 +2,153 @@
...
+fields described above) and grows downward through the remainder of
+the top page.
```
syntactically: Technically I think it's actually 2 pages used for this now.

docs/paging-development-notes.md
```diff
@@ -90,20 +177,26 @@ calls, i.e. there may be no calls in flight at the time of
...
-virtual address that it wants to have this sort of bookkeeping
-information mapped at, and the snapshot creation process treats that
-address specially writing out a manifest
-
-Q: how do we want to manage buffer
-allocation/freeing/reallocation/etc? Maybe A: for now we will mostly
-ignore because we only need 1-2 buffers inflight at a time. We can
-emulate the current paradigm by recreating a new buffer out of the
-free space in the original buffer on call, etc etc.
+I/O buffers are statically allocated at the bottom of the scratch
```
syntactically: Let's rewrite this to be a little bit more high-level about how buffer management should work, plus details on how it is done right now (I believe there is also some discussion about buffer management in #1112).

docs/paging-development-notes.md
```diff
@@ -90,20 +177,26 @@ calls, i.e. there may be no calls in flight at the time of
...
++-------------------------------------------+
+|           Free Scratch Memory             |
++-------------------------------------------+
+|                Output Data                |
++-------------------------------------------+
+|                Input Data                 |
++-------------------------------------------+ (scratch base)
+```
+
+The minimum scratch size (`min_scratch_size()`) accounts for these
```
syntactically: I'm not sure if this is a useful comment to have in this file. If it's useful to have the details of the min scratch size calculation duplicated in here (my first instinct would be to just link to the layout.rs file that has the calculation explained in detail), it should either be a general statement (e.g. "some architecture-specific processor structures that will be written during initialisation" or explicitly split out for each supported architecture).

docs/paging-development-notes.md
```diff
@@ -113,11 +206,10 @@ which simply map the segments of that ELF to the appropriate places in
 virtual memory.  If the ELF has segments whose virtual addresses
 overlap with the scratch map, an error will be returned.
 
-The initial stack pointer will point to the top of the second-highest
-page of the scratch map, but this should usually be changed by early
-init code in the guest, since it will otherwise be difficult to detect
-collisions between the guest stack and the scratch physical page
-allocator.
+In the current startup path, the host enters the guest with
+`RSP` pointing to the exception stack. Early guest init then
```
syntactically: Please retain the less-architecture-dependent language of "the stack pointer" rather than "rsp", and consider whether or not it makes sense to retain any useful points/implications of the note about exception stack overflows being difficult to detect (thanks for the rest of the rewrite of this section, which is good).

docs/paging-development-notes.md
jsturtevant: removed this diagram and put a pointer to the source code in

docs/paging-development-notes.md
jsturtevant: removed this for a pointer to the src code diagram

docs/paging-development-notes.md
jsturtevant: moved back to the previous version

docs/paging-development-notes.md
jsturtevant: went back to the original

docs/paging-development-notes.md
jsturtevant: got it, I think i got confused here with which was which but make sense now.

docs/paging-development-notes.md
jsturtevant: updated

docs/paging-development-notes.md
jsturtevant: went back to currently.

docs/paging-development-notes.md
jsturtevant: makes sense

docs/paging-development-notes.md
jsturtevant: just left it with "grows downward"

docs/paging-development-notes.md
jsturtevant: updated

docs/paging-development-notes.md
jsturtevant: done

docs/paging-development-notes.md
jsturtevant: makes sense

docs/paging-development-notes.md
jsturtevant: i made it a pointer to the src

docs/paging-development-notes.md
jsturtevant: updated

jsturtevant:
going to merge this for the Release, feel free to open up follow ups.

PR #1306, opened 2026-03-12
Use WHvResetPartition on Windows instead of manual vCPU state reset

ludfjig:
needs #1319 
Replace the manual set_regs/set_debug_regs/reset_xsave calls in reset_vcpu on Windows with a single WHvResetPartition API call. This resets a bunch of state that wasn't reset before, including msrs, apic/lapic, interruptrs, vmcs/b internals, etc.

This will clear all guest visible state, at the expense of a little bit performance (rougly +~50 micro seconds compared to before on my local machine). `WHvResetPartition` is an internal API available from `Win11 21H2`, but qemu already uses it so I assume it's stable enough. `WHvResetPartition` has been shipped since Windows 11 21H2 (original Windows 11 release, October 2021) and Windows Server 2022 (released August 2021).

Our windows 2022 CI can't handle this it seems like, but our windows 2025 works


ludfjig:
closed in favor of #1360

PR #1313, opened 2026-03-16
fix: make map types public

simongdavies:
This pull request fixes visibility in the memory region handling code in `src/hyperlight_host/src/mem/memory_region.rs` for hyperlight-wasm use.

**Visibility improvements:**

* Made the `MemoryRegionType` enum public, allowing it to be used outside of the current crate.
* Made the `SurrogateMapping` enum public, so it can be accessed from other modules or crates.
* Made the `surrogate_mapping` method on `MemoryRegionType` public, enabling external code to determine the surrogate mapping for a given memory region type.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/mod.rs
@@ -31,7 +31,7 @@ pub(crate) mod surrogate_process;
 pub(crate) mod surrogate_process_manager;
 /// Safe wrappers around windows types like `PSTR`
 #[cfg(target_os = "windows")]
-pub(crate) mod wrappers;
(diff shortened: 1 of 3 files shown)


jsturtevant:
Need to make the field on MemoryRegion pub as well:

```
      let rgn = MemoryRegion {
    |                   ------------ in this type
...
127 |             region_type: MemoryRegionType::Heap,
    |             ^^^^^^^^^^^ private field
```

Maybe a integration test that uses `map_region` would be a good addtion?  I should have done this when we ran into it before https://github.com/hyperlight-dev/hyperlight/pull/1293

jsturtevant:
verified the latest changes work in HL-wasm

Issue #1286, opened 2026-03-07
Consider moving the hostfunctions definitions from Hyperlight common


jsturtevant:
https://github.com/hyperlight-dev/hyperlight-wasm/pull/389#discussion_r2890046140

>I guess the only place where these structs cross the host/guest boundary is now hlwasm/hljs downstreams, which makes it feel a little weird that they are still defined in hlcommon. I don't think we need to do anything about this right now since IIRC it's pretty urgent to get this in so we can prepare for release, but perhaps it's worth thinking about...

PR #1314, opened 2026-03-16
Re-export host function types through hyperlight_host

jsturtevant:
fixes: #1286

Host function definition types (HostFunctionDefinition, HostFunctionDetails, ParameterType) are currently only accessible via hyperlight_common, even though they primarily serve the host-side API. This makes it awkward for downstream consumers (e.g., hyperlight-wasm, hyperlight-js) that need to import these types alongside other host constructs.


ludfjig:
What's this needed for? And maybe the PR title could be more accurate

jsturtevant:
> What's this needed for? And maybe the PR title could be more accurate

I added a description, For the title I also made a small update but not sure if there is something clearer.

PR #1315, opened 2026-03-17
CoW follow-up work

syntactically:
This starts to move to actually take advantage of the fact that snapshots are now RO by not copying them all the time (although they still aren't loadable-from-disk). It also includes a few other tweaks to improve snapshot creation.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -47,6 +47,9 @@
                   chmod +x "$ldlld"
                 fi
               '');
+            passthru = (oA.passthru or {}) // {
+              toolchainVersionAttrs = args;
+            };
           })) // {
             targetPlatforms = [ "x86_64-linux" ];
             badTargetPlatforms = [ ];
@@ -93,9 +96,9 @@
               sha256 = "sha256-sqSWJDUxc+zaz1nBWMAJKTAGBuGWP25GCftIOlCEAtA=";
             };
             nightly = {
-              date = "2026-01-19";
+              date = "2026-02-27";
               channel = "nightly";
-              sha256 = "sha256-Ye65U/qzilPLte800N5oxFOY96shgG8bST8dbrF6Qh0=";
+              sha256 = "sha256-5twI9QsrPl0ryOZ4POGYAivSeI08jgmWnv0wVvzbjcE=";
             };
             "1.89" = {
               date = "2025-08-07";
@@ -191,8 +194,10 @@
           # nothing as well.
           rustup-like-wrapper = name: pkgs.writeShellScriptBin name
             (let
-              clause = name: toolchain:
-                "+${name}) base=\"${toolchain.rust}\"; shift 1; ;;";
+              clause = name: toolchain: ''
+                +${name}) base="${toolchain.rust}"; shift 1; ;;
+                +${name}-${toolchain.rust.toolchainVersionAttrs.date}) base="${toolchain.rust}"; shift 1; ;;
+              '';
               clauses = lib.strings.concatStringsSep "\n"
                 (lib.mapAttrsToList clause toolchains);
             in ''


src/hyperlight_host/src/hypervisor/gdb/mod.rs
```diff
@@ -95,18 +95,11 @@ pub enum DebugMemoryAccessError {
     LockFailed(&'static str, u32, String),
     #[error("Failed to translate guest address {0:#x}")]
     TranslateGuestAddress(u64),
+    #[error("Failed to write to read-only region")]
+    WriteToReadOnly(),
```
ludfjig: nit: unnecessary parens

src/hyperlight_host/src/sandbox/snapshot.rs
```diff
@@ -732,19 +657,16 @@ mod tests {
...
         mgr.shared_mem
-            .with_exclusivity(|e| assert_eq!(&e.as_slice()[0..pattern_a.len()], &pattern_a[..]))
+            .with_contents(|contents| assert_eq!(&contents[0..pattern_a.len()], &pattern_a[..]))
             .unwrap();
 
         // Restore snapshot B
         mgr.restore_snapshot(&snapshot_b).unwrap();
         mgr.shared_mem
-            .with_exclusivity(|e| assert_eq!(&e.as_slice()[0..pattern_b.len()], &pattern_b[..]))
+            .with_contents(|contents| assert_eq!(&contents[0..pattern_a.len()], &pattern_b[..]))
```
ludfjig: nit: I think this should be `pattern_b.len()`

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -438,16 +423,18 @@ impl SandboxMemoryManager<HostSharedMemory> {
     pub(crate) fn restore_snapshot(
         &mut self,
         snapshot: &Snapshot,
-    ) -> Result<(Option<GuestSharedMemory>, Option<GuestSharedMemory>)> {
-        let gsnapshot = if self.shared_mem.mem_size() == snapshot.mem_size() {
+    ) -> Result<(
+        Option<SnapshotSharedMemory<GuestSharedMemory>>,
+        Option<GuestSharedMemory>,
+    )> {
+        let gsnapshot = if *snapshot.memory() == self.shared_mem {
```
ludfjig, line 494: Would you mind adding a small comment about what this means semantically, and/or a comment in the caller of `restore_snapshot` maybe as well?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -108,6 +230,9 @@ pub(crate) struct SandboxMemoryLayout {
     // The size of the scratch region in physical memory; note that
     // this will appear under the top of physical memory.
     scratch_size: usize,
+    // The size of the scratch region in physical memory; note that
+    // this will appear somewhere near the base of physical memory.
+    snapshot_size: usize,
 }
```
ludfjig: I think this comment got incorrectly copy-pasted.

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -68,16 +68,138 @@ use tracing::{Span, instrument};
...
+        let ss: &[u8] = &self.as_ref()[offset..];
+        let end = offset + slice.len();
+        if end > ss.len() {
+            return Err(new_error!(
+                "Attempt to read up to {} in memory of size {}",
+                offset + slice.len(),
+                self.as_ref().len()
+            ));
+        }
+        slice.copy_from_slice(&ss[offset..end]);
```
ludfjig, line 175: i think `ss` already had `offset` applied, and this applies it again. Should it be somethign like
```rust
let ss: &[u8] = &self.as_ref()[offset..];
if slice.len() > ss.len() { ... }
slice.copy_from_slice(&ss[..slice.len()]);
```
?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1990,3 +1989,86 @@ mod tests {
...
+/// A ReadonlySharedMemory is a different kind of shared memory,
+/// separate from the exclusive/host/guest lifecycle, used to
+/// represent read-only mappings of snapshot pages into the guest
+/// efficiently.
+#[derive(Clone, Debug)]
+pub struct ReadonlySharedMemory {
+    region: Arc<HostMapping>,
+}
+unsafe impl Send for ReadonlySharedMemory {}
+unsafe impl Sync for ReadonlySharedMemory {}
```
ludfjig, line 2021: could we have safety comments for these 2?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -68,16 +68,138 @@ use tracing::{Span, instrument};
...
+            BaseGpaRegion::Snapshot(sn) => sn,
+            BaseGpaRegion::Scratch(sc) => sc,
+            BaseGpaRegion::Mmap(r) => unsafe {
+                #[allow(clippy::useless_conversion)]
+                let host_region_base: usize = r.host_region.start.into();
+                #[allow(clippy::useless_conversion)]
+                let host_region_end: usize = r.host_region.end.into();
+                let len = host_region_end - host_region_base;
+                std::slice::from_raw_parts(host_region_base as *const u8, len)
+            },
```
ludfjig, line 129: Could we have a safety comment here too

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -525,68 +659,70 @@ impl SandboxMemoryLayout {
...
     /// from this function.
     #[instrument(err(Debug), skip_all, parent = Span::current(), level= "Trace")]
-    pub(crate) fn write(
-        &self,
-        shared_mem: &mut ExclusiveSharedMemory,
-        guest_offset: usize,
-        //TODO: Unused remove
-        _size: usize,
-    ) -> Result<()> {
+    pub(crate) fn write_peb(&self, mem: &mut [u8]) -> Result<()> {
```
ludfjig, line 734: nit: doc comment still mentions shared_mem

ludfjig (commented):
LGTM! Note I notcied `cargo test --features gdb -p hyperlight-host -- hypervisor::gdb::tests::mem_access_tests::test_mem_access_write` fails, are we fine with this? Unsure if it fails on `main`.

Btw this is awesome
```
snapshots/restore/large time:   [56.654 µs 58.220 µs 60.077 µs]
                        change: [−99.864% −99.829% −99.779%] (p = 0.00 < 0.05)
                        Performance has improved.
```

src/hyperlight_host/src/mem/layout.rs
ludfjig: btw does `snapshot_size + scratch_size` account for the size of the entire sandbox now?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -65,6 +65,7 @@ fn mapping_kind_to_flags(kind: &MappingKind) -> (MemoryRegionFlags, MemoryRegion
             }
             (flags, MemoryRegionType::Scratch)
         }
+        MappingKind::Unmapped => (MemoryRegionFlags::empty(), MemoryRegionType::Snapshot),
```
jsturtevant, line 71: is snapshot the right type here?  if its not mapped is it a snapshot?

src/hyperlight_host/src/sandbox/snapshot.rs
```diff
@@ -287,32 +278,16 @@ fn filtered_mappings<'a>(
...
-    let (mem, off) = access_gpa(snap, scratch, scratch_size, gpa)?;
-    if off + PAGE_SIZE <= mem.as_slice().len() {
-        Some(&mem.as_slice()[off..off + PAGE_SIZE])
-    } else {
-        None
-    }
+    let resolved = layout
+        .resolve_gpa(gpa, regions)?
+        .with_memories(snap, scratch);
+    Some(&resolved.as_ref()[0..PAGE_SIZE])
```
jsturtevant: do we need the bounds check here?   the previous code had: 

if off + PAGE_SIZE <= mem.as_slice().len() {

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -597,6 +733,37 @@ impl SandboxMemoryLayout {
...
+            });
+        } else if gpa >= SandboxMemoryLayout::BASE_ADDRESS as u64
+            && gpa < SandboxMemoryLayout::BASE_ADDRESS as u64 + self.snapshot_size as u64
+        {
+            return Some(ResolvedGpa {
+                offset: gpa as usize - SandboxMemoryLayout::BASE_ADDRESS,
+                base: BaseGpaRegion::Snapshot(()),
+            });
+        }
+        for rgn in mmap_regions {
```
jsturtevant, line 844: just checking here, this looks like it got moved from snapshot, there it was mmap before scratch/base, here we scan scratch/base first. Is that an intentional change?

syntactically: They really shouldn't overlap (I think the HV would error if we tried to make overlapping ones) so it doesn't really matter. The choice before was basically arbitrary and after rewriting the code this way I think it's a little cleaner to focus on the "main cases" like this.

src/hyperlight_host/src/mem/mgr.rs
syntactically: This won't ever be actually used in practice, since `virt_to_phys` will simply not return a mapping at all for unmapped areas. I think it's fine to just leave this case returning sensible-ish defaults so that we don't have to either panic, introduce fallibility to the function for no real reason, or do a bigger refactor to split the MappingKinds used in map vs virt_to_phys.

src/hyperlight_host/src/mem/layout.rs
syntactically: It's not the best tbh but it matches what was happening (without comments) in several other places that was just consolidated here. I added a comment with some thin justification. I would like to try to improve more safety stuff around the mmap'd MemoryRegions at some point, but not mixed in with all this.

src/hyperlight_host/src/mem/mgr.rs
syntactically: Added a little comment; hopefully it's along the lines of what you were looking for?

src/hyperlight_host/src/sandbox/snapshot.rs
syntactically: Fixed

src/hyperlight_host/src/hypervisor/gdb/mod.rs
syntactically: Fixed

src/hyperlight_host/src/mem/layout.rs
syntactically: Fixed the comment to not refer to scratch, thanks.

I'm not sure that it really makes sense to talk about "the size of a sandbox" anymore---the scratch, snapshot, and mmap regions are all very different things (well the mmap regions are a lot like the snapshot) with different properties around e.g. how much they can be shared, etc.

src/hyperlight_host/src/sandbox/snapshot.rs
syntactically: Good catch---I thought this bounds check was moved into as_ref(), but it doesn't get the PAGE_SIZE upper end of it (and actually the original check got lost in rebases...)

src/hyperlight_host/src/mem/layout.rs
syntactically: Fixed

src/hyperlight_host/src/mem/shared_mem.rs
syntactically: Added some comments along the same lines of what we have on the other SharedMemories.

src/hyperlight_host/src/mem/layout.rs
syntactically: Thanks, fixed I think (although the other direction, so that the bounds check is more useful)

syntactically:
> LGTM! Note I notcied `cargo test --features gdb -p hyperlight-host -- hypervisor::gdb::tests::mem_access_tests::test_mem_access_write` fails, are we fine with this? Unsure if it fails on `main`.

As far as I can tell this test was causing UB (causing a write to happen to an mmap region, which other threads may be reading from based on the contract we documented for mmap regions), so I think it is correct to fail. We could get rid of / rewrite it, but if it already isn't in CI for some reason I'd rather do that later.

src/hyperlight_host/src/mem/mgr.rs
jsturtevant: Works for me

syntactically:
@jsturtevant @simongdavies Rebasing this onto main necessitated moving the changes from #1322 behind nanvix-unstable, because the snapshot is readonly by the time that that's doing writes. Please take another quick look before I merge this?

PR #1317, opened 2026-03-17
fix: host_bindgen! returns Result instead of panicking on guest errors

### LINKED ISSUE #1316: `host_bindgen!` generated code panics on guest errors instead of returning `Result`
### What happened?

The `host_bindgen!` macro generates export wrapper functions that `panic!` when `Callable::call()` returns `Err` (e.g., guest abort, trap, timeout). This crashes the host process. The generated code should return `Result` instead, allowing the host to handle guest failures gracefully.

### Steps to Reproduce

1. Host calls `failable.will_trap()` — the macro-generated export wrapper
2. Inside the wrapper, `Callable::call()` invokes the guest function
3. The guest panics → hyperlight guest runtime aborts the VM
4. `Callable::call()` returns `Err(GuestAborted(...))`
5. **The macro-generated code panics**: `let Ok(ret) = ret else { panic!(...) };`
6. The host thread's stack unwinds → the host process crashes

Add a function to the existing `witguest` test that deliberately traps, then call it from the host test and observe the panic.

### Step 1: Add to `guest.wit`

```wit
interface failable {
  /// This function will deliberately trap in the guest
  will-trap: func() -> string;
}

world test {
  // ... existing imports/exports ...
  export failable;
}
```

### Step 2: Add guest implementation in `witguest/src/main.rs`

```rust
impl test::wit::Failable for Guest {
    fn will_trap(&mut self) -> alloc::string::String {
        // Deliberately crash the guest VM.
        // This simulates any guest failure: out of memory, stack overflow,
        // calling an unimplemented import, untrusted code doing bad things, etc.
        panic!("deliberate guest crash")
    }
}

// Update TestExports to include the new interface:
impl test::wit::TestExports<Host> for Guest {
    // ... existing ...
    type Failable = Self;
    fn failable(&mut self) -> &mut Self { self }
}
```

### Step 3: Add host test in `wit_test.rs`

```rust
#[test]
fn test_guest_trap_should_not_panic_host() {
    // This test demonstrates that a guest trap (e.g., unreachable instruction)
    // causes the host to panic, crashing the entire test process.
    //
    // Expected: the call should return a Result::Err that the host can handle.
    // Actual: the macro-generated code panics, killing the host.
    
    let mut sb = sb();
    let failable = sb.failable();
    
    // This will panic with:
    //   "bad return from guest Err(GuestAborted(...))"
    // instead of returning an Err that we could handle.
    let result = failable.will_trap();
    // ^ never reaches here
    
    // What we WANT to be able to write:
    // let result = failable.try_will_trap();  // returns Result<String, Error>
    // assert!(result.is_err());
}
```

### Step 4: Run and observe

```
$ cargo test test_guest_trap_should_not_panic_host

thread 'test_guest_trap_should_not_panic_host' panicked at
'bad return from guest Err(GuestAborted(15, "Exception: GeneralProtectionFault ..."))'
```

### Expected Results
The generated trait functions should return `Result<T, E>` instead of `T` when calling into the guest, so the host can handle guest failures:

```rust
let result = failable.will_trap(); // Returns Result<String, HyperlightError>
match result {
    Ok(s) => println!("guest returned: {s}"),
    Err(e) => println!("guest crashed: {e}"),  // host stays alive
}
```

### Actual Results

TODO: What actually happened?

### Versions and Environment

#### OS Version

Linux:
```console
```

Windows (PowerShell):
```powershell
cmd /c ver
```

#### Hypervisor

Linux:
```console
```

Windows (Admin PowerShell):
```powershell
```

### Extra Info

[`src/hyperlight_component_util/src/host.rs` line 69](https://github.com/hyperlight-dev/hyperlight/blob/c9a1a4cefce6a085af92e375121e9f1b1c8deeec/src/hyperlight_component_util/src/host.rs#L69):

```rust
let ::std::result::Result::Ok(#ret) = #ret else { panic!("bad return from guest {:?}", #ret) };
```

jsturtevant:
fixes: #1316

`host_bindgen!` export wrappers no longer panic when `Callable::call()` returns `Err` (guest abort/trap/timeout). Host callers now receive the `HyperlightError`.

## Breaking change

Host-side guest export traits are now fallible and use a `*Exports` suffix. Host functions imported by the guest keep the original WIT signatures.

Example WIT:

```wit
world test {
  import roundtrip;
  export roundtrip;
}

interface roundtrip {
  echo: func(x: string) -> string;
}
```

Generated/used Rust shape:

```rust
// Guest -> host import implementation: unchanged.
impl test::wit::Roundtrip for Host {
    fn echo(&mut self, x: String) -> String { x }
}

// Host -> guest export call: renamed trait and fallible result.
use test::wit::{RoundtripExports, TestExports};

let value: Result<String, HyperlightError> = sandbox.roundtrip().echo("hi".into());
```

### DIFF
--- a/src/hyperlight_component_util/src/emit.rs
+++ b/src/hyperlight_component_util/src/emit.rs
@@ -149,6 +149,10 @@ pub struct Trait {
     pub tvs: BTreeMap<Ident, (Option<u32>, TokenStream)>,
     /// Raw tokens of the contents of the trait
     pub items: TokenStream,
+    /// Trait items keyed by their WIT names. A resource can be reached
+    /// more than once, e.g. via both the import-side and the export-side
+    /// trait of an instance, and its trait is shared.
+    pub extern_decls: BTreeMap<String, TokenStream>,
 }
 impl Trait {
     pub fn new() -> Self {
@@ -156,6 +160,7 @@ impl Trait {
             supertraits: BTreeMap::new(),
             tvs: BTreeMap::new(),
             items: TokenStream::new(),
+            extern_decls: BTreeMap::new(),
         }
     }
     /// Collect the component tyvar indices that correspond to the
@@ -217,8 +222,9 @@ impl Trait {
             .collect::<Vec<_>>();
         let tvs = self.tv_toks();
         let items = &self.items;
+        let extern_decls = self.extern_decls.values();
         quote! {
-            pub trait #n #tvs #trait_colon #(#supertraits)+* { #items }
+            pub trait #n #tvs #trait_colon #(#supertraits)+* { #items #(#extern_decls)* }
         }
     }
 }
@@ -230,18 +236,21 @@ impl Trait {
 pub struct Mod {
     pub submods: BTreeMap<Ident, Mod>,
     pub items: TokenStream,
+    /// Definitions keyed by their WIT names. An instance can be reached
+    /// more than once, e.g. as both an import and an export of a component,
+    /// and its helper module is shared.
+    pub extern_decls: BTreeMap<String, TokenStream>,
     pub traits: BTreeMap<Ident, Trait>,
     pub impls: BTreeMap<(Vec<Ident>, Ident), TokenStream>,
-    pub emitted_type_names: BTreeSet<Ident>,
 }
 impl Mod {
     pub fn empty() -> Self {
         Self {
             submods: BTreeMap::new(),
             items: TokenStream::new(),
+            extern_decls: BTreeMap::new(),
             traits: BTreeMap::new(),
             impls: BTreeMap::new(),
-            emitted_type_names: BTreeSet::new(),
         }
     }
     /// Get a reference to a sub-module, creating it if necessary
@@ -295,6 +304,9 @@ impl Mod {
         }
         for (n, mut t) in self.traits {
             tt.extend(t.into_tokens(n));
+        }
+        for (_, decl) in self.extern_decls {
+            tt.extend(decl);
         }
         tt.extend(self.items);
         for ((ns, i), t) in self.impls {
--- a/src/hyperlight_component_util/src/emit.rs
+++ b/src/hyperlight_component_util/src/emit.rs
@@ -232,6 +232,7 @@ pub struct Mod {
     pub items: TokenStream,
     pub traits: BTreeMap<Ident, Trait>,
     pub impls: BTreeMap<(Vec<Ident>, Ident), TokenStream>,
+    pub emitted_type_names: BTreeSet<Ident>,
 }
 impl Mod {
     pub fn empty() -> Self {
@@ -240,6 +241,7 @@ impl Mod {
             items: TokenStream::new(),
             traits: BTreeMap::new(),
             impls: BTreeMap::new(),
+            emitted_type_names: BTreeSet::new(),
         }
     }
     /// Get a reference to a sub-module, creating it if necessary
--- a/src/hyperlight_component_util/src/host.rs
+++ b/src/hyperlight_component_util/src/host.rs
@@ -47,6 +47,7 @@ fn emit_export_extern_decl<'a, 'b, 'c>(
                         .map(|p| rtypes::emit_func_param(s, p))
                         .collect::<Vec<_>>();
                     let result_decl = rtypes::emit_func_result(s, &ft.result);
+                    let result_decl = quote! { ::std::result::Result<#result_decl, ::hyperlight_host::error::HyperlightError> };
                     let hln = emit_fn_hl_name(s, ed.kebab_name);
                     let ret = format_ident!("ret");
                     let marshal = ft
@@ -67,11 +68,11 @@ fn emit_export_extern_decl<'a, 'b, 'c>(
                                 #hln,
                                 marshalled,
                             );
-                            let ::std::result::Result::Ok(#ret) = #ret else { panic!("bad return from guest {:?}", #ret) };
+                            let #ret = #ret?;
                             #[allow(clippy::unused_unit)]
                             let mut rts = self.rt.lock().unwrap();
                             #[allow(clippy::unused_unit)]
-                            #unmarshal
+                            Ok(#unmarshal)
                         }
                     }
                 }
@@ -122,7 +123,7 @@ fn emit_export_instance<'a, 'b, 'c>(s: &'c mut State<'a, 'b>, wn: WitName, it: &
 
     let ns = wn.namespace_path();
     let nsi = wn.namespace_idents();
-    let trait_name = kebab_to_type(wn.name);
+    let trait_name = kebab_to_exports_name(wn.name);
     let r#trait = s.r#trait(&nsi, trait_name.clone());
     let tvs = r#trait
         .tvs
--- a/src/hyperlight_component_util/src/host.rs
+++ b/src/hyperlight_component_util/src/host.rs
@@ -47,7 +47,6 @@ fn emit_export_extern_decl<'a, 'b, 'c>(
                         .map(|p| rtypes::emit_func_param(s, p))
                         .collect::<Vec<_>>();
                     let result_decl = rtypes::emit_func_result(s, &ft.result);
-                    let result_decl = quote! { ::std::result::Result<#result_decl, ::hyperlight_host::error::HyperlightError> };
                     let hln = emit_fn_hl_name(s, ed.kebab_name);
                     let ret = format_ident!("ret");
                     let marshal = ft
@@ -60,17 +59,16 @@ fn emit_export_extern_decl<'a, 'b, 'c>(
                         fn #n(&mut self, #(#param_decls),*) -> #result_decl {
                             let mut to_cleanup = Vec::<Box<dyn Drop>>::new();
                             let marshalled = {
-                                let mut rts = self.rt.lock().unwrap();
+                                let mut rts = self.rt.lock()?;
                                 #[allow(clippy::unused_unit)]
                                 (#(#marshal,)*)
                             };
                             let #ret = ::hyperlight_host::sandbox::Callable::call::<::std::vec::Vec::<u8>>(&mut self.sb,
                                 #hln,
                                 marshalled,
-                            );
-                            let #ret = #ret?;
+                            )?;
                             #[allow(clippy::unused_unit)]
-                            let mut rts = self.rt.lock().unwrap();
+                            let mut rts = self.rt.lock()?;
                             #[allow(clippy::unused_unit)]
                             Ok(#unmarshal)
                         }
@@ -167,7 +165,7 @@ impl SelfInfo {
             orig_id,
             type_id: vec![format_ident!("I")],
             inner_preamble: quote! {
-                let mut #inner_id = #outer_id.lock().unwrap();
+                let mut #inner_id = #outer_id.lock()?;
                 let mut #inner_id = ::std::ops::DerefMut::deref_mut(&mut #inner_id);
             },
             outer_id,
@@ -246,7 +244,7 @@ fn emit_import_extern_decl<'a, 'b, 'c>(
                 let #outer_id = #orig_id.clone();
                 let captured_rts = rts.clone();
                 sb.register_host_function(#hln, move |#(#pds),*| {
-                    let mut rts = captured_rts.lock().unwrap();
+                    let mut rts = captured_rts.lock()?;
                     #inner_preamble
                     let #ret = #callname(
                         ::std::borrow::BorrowMut::<#(#type_id)::*>::borrow_mut(
--- a/src/hyperlight_component_util/src/rtypes.rs
+++ b/src/hyperlight_component_util/src/rtypes.rs
@@ -674,6 +674,11 @@ fn emit_extern_decl<'a, 'b, 'c>(
                         .map(|p| emit_func_param(&mut s, p))
                         .collect::<Vec<_>>();
                     let result = emit_func_result(&mut s, &ft.result);
+                    let result = if !s.is_guest && s.is_export {
+                        quote! { ::std::result::Result<#result, ::hyperlight_host::error::HyperlightError> }
+                    } else {
+                        result
+                    };
                     quote! {
                         fn #n(&mut self, #(#params),*) -> #result;
                     }
@@ -696,12 +701,22 @@ fn emit_extern_decl<'a, 'b, 'c>(
                         }
                         ResourceItemName::Method(n) => {
                             let result = emit_func_result(&mut sv, &ft.result);
+                            let result = if !sv.is_guest && sv.is_export {
+                                quote! { ::std::result::Result<#result, ::hyperlight_host::error::HyperlightError> }
+                            } else {
+                                result
+                            };
                             sv.cur_trait().items.extend(quote! {
                                 fn #n(&mut self, #(#params),*) -> #result;
                             });
                         }
                         ResourceItemName::Static(n) => {
                             let result = emit_func_result(&mut sv, &ft.result);
+                            let result = if !sv.is_guest && sv.is_export {
+                                quote! { ::std::result::Result<#result, ::hyperlight_host::error::HyperlightError> }
+                            } else {
+                                result
+                            };
                             sv.cur_trait().items.extend(quote! {
                                 fn #n(&mut self, #(#params),*) -> #result;
                             });
@@ -724,6 +739,9 @@ fn emit_extern_decl<'a, 'b, 'c>(
             ) -> TokenStream {
                 let id = kebab_to_type(ed.kebab_name);
                 let mut s = s.helper();
+                if !s.cur_mod().emitted_type_names.insert(id.clone()) {
+                    return TokenStream::new();
+                }
 
                 let t = emit_defined(&mut s, v, id, t);
                 s.cur_mod().items.extend(t);
@@ -745,6 +763,9 @@ fn emit_extern_decl<'a, 'b, 'c>(
                         let rn = kebab_to_type(ed.kebab_name);
                         s.add_helper_supertrait(rn.clone());
                         let mut s = s.helper();
+                        if !s.cur_mod().emitted_type_names.insert(rn.clone()) {
+                            return quote! {};
+                        }
                         s.cur_trait = Some(rn.clone());
                         s.cur_trait().items.extend(quote! {
                             type T: ::core::marker::Send;
@@ -762,7 +783,12 @@ fn emit_extern_decl<'a, 'b, 'c>(
             emit_instance(&mut s, wn.clone(), it);
 
             let nsids = wn.namespace_idents();
-            let repr = s.r#trait(&nsids, kebab_to_type(wn.name));
+            let trait_name = if !s.is_guest && s.is_export {
+                kebab_to_exports_name(wn.name)
+            } else {
+                kebab_to_type(wn.name)
+            };
+            let repr = s.r#trait(&nsids, trait_name.clone());
             let vs = if !repr.tvs.is_empty() {
                 let vs = repr.tvs.clone();
                 let tvs = vs
@@ -780,11 +806,10 @@ fn emit_extern_decl<'a, 'b, 'c>(
             };
             let rp = s.root_path();
             let tns = wn.namespace_path();
-            let trait_tn = kebab_to_type(wn.name);
             let trait_bound = if tns.is_empty() {
-                quote! { #rp #trait_tn }
+                quote! { #rp #trait_name }
             } else {
-                quote! { #rp #tns::#trait_tn }
+                quote! { #rp #tns::#trait_name }
             };
             quote! {
                 type #member_tn: #trait_bound #vs;
@@ -803,7 +828,11 @@ fn emit_instance<'a, 'b, 'c>(s: &'c mut State<'a, 'b>, wn: WitName, it: &'c Inst
     tracing::debug!("emitting instance {:?}", wn);
     let mut s = s.with_cursor(wn.namespace_idents());
 
-    let name = kebab_to_type(wn.name);
+    let name = if !s.is_guest && s.is_export {
+        kebab_to_exports_name(wn.name)
+    } else {
+        kebab_to_type(wn.name)
+    };
 
     s.cur_helper_mod = Some(kebab_to_namespace(wn.name));
     s.cur_trait = Some(name.clone());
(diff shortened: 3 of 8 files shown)


ludfjig (commented):
I think I am fine with this, but I suspect this is intentional. I think we generally want to avoid Result types as much as possible where it makes sense, but since this matches the current guest calling convention maybe let's defer that discussion to later

jsturtevant:
> I think I am fine with this, but I suspect this is intentional. I think we generally want to avoid Result types as much as possible where it makes sense, but since this matches the current guest calling convention maybe let's defer that discussion to later

Open to other ways of addressing this I guess but a panic in the guest shouldn't panic this host.  It seemed practical to match existing behavoir

syntactically:
I agree with @ludfjig that, generally speaking, the semantic gap of lifting everything from the interface into Option isn't particularly attractive. However, I do see the need to deal with errors that fundamentally can happen during execution on the host side.

Is there a reason that this is also applied to the host functions imported by the guest? I think that generally speaking it is the _ability to export a function implementing an interface, but actually implement a different interface (i.e. one with more optionality)_ that is the most semantically damaging---so the implicit `Result`s are particularly unattractive on guest->host function calls (and to some extent in the implementation of guest functions, although the ability to trap/panic/etc is perhaps equivalent). It seems like splitting this (so that host->guest function calls can return errors, but host functions can't be implemented by returning errors) is what is common in the prior art e.g. the wasmtime host bindgen. Is it just about continuing to be able to use the same trait for the import and export versions of an interface? Maybe a type parameter to track whether this is an imported or exported usage would make sense?

Let me know what you think about the usability of that. As I think about this more, from the wasm perspective especially, I do think that having the `Result`s on host function calls is slightly unattractive, but it's not quite as bad as I would have guessed, since from wasm's perspective, a (wasm) import that's called always does in fact have the option of causing the entire world to be torn town via purposefully causing a trap, and this is perhaps in some ways similar behaviour.

jsturtevant:
> Is there a reason that this is also applied to the host functions imported by the guest?

This was to match the current non-wit implementation. We had improved the situation with errors in https://github.com/hyperlight-dev/hyperlight/pull/868.  

>I think that generally speaking it is the ability to export a function implementing an interface, but actually implement a different interface (i.e. one with more optionality) that is the most semantically damaging---so the implicit Results are particularly unattractive on guest->host function calls (and to some extent in the implementation of guest functions, although the ability to trap/panic/etc is perhaps equivalent). It seems like splitting this (so that host->guest function calls can return errors, but host functions can't be implemented by returning errors) is what is common in the prior art e.g. the wasmtime host bindgen. 

Agree with the general sentiment that's its not ideal.   I am not so sure we don't want host function to return errors, it seems useful to me to be able to say something went wrong across the boundry.  Be interesting in what others think

jsturtevant:
> Agree with the general sentiment that's its not ideal. I am not so sure we don't want host function to return errors, it seems useful to me to be able to say something went wrong across the boundry. Be interesting in what others think

actually, was thinking on this some more and WIT does have the ability capture "result" types so we probalby don't need this on the guest side (I don't see the equivalent `::std::result::Result::Ok(#ret) = #ret else { panic!("bad return from guest {:?}", #ret) };`  on the Guest->Host calls.  Maybe there is another way to handle that too... I am open to ideas just really think we shouldn't panic if guest traps.

syntactically:
>  just really think we shouldn't panic if guest traps

I'm with you on that one :)

> This was to match the current non-wit implementation. We had improved the situation with errors in https://github.com/hyperlight-dev/hyperlight/pull/868.

> I am not so sure we don't want host function to return errors, it seems useful to me to be able to say something went wrong across the boundry.

I'm still not totally convinced on this, although I can see the arguments. Can you elaborate a little more on the use cases you see?  If we did allow this, I think it's important that it not be semantically visible in the guest, which shouldn't know whether its imported component is implemented across a hyperlight-host boundary or by any other component, so we would need to propagate the error return up, and I would think that it would probably make sense to poison the sandbox (probably by panicking the guest?).  If there's not a super strong use case, having the optionality just on the host->guest calls and not the hostcall returns seems like it is a starting point that gets rid of the main issue and is minimally invasive?

> actually, was thinking on this some more and WIT does have the ability capture "result" types so we probalby don't need this on the guest side (I don't see the equivalent ::std::result::Result::Ok(#ret) = #ret else { panic!("bad return from guest {:?}", #ret) }; on the Guest->Host calls. Maybe there is another way to handle that too...

There is an unwrap on the host call result [here](https://github.com/hyperlight-dev/hyperlight-wasm/blob/main/src/hyperlight_wasm_macro/src/wasmguest.rs#L90).

Precisely the thing that I want to be careful about here is conflating WIT-level semantically-there-in-the-API optionality  with this unconditionally-added-everywhere Hyperlight-level optionality.  As I said in the paragraph above, even if we did add this for host function return types, I would not want to see it meaningfully exposed to the guest code.

jsturtevant:
>I am not so sure we don't want host function to return errors, it seems useful to me to be able to say something went wrong across the boundry.
>
> I'm still not totally convinced on this, although I can see the arguments. Can you elaborate a little more on the use cases you see? If we did allow this, I think it's important that it not be semantically visible in the guest, which shouldn't know whether its imported component is implemented across a hyperlight-host boundary or by any other component, so we would need to propagate the error return up, and I would think that it would probably make sense to poison the sandbox (probably by panicking the guest?). If there's not a super strong use case, having the optionality just on the host->guest calls and not the hostcall returns seems like it is a starting point that gets rid of the main issue and is minimally invasive?

Yes I am in agree that we shouldn't do this now.  I'm going to revisit this and see if I can come up with something different for consideration.

ludfjig:
@jsturtevant should we close this one?

jsturtevant:
We need a version of this, let me see if I can fix it up this week

src/hyperlight_host/tests/wit_test.rs
```diff
@@ -416,14 +423,45 @@ mod wit_test {
...
+
+    #[test]
+    fn test_guest_call_error_returns_error() {
+        let mut sb = failing_sb();
+        let err = sb.roundtrip().roundtrip_bool(true).unwrap_err();
+        assert!(matches!(
+            err,
+            HyperlightError::Error(msg) if msg == "synthetic guest call failure"
+        ));
+    }
```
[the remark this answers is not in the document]
jsturtevant: updated

src/hyperlight_component_util/src/rtypes.rs
```diff
@@ -729,7 +750,12 @@ fn emit_extern_decl<'a, 'b, 'c>(
             emit_instance(&mut s, wn.clone(), it);
 
             let nsids = wn.namespace_idents();
-            let repr = s.r#trait(&nsids, kebab_to_type(wn.name));
+            let trait_name = if !s.is_guest && s.is_export {
+                kebab_to_exports_name(wn.name)
+            } else {
+                kebab_to_type(wn.name)
+            };
+            let repr = s.r#trait(&nsids, trait_name.clone());
```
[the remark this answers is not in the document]
jsturtevant: updaed description

jsturtevant:
I refreshed this.  I wasn't able to find a way to not do this on host side but it is minimally invasive I believe.  On the guest side we can use wit interface as discussed above.

src/hyperlight_component_util/src/rtypes.rs
```diff
@@ -641,6 +641,11 @@ fn emit_extern_decl<'a, 'b, 'c>(
                         .map(|p| emit_func_param(&mut s, p))
                         .collect::<Vec<_>>();
                     let result = emit_func_result(&mut s, &ft.result);
+                    let result = if !s.is_guest && s.is_export {
+                        quote! { ::std::result::Result<#result, ::hyperlight_host::error::HyperlightError> }
+                    } else {
+                        result
+                    };
```
danbugs: Might want a helper for this? It seems to repeat enough times.

src/hyperlight_component_util/src/host.rs
```diff
@@ -66,11 +67,11 @@ fn emit_export_extern_decl<'a, 'b, 'c>(
                                 #hln,
                                 marshalled,
                             );
-                            let ::std::result::Result::Ok(#ret) = #ret else { panic!("bad return from guest {:?}", #ret) };
+                            let #ret = #ret?;
                             #[allow(clippy::unused_unit)]
                             let mut rts = self.rt.lock().unwrap();
```
danbugs: This will still panic--fine?

danbugs (commented):
Mostly LGTM. Just minor nits.

src/hyperlight_component_util/src/emit.rs
```diff
@@ -124,6 +124,7 @@ pub struct Mod {
     pub items: TokenStream,
     pub traits: BTreeMap<Ident, Trait>,
     pub impls: BTreeMap<(Vec<Ident>, Ident), TokenStream>,
+    pub emitted_type_names: BTreeSet<Ident>,
```
syntactically: I don't follow what this is being used for?

src/hyperlight_component_util/src/rtypes.rs
syntactically: I think it should be pushed into `emit_func_result`, whose semantic job is to centralize the logic for converting a C-M result type into a Rust type for bindings generation.

 Are there any times where we use that where we don't want it?

src/hyperlight_component_util/src/rtypes.rs
```diff
@@ -691,6 +706,9 @@ fn emit_extern_decl<'a, 'b, 'c>(
             ) -> TokenStream {
                 let id = kebab_to_type(ed.kebab_name);
                 let mut s = s.helper();
+                if !s.cur_mod().emitted_type_names.insert(id.clone()) {
```
syntactically: What is this being used for? It seems like it will simply drop some definitions, which doesn't seem ideal?

syntactically (dismissed):
Just to make sure I understand, is the point of adding the `Exports` onto the end of the trait name on the host side just to make sure that if the same interface is used as an import and an export, the import and export don't collide?

Is there any barrier to centralizing the logic about the function result types in `emit_func_result`, whose job it usually is to decide the function result types?

src/hyperlight_component_util/src/host.rs
jsturtevant: IIRC, we discusssed these panics in the orignal imp and its something that *should* never happen unless we messed something up but now that we have the ability to bubble it up. let me try this.

src/hyperlight_component_util/src/rtypes.rs
jsturtevant: not that i know of, i will take a look at pushing it there

jsturtevant:
> Just to make sure I understand, is the point of adding the `Exports` onto the end of the trait name on the host side just to make sure that if the same interface is used as an import and an export, the import and export don't collide?

yes, now that they are not the same shape, I needed a way to make those distinct.

src/hyperlight_component_util/src/host.rs
jsturtevant: fixed

src/hyperlight_component_util/src/rtypes.rs
jsturtevant: this is not in the right spot, so its not duplicated

src/hyperlight_component_util/src/emit.rs
jsturtevant: I've added a comment to address this and re-worked it.  When the import and export where sharing resources it ended up with duplicates. This is tracking that so they are shared with out duplication

```
emit_instance  walks  roundtrip 's decls twice and writing them both:

line  20:            pub struct Testrecord {     <- from the import walk
line  78:            pub struct Testrecord {     <- from the export walk
```

src/hyperlight_component_util/src/rtypes.rs
jsturtevant: updated to be clearer I hope

syntactically (commented):
I think that the actual panic-related changes look pretty good to me now.

>  When the import and export where sharing resources it ended up with duplicates. This is tracking that so they are shared with out duplication

Just to clarify, is this something new brought on by the changes in this PR, or pre-existing behaviour that you happened upon and wanted to fix at the same time? I was assuming the former & therefore confused, but if the latter it makes sense, and i did manage to reproduce it locally before the change, so I'm guessing it's that?

I am still not convinced by this implementation, which feels a bit invasive/global to me.

Among other things, If I am reading it correctly, I think that, if you had an import and an export, or two imports or two exports, with the same wit-style package/etc names but different contents, it would sort of smash them together and make something not usable as either.

I would expect at a first glance that this is only an issue in one place: the instance case of `rtypes::emit_extern_decl`. The reasoning for that is:
- For all other kinds of externs that we support in the bindgen, they are emitted "in place" as members of the type/trait for the parent instance/component with a name based on the externname, which validation already ensures does not collide.
- Only in the Instance case do we rename the type over to a namespace based on the wit package parsing we do of its name, which I think is where it becomes possible for that to happen twice, which is the fundamental problem here?

Did you run into it anywhere else? If not, even when mitigating it at this layer, I think there should be a more contained approach, where in the `emit::State` we keep track of which wit-names we have already generated an instance for, and check/maintain that list only in the one place in `rtypes::emit_extern_decl`.  At least, the test you added passes with the attached patch.

[0001-hyperlight_component_util-Do-not-emit-duplicate-inst.patch](https://github.com/user-attachments/files/30551945/0001-hyperlight_component_util-Do-not-emit-duplicate-inst.patch)

It feels a little unsatisfying to mitigate it at the codegen level here at all, because in some sense we are dealing with some duplication that we should be aware of in a phase where we have a rich semantic structure, but instead are dealing with it at a layer where everything is very syntactic and ugly and the modifications we can make and the sanity-checking we can do is very limited. For example, there is no easy, sensible, semantic way at this level to make sure that the instances we do not emit are "equal" to those that we do.

Another approach would be to go up a level and have a sanity check at the etypes level when we have the types in a nice form. There we could
- scan through the import/export tree to see if there are any instances which have the same wit-style package/namespace name; and
- if there are, use the functions from the `subtype` module to check them for equality; however, this will not be totally trivial, because:
  + The subtyping functions will need to be extended to include the definition of instance subtyping from the spec pdf; and
  + There is not a natural single typing context in which we can check both of them, so we may have to synthesize one. There is some complexity here due to the way that the `deftype`s for the instances will be the same `QualifiedInstance`, but the things we actually have in the imports/exports have had their universal quantifiers stripped off them to make other typechecking easier. This is a bit of a problem, actually, which I am not sure how to resolve within a few minutes of thinking.

syntactically:
It might make sense to merge just the panic changes from this and then make a new PR for the other changes where we can deliberate a little bit more the best way to solve them?

jsturtevant:
Sounds good. I did look at doing something higher up but was a much more evasive set of changes and felt much more challenging to review.  Splitting it out as a separate issues/pr makes sense to me.

jsturtevant:
>Just to clarify, is this something new brought on by the changes in this PR, or pre-existing behaviour that you happened upon and wanted to fix at the same time? I was assuming the former & therefore confused, but if the latter it makes sense, and i did manage to reproduce it locally before the change, so I'm guessing it's that?

ah, So I just tried removing the deduplication code and found that it is needed.  We are confused because both of these are true 😓 

The inline test I added fails on main with no changes **_and_** once I move the interface from `Roundtrip` -> `RoundtripExports` (because they are two different shapes now) it starts failing with these changes.

> It might make sense to merge just the panic changes from this and then make a new PR for the other changes where we can deliberate a little bit more the best way to solve them?

So I don't think we can do this.  We could either:

- make the "hack" I did here at the code gen level and open an issue to fix more wholistically 
- make the more generic changes to up a layer, making these changes a bit more extensive

syntactically:
(I apologise in advance for the length of this reply. As Pascal said, I did not have time to be brief.)

> ah, So I just tried removing the deduplication code and found that it is needed. We are confused because both of these are true 😓
> 
> The inline test I added fails on main with no changes **_and_** once I move the interface from `Roundtrip` -> `RoundtripExports` (because they are two different shapes now) it starts failing with these changes.

Aha, you are totally right---I had forgotten, but there is _already_ a stupid hack to deal with this at the token emission layer (which I guess I was equally unsatisfied with when I wrote, but did not have time to fix at the time) [here](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_component_util/src/rtypes.rs#L810).   Now, your change to make the exports trait name different breaks that version of the hack because the trait object is different, so the check for whether we have visited this trait before or not breaks. And it is, or was, one of the unsolved semantic questions around how the wit bindgen should in general work, because iirc plans for subtyping mean that it's totally reasonably for language toolchains to emit imports of stripped-down versions of the interface missing items they don't use---if we have several imports of disjoint subsets, how do we unify them into one? The machinery to do that at the higher level is not quite there in this implementation. Despite what I wrote in that comment about a slightly less hacky (than the old hack) way of doing it, I do think that it is nicer to do this in some semantically meaningful place at the etypes layer.  However, it is nice validation that the approach in the comment for a middle-hackiness way of doing it (probably due to the formalism of the wit translation layer overall being incomplete, and so it being unclear how to do it at the etypes level) actually I think lines up pretty well with the approach you suggested here.

In that case, the patch I suggested above probably has the inverse issue of being too conservative and preventing the emission of the `Exports` side trait indiscriminately.

Given that there is already a (arguably more) hacky solution to this issue in the current code, I don't think it makes sense to delay releasing the `Result`-ification of the host side of the trait until we solve that problem in general.

I'm not convinced by the semantics of using the same resource trait on both the import and the export side, since the resource trait includes functions that operate on the resource, so I think it needs the same `Result`-transform when the functions are implemented by the guest. I think there are some other places where resources on the guest side are explicitly not-implemented, but it would be nice to make sure that all the generic code here is not dependent on that, so that it does not become an issue later. So it may be that we need to make the resource traits also have an Export version.

Having to thread around duplicate `Exports` copies of traits and resource impls and everything everywhere feels a little bit unpleasant. Another thing that we could do is get rid of the separate export trait definitions entirely, and replace them with a type-parameter. (I'm undecided either way on which of these is more elegant, but I wanted to call attention to the possibility---maybe you have an opinion?). Something like (but with better naming) (assuming we added a function `test: Item -> ()` to the `shared_interface_bindings` example:
```rust
// todo figure out a structure where we can be sure this namespace won't conflict with generated bindings, or get it from an import, or something
mod bindgen_internals {
  trait Positivity {
    type NegativeOfThis: Positivity;
    type FnReturnWrapper<T>;
  }
  struct Positive;
  struct Negative;
  impl Positivity for Positive {
    type NegativeOfThis = Negative;
    // host side
    type FnReturnWrapper<T> = Results<T>;
    // guest side
    type FnReturnWrapper<T> = T;
  }
  impl Positivity for Negative {
    type NegativeOfThis = Positive;
    type FnReturnWrapper<T> = T;
  }
}
// Code we will generate
mod shared {
  struct Item {
    value: u32,
  }
}
trait Shared<P: Positivity> {
  fn test(shared::Item) -> P::FnReturnWrapper<()>;
}
trait SharedWorldImports<P: Positivity> {
  type Shared: Shared<P>;
  fn shared() -> Self::Shared;
}
trait SharedWorldExports<P: Positivity> {
  type Shared: Shared<P>;
  fn shared() -> Self::Shared;
}
trait SharedWorld<P: Positivity> {
  type Exports<I: SharedWorldImports<P::NegativeOfThis> + 'static>: SharedWorldExports<P + 'static>;
  fn instantiate<I: SharedWorldImports<P::NegativeOfThis> + 'static>(self, imports: I) -> Self::Exports<I>;
}
impl SharedWorld<Positive> for ... {
}
```
If we did this, we would only need to emit one instance, and the old terrible hack would still work.

(Actually, thinking about this a little bit more: `State::is_export` is meant to track the same kind of positivity information, but does so at generation time. But, for the exact reason that we are seeing here, this is probably wrong, and would result in issues if the same item that relied on it occurred in both import and export positions.  So, if we went this route, we would want to remove `State::is_export` (or at least ban its use from rtypes, since the same issues do not apply in the wrapper registration/etc) and replace the one use of it with another associated type on `Positivity`. That's almost certainly a latent bug that we have not run into only because of the lack of support for resources defined on the host side, which would be needed to import/export the same resource & so run into this)

Apart from the possibility of changing to the approach above, I have looked in some more detail at your approach and I do think that is more-or-less "correct" for doing something at this level, save for the issue with the resource traits. However, I do have a few suggestions:
- I think that `Trait::items` could be re-typed to be directly the map from extern names to tokens, since every thing generated in a trait is directly tied to one of these exports, which would be preferable
- I also think that it would be preferable to have a helper function on `Trait` to handle the insertion which checked the existing value and raised a warning/error if the token sequences were different.
- It's a bit of a shame that `Mod` still needs both its existing unstructured `items` (to support all the other things that get shoved in there) and may also need a new thing for keeping track of "items that are due to an extern name in an instance for which this is a helper module containing type definitions". If we did not do the "checking that every time we encountered something we get the same tokens", we could avoid the need to add anything to `Mod` by simply not calling `go_defined` in the case that we had already processed this extern. That might be worth it for the simplification, but I will leave it up to your taste
- There are a number of places where some other state of the trait is manipulated. If replacing the existing hack with trying to merge a bit better (which I think is worth doing if we do it this way, although not if we switch to the type parameter tracking positivity approach above), would you mind taking a sanity check that none of those have the chance to override things they shouldn't when run the "second time"? For example, [this line](https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_component_util/src/rtypes.rs#L712) looks at first glance like it might override some bounds with an empty token stream---but it's possible that in practice it will not, so it's worth thinking about a little bit more and adding comments with rationale anywhere there is something that looks concerning at first glance.
- Finally, if you do decide to change the detection hack rather than thread the type parameter around, I think it would be good to factor that out into its own commit (that we either merge without squash or put up in another PR), with a detailed commit comment explaining why the old detection hack is becoming untenable and why the various decisions in the new one were made. This is a fairly interesting change on its own (more interesting than the actual point here) and I am sure I and others in the future will be extremely happy to have the `git blame` motivation for it cleanly there.

jsturtevant:
>Another thing that we could do is get rid of the separate export trait definitions entirely, and replace them with a type-parameter. (I'm undecided either way on which of these is more elegant, but I wanted to call attention to the possibility---maybe you have an opinion?). 

I actually had an implementation like this at one point :-)  I didn't push it because it felt much heavier change but that is before I understood that we had a lot other things causing issues here.  

I've tried both again and believe the trait approach is cleaner, after making the requested changes to the code gen/Export naming approach which kind of spiraled out of control

jsturtevant:
I pushed https://github.com/hyperlight-dev/hyperlight/pull/1709 instead of overwriting and loosing this context for now.  I named them a little differently for our particular context (positivity and Negative were hard for me to track as I was reading and Export/Import essentially track the existing ideas in the code.  I also left out the `NegativeOfThis` as we don't have a use case for it right now (that i am aware of).

jsturtevant:
closing in favor of https://github.com/hyperlight-dev/hyperlight/pull/1709 which we've decided is a better approach in general thought needs a bit more work

Issue #1290, opened 2026-03-09
Linux `map_file_cow` leaks VMA entries (missing `munmap` on success path)


simongdavies:
## Summary

`MultiUseSandbox::map_file_cow` on Linux calls `libc::mmap(MAP_PRIVATE)` but never calls `munmap` on the success path. Each call leaks one kernel VMA (`vm_area_struct`), which counts against the per-process `vm.max_map_count` limit (default: 65530).

**Physical memory is NOT leaked** — all mmaps of the same file share the kernel page cache. The leak is purely a VMA slot leak (~200 bytes of kernel metadata per entry). However, a long-running service that churns sandboxes with `map_file_cow` will eventually hit `vm.max_map_count` and subsequent `mmap` calls will fail with `ENOMEM`.

The `vm.max_map_count` limit is a kernel-wide sysctl, not a per-process resource limit — it cannot be raised with `ulimit`, only via `sudo sysctl -w vm.max_map_count=<value>`.

## What's leaked per call

| Resource | Leaked? | Impact |
|----------|---------|--------|
| Kernel VMA (`vm_area_struct`) | **Yes** — ~200 bytes | Counts against `vm.max_map_count` (default 65530) |
| Virtual address space | **Yes** — one page-aligned region | Negligible on 64-bit (128TB user VA space) |
| Physical RAM (page cache) | **No** — shared across all mmaps of the same file | Kernel reclaims under memory pressure |
| File descriptor | **No** — `File` is dropped (fd closed) after `mmap`; kernel holds inode ref internally | Released on `munmap` |

## Root Cause

The raw `*mut c_void` pointer returned by `libc::mmap` is cast to `usize` and stored in `MemoryRegion.host_region` — a plain integer with no ownership or `Drop` semantics. Nothing in the cleanup chain calls `munmap`:

- `HyperlightVm::unmap_region` — removes the KVM/MSHV slot only, does not `munmap`
- `HyperlightVm::Drop` — only calls `interrupt_handle.set_dropped()`
- `MultiUseSandbox` — has no `Drop` impl
- Snapshot `restore()` — calls `unmap_region` for stale regions (same problem)

The error path in `map_file_cow` correctly calls `munmap`, confirming the success path omission is an oversight.

## Proposed Fix

[...]

Issue #700, opened 2025-07-08
Allow host mappings to have destructors


syntactically:
We should provide some support for ensuring that when host mapping objects are no longer needed, they have a chance to run a destructor.  This is particularly important as, currently, `map_file_cow` has no way to ensure that its mappings are cleaned up, so it leaks a bit of virtual memory whenever it is run.

ludfjig:
Since `map_file_cow` currently only calls `map_region`, could we remove `map_file_cow`, and have the caller just use `map_region` and keep track of the handle themselves?

jsturtevant:
from #1290:

```
Root Cause
The raw *mut c_void pointer returned by libc::mmap is cast to usize and stored in MemoryRegion.host_region — a plain integer with no ownership or Drop semantics. Nothing in the cleanup chain calls munmap:

HyperlightVm::unmap_region — removes the KVM/MSHV slot only, does not munmap
HyperlightVm::Drop — only calls interrupt_handle.set_dropped()
MultiUseSandbox — has no Drop impl
Snapshot restore() — calls unmap_region for stale regions (same problem)
The error path in map_file_cow correctly calls munmap, confirming the success path omission is an oversight.

Proposed Fix
Introduce a Linux OwnedFileMapping struct that stores the mmap pointer and size with a Drop impl calling libc::munmap. Track these in a Vec on MultiUseSandbox and clean up in restore() (when regions are unmapped during snapshot restore) and on MultiUseSandbox::Drop (catch-all for remaining mappings).
```

syntactically:
The correct solution here is probably that `map_region` instead of taking a `MemoryRegion` directly takes some `where for<'a> &'a T: Into<MemoryRegion> + Clone` object that it holds onto and drops at the appropriate times. When `map_file_cow` calls `map_region`, it uses some (arch-specific) `FileMapping` structure that implements that properly.  That would also let you write code that shared one mapping between multiple sandboxes and use it on both Linux and Windows without disaster striking from the strange lifecycle management of the region on Windows.

PR #1320, opened 2026-03-18
Implement map_file_cow before a sandbox is initialised

simongdavies:
### Overview

This pull request implements `map_file_cow` on `UninitializedSandbox` so file mappings can be created before a VM exists.

It refactors the file mapping logic for guest sandboxes, splitting the process into two distinct phases: host-side preparation and VM-side application. The main goal is to allow file mappings to be prepared before the VM exists and then applied any later when the VM is created (i.e., during `evolve`). 

### Changes

- File Mapping Refactor and Modularization:
  - Added `file_mapping` module and `PreparedFileMapping` abstraction for two-phase file mapping (host-side preparation and VM-side application), with safe cleanup and platform-specific handling.
  - Refactored `MultiUseSandbox::map_file_cow` to use the new abstraction, removing duplicated platform-specific code.

- Support for Deferred File Mapping:
  - Added `pending_file_mappings` to `UninitializedSandbox` and a new `map_file_cow` method to defer mapping until VM creation. 

- Code Cleanup and Dependency Updates:
  - Removed obsolete imports and platform-specific code from `initialized_multi_use.rs`, consolidating logic in `file_mapping`. 

- Test and API Adaptations:
  - Updated call sites and tests to use the new file mapping API and pass `page_size` as needed. 
  
  
Also fixes a bug where the `mmap` memory access flags were too permissive (remove `wx` for KVM and `x` for mshv)

### DIFF
--- a/src/hyperlight_host/src/sandbox/file_mapping.rs
+++ b/src/hyperlight_host/src/sandbox/file_mapping.rs
@@ -1,0 +1,326 @@
+/*
+Copyright 2025  The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Host-side file mapping preparation for [`map_file_cow`].
+//!
+//! This module splits the file mapping operation into two phases:
+//! - **Prepare** ([`prepare_file_cow`]): performs host-side OS calls
+//!   (open file, create mapping) without requiring a VM.
+//! - **Apply**: performed by the caller (either [`MultiUseSandbox::map_file_cow`]
+//!   or [`evolve_impl_multi_use`]) to map the prepared region into
+//!   the guest via [`HyperlightVm::map_region`].
+//!
+//! This separation allows [`UninitializedSandbox`] to accept
+//! `map_file_cow` calls before the VM exists, deferring the VM-side
+//! work until [`evolve()`].
+
+use std::ffi::c_void;
+use std::path::Path;
+
+use tracing::{Span, instrument};
+
+#[cfg(target_os = "windows")]
+use crate::HyperlightError;
+#[cfg(target_os = "windows")]
+use crate::hypervisor::wrappers::HandleWrapper;
+#[cfg(target_os = "windows")]
+use crate::mem::memory_region::{HostRegionBase, MemoryRegionKind};
+use crate::mem::memory_region::{MemoryRegion, MemoryRegionFlags, MemoryRegionType};
+use crate::{Result, log_then_return};
+
+/// A prepared (host-side) file mapping ready to be applied to a VM.
+///
+/// Created by [`prepare_file_cow`]. The host-side OS resources (file
+/// mapping handle + view on Windows, mmap on Linux) are held here
+/// until consumed by the VM-side apply step.
+///
+/// If dropped without being consumed, the `Drop` impl releases all
+/// host-side resources — preventing leaks when an
+/// [`UninitializedSandbox`] is dropped without evolving or when
+/// apply fails.
+#[must_use = "holds OS resources that leak if discarded — apply to a VM or let Drop clean up"]
+pub(crate) struct PreparedFileMapping {
+    /// The guest address where this file should be mapped.
+    pub(crate) guest_base: u64,
+    /// The page-aligned size of the mapping in bytes.
+    pub(crate) size: usize,
+    /// Host-side OS resources. `None` after successful consumption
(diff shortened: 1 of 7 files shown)


src/hyperlight_host/src/sandbox/file_mapping.rs
```diff
@@ -0,0 +1,294 @@
...
+        let size = size.div_ceil(page_size) * page_size;
+        let base = unsafe {
+            libc::mmap(
+                std::ptr::null_mut(),
+                size,
+                libc::PROT_READ | libc::PROT_WRITE | libc::PROT_EXEC,
+                libc::MAP_PRIVATE,
+                file.as_raw_fd(),
+                0,
+            )
```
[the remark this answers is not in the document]
simongdavies: This is curent behaviour but I think this only needs to be READ.

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
```diff
@@ -71,14 +77,40 @@ pub(super) fn evolve_impl_multi_use(u_sbox: UninitializedSandbox) -> Result<Mult
...
 
     #[cfg(target_os = "linux")]
     setup_signal_handlers(&u_sbox.config)?;
 
+    // Apply any file mappings that were prepared before evolve.
+    // This must happen before vm.initialise() so the mapped data is
+    // visible to the guest's init code.
+    //
+    // RAII pattern: we collect successfully-mapped PreparedFileMappings
+    // and only mark_consumed() after ALL succeed. On early ?-return,
```
danbugs: Aren't we marking each one consumed immediately instead of after all suceeding?

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
```diff
@@ -71,14 +77,40 @@ pub(super) fn evolve_impl_multi_use(u_sbox: UninitializedSandbox) -> Result<Mult
...
+    // the Vec drops and each PreparedFileMapping::drop cleans up the
+    // host-side OS resources.
+    //
+    // SAFETY: On failure, host resources for already-mapped regions are
+    // released by Drop before the VM is torn down. No guest code
+    // executes between the error and VM drop, so the transient stale
+    // guest-physical mappings are harmless.
+    let pending = u_sbox.pending_file_mappings;
+    let mut mapped: Vec<super::file_mapping::PreparedFileMapping> =
```
danbugs: Isn't this vec not really used?

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
jsturtevant: Agreed this comment doesn't seem to align with the implementation below

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
```diff
@@ -71,14 +77,40 @@ pub(super) fn evolve_impl_multi_use(u_sbox: UninitializedSandbox) -> Result<Mult
...
+        Vec::with_capacity(pending.len());
+    for mut prepared in pending {
+        let region = prepared.to_memory_region()?;
+        unsafe { vm.map_region(&region) }.map_err(|e| {
+            crate::HyperlightError::HyperlightVmError(HyperlightVmError::MapRegion(e))
+        })?;
+        // Mapping succeeded: transfer host resource ownership to the VM
+        // immediately so Drop won't release them on later failures.
+        prepared.mark_consumed();
```
jsturtevant, line 108: what happens if `vm.initialise` below fails?

jsturtevant: now that the region has been passed to `map_region` does it do the clean up?

src/hyperlight_host/src/sandbox/file_mapping.rs
```diff
@@ -0,0 +1,309 @@
...
+                    handle_size: self.size,
+                    offset: 0,
+                };
+                let host_end =
+                    <crate::mem::memory_region::HostGuestMemoryRegion as MemoryRegionKind>::add(
+                        host_base, self.size,
+                    );
+                Ok(MemoryRegion {
+                    host_region: host_base..host_end,
+                    guest_region: self.guest_base as usize..self.guest_base as usize + self.size,
```
jsturtevant: should we have a check_add here?

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
simongdavies: yes this was from a previous iteration I will clean it up

src/hyperlight_host/src/sandbox/uninitialized_evolve.rs
simongdavies: It does on windows, for Linux see #1290 and #700 , it doesn't alter pre-existing behaviour

jsturtevant: Thanks for the info.  Seems to be same use case then this works for now

src/hyperlight_host/src/sandbox/file_mapping.rs
simongdavies: Done

PR #1322, opened 2026-03-18
feat: pass file mapping info via PEB with label support

simongdavies:
Add FileMappingInfo struct (guest_addr, size, label) and file_mappings field to HyperlightPEB so file mapping metadata is communicated to the guest through the PEB. Space for MAX_FILE_MAPPINGS (32) entries is statically reserved after the PEB struct to avoid dynamic layout changes.

- Add label parameter to map_file_cow (optional, defaults to filename)
- Add shared memory overlap validation (full mapped range)
- Add inter-mapping overlap detection
- Add write_file_mapping_entry on SandboxMemoryManager
- Add MAX_FILE_MAPPINGS limit enforcement at registration time
- Update PEB region sizing in get_memory_regions and layout tests

### DIFF
--- a/src/hyperlight_host/src/mem/layout.rs
+++ b/src/hyperlight_host/src/mem/layout.rs
@@ -63,7 +63,7 @@ limitations under the License.
 use std::fmt::Debug;
 use std::mem::{offset_of, size_of};
 
-use hyperlight_common::mem::{GuestMemoryRegion, HyperlightPEB, PAGE_SIZE_USIZE};
+use hyperlight_common::mem::{HyperlightPEB, PAGE_SIZE_USIZE};
 use tracing::{Span, instrument};
 
 use super::memory_region::MemoryRegionType::{Code, Heap, InitData, Peb};
@@ -92,6 +92,7 @@ pub(crate) struct SandboxMemoryLayout {
     peb_output_data_offset: usize,
     peb_init_data_offset: usize,
     peb_heap_data_offset: usize,
+    peb_file_mappings_offset: usize,
 
     guest_heap_buffer_offset: usize,
     init_data_offset: usize,
@@ -140,6 +141,10 @@ impl Debug for SandboxMemoryLayout {
             .field(
                 "Guest Heap Offset",
                 &format_args!("{:#x}", self.peb_heap_data_offset),
+            )
+            .field(
+                "File Mappings Offset",
+                &format_args!("{:#x}", self.peb_file_mappings_offset),
             )
             .field(
                 "Guest Heap Buffer Offset",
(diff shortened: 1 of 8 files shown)


src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -545,33 +545,98 @@ impl MultiUseSandbox {
...
-        // Successfully mapped — transfer host resource ownership to
-        // the VM layer.
         let size = prepared.size as u64;
+
+        // Record the mapping metadata in the PEB so the guest can
+        // discover it. Must happen before mark_consumed() so that if
+        // this fails, Drop still cleans up the host resources.
+        self.mem_mgr
+            .write_file_mapping_entry(prepared.guest_base, size, &prepared.label)?;
+
```
jsturtevant: If `write_file_mapping_entry` fails here, `prepared` hasn't been consumed yet so its `Drop` frees the host-side file resources — but the VM already holds a mapping from `map_region` above, leaving it with a dangling backing. The evolve path in `uninitialized_evolve.rs` avoids this by calling `mark_consumed()` immediately after `map_region`, then `write_file_mapping_entry`. Consider matching that order here.

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -351,6 +351,60 @@ impl SandboxMemoryManager<ExclusiveSharedMemory> {
...
+                MAX_FILE_MAPPINGS,
+            ));
+        }
+
+        // Write the entry into the next available slot.
+        let entry_offset = self.layout.get_file_mappings_array_offset()
+            + current_count * std::mem::size_of::<FileMappingInfo>();
+        self.shared_mem.write::<u64>(entry_offset, guest_addr)?;
+        self.shared_mem.write::<u64>(entry_offset + 8, size)?;
+        self.shared_mem.copy_from_slice(label, entry_offset + 16)?;
```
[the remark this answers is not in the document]
simongdavies: fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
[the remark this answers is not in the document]
simongdavies: done

simongdavies: fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -545,36 +545,107 @@ impl MultiUseSandbox {
...
+                    "map_file_cow: guest address overflow: {:#x} + {:#x}",
+                    guest_base, prepared.size
+                ))
+            })?;
+        if guest_base < shared_end && mapping_end > base_addr {
+            return Err(crate::HyperlightError::Error(format!(
+                "map_file_cow: mapping [{:#x}..{:#x}) overlaps sandbox shared memory [{:#x}..{:#x})",
+                guest_base, mapping_end, base_addr, shared_end,
+            )));
+        }
```
danbugs, line 606: nit: aren't these checks redundant w/ the check below w/ `self.vm.get_mapped_regions()`?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -588,6 +631,18 @@ impl SandboxMemoryLayout {
         shared_mem.write_u64(self.get_heap_size_offset(), self.heap_size.try_into()?)?;
         shared_mem.write_u64(self.get_heap_pointer_offset(), addr)?;
 
+        // Set up the file_mappings descriptor in the PEB.
+        // - The `size` field holds the number of valid FileMappingInfo
+        //   entries currently written (initially 0 — entries are added
```
danbugs, line 636: question: so size doesn't hold the size in bytes, yes? We're repurposing the field to hold entry counts from what I understand?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -545,36 +545,107 @@ impl MultiUseSandbox {
...
+        // freeing it here would leave a dangling backing.
         prepared.mark_consumed();
         self.mem_mgr.mapped_rgns += 1;
 
+        // Record the mapping metadata in the PEB. If this fails the VM
+        // still holds a valid mapping but the PEB won't list it — the
+        // limit was already pre-checked above so this should not fail
+        // in practice.
+        self.mem_mgr
+            .write_file_mapping_entry(prepared.guest_base, size, &prepared.label)?;
```
danbugs, line 647: So, by this point, we already have mark_consumed called and mapped_rngs has been incremented, meaning the VM holds a valid mapping, but we haven't added it to the PEB yet. `write_file_mapping_entry` is fallible, but I see above you said it should not fail due to pre-checks. Just making sure--is there nothing else that could fail in this operation (e.g., bounds error?) that could get us in a bad state?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: If `vm.get_mapped_regions()` covers the entire range then yes they are redundant

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
simongdavies: In theory at this stage this function is infallible , but the worst that could happen is that we dont record a mapping that we have made, things would still get cleaned up correctly.

src/hyperlight_host/src/mem/layout.rs
simongdavies: Effectively yes, the size is fixed, I could add the index as an additional field and retain the size to make it absolutely clear, if someone were to add more data here and use the value in here for layout or calculation it would be wrong

PR #1327, opened 2026-03-19
fix: add wasmtime flags! macro support for WIT flags types

### LINKED ISSUE #1318: wasm_guest_bindgen! macro doesn't support WIT flags types (needed for wasi:filesystem)
### What happened?

The wasm_guest_bindgen! macro in hyperlight_wasm_macro fails to compile when the WIT world includes flags type definitions (e.g. from wasi:filesystem/types@0.2.0).

Error:
```
error[E0277]: the trait bound `DescriptorFlags: wasmtime::component::Lower` is not satisfied
error[E0277]: the trait bound `PathFlags: wasmtime::component::Lift` is not satisfied
error[E0277]: the trait bound `OpenFlags: wasmtime::component::Lift` is not satisfied
```

WIT that triggers the error:

```
flags descriptor-flags {
  read,
  write,
  file-integrity-sync,
  data-integrity-sync,
  requested-write-sync,
  mutate-directory,
}

flags path-flags {
  symlink-follow,
}

flags open-flags {
  create,
  directory,
  exclusive,
  truncate,
}
```

### Steps to Reproduce

Reproduction

1. Create a WIT with `wasi:filesystem/types` including `flags descriptor-flags { ... }`
2. Compile with `wasm-tools component wit ... -w -o world.wasm`
3. Build with `WIT_WORLD=world.wasm cargo build -p hyperlight-wasm`
4. Runtime sub-build fails with the `Lift`/`Lower` errors above

### Expected Results

Without `flags` support, it's impossible to use `wasi:filesystem` with hyperlight-wasm component model support.

### Actual Results

TODO: What actually happened?

### Versions and Environment

#### OS Version

Linux:
```console
```

Windows (PowerShell):
```powershell
cmd /c ver
```

#### Hypervisor

Linux:
```console
```

Windows (Admin PowerShell):
```powershell
```

### Extra Info

- `hyperlight-wasm-sockets-example` — works around by omitting `flags` from WIT
- `hyperlight-wasm-http-example` — avoids the issue by not importing `wasi:filesystem` at all

jsturtevant:
When `is_wasmtime_guest` is true, emit `wasmtime::component::flags!` macro invocations instead of plain structs with bool fields. This enables WIT flags types (e.g. from `wasi:filesystem`) to satisfy the `wasmtime::component::Lift` and `wasmtime::component::Lower` trait bounds. See https://docs.rs/wasmtime/latest/wasmtime/component/macro.flags.html

The non-Wasmtime code path used by `host_bindgen!` and Hyperlight `guest_bindgen!` in this repo is unchanged. The Wasmtime path is used by downstream Wasmtime guest-bindgen consumers.

### WIT flags input

The tests use the existing WIT flags in `src/tests/rust_guests/witguest/guest.wit`:

```wit
interface roundtrip {
    flags smallflags {
        flag-a,
        flag-b,
        flag-c,
    }

    roundtrip-flags-small: func(x: smallflags) -> smallflags;
}
```

### Expected non-Wasmtime Rust output

`host_bindgen!` and Hyperlight `guest_bindgen!` keep generating a bool-field struct:

```rust
#[derive(Debug, Clone, PartialEq)]
pub struct Smallflags {
    pub flag_a: bool,
    pub flag_b: bool,
    pub flag_c: bool,
}
```

### Expected Wasmtime guest Rust output

The Wasmtime guest generator emits the real Wasmtime flags macro and preserves the original WIT names on each constant:

```rust
::wasmtime::component::flags! {
    Smallflags {
        #[component(name = "flag-a")]
        const FLAG_A;

        #[component(name = "flag-b")]
        const FLAG_B;

        #[component(name = "flag-c")]
        const FLAG_C;
    }
}
```

### Expected marshal/unmarshal output

Because the type produced by `wasmtime::component::flags!` is bitflag-style, the Hyperlight-generated Wasmtime guest marshal/unmarshal code cannot access fields like `value.flag_a`. Note: I did consider just porting over to useing bitflag-style everywhere but that's a much bigger change.

For non-Wasmtime bindings, Hyperlight still generates code that reads and writes bool fields:

```rust
// Unmarshal bytes into the bool-field struct.
(
    Smallflags {
        flag_a: (bytes[0] >> 0) & 0x1 == 1,
        flag_b: (bytes[0] >> 1) & 0x1 == 1,
        flag_c: (bytes[0] >> 2) & 0x1 == 1,
    },
    1,
)

// Marshal the bool-field struct into bytes.
let mut bytes = [0; 1];
bytes[0] |= (if value.flag_a { 1 } else { 0 }) << 0;
bytes[0] |= (if value.flag_b { 1 } else { 0 }) << 1;
bytes[0] |= (if value.flag_c { 1 } else { 0 }) << 2;
alloc::vec::Vec::from(bytes)
```

For Wasmtime guest bindings, Hyperlight now generates code that uses the Wasmtime flags API:

```rust
// Unmarshal bytes into the Wasmtime flags value.
{
    let mut value_flags = Smallflags::empty();
    if (bytes[0] >> 0) & 0x1 == 1 {
        value_flags |= Smallflags::FLAG_A;
    }
    if (bytes[0] >> 1) & 0x1 == 1 {
        value_flags |= Smallflags::FLAG_B;
    }
    if (bytes[0] >> 2) & 0x1 == 1 {
        value_flags |= Smallflags::FLAG_C;
    }
    (value_flags, 1)
}

// Marshal the Wasmtime flags value into bytes.
let mut bytes = [0; 1];
bytes[0] |= (if value.contains(Smallflags::FLAG_A) { 1 } else { 0 }) << 0;
bytes[0] |= (if value.contains(Smallflags::FLAG_B) { 1 } else { 0 }) << 1;
bytes[0] |= (if value.contains(Smallflags::FLAG_C) { 1 } else { 0 }) << 2;
alloc::vec::Vec::from(bytes)
```

Closes #1318

### DIFF
--- a/src/hyperlight_component_util/src/hl.rs
+++ b/src/hyperlight_component_util/src/hl.rs
@@ -18,7 +18,7 @@ use itertools::Itertools;
 use proc_macro2::{Ident, TokenStream};
 use quote::{format_ident, quote};
 
-use crate::emit::{ResolvedBoundVar, State, kebab_to_cons, kebab_to_var};
+use crate::emit::{ResolvedBoundVar, State, kebab_to_cons, kebab_to_flags_const, kebab_to_var};
 use crate::etypes::{self, Defined, Handleable, Tyvar, Value};
 use crate::rtypes;
 
@@ -91,16 +91,37 @@ pub fn emit_hl_unmarshal_toplevel_value(
         }
         Value::Flags(ns) => {
             let bytes = usize::div_ceil(ns.len(), 8);
+            let result_var = format_ident!("{}_flags", id);
             let fields = ns.iter().enumerate().map(|(i, n)| {
                 let byte_offset = i / 8;
                 let bit_offset = i % 8;
-                let fieldid = kebab_to_var(n.name);
-                quote! {
-                    #fieldid: (#id[#byte_offset] >> #bit_offset) & 0x1 == 1,
+                let is_set = quote! { (#id[#byte_offset] >> #bit_offset) & 0x1 == 1 };
+                if s.is_wasmtime_guest {
+                    let const_name = kebab_to_flags_const(n.name);
+                    quote! {
+                        if #is_set {
+                            #result_var |= #tname::#const_name;
+                        }
+                    }
+                } else {
+                    let fieldid = kebab_to_var(n.name);
+                    quote! {
+                        #fieldid: #is_set,
+                    }
                 }
             });
-            quote! {
-                (#tname { #(#fields)* }, #bytes)
+            if s.is_wasmtime_guest {
+                quote! {
+                    {
+                        let mut #result_var = #tname::empty();
+                        #(#fields)*
+                        (#result_var, #bytes)
+                    }
+                }
+            } else {
+                quote! {
+                    (#tname { #(#fields)* }, #bytes)
+                }
             }
         }
         Value::Variant(vcs) => {
@@ -434,9 +455,15 @@ pub fn emit_hl_marshal_toplevel_value(
                 .map(|(i, n)| {
                     let byte_offset = i / 8;
                     let bit_offset = i % 8;
-                    let fieldid = kebab_to_var(n.name);
+                    let is_set = if s.is_wasmtime_guest {
+                        let const_name = kebab_to_flags_const(n.name);
+                        quote! { #id.contains(#tname::#const_name) }
+                    } else {
+                        let fieldid = kebab_to_var(n.name);
+                        quote! { #id.#fieldid }
+                    };
                     quote! {
-                        bytes[#byte_offset] |= (if #id.#fieldid { 1 } else { 0 }) << #bit_offset;
+                        bytes[#byte_offset] |= (if #is_set { 1 } else { 0 }) << #bit_offset;
                     }
                 })
                 .collect::<Vec<_>>();
--- a/src/hyperlight_component_util/tests/wasmtime_guest_codegen.rs
+++ b/src/hyperlight_component_util/tests/wasmtime_guest_codegen.rs
@@ -1,0 +1,47 @@
+/*
+Copyright 2026 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use hyperlight_component_util::{emit, guest, rtypes, util};
+
+#[test]
+fn wasmtime_guest_codegen_emits_wasmtime_flags_macro() {
+    let generated = util::read_wit_type_from_file(
+        "../tests/rust_guests/witguest/interface.wasm",
+        None,
+        |kebab_name, ct| {
+            // Mirrors the hyperlight-wasm guest-bindgen expansion path:
+            // https://github.com/hyperlight-dev/hyperlight-wasm/blob/81e72f5920ebc23584097abfe24d05a40bf084cc/src/hyperlight_wasm_macro/src/lib.rs#L37-L38
+            emit::run_state(true, true, |s| {
+                rtypes::emit_toplevel(s, &kebab_name, ct);
+                guest::emit_toplevel(s, &kebab_name, ct);
+            })
+        },
+    );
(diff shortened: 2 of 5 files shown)


jsturtevant:
moved this to draft, I was integrating into hyperlight-wasm and found the usage of flags needs to be slightly different. than this.  will push an update shortly

src/tests/rust_guests/witguest/bindgen-test-cases/world.wit
```diff
@@ -5,9 +5,15 @@ world bindgen-test-cases {
 }
 
 interface executor {
+    flags execution-flags {
+        fast-path,
+        guest-visible,
```
jsturtevant: i still don't know what these really mean, could we just call them option a and b

src/hyperlight_component_util/tests/wasmtime_guest_codegen.rs
```diff
@@ -0,0 +1,47 @@
...
+    assert!(generated.contains("Smallflags {"));
+    assert!(generated.contains("\"flag-a\""));
+    assert!(generated.contains("const FLAG_A;"));
+    assert!(generated.contains("\"flag-b\""));
+    assert!(generated.contains("const FLAG_B;"));
+    assert!(generated.contains("\"flag-c\""));
+    assert!(generated.contains("const FLAG_C;"));
+    assert!(!generated.contains("pub flag_a: bool"));
+    assert!(!generated.contains("pub flag_b: bool"));
+    assert!(!generated.contains("pub flag_c: bool"));
```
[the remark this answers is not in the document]
jsturtevant: eh, I considered this but feels not quite right. There isn't really a great way to test this e2e in this repo since it really requires a whole wasmtime guest.  What do others think?

src/hyperlight_component_util/src/hl.rs
```diff
@@ -91,16 +91,37 @@ pub fn emit_hl_unmarshal_toplevel_value(
...
+                        #fieldid: (#id[#byte_offset] >> #bit_offset) & 0x1 == 1,
+                    }
+                });
+                quote! {
+                    (#tname { #(#fields)* }, #bytes)
                 }
-            });
-            quote! {
-                (#tname { #(#fields)* }, #bytes)
             }
```
ludfjig: nit: could potentially be simplified to something like this

```suggestion
            let result_var = format_ident!("{}_flags", id);
            let fields = ns.iter().enumerate().map(|(i, n)| {
                let byte_offset = i / 8;
                let bit_offset = i % 8;
                let is_set = quote! { (#id[#byte_offset] >> #bit_offset) & 0x1 == 1 };
                if s.is_wasmtime_guest {
                    let const_name = kebab_to_flags_const(n.name);
                    quote! {
                        if #is_set {
                            #result_var |= #tname::#const_name;
                        }
                    }
                } else {
                    let fieldid = kebab_to_var(n.name);
                    quote! {
                        #fieldid: #is_set,
                    }
                }
            });
            if s.is_wasmtime_guest {
                quote! {
                    {
                        let mut #result_var = #tname::empty();
                        #(#fields)*
                        (#result_var, #bytes)
                    }
                }
            } else {
                quote! {
                    (#tname { #(#fields)* }, #bytes)
                }
            }
```

src/hyperlight_component_util/src/hl.rs
```diff
@@ -428,22 +449,42 @@ pub fn emit_hl_marshal_toplevel_value(
...
+                            bytes[#byte_offset] |= (if #id.#fieldid { 1 } else { 0 }) << #bit_offset;
+                        }
+                    })
+                    .collect::<Vec<_>>();
+                quote! {
+                    let mut bytes = [0; #bytes];
+                    #(#fields)*
+                    alloc::vec::Vec::from(bytes)
+                }
             }
```
ludfjig: nit: could potentially be simplified tos omething like this

```suggestion
            let fields = ns
                .iter()
                .enumerate()
                .map(|(i, n)| {
                    let byte_offset = i / 8;
                    let bit_offset = i % 8;
                    let is_set = if s.is_wasmtime_guest {
                        let const_name = kebab_to_flags_const(n.name);
                        quote! { #id.contains(#tname::#const_name) }
                    } else {
                        let fieldid = kebab_to_var(n.name);
                        quote! { #id.#fieldid }
                    };
                    quote! {
                        bytes[#byte_offset] |= (if #is_set { 1 } else { 0 }) << #bit_offset;
                    }
                })
                .collect::<Vec<_>>();
            quote! {
                let mut bytes = [0; #bytes];
                #(#fields)*
                alloc::vec::Vec::from(bytes)
            }
```

ludfjig (dismissed):
LGTM!

src/hyperlight_component_util/src/hl.rs
jsturtevant: thanks that's much cleaner!

ludfjig (approved):
lgtm

PR #1331, opened 2026-03-20
fix: handle empty namespace paths in component codegen

### LINKED ISSUE #1329: wit: ﻿empty namespaces
### What happened?

When a WIT interface name has no package namespace (e.g. a plain `"exports"` interface rather than `wasi:http/types`), the code generator produces invalid Rust paths with a leading `::` separator:

```rust
// Generated (broken):
impl<I: ...> ::Types<...> for ...

// Should be:
impl<I: ...> Types<...> for ...
```

### Steps to Reproduce

Use a WIT world with unqualified interface names (no package prefix):

```wit
world root {
  import wasi:filesystem/types@0.2.0;
  export executor;  // no package namespace
}
```

### Expected Results

Generates valid Rust: `impl<I: ...> Types<...> for ...`

### Actual Results

TODO: What actually happened?

### Versions and Environment

#### OS Version

Linux:
```console
```

Windows (PowerShell):
```powershell
cmd /c ver
```

#### Hypervisor

Linux:
```console
```

Windows (Admin PowerShell):
```powershell
```

### Extra Info

jsturtevant:
Fixes #1329.

This updates component code generation so WIT interfaces without a namespace emit local Rust paths instead of paths with an invalid leading separator.

It also adds the shared `bindgen-test-cases` WIT fixture and CI artifact plumbing so stacked bindgen-focused PRs can extend the same fixture instead of adding one-off wasm artifacts. This is an simple with test case but it gets tricky to add all of the examples there so split this out to catch all the edge cases.

### DIFF
--- a/.github/workflows/dep_build_guests.yml
+++ b/.github/workflows/dep_build_guests.yml
@@ -113,3 +113,12 @@ jobs:
           path: src/tests/rust_guests/witguest/twoworlds.wasm
           retention-days: 1
           if-no-files-found: error
+
+      - name: Upload bindgen-test-cases.wasm
+        if: inputs.config == 'debug'
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a  # v7.0.1
+        with:
+          name: bindgen-test-cases-wasm
+          path: src/tests/rust_guests/witguest/bindgen-test-cases.wasm
+          retention-days: 1
+          if-no-files-found: error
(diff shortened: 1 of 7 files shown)


dblnz (commented):
Great to see this change!
It's been bugging me for some time when working on component debugging .

I am still new to this part of the codebase, maybe someone with more insight can have a look.

andreiltd (approved):

ludfjig (approved):

PR #1338, opened 2026-03-24
moving a few more modules to tracing crate from log crate

vshailesh:


vshailesh:
except for tsc invariant functionality, I have tested other things on local machine.

dblnz:
It looks like CI is failing because of an unrelated problem. I've created issue #1341 to track it.
After a re-run, the CI checks passed. We'll continue monitoring the issue

PR #1339, opened 2026-03-24
feat: Windows SHA-stamped surrogate filename and configurable surrogate pool size

simongdavies:
Fixes hyperlight surrogate process manager on Windows:

If two different implementations of hyperlight are used by a single host (e.g. hyperlight and hyperlight-js) then they may (even if this is highly unlikely) have different versions of the hyperlight surrogate binary (if the hyperlight versions are different). More likely is that they use the same version, when this happens, then the second implementation will try to overwrite the in already in use surrogate process exe.

If there are multiple implementations of hyperlight each with their own surrogate process manager then under current behavior each one will spin up 512 surrogate process, not only does this waste resources and take time it also means that there will be 1024 which is more that can be supported in a single process.

The issue with file copying prevents hyperagent from running on Windows (as it uses both hyperlight and hyperlight-js). It also does not need the overhead of 512 surrogate processes.

There are other scenarios where hyperlight may be used where this upfront creation of surrogate processes is both unnecessary and wasteful. 

This PR introduces the following changes to deal with this:

- Use SHA-256 content hash in surrogate binary filename (hyperlight_surrogate_{sha8}.exe) to eliminate cross-version ACCESS_DENIED race when multiple hyperlight versions coexist.

- Add HYPERLIGHT_INITIAL_SURROGATES env var (1-512, default 512) to control how many surrogate processes are pre-created at startup.

- Add HYPERLIGHT_MAX_SURROGATES env var (>=initial, <=512, default 512) as hard cap with on-demand CAS growth when pool is exhausted.

- Add tests for env var parsing (with #[serial] for thread safety) and locked-file extraction resilience.

- Update surrogate development notes documentation.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
+++ b/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
@@ -19,10 +19,11 @@ use std::fs::File;
 use std::io::Write;
 use std::mem::size_of;
 use std::path::{Path, PathBuf};
+use std::sync::atomic::{AtomicUsize, Ordering};
 
-use crossbeam_channel::{Receiver, Sender, unbounded};
+use crossbeam_channel::{Receiver, Sender, TryRecvError, unbounded};
 use rust_embed::RustEmbed;
-use tracing::{Span, error, info, instrument};
+use tracing::{Span, error, info, instrument, warn};
 use windows::Win32::Foundation::HANDLE;
 use windows::Win32::Security::SECURITY_ATTRIBUTES;
 use windows::Win32::System::JobObjects::{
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -49,14 +50,70 @@ use crate::{Result, log_then_return, new_error};
...
+const MAX_SURROGATES_ENV_VAR: &str = "HYPERLIGHT_MAX_SURROGATES";
+
+/// Returns the on-disk filename for the surrogate binary, incorporating the
+/// first 8 hex characters of the SHA-256 of the embedded binary so that
+/// different hyperlight versions produce different filenames and can coexist
+/// without file-deletion races.
+fn surrogate_binary_name() -> Result<String> {
+    let exe = Asset::get(EMBEDDED_SURROGATE_NAME)
+        .ok_or_else(|| new_error!("could not find embedded surrogate binary"))?;
+    let sha = sha256::digest(exe.data.as_ref());
```
ludfjig: Have you considered making this random instead? Then it would additionally cover the case where you run 2 hyperlight of the same versino

simongdavies: This is already covered since we no longer fail if the copy cannot proceed, the challenge with random names is two fold ,  first it will potentially fill a disk up over time, at least it will leave a lot of redundant files around, second its poentiallymore likely to trigger AV, there are a couple more issues here that need to be solved that I will open issues for so I think for now I will leave it as is.

simongdavies:
> Did you consider making these knobs on SandboxConfigurtion instead?

Yes, but since I wanted the values to apply equally to any surrogate process managers in the process there isn't a place to put it, the values are also not SandboxConfiguration, I think maybe when we revise the API we can look at this again

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
simongdavies: See #1342

ludfjig:
> > Did you consider making these knobs on SandboxConfigurtion instead?
> 
> Yes, but since I wanted the values to apply equally to any surrogate process managers in the process there isn't a place to put it, the values are also not SandboxConfiguration, I think maybe when we revise the API we can look at this again

Hmm not sure I understand this? You really want the config per process not per sandbox?

simongdavies:
> > > Did you consider making these knobs on SandboxConfigurtion instead?
> > 
> > 
> > Yes, but since I wanted the values to apply equally to any surrogate process managers in the process there isn't a place to put it, the values are also not SandboxConfiguration, I think maybe when we revise the API we can look at this again
> 
> Hmm not sure I understand this? You really want the config per process not per sandbox?

The config cannot be per sandbox , all sandboxes in a single hyperlight implementation share the same surrogate process manager and surrogate processes

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -622,4 +770,161 @@ mod tests {
...
+
+        // File is still present and unchanged.
+        assert!(path.exists(), "surrogate binary should still exist");
+
+        // Release the file handle.
+        drop(file_lock);
+
+        // Verify the name is deterministic (same embedded binary → same SHA).
+        let binary_name_2 = surrogate_binary_name().expect("second call should also succeed");
+        assert_eq!(
```
jsturtevant, line 841: not sure what this tests, seem like it would make more sense to see if we can test if `ensure_surrogate_process_exe` passes/fails if there isn't a lock? or that we create a new one if the previous one got deleted?

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -247,35 +400,27 @@ fn get_surrogate_process_dir() -> Result<PathBuf> {
...
-                embedded_file_sha,
-                file_on_disk_sha,
-                &surrogate_process_path.display()
-            );
-            std::fs::remove_file(p)?;
-        }
-    }
-
+    // The filename already contains the SHA, so if the file exists it is
+    // guaranteed to have the correct content — no need to re-verify.
```
jsturtevant: what we lost here is that if the write fails below, we will have a file on disk with the same name so we will skip but it could be corrupted since the write failed half way through. Not critical probably but sha check prevented this previously

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -49,14 +50,94 @@ use crate::{Result, log_then_return, new_error};
...
+    let mut initial = raw_initial
+        .map(|n| n.clamp(1, HARD_MAX_SURROGATE_PROCESSES))
+        .unwrap_or(HARD_MAX_SURROGATE_PROCESSES);
+
+    let max = raw_max
+        .map(|n| n.clamp(1, HARD_MAX_SURROGATE_PROCESSES))
+        .unwrap_or(HARD_MAX_SURROGATE_PROCESSES);
+
+    // Max is authoritative — you can't pre-create more than the limit allows.
+    if initial > max {
```
dblnz: Nit: Maybe clamping `initial` to `1 .. =max` after clamping `max` to remove the need of this if?

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
simongdavies: done

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
simongdavies: done

simongdavies:
> LGTM. nit: there's a potential TOCTOU issue with `if !p.exists() { ...` where another process can create the file inbetween the check. Could be fixed with something like https://doc.rust-lang.org/std/fs/struct.File.html#method.create_new, but up to you, not critical.
> 
> We also seem to be using both `blake3` and `sha256`, maybe we can remove 1 of these dependencies? Not blocking this PR though

Fixed

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
simongdavies: Good idea, done.

PR #1340, opened 2026-03-24
feat: increase MAX_MEMORY_SIZE from ~1 GiB to ~16 GiB

simongdavies:
The previous limit of 0x4000_0000 - BASE_ADDRESS (~1 GiB) was arbitrary and insufficient for some Nnanvix related cases. Bump to (16 GiB - BASE_ADDRESS) which is large enough for most use cases while still preventing accidental resource exhaustion.

The BASE_ADDRESS subtraction is preserved for consistency with the original definition.

Update test to validate the new ~16 GiB boundary.

### DIFF
--- a/src/hyperlight_host/src/mem/layout.rs
+++ b/src/hyperlight_host/src/mem/layout.rs
@@ -170,13 +170,10 @@ impl Debug for SandboxMemoryLayout {
 impl SandboxMemoryLayout {
     /// The maximum amount of memory a single sandbox will be allowed.
     ///
-    /// Currently, both the scratch region and the snapshot region are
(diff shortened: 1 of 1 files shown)


simongdavies:
> LGTM! Could we add a unit test that exercises this too by creating a sandbox slightly smaller than 16GB and running it?

One reason I didn't do this is that I wasnt 100% sure how much memory would actually get allocated if I did this and if it would cause tests to fail on dev machines..

ludfjig:
oh yeah maybe this would actually allocate until cow work is complete

ludfjig (approved):

PR #1360, opened 2026-04-03
Use WHvResetPartition on windows

ludfjig:
Addresses #791 for windows (WHP), by resetting all guest visible state on the partition. Resets more vm state on `restore()` at the expense of some performance (roughly ~100micros). MSHV is planned to use the same hv call, but doesn't exist in mshv crate yet, although dom0 kernel should have support already. KVM doesn't have a nice api like this, so will still manual work for KVM.

Note: the second commit is a workaround for a hyper-v bug

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
@@ -40,9 +40,9 @@ use crate::hypervisor::gdb::{
 };
 #[cfg(gdb)]
 use crate::hypervisor::gdb::{DebugError, DebugMemoryAccessError};
-use crate::hypervisor::regs::{
-    CommonDebugRegs, CommonFpu, CommonRegisters, CommonSpecialRegisters,
-};
+#[cfg(not(target_os = "windows"))]
+use crate::hypervisor::regs::CommonDebugRegs;
+use crate::hypervisor::regs::{CommonFpu, CommonRegisters, CommonSpecialRegisters};
 #[cfg(not(gdb))]
 use crate::hypervisor::virtual_machine::VirtualMachine;
 #[cfg(kvm)]
@@ -335,23 +335,37 @@ impl HyperlightVm {
     }
 
     /// Resets the following vCPU state:
-    /// - General purpose registers
-    /// - Debug registers
-    /// - XSAVE (includes FPU/SSE state with proper FCW and MXCSR defaults)
-    /// - Special registers (restored from snapshot, with CR3 updated to new page table location)
+    /// - On Windows: calls WHvResetPartition (resets all per-VP state including
+    ///   GP registers, debug registers, XSAVE, MSRs, APIC, etc.)
+    /// - On Linux: explicitly resets GP registers, debug registers, and XSAVE
+    ///
+    /// This does NOT restore special registers (except on windows). Call `restore_sregs` separately
+    /// after memory mappings are established.
     // TODO: check if other state needs to be reset
-    pub(crate) fn reset_vcpu(
+    pub(crate) fn reset_vm_state(&mut self) -> std::result::Result<(), RegisterError> {
+        #[cfg(target_os = "windows")]
+        self.vm.reset_partition()?;
+
+        #[cfg(not(target_os = "windows"))]
+        {
+            self.vm.set_regs(&CommonRegisters {
+                rflags: 1 << 1, // Reserved bit always set
+                ..Default::default()
+            })?;
+            self.vm.set_debug_regs(&CommonDebugRegs::default())?;
+            self.vm.reset_xsave()?;
+        }
+
+        Ok(())
+    }
+
+    /// Restores special registers from snapshot with CR3 updated to the
+    /// new page table location.
+    pub(crate) fn restore_sregs(
         &mut self,
         cr3: u64,
         sregs: &CommonSpecialRegisters,
     ) -> std::result::Result<(), RegisterError> {
-        self.vm.set_regs(&CommonRegisters {
-            rflags: 1 << 1, // Reserved bit always set
-            ..Default::default()
-        })?;
-        self.vm.set_debug_regs(&CommonDebugRegs::default())?;
-        self.vm.reset_xsave()?;
-
         #[cfg(not(feature = "nanvix-unstable"))]
         {
             // Restore the full special registers from snapshot, but update CR3
@@ -885,7 +899,9 @@ mod tests {
     use super::*;
     #[cfg(kvm)]
     use crate::hypervisor::regs::FP_CONTROL_WORD_DEFAULT;
-    use crate::hypervisor::regs::{CommonSegmentRegister, CommonTableRegister, MXCSR_DEFAULT};
+    use crate::hypervisor::regs::{
+        CommonDebugRegs, CommonSegmentRegister, CommonTableRegister, MXCSR_DEFAULT,
+    };
     use crate::hypervisor::virtual_machine::VirtualMachine;
     use crate::mem::layout::SandboxMemoryLayout;
     use crate::mem::memory_region::{GuestMemoryRegion, MemoryRegionFlags};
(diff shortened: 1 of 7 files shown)


src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -335,23 +335,37 @@ impl HyperlightVm {
...
-    /// - Debug registers
-    /// - XSAVE (includes FPU/SSE state with proper FCW and MXCSR defaults)
-    /// - Special registers (restored from snapshot, with CR3 updated to new page table location)
+    /// - On Windows: calls WHvResetPartition (resets all per-VP state including
+    ///   GP registers, debug registers, XSAVE, MSRs, APIC, etc.)
+    /// - On Linux: explicitly resets GP registers, debug registers, and XSAVE
+    ///
+    /// This does NOT restore special registers (except on windows). Call `restore_sregs` separately
+    /// after memory mappings are established.
     // TODO: check if other state needs to be reset
```
simongdavies, line 342: Is this TODO still valid (presumably not on Windows) is there an issue to track this

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -335,23 +335,37 @@ impl HyperlightVm {
...
     /// Resets the following vCPU state:
-    /// - General purpose registers
-    /// - Debug registers
-    /// - XSAVE (includes FPU/SSE state with proper FCW and MXCSR defaults)
-    /// - Special registers (restored from snapshot, with CR3 updated to new page table location)
+    /// - On Windows: calls WHvResetPartition (resets all per-VP state including
+    ///   GP registers, debug registers, XSAVE, MSRs, APIC, etc.)
+    /// - On Linux: explicitly resets GP registers, debug registers, and XSAVE
+    ///
+    /// This does NOT restore special registers (except on windows). Call `restore_sregs` separately
```
simongdavies, line 340: This seems a bit fragile, why can't we call `restore_sregs` from here?

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -1206,16 +1222,44 @@ mod tests {
...
+            // WHvResetPartition sets x86 power-on reset values
+            // (Intel SDM Vol. 3, Table 10-1)
+            let expected = CommonRegisters {
+                rip: 0xFFF0,   // Reset vector
+                rdx: regs.rdx, // CPUID signature (CPU-dependent)
+                rflags: 0x2,   // Reserved bit 1
+                ..Default::default()
+            };
+            assert_ne!(
+                regs.rdx, 0x4444444444444444,
```
simongdavies, line 1255: What's the significance of this value?

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -107,7 +107,7 @@ pub(crate) enum HypervisorType {
 /// Minimum XSAVE buffer size: 512 bytes legacy region + 64 bytes header.
 /// Only used by MSHV and WHP which use compacted XSAVE format and need to
```
simongdavies, line 108: Comment needs updating

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -87,6 +86,145 @@ fn release_file_mapping(view_base: *mut c_void, mapping_handle: HandleWrapper) {
...
 
+/// Workaround for a Hyper-V bug on AMD SVM where the `ObScrubPartition`
+/// path (invoked by `WHvResetPartition`) does not flush the Nested Page
+/// Table (NPT) TLB. After reset, stale GPA-to-HPA translations from
+/// the previous guest execution persist, causing the guest to read
+/// wrong physical memory. Intel is unaffected.
+///
+/// This bug most commonly surfaces as the `interrupt_same_thread_no_barrier`
+/// integration test failing on AMD Windows hosts. @ludfjig has verified
+/// the fix works on Windows Insider builds; this workaround can be
```
simongdavies, line 130: Can we get an issue to track this? Even better is there a way we can have a test that fails when this is fixed in windows? That will force us then to gate this code on a specific windows version or lower since I think the comment is wrong, we will need this to endure but will only need to apply it on impacted versions of windows

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -87,6 +86,145 @@ fn release_file_mapping(view_base: *mut c_void, mapping_handle: HandleWrapper) {
...
+/// removed once that fix ships in a released version of Windows.
+///
+/// The workaround unmaps and remaps a dummy page after each
+/// `WHvResetPartition`, which forces the hypervisor to invalidate
+/// NPT entries.
+mod npt_flush {
```
simongdavies, line 136: can we also gate this on affected CPUs only?

simongdavies (changes requested):
Looks good , just a few comments, it would be good to tighten up the AMD fix targeting and also see if we can get rid of the need to call sregs reset independently

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -335,23 +335,37 @@ impl HyperlightVm {
...
-    /// - Special registers (restored from snapshot, with CR3 updated to new page table location)
+    /// - On Windows: calls WHvResetPartition (resets all per-VP state including
+    ///   GP registers, debug registers, XSAVE, MSRs, APIC, etc.)
+    /// - On Linux: explicitly resets GP registers, debug registers, and XSAVE
+    ///
+    /// This does NOT restore special registers (except on windows). Call `restore_sregs` separately
+    /// after memory mappings are established.
     // TODO: check if other state needs to be reset
-    pub(crate) fn reset_vcpu(
+    pub(crate) fn reset_vm_state(&mut self) -> std::result::Result<(), RegisterError> {
```
syntactically: It feels a bit of a shame to have this (ought-to-be-mostly-platform-independent) code making the call about whether or not to use `reset_partition`. Perhaps it would make more sense to either have a `vm.reset_partition_supported()` or have `vm.reset_partition()` return a particular failure value that we can see and fall back to the alternative approach here? (I assume the reason why the fallback is here and not in each individual HV driver is that the fallback should be more-or-less the same on all platforms of the same architecture).

That adds a bit of dynamic dispatch and a runtime branch, though---but I am not sure the performance overhead will be noticeably since it seems like this is a pretty heavy operation.  If it is, I've actually been wondering for a bit why we use the `dyn VirtualMachine` here---could we push the `dyn` up a layer or two and make `HyperlightVm<T: VirtualMachine>`? I imagine that that would open up a bunch of opportunities for inlining in general.

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -791,6 +858,40 @@ impl VirtualMachine for WhpVm {
...
+
+            // Flush NPT TLB (AMD SVM workaround, see npt_flush module).
+            self.npt_flush
+                .flush(self.partition, &self.surrogate_process)?;
+
+            // WHvResetPartition resets LAPIC to power-on defaults.
+            // Re-initialize it when LAPIC emulation is active.
+            //
+            // set_sregs filters out APIC_BASE (the snapshot sregs
+            // default it to 0, which would disable the LAPIC). This
```
syntactically, line 1041: Would just making sure that the snapshot captures `APIC_BASE` (and that Snapshot::from_file sets it to the architectural default value when the hw-interrupts feature is on) work?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -287,6 +287,12 @@ impl MultiUseSandbox {
             return Err(SnapshotSandboxMismatch);
         }
 
+        // Reset VM/vCPU state before memory restore.
+        self.vm.reset_vm_state().map_err(|e| {
```
syntactically, line 523: I assume that the removal of the sregs reset from `reset_vcpu` when it was renamed is related to moving this in front of the memory restore? Is there a particular reason that it should happen before?

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
ludfjig: I'd like that, but we cannot currently do this since we only know which one of kmv/mshv is available at runtime. Maybe having the mshv3/kvm cargo features be mutually exclusive could help with this (although mutually exclusive featuers are discouraged). Agree with the first part though

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: For some reason it only works when done before which is why I did it this way, I'm not sure why but I will find out before merging this

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
syntactically: What I meant for the latter was that we could add a `HyperlightVm` trait and have the sandboxes hold a `dyn HyperlightVm` which at runtime is e.g. `HyperlightVm<KvmVm>`, on the assumption (to be checked) that having the annoying dynamic dispatch be at that higher level would be beneficial.

jsturtevant:
closing since we are tracking the update for this in https://github.com/hyperlight-dev/hyperlight/issues/1682 when MSHV gets the hv_scrub feature and the reset completed in a different way here https://github.com/hyperlight-dev/hyperlight/issues/791

PR #1366, opened 2026-04-09
mshv: use VP register page for RIP/RAX writes in run_vcpu

ludfjig:
Replace set_reg() hypercalls with direct VP register page writes for the RIP increment on IO exits and the RAX write on IO-in (hw-interrupts). The VP register page is a shared memory page between userspace and the hypervisor that is picked up automatically on the next DISPATCH_VP, eliminating one hypercall per IO exit.

## Benchmark Results (MSHV)

| Benchmark | Before | After | Change |
|---|--:|--:|---|
| guest_calls/call/default | 51.4 µs | 46.0 µs | **-10.3%** |
| guest_calls/call/small | 54.2 µs | 38.6 µs | **-28.5%** |
| guest_calls/call/medium | 55.6 µs | 39.6 µs | **-28.7%** |
| guest_calls/call/large | 58.8 µs | 40.7 µs | **-30.7%** |
| guest_calls/call_with_restore/default | 183.1 µs | 157.3 µs | **-14.0%** |
| guest_calls/call_with_restore/small | 193.6 µs | 170.6 µs | **-14.1%** |
| guest_calls/call_with_host_function/default | 108.3 µs | 67.5 µs | **-37.6%** |
| guest_calls/call_with_host_function/small | 95.8 µs | 70.4 µs | **-26.3%** |
| guest_calls/call_with_host_function/medium | 94.9 µs | 71.5 µs | **-24.4%** |
| guest_calls/call_with_host_function/large | 99.8 µs | 74.3 µs | **-25.5%** |

### DIFF
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
@@ -30,12 +30,14 @@ use mshv_bindings::{
     hv_message_type, hv_message_type_HVMSG_GPA_INTERCEPT, hv_message_type_HVMSG_UNMAPPED_GPA,
     hv_message_type_HVMSG_X64_HALT, hv_message_type_HVMSG_X64_IO_PORT_INTERCEPT,
     hv_partition_property_code_HV_PARTITION_PROPERTY_SYNTHETIC_PROC_FEATURES,
-    hv_partition_synthetic_processor_features, mshv_create_partition_v2, mshv_user_mem_region,
+    hv_partition_synthetic_processor_features, hv_register_assoc,
+    hv_register_name_HV_X64_REGISTER_RIP, hv_register_value, mshv_create_partition_v2,
+    mshv_user_mem_region,
 };
 #[cfg(feature = "hw-interrupts")]
 use mshv_bindings::{
     HV_X64_REGISTER_CLASS_GENERAL, hv_interrupt_type_HV_X64_INTERRUPT_TYPE_FIXED,
-    hv_register_assoc, hv_register_value,
+    hv_register_name_HV_X64_REGISTER_RAX, set_gp_regs_field_ptr,
 };
 #[cfg(feature = "hw-interrupts")]
 use mshv_ioctls::InterruptRequest;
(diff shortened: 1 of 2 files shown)


src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
```diff
@@ -219,16 +218,21 @@ impl VirtualMachine for MshvVm {
...
+                            // mshv, unlike kvm, does not automatically increment RIP.
+                            {
+                                let page = self
+                                    .vcpu_fd
+                                    .get_vp_reg_page()
+                                    .ok_or(RunVcpuError::NoVpRegisterPage)?;
+                                // SAFETY: The register page is a valid mmap'd page
+                                // from create_vcpu and is always populated after a
+                                // vcpu run returns an intercept message.
+                                unsafe {
```
simongdavies, line 231: Its a pity `RegisterPage` doesn't have safe accessors for all the data, might be worth an upstream PR?

simongdavies (dismissed):
This is a great update Thanks!!

src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
syntactically: It looks like upstream code uses `mshv_bindings::set_gp_regs_field_ptr`, e.g. [here](https://docs.rs/mshv-ioctls/latest/src/mshv_ioctls/ioctls/vcpu.rs.html#380-381)---so perhaps that at least is worth using

src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
```diff
@@ -219,16 +218,21 @@ impl VirtualMachine for MshvVm {
...
-                                    },
-                                    ..Default::default()
-                                }])
-                                .map_err(|e| RunVcpuError::IncrementRip(e.into()))?;
+                            // mshv, unlike kvm, does not automatically increment RIP.
+                            {
+                                let page = self
+                                    .vcpu_fd
+                                    .get_vp_reg_page()
+                                    .ok_or(RunVcpuError::NoVpRegisterPage)?;
```
syntactically: It looks like a lot of the code in mshv-ioctls uses `VcpuFd::is_valid_vp_reg_page` to check that the hypervisor has set some validity state saying that the vp reg page access works---do we need to do that as well?

syntactically (commented):
Mostly LGTM, but some minor comments.

src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
ludfjig: Switched to using that where possible (not for RIP unfortunately)

src/hyperlight_host/src/hypervisor/virtual_machine/mshv/x86_64.rs
ludfjig: Added that for safety, falls back to old behavior now if not valid instead of erroring out

simongdavies (approved):

danbugs (approved):
LGTM. Is there no WHP equivalent?

PR #1368, opened 2026-04-10

andreiltd:
https://github.com/andreiltd/hyperlight/blob/tandr/rng-rfc/proposals/0001-rng-buf/README.md


ludfjig:
@andreiltd are you ok with closing this one since it probably won't get merged?

PR #1376, opened 2026-04-14
feat(clippy): add disallowed macros for asserts in clippy.toml 

simongdavies:
This pull request introduces stricter linting for assertion macros in release builds.

**Linting and Build Configuration Changes:**

* Added a `clippy.toml` configuration disallowing the use of `assert!`, `assert_eq!`, and `assert_ne!` macros in release builds, with explanations for each.
* Updated crate-level attributes in `lib.rs` to globally suppress the `clippy::disallowed_macros` lint, then selectively re-enable it for release host builds (with `std` feature), while allowing asserts for guest targets and in test/debug builds. This is thoroughly documented in comments.

### DIFF
--- a/src/hyperlight_common/clippy.toml
+++ b/src/hyperlight_common/clippy.toml
@@ -1,0 +1,5 @@
+disallowed-macros = [
+    { path = "std::assert", reason = "no asserts in release builds" },
+    { path = "std::assert_eq", reason = "no asserts in release builds" },
+    { path = "std::assert_ne", reason = "no asserts in release builds" },
+]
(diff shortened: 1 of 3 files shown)


dblnz (approved):
I see the changes only affect `hyperlight-common`, shouldn't we do the same thing for the other crates used on the host?

simongdavies:
> I see the changes only affect `hyperlight-common`, shouldn't we do the same thing for the other crates used on the host?

The hyperlight-host crate should already have this, are there others we need?

PR #1379, opened 2026-04-15
fix: zero-fill NOBITS sections in ELF loader

ludfjig:
Some linkers emit PT_LOAD segments where filesz == memsz but contain .bss sections whose VMA range overlaps with file bytes from unrelated sections. The loader copies the full segment verbatim, leaving .bss with stale data instead of zeros.

Collect NOBITS section ranges (excluding .tbss) during ELF parsing and zero-fill them after loading PT_LOAD segments.

This is needed for unikraft

### DIFF
--- a/src/hyperlight_host/src/mem/elf.rs
+++ b/src/hyperlight_host/src/mem/elf.rs
@@ -21,10 +21,7 @@ use std::sync::Arc;
 use goblin::elf::reloc::{R_AARCH64_NONE, R_AARCH64_RELATIVE};
 #[cfg(target_arch = "x86_64")]
 use goblin::elf::reloc::{R_X86_64_NONE, R_X86_64_RELATIVE};
-use goblin::elf::{Elf, ProgramHeaders, Reloc};
-#[cfg(feature = "nanvix-unstable")]
-use goblin::elf32::program_header::PT_LOAD;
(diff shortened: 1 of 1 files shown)


src/hyperlight_host/src/mem/elf.rs
```diff
@@ -206,6 +220,21 @@ impl ElfInfo {
                 .copy_from_slice(&self.payload[payload_offset..payload_offset + payload_len]);
             target[start_va + payload_len..start_va + phdr.p_memsz as usize].fill(0);
         }
+        // Zero-fill NOBITS sections (e.g. .bss) that were not already
+        // covered by the filesz < memsz zeroing above.
+        for &(addr, size) in &self.nobits_ranges {
+            let sh_start = (addr - base_va) as usize;
```
dblnz, line 226: Nit: I am not sure if it can happen, but is this guaranteed to not fail? Maybe a checked sub is better here

danbugs:
I think this might still be worth to do from a correctiveness standpoint. But it seems that, at some point, this stopped being a requirement for Unikraft–see [hyperlight-unikraft](https://github.com/danbugs/hyperlight-unikraft) running off unmodified latest HL.

danbugs:
Won't merge rn. Focusing on https://github.com/hyperlight-dev/hyperlight/pull/1385

PR #1380, opened 2026-04-15
Remove HyperlightPEB file mappings

ludfjig:
Removes file_mappings which instead can be replaced by init_blob.

depends on #1379 first, will mark ready after


danbugs:
LGTM. dev Nanvix today does use `file_mappings`, but, in the current on-going refactor work, the RAMFS location (i.e., what it uses file mappings for) is computed as `shared_mem_end` from `init_data.ptr` + `init_data.size` + `page-table area`.

danbugs:
Won't merge rn. Focusing on https://github.com/hyperlight-dev/hyperlight/pull/1385.

PR #1382, opened 2026-04-16
feat(virtq): add packed virtio ring primitives

andreiltd:
Add low-level packed virtqueue ring implementation in `hyperlight_common::virtq`, based on the virtio packed ring format.

This is split from: https://github.com/hyperlight-dev/hyperlight/pull/1368 and does not include any actual plumbing for guest/host communication.

Useful materials:
- HIP: https://github.com/andreiltd/hyperlight/blob/tandr/rng-rfc/proposals/0001-rng-buf/README.md
- spec: https://docs.oasis-open.org/virtio/virtio/v1.3/csd01/virtio-v1.3-csd01.html#x1-720008
- redhat docs: https://www.redhat.com/en/blog/packed-virtqueue-how-reduce-overhead-virtio

### DIFF
--- a/src/hyperlight_common/src/virtq/access.rs
+++ b/src/hyperlight_common/src/virtq/access.rs
@@ -1,0 +1,167 @@
+/*
+Copyright 2026  The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Memory Access Traits for Virtqueue Operations
+//!
+//! This module defines the [`MemOps`] trait that abstracts memory access patterns
+//! required by the virtqueue implementation. This allows the virtqueue code to
+//! work with different memory backends e.g. Host vs Guest.
+
+use alloc::sync::Arc;
+
+use bytemuck::Pod;
+
+/// Backend-provided memory access for virtqueue.
+///
+/// # Safety
+///
+/// Implementations must ensure that:
+/// - Pointers passed to methods are valid for the duration of the call
+/// - Memory ordering guarantees are upheld as documented
+/// - Reads and writes don't cause undefined behavior (alignment, validity)
+///
+/// [`RingProducer`]: super::RingProducer
+/// [`RingConsumer`]: super::RingConsumer
+pub trait MemOps {
+    type Error;
+
+    /// Read bytes from physical memory.
+    ///
+    /// Used for reading buffer contents pointed to by descriptors.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - Guest physical address to read from
+    /// * `dst` - Destination buffer to fill
+    ///
+    /// # Returns
+    ///
+    /// Number of bytes actually read (should equal `dst.len()` on success).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `paddr` is valid and points to at least `dst.len()` bytes.
+    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<usize, Self::Error>;
+
+    /// Write bytes to physical memory.
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - address to write to
+    /// * `src` - Source data to write
+    ///
+    /// # Returns
+    ///
+    /// Number of bytes actually written (should equal `src.len()` on success).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `paddr` is valid and points to at least `src.len()` bytes.
+    fn write(&self, addr: u64, src: &[u8]) -> Result<usize, Self::Error>;
+
+    /// Load a u16 with acquire semantics.
+    ///
+    /// # Safety
+    ///
+    /// `addr` must translate to a valid, aligned `AtomicU16` in shared memory.
+    fn load_acquire(&self, addr: u64) -> Result<u16, Self::Error>;
+
+    /// Store a u16 with release semantics.
+    ///
+    /// # Safety
+    ///
+    /// `addr` must translate to a valid `AtomicU16` in shared memory.
+    fn store_release(&self, addr: u64, val: u16) -> Result<(), Self::Error>;
+
+    /// Get a direct read-only slice into shared memory.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure:
+    /// - `addr` is valid and points to at least `len` bytes.
+    /// - The memory region is not concurrently modified for the lifetime of
+    ///   the returned slice. Caller must uphold this via protocol-level
+    ///   synchronisation, e.g. descriptor ownership transfer.
+    ///
+    /// See also [`BufferOwner`]: super::BufferOwner
+    unsafe fn as_slice(&self, addr: u64, len: usize) -> Result<&[u8], Self::Error>;
+
+    /// Get a direct mutable slice into shared memory.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure:
+    /// - `addr` is valid and points to at least `len` bytes.
+    /// - No other references (shared or mutable) to this memory region exist
+    ///   for the lifetime of the returned slice.
+    /// - Protocol-level synchronisation (e.g. descriptor ownership) guarantees
+    ///   exclusive access.
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn as_mut_slice(&self, addr: u64, len: usize) -> Result<&mut [u8], Self::Error>;
+
+    /// Read a Pod type at the given pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `addr` is valid, aligned, and translates to initialized memory.
+    fn read_val<T: Pod>(&self, addr: u64) -> Result<T, Self::Error> {
+        let mut val = T::zeroed();
+        let bytes = bytemuck::bytes_of_mut(&mut val);
+
+        self.read(addr, bytes)?;
+        Ok(val)
+    }
+
+    /// Write a Pod type at the given pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller ensures that `ptr` is valid.
+    fn write_val<T: Pod>(&self, addr: u64, val: T) -> Result<(), Self::Error> {
+        let bytes = bytemuck::bytes_of(&val);
+        self.write(addr, bytes)?;
+        Ok(())
+    }
+}
+
+impl<T: MemOps> MemOps for Arc<T> {
+    type Error = T::Error;
+
+    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<usize, Self::Error> {
+        (**self).read(addr, dst)
+    }
+
+    fn write(&self, addr: u64, src: &[u8]) -> Result<usize, Self::Error> {
+        (**self).write(addr, src)
+    }
+
+    fn load_acquire(&self, addr: u64) -> Result<u16, Self::Error> {
+        (**self).load_acquire(addr)
+    }
+
+    fn store_release(&self, addr: u64, val: u16) -> Result<(), Self::Error> {
+        (**self).store_release(addr, val)
+    }
+
+    unsafe fn as_slice(&self, addr: u64, len: usize) -> Result<&[u8], Self::Error> {
+        unsafe { (**self).as_slice(addr, len) }
+    }
+
+    #[allow(clippy::mut_from_ref)]
+    unsafe fn as_mut_slice(&self, addr: u64, len: usize) -> Result<&mut [u8], Self::Error> {
+        unsafe { (**self).as_mut_slice(addr, len) }
+    }
+}
--- a/src/hyperlight_common/src/virtq/access.rs
+++ b/src/hyperlight_common/src/virtq/access.rs
@@ -53,7 +53,7 @@ pub trait MemOps {
     ///
     /// # Safety
     ///
-    /// The caller must ensure `paddr` is valid and points to at least `dst.len()` bytes.
+    /// The caller must ensure `addr` is valid and points to at least `dst.len()` bytes.
     fn read(&self, addr: u64, dst: &mut [u8]) -> Result<usize, Self::Error>;
 
     /// Write bytes to physical memory.
@@ -69,7 +69,7 @@ pub trait MemOps {
     ///
     /// # Safety
     ///
-    /// The caller must ensure `paddr` is valid and points to at least `src.len()` bytes.
+    /// The caller must ensure `addr` is valid and points to at least `src.len()` bytes.
     fn write(&self, addr: u64, src: &[u8]) -> Result<usize, Self::Error>;
 
     /// Load a u16 with acquire semantics.
--- a/src/hyperlight_common/src/virtq/access.rs
+++ b/src/hyperlight_common/src/virtq/access.rs
@@ -29,13 +29,17 @@ use bytemuck::Pod;
 /// # Safety
 ///
 /// Implementations must ensure that:
-/// - Pointers passed to methods are valid for the duration of the call
-/// - Memory ordering guarantees are upheld as documented
-/// - Reads and writes don't cause undefined behavior (alignment, validity)
+/// - Addresses accepted by these methods are translated according to the
+///   backend's memory model.
+/// - Invalid or inaccessible addresses are reported with `Self::Error` rather
+///   than causing undefined behavior.
+/// - Memory ordering guarantees are upheld as documented.
+/// - Typed reads/writes and atomic operations honor alignment and initialized
+///   memory requirements for the translated addresses.
 ///
 /// [`RingProducer`]: super::RingProducer
 /// [`RingConsumer`]: super::RingConsumer
-pub trait MemOps {
+pub unsafe trait MemOps {
     type Error;
 
     /// Read bytes from physical memory.
@@ -47,9 +51,8 @@ pub trait MemOps {
     /// * `addr` - Guest physical address to read from
     /// * `dst` - Destination buffer to fill
     ///
-    /// # Safety
-    ///
-    /// The caller must ensure `addr` is valid and points to at least `dst.len()` bytes.
+    /// Implementations must return an error if `addr` cannot be read for
+    /// at least `dst.len()` bytes.
     fn read(&self, addr: u64, dst: &mut [u8]) -> Result<(), Self::Error>;
 
     /// Write bytes to physical memory.
@@ -59,23 +62,20 @@ pub trait MemOps {
     /// * `addr` - address to write to
     /// * `src` - Source data to write
     ///
-    /// # Safety
-    ///
-    /// The caller must ensure `addr` is valid and points to at least `src.len()` bytes.
+    /// Implementations must return an error if `addr` cannot be written for
+    /// at least `src.len()` bytes.
     fn write(&self, addr: u64, src: &[u8]) -> Result<(), Self::Error>;
 
     /// Load a u16 with acquire semantics.
     ///
-    /// # Safety
-    ///
-    /// `addr` must translate to a valid, aligned `AtomicU16` in shared memory.
+    /// Implementations must return an error if `addr` does not translate to a
+    /// valid, aligned `AtomicU16` in shared memory.
     fn load_acquire(&self, addr: u64) -> Result<u16, Self::Error>;
 
     /// Store a u16 with release semantics.
     ///
-    /// # Safety
-    ///
-    /// `addr` must translate to a valid `AtomicU16` in shared memory.
+    /// Implementations must return an error if `addr` does not translate to a
+    /// valid, aligned `AtomicU16` in shared memory.
     fn store_release(&self, addr: u64, val: u16) -> Result<(), Self::Error>;
 
     /// Get a direct read-only slice into shared memory.
@@ -106,9 +106,8 @@ pub trait MemOps {
 
     /// Read a Pod type at the given pointer.
     ///
-    /// # Safety
-    ///
-    /// The caller must ensure `addr` is valid, aligned, and translates to initialized memory.
+    /// Implementations must return an error if `addr` is not valid, aligned,
+    /// and initialized for `T`.
     fn read_val<T: Pod>(&self, addr: u64) -> Result<T, Self::Error> {
         let mut val = T::zeroed();
         let bytes = bytemuck::bytes_of_mut(&mut val);
@@ -119,9 +118,8 @@ pub trait MemOps {
 
     /// Write a Pod type at the given pointer.
     ///
-    /// # Safety
-    ///
-    /// The caller ensures that `ptr` is valid.
+    /// Implementations must return an error if `addr` is not valid and aligned
+    /// for `T`.
     fn write_val<T: Pod>(&self, addr: u64, val: T) -> Result<(), Self::Error> {
         let bytes = bytemuck::bytes_of(&val);
         self.write(addr, bytes)?;
@@ -129,7 +127,9 @@ pub trait MemOps {
     }
 }
 
-impl<T: MemOps> MemOps for Arc<T> {
+// SAFETY: Arc delegates all memory operations to the wrapped backend, preserving
+// that backend's MemOps contract.
+unsafe impl<T: MemOps> MemOps for Arc<T> {
     type Error = T::Error;
 
     fn read(&self, addr: u64, dst: &mut [u8]) -> Result<(), Self::Error> {
--- a/src/hyperlight_common/src/virtq/access.rs
+++ b/src/hyperlight_common/src/virtq/access.rs
@@ -47,14 +47,10 @@ pub trait MemOps {
     /// * `addr` - Guest physical address to read from
     /// * `dst` - Destination buffer to fill
     ///
-    /// # Returns
-    ///
-    /// Number of bytes actually read (should equal `dst.len()` on success).
-    ///
     /// # Safety
     ///
     /// The caller must ensure `addr` is valid and points to at least `dst.len()` bytes.
-    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<usize, Self::Error>;
+    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<(), Self::Error>;
 
     /// Write bytes to physical memory.
     ///
@@ -63,14 +59,10 @@ pub trait MemOps {
     /// * `addr` - address to write to
     /// * `src` - Source data to write
     ///
-    /// # Returns
-    ///
-    /// Number of bytes actually written (should equal `src.len()` on success).
-    ///
     /// # Safety
     ///
     /// The caller must ensure `addr` is valid and points to at least `src.len()` bytes.
-    fn write(&self, addr: u64, src: &[u8]) -> Result<usize, Self::Error>;
+    fn write(&self, addr: u64, src: &[u8]) -> Result<(), Self::Error>;
 
     /// Load a u16 with acquire semantics.
     ///
@@ -140,11 +132,11 @@ pub trait MemOps {
 impl<T: MemOps> MemOps for Arc<T> {
     type Error = T::Error;
 
-    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<usize, Self::Error> {
+    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<(), Self::Error> {
         (**self).read(addr, dst)
     }
 
-    fn write(&self, addr: u64, src: &[u8]) -> Result<usize, Self::Error> {
+    fn write(&self, addr: u64, src: &[u8]) -> Result<(), Self::Error> {
         (**self).write(addr, src)
     }
 
--- a/src/hyperlight_common/src/virtq/desc.rs
+++ b/src/hyperlight_common/src/virtq/desc.rs
@@ -27,6 +27,9 @@ use super::MemOps;
 
 bitflags! {
     /// Descriptor flags as defined by VIRTIO specification.
+    ///
+    /// Note: The implementation never follows the indirect-table interpretation,
+    /// so INDIRECT bit is effectively ignored.
     #[repr(transparent)]
     #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
     pub struct DescFlags: u16 {
@@ -90,7 +93,7 @@ impl Descriptor {
         DescFlags::from_bits_truncate(self.flags)
     }
 
-    /// Did the guest mark this descriptor in the current guest round?
+    /// Did the driver make this descriptor available in the current driver round?
     #[inline]
     pub fn is_avail(&self, wrap: bool) -> bool {
         let f = self.flags();
@@ -99,7 +102,7 @@ impl Descriptor {
         avail == wrap && used != wrap
     }
 
-    /// Did the host mark this descriptor used in the current host round?
+    /// Did the device mark this descriptor used in the current device round?
     #[inline]
     pub fn is_used(&self, wrap: bool) -> bool {
         let f = self.flags();
--- a/src/hyperlight_common/src/virtq/desc.rs
+++ b/src/hyperlight_common/src/virtq/desc.rs
@@ -1,0 +1,331 @@
+/*
+Copyright 2026  The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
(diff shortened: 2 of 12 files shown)


src/hyperlight_common/src/virtq/access.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// * `addr` - Guest physical address to read from
+    /// * `dst` - Destination buffer to fill
+    ///
+    /// # Returns
+    ///
+    /// Number of bytes actually read (should equal `dst.len()` on success).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `paddr` is valid and points to at least `dst.len()` bytes.
```
dblnz: Nit:
```suggestion
    /// The caller must ensure `addr` is valid and points to at least `dst.len()` bytes.
```

src/hyperlight_common/src/virtq/access.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// * `addr` - address to write to
+    /// * `src` - Source data to write
+    ///
+    /// # Returns
+    ///
+    /// Number of bytes actually written (should equal `src.len()` on success).
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `paddr` is valid and points to at least `src.len()` bytes.
```
dblnz: Nit:
```suggestion
    /// The caller must ensure `addr` is valid and points to at least `src.len()` bytes.
```

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+        })
+    }
+
+    /// Write a descriptor to memory with release semantics for flags at the given base pointer
+    ///
+    /// This is the primary synchronization point for publishing descriptors.
+    ///
+    /// # Invariant
+    ///
+    /// The caller must ensure that `base` is valid for writes of Descriptor
```
dblnz: I think `base` has been renamed

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+}
+
+impl DescTable {
+    pub const DEFAULT_LEN: usize = 256;
+
+    /// Create a descriptor table from shared memory.
+    ///
+    /// # Safety
+    ///
+    /// - `base` must be valid for reads and writes of `size` descriptors
```
dblnz:
```suggestion
    /// - `base_addr` must be valid for reads and writes of `size` descriptors
```

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// - `base` must be properly aligned for `Descriptor`
```
dblnz:
```suggestion
    /// - `base_addr` must be properly aligned for `Descriptor`
```

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// - `size` must not exceed `u16::MAX`
+    /// - memory must remain valid for the lifetime of this table
```
dblnz, line 208: Is this "ensured" by the way it's going to be used (by design)?

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub unsafe fn from_raw_parts(base_addr: u64, size: usize) -> Self {
+        assert!(base_addr.is_multiple_of(Descriptor::ALIGN as u64));
+        assert!(size <= u16::MAX as usize);
```
dblnz: Why not make `size` a `u16` in `DescTable`? I see there are some casts to `u16`. Maybe there would be more casts to `usize` if this is changed?

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+    pub fn desc_addr(&self, idx: u16) -> Option<u64> {
+        if idx >= self.size as u16 {
+            return None;
+        }
+
+        Some(self.base_addr + (idx as u64 * Descriptor::SIZE as u64))
+    }
+
+    /// Get number of descriptors in table
+    pub fn len(&self) -> usize {
```
dblnz, line 226: This is a bit confusing.
I see the `size` field is set in `from_raw_parts` which seems to suggest it is a size in bytes, not a number of descriptors.

src/hyperlight_common/src/virtq/event.rs
dblnz, line 1: This comment is for the other files also: maybe doing some checked adds to ensure correct result for the read_acquire/write_release APIs

src/hyperlight_common/src/virtq/event.rs
```diff
@@ -0,0 +1,117 @@
...
+        /// Notify only at specific descriptor (EVENT_IDX mode).
+        const DESC = 0x2;
+    }
+}
+
+/// Event suppression structure for controlling notifications.
+#[repr(C)]
+#[derive(Clone, Copy, Debug, Pod, Zeroable, PartialEq, Eq, Hash)]
+pub struct EventSuppression {
+    // bits 0-14: offset, bit 15: wrap
```
dblnz: Clippy might complain about lack of `///` comment per field, I think

src/hyperlight_common/src/virtq/event.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub off_wrap: u16,
```
dblnz: Are the fields necessary to be `pub`? Is the logic using it outside of `hyperlight_common`?

src/hyperlight_common/src/virtq/ring.rs
dblnz, line 1: Do all these defined types need to be `pub`?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3459 @@
...
+/// Type-state: Can add writable buffers (no more readables allowed)
+pub struct Writable;
+
+/// A builder for buffer chains using type-state to enforce readable/writable order.
+///
+/// Upholds invariants: at least one buffer must be present in the chain,
+/// and readable buffers must be added before writable buffers.
+#[derive(Debug, Default)]
+pub struct BufferChainBuilder<T> {
+    elems: SmallVec<[BufferElement; 16]>,
```
dblnz, line 208: Maybe documenting this 16 inline capacity as an optimization for the user to be aware

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3459 @@
...
+        Self {
+            head: 0,
+            size: size as u16,
+            wrap: true,
+        }
+    }
+
+    /// Advance to next position, wrapping around and toggling wrap counter if needed
+    #[inline]
+    pub(crate) fn advance(&mut self) {
```
dblnz: Could this be called concurrently on the host?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3459 @@
...
+        self.head += 1;
+        if self.head >= self.size {
+            self.head = 0;
+            self.wrap = !self.wrap;
+        }
+    }
+
+    /// Advance by n positions using modular arithmetic.
+    #[inline]
+    pub(crate) fn advance_by(&mut self, n: u16) {
```
dblnz: Nit: Maybe have `advance` be `advance_by(1)`

dblnz (commented):
Great stuff, @andreiltd !
I left a few comments, most are small remarks, others are things I might have missed.
This is part 1 of my review, I still have some things left to look at that I plan on doing tomorrow.

src/hyperlight_common/src/virtq/desc.rs
andreiltd: It's a zero sum game, the way it is now requires slightly less casts.

src/hyperlight_common/src/virtq/desc.rs
andreiltd: Changed internal field from `size` to `len`.

src/hyperlight_common/src/virtq/desc.rs
andreiltd: The `virtq` lives in shared memory region controlled by host - it is static for the guest and special region for the host.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: I'm not sure if I follow but normal rust rules would apply to call this function concurrently. But this just modify in memory state of cursor, so there is never need to call that function outside `ring.rs`. The reason it's not private is that restoring state from snapshot is messing with cursors.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: The reason both exists is that they deploy different optimizations so merging implementations will defeat the purpose for which they were split in the first place.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: I did audit and it looks like yes, if we want to keep Ring public (which I believe is useful) the types are leaked by ring API, so they need to be public as well.

src/hyperlight_common/src/virtq/ring.rs
dblnz: Got it, thanks

src/hyperlight_common/src/virtq/ring.rs
dblnz: got it, thanks!

src/hyperlight_common/src/virtq/event.rs
andreiltd: That's a good point, my worry is that these are the hottest paths in the virtq as every submit and poll goes through them. The const offsets are stable, so I would rather add checks to table/layout constructors that the base address + total size does not overflow. wdyt?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3459 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The descriptor ID assigned to this buffer, for matching with completions.
+    ///
+    /// # Errors
+    ///
+    /// - [`RingError::WouldBlock`] - No free descriptor slots
+    /// - [`RingError::OutOfMemory`] - No free descriptor IDs (internal error)
+    /// - [`RingError::InvalidState`] - ID tracking corrupted (internal error)
```
dblnz, line 525: This API can also return `RingError::MemError`

src/hyperlight_common/src/virtq/access.rs
```diff
@@ -0,0 +1,167 @@
...
+    /// Read a Pod type at the given pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `addr` is valid, aligned, and translates to initialized memory.
+    fn read_val<T: Pod>(&self, addr: u64) -> Result<T, Self::Error> {
+        let mut val = T::zeroed();
+        let bytes = bytemuck::bytes_of_mut(&mut val);
+
+        self.read(addr, bytes)?;
```
jprendes, line 115: The `read` method returns `Result<usize, Error>` but this discards the read count. If `read` returns less than the full length, this could silently produce corrupted data.

src/hyperlight_common/src/virtq/access.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    /// Write a Pod type at the given pointer.
+    ///
+    /// # Safety
+    ///
+    /// The caller ensures that `ptr` is valid.
+    fn write_val<T: Pod>(&self, addr: u64, val: T) -> Result<(), Self::Error> {
+        let bytes = bytemuck::bytes_of(&val);
+        self.write(addr, bytes)?;
```
jprendes, line 125: The `write` method returns `Result<usize, Error>` but this discards the write count. If `write` returns less than the full length, this could silently produce corrupted data.

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3459 @@
...
+        &self.elems[..self.split]
+    }
+
+    /// Get writable buffers in chain
+    pub fn writables(&self) -> &[BufferElement] {
+        &self.elems[self.split..]
+    }
+
+    /// Get total number of buffers in chain
+    // Note: buffer chain cannot be empty by construction
```
jprendes, line 369: Is this true?
What if I do `BufferChain::default()`? Maybe we don't need to derive `Default`?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3459 @@
...
+        mem: TestMem,
+        layout: Layout,
+    }
+
+    fn align_up(val: usize, align: usize) -> usize {
+        (val + align - 1) & !(align - 1)
+    }
+
+    impl OwnedRing {
+        pub fn new(size: usize) -> Self {
```
jprendes, line 1485: There's a chain that least to `size` must be a power of 2, can we add that to the doc comments? Do we need a runtime check for that?

src/hyperlight_common/src/virtq/access.rs
andreiltd: That's a good catch. Thinking about this more, the truncated reads and writes should be hard errors on virtq level so I lean towards changing it to return `()` instead of size.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Removed default.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: I've added checks and comments to `Layout` constructor. The `OwnedRing` is only for tests.

src/hyperlight_common/src/virtq/event.rs
andreiltd: I've added checks in `Layout` constructor.

src/hyperlight_common/src/virtq/access.rs
andreiltd: Fixed.

src/hyperlight_common/src/virtq/access.rs
andreiltd: Fixed the same way as `read`.

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        mut self,
+        elements: impl IntoIterator<Item = impl Into<BufferElement>>,
+    ) -> Self {
+        for elem in elements {
+            self.elems.push(elem.into());
+            self.split += 1;
+        }
+
+        self
+    }
```
ludfjig, line 247: should we set `elem.writeable = false` here in case a caller passes writeable? Or error out?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+
+    /// Add multiple readable buffers from an iterator.
+    ///
+    /// This transitions to Writable state so no more readable buffers can be added.
+    pub fn writables(
+        mut self,
+        elements: impl IntoIterator<Item = impl Into<BufferElement>>,
+    ) -> BufferChainBuilder<Writable> {
+        for elem in elements {
+            self.elems.push(elem.into());
```
ludfjig: same here but vice versa, shouldwe set `writable = true`?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        self.id_free.extend(0..size as u16);
+        self.id_num.iter_mut().for_each(|n| *n = 0);
+        self.event_flags_shadow = EventFlags::ENABLE;
+    }
+
+    /// Reset the ring to the "N slots submitted, none completed" state.
+    ///
+    /// `ids` contains the descriptor IDs that are in-flight.
+    /// Sets cursors, counters, and `id_num` accordingly. The chain lengths are all set to 1.
+    pub fn reset_prefilled(&mut self, ids: &[u16]) {
```
ludfjig, line 918: if a ring has size 8 and we call `reset_prefilled(&[4,5,6,7])`, and call  `producer.submit_one(addr, len, false)`, it will fail because `self.id_free` is empty, I'm not sure this is correct?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        let readables = elements.len() - writables;
+
+        // Since driver wrote the same id everywhere, head_desc.id is valid.
+        let id = head_desc.id;
+        if (id as usize) >= self.id_num.len() {
+            return Err(RingError::InvalidState);
+        }
+
+        // Record chain length for later used submission
+        self.id_num[id as usize] = chain_len;
```
ludfjig: this is overwritten unconditionally, should we check that this id is not 0 like the producer already does:

```rust
// We should never reuse an ID that is still outstanding
if self.id_num[id as usize] != 0 {
    return Err(RingError::InvalidState);
}
```

src/hyperlight_common/src/virtq/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+    /// Driver-written event suppression area in shared memory.
+    pub drv_evt_addr: u64,
+    /// Device-written event suppression area in shared memory.
+    pub dev_evt_addr: u64,
+}
+
+#[inline]
+const fn align_up(val: usize, align: usize) -> usize {
+    (val + align - 1) & !(align - 1)
+}
```
ludfjig, line 89: nit: maybe `next_multiple_of()` can be used instead?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+    /// Create a new producer from a memory layout and accessor.
+    pub fn new(layout: Layout, mem: M) -> Self {
+        let size = layout.desc_table_len as usize;
+        let raw = layout.desc_table_addr;
+
+        // SAFETY: layout is valid
+        let table = unsafe { DescTable::from_raw_parts(raw, size) };
+        let cursor = RingCursor::new(size);
+
+        const DEFAULT_LEN: usize = DescTable::default_len();
```
ludfjig, line 485: Layout has all-public fields but only one constructor so this safety comment doens't hold since you can make a super unsafe Layout with safe code. We probably want to make Layout's fields private so the only construction path goes through the existing unsafe `from_base`.

src/hyperlight_common/src/virtq/access.rs
```diff
@@ -0,0 +1,159 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `addr` - Guest physical address to read from
+    /// * `dst` - Destination buffer to fill
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure `addr` is valid and points to at least `dst.len()` bytes.
+    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<(), Self::Error>;
```
ludfjig: Should all methods of MemOps be unsafe given that they have safety comments?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        desc.mark_avail(head_wrap);
+
+        let addr = self
+            .desc_table
+            .desc_addr(head_idx)
+            .ok_or(RingError::InvalidState)?;
+
+        // Release publish
+        desc.write_release(&self.mem, addr)
+            .map_err(|_| RingError::MemError)?;
```
ludfjig: nit: lot's of map_err lose their context when going to `RingError::MemError`, maybe can parametrize `RingError<E>` on generic `E` which is some M::Error? Not sure, feel free to ignore

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+bitflags! {
+    /// Descriptor flags as defined by VIRTIO specification.
+    #[repr(transparent)]
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+    pub struct DescFlags: u16 {
+        /// This marks a buffer as continuing via the next field.
+        const NEXT     = 1 << 0;
+        /// This marks a buffer as device write-only (otherwise device read-only).
+        const WRITE    = 1 << 1;
+        /// This means the buffer contains a list of buffer descriptors (unsupported here).
```
ludfjig, line 40: nit: should we add runtimes check for this unsupported?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+
+        if total_descs == 1 {
+            let elem = chain.elems()[0];
+            return self.submit_one(elem.addr, elem.len, elem.writable);
+        }
+
+        let head_idx = self.avail_cursor.head();
+        let head_wrap = self.avail_cursor.wrap();
+
+        let id = self.id_free.pop().ok_or(RingError::InvalidState)?;
```
ludfjig: nit: this uses `RingError::InvalidState` but submit_one uses another error for the same error condition

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+    split: usize,
+}
+
+impl BufferChain {
+    /// Get all buffer elements in the chain.
+    pub fn elems(&self) -> &[BufferElement] {
+        self.elems.as_slice()
+    }
+
+    /// Get writable buffers in chain
```
ludfjig: typo writable -> readable

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        if chain_len == 0 {
+            return Err(RingError::InvalidState);
+        }
+
+        let idx = self.used_cursor.head();
+        let wrap = self.used_cursor.wrap();
+
+        // addr is unused for used descriptor according to packed-virtqueue spec
+        let mut used_desc = Descriptor::new(0, 0, id, DescFlags::empty());
+        used_desc.len = written_len;
```
ludfjig: why not pass `written_len` to `new(..)` directly?

ludfjig (dismissed):
This PR is very clean, big fan!

I know we don't really have policy on asserts vs errors etc in hyperlight (I personally like asserts...), but ring.rs has 3 debug_assert! that could be useful in release mode as well, and they're already in methods that return Results. Do you think it makes sense to convert them to errros instead?

src/hyperlight_common/src/virtq/access.rs
```diff
@@ -0,0 +1,159 @@
...
+/// # Safety
+///
+/// Implementations must ensure that:
+/// - Pointers passed to methods are valid for the duration of the call
+/// - Memory ordering guarantees are upheld as documented
+/// - Reads and writes don't cause undefined behavior (alignment, validity)
+///
+/// [`RingProducer`]: super::RingProducer
+/// [`RingConsumer`]: super::RingConsumer
+pub trait MemOps {
```
simongdavies: Should the trait be unsafe?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        });
+
+        BufferChainBuilder {
+            elems: self.elems,
+            split: self.split,
+            marker: PhantomData,
+        }
+    }
+
+    /// Add multiple readable buffers from an iterator.
```
simongdavies: Should this be writeable?

```suggestion
    /// Add multiple writeable buffers from an iterator.
```

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+
+            ring_need_event(off, new.head(), old.head())
+        }
+        // treat as disabled if invalid
+        _ => false,
+    }
+}
+
+#[inline(always)]
+pub fn ring_need_event(event_idx: u16, new: u16, old: u16) -> bool {
```
simongdavies: does this need to be pub?

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -0,0 +1,331 @@
...
+    /// Did the host mark this descriptor used in the current host round?
+    #[inline]
+    pub fn is_used(&self, wrap: bool) -> bool {
+        let f = self.flags();
+        let avail = f.contains(DescFlags::AVAIL);
+        let used = f.contains(DescFlags::USED);
+        avail == wrap && used == wrap
+    }
+
+    /// Is this descriptor writable by the device?
```
simongdavies, line 114: The comments here use the terms "guest" "host" and "device" and "driver" I read the HIP again at https://github.com/andreiltd/hyperlight/blob/tandr/rng-rfc/proposals/0001-rng-buf/README.md#design-details

but still wasn't sure how to interpret these terms here? The HIP suggests that the driver is always the guest and the device is always the host but then goes on to say that for bi-directional communication these roles are reversed , I think it might help us in the future to define explicitly what we are referring to in the code comments.

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3447 @@
...
+        }
+    }
+
+    /// Owns the descriptor table and event suppression structures
+    pub struct OwnedRing {
+        mem: TestMem,
+        layout: Layout,
+    }
+
+    fn align_up(val: usize, align: usize) -> usize {
```
simongdavies: This is a duplicate of the function in `mod.rs`

src/hyperlight_common/src/virtq/event.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Create a new event suppression with the given offset/wrap and flags.
+    pub fn new(off_wrap: u16, flags: EventFlags) -> Self {
+        Self {
+            off_wrap,
+            flags: flags.bits(),
+        }
+    }
+
+    /// Get the event flags.
+    pub fn flags(&self) -> EventFlags {
```
simongdavies, line 74: Can we make the magic numbers for Flags mask, off, wrap consts please?

simongdavies (commented):
Great stuff a few minor comments but looks good to me

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Done!

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Updated!

src/hyperlight_common/src/virtq/ring.rs
andreiltd: I've added normalization, I would rather leave the builder methods infallible.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Good catch! Fixed and added regression test.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Good catch! Fixed and added regression test.

src/hyperlight_common/src/virtq/mod.rs
andreiltd: Agree and fixed.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Fixed!

src/hyperlight_common/src/virtq/access.rs
andreiltd: Yeah that's a valid point. I did make `MemOps` an unsafe trait and kept only `as_slice`/`as_mut_slice` as unsafe methods because they return references with aliasing. I also updated docs on safety contract.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: I tried this initially yes, and make it generic creates much more API churn especially for the layers on top of virtq. But let me think a bit more how we could preserve context without making ring error generic.

src/hyperlight_common/src/virtq/desc.rs
andreiltd: We could, but I think it's perfectly safe if we just ignore this bit (treat as reserved) and document that current implementation does not follow direct tables. The reason I'm skeptical to add runtime checks is that buffer chains are processed in the loop so we will need the check inside the loop for no real benefits.

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Fixed!

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Obviously, yes 🙂

src/hyperlight_common/src/virtq/access.rs
andreiltd: Agree! I've changed that and added some safety contract docs.

src/hyperlight_common/src/virtq/desc.rs
andreiltd: Yes, HIP needs to be updated. Initial version of HIP assumed the roles are swapped depending on direction -- but that idea has been later discarded following allocation discussion in the heap. The most recent agreement is: driver - always a guest, device - always host. I also unified naming here so that it refers to virtio  semantics so driver/device.

src/hyperlight_common/src/virtq/event.rs
andreiltd: Fixed!

src/hyperlight_common/src/virtq/ring.rs
andreiltd: I've added  more context to the `MemError` variant:

```rust
    #[error("Backend memory error while {op} at address 0x{addr:x}, len {len}")]
    MemError {
        /// Memory operation that failed.
        op: MemOp,
        /// Address passed to the memory backend.
        addr: u64,
        /// Number of bytes requested for the operation.
        len: usize,
    },
```

I guess the alternative would be to use type erasure like `Box<dyn Error>` but that will require extra bounds on `M::Error`: `Error + Send + Sync + 'static` and will allocate on error path?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3580 @@
...
+            Self {
+                inner: Arc::new(TestMemInner {
+                    storage: UnsafeCell::new(storage),
+                    base_addr,
+                }),
+            }
+        }
+
+        /// Get a pointer with proper provenance for the given address
+        fn ptr_for_addr(&self, addr: u64) -> *mut u8 {
```
danbugs, line 1411: Should we have bounds checks here?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3580 @@
...
+        if self.head >= self.size {
+            self.head = 0;
+            self.wrap = !self.wrap;
+        }
+    }
+
+    /// Advance by n positions using modular arithmetic.
+    #[inline]
+    fn advance_by(&mut self, n: u16) {
+        let new = self.head + n;
```
danbugs, line 411: Might be worth to add a debug_assert here? `debug_assert!(self.head.checked_add(n).is_some())`?

ludfjig (dismissed):
LGTM, not sure if you saw this comment but not blocking

>I know we don't really have policy on asserts vs errors etc in hyperlight (I personally like asserts...), but ring.rs has 3 debug_assert! that could be useful in release mode as well, and they're already in methods that return Results. Do you think it makes sense to convert them to errros instead?

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -0,0 +1,3580 @@
...
+        &self.elems[..self.split]
+    }
+
+    /// Get writable buffers in chain
+    pub fn writables(&self) -> &[BufferElement] {
+        &self.elems[self.split..]
+    }
+
+    /// Get total number of buffers in chain
+    // Note: buffer chain cannot be empty by construction
```
danbugs, line 369: But BufferChain is `pub`. Couldn't we just make it empty?

danbugs (dismissed):
Only minor NITs. Really great work, @andreiltd !

src/hyperlight_common/src/virtq/ring.rs
andreiltd: It can only be constructed via `ChainBuilder` that will reject an empty chain. That is consistent with spec:

> A buffer consists of zero or more device-readable physically-contiguous elements followed by zero or more physically-contiguous device-writable elements (each buffer has at least one element).

src/hyperlight_common/src/virtq/ring.rs
andreiltd: This is just a test utility and all the tests are run through `miri` as well so any bounds violation should be catched.

andreiltd:
> LGTM, not sure if you saw this comment but not blocking
> 
> > I know we don't really have policy on asserts vs errors etc in hyperlight (I personally like asserts...), but ring.rs has 3 debug_assert! that could be useful in release mode as well, and they're already in methods that return Results. Do you think it makes sense to convert them to errros instead?

Hey @ludfjig, sorry - there were few back and forth with usage of asserts so I thought I already addressed that - we had CI checks that banned usage of assert macro at some point but that has been now relaxed. I converted the asserts in question into invalid state errors - I think it makes sense for dealing with malicious driver. Thanks for pointing that out!

https://github.com/hyperlight-dev/hyperlight/pull/1382/commits/032ac0b3a2816a8fe2d483289cad724748a52473

dblnz (approved):
Great work! I have nothing else to add.

jprendes (approved):

ludfjig (approved):

PR #1383, opened 2026-04-16
Re-export guest types and error utilities from `hyperlight_guest_bin`

jprendes:
## Summary

This PR re-exports common types and error utilities from `hyperlight_guest_bin`, allowing users to build guest binaries with a single dependency on `hyperlight_guest_bin` instead of requiring an explicit dependency on both `hyperlight_guest` and `hyperlight_guest_bin`.

## Changes

- Re-export `ErrorCode` from `hyperlight_guest::error`
- Add new `types` module to `hyperlight_guest` that re-exports `FunctionCall`, `ParameterType`, `ParameterValue`, `ReturnType`, and `ReturnValue`
- Add new `error` module to `hyperlight_guest_bin` that re-exports error types and the `bail`/`ensure` macros

## Motivation

Simplifies the dependency graph for guest binary authors by allowing them to depend only on `hyperlight_guest_bin`.

### DIFF
--- a/src/hyperlight_guest/src/error.rs
+++ b/src/hyperlight_guest/src/error.rs
@@ -17,7 +17,7 @@ limitations under the License.
 use alloc::format;
 use alloc::string::{String, ToString as _};
 
-use hyperlight_common::flatbuffer_wrappers::guest_error::ErrorCode;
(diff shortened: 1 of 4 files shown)


dblnz (approved):
LGTM!

PR #1386, opened 2026-04-17
fix(whp): use NULL DACL for map_file_cow file mapping sections

danbugs:
## Summary

- File-backed sections created with the default DACL via `CreateFileMappingW` fail with `ERROR_ACCESS_DENIED` when mapped into a surrogate process via `MapViewOfFileNuma2` on modern Windows
- Fix: create the section with a NULL DACL security descriptor granting unrestricted access
- The mapping remains file-backed and zero-copy (`FILE_MAP_READ`, no data copied)
- Adds a `map-file-cow-test` example that exercises the full lifecycle (create → map_file_cow → evolve → call) — fails before this change on Windows, succeeds after

### DIFF
--- a/src/hyperlight_host/src/sandbox/file_mapping.rs
+++ b/src/hyperlight_host/src/sandbox/file_mapping.rs
@@ -295,6 +295,7 @@ pub(crate) fn prepare_file_cow(
         use std::os::windows::io::AsRawHandle;
 
         use windows::Win32::Foundation::HANDLE;
+        use windows::Win32::Security::{PSECURITY_DESCRIPTOR, SECURITY_ATTRIBUTES};
         use windows::Win32::System::Memory::{
             CreateFileMappingW, FILE_MAP_READ, MapViewOfFile, PAGE_READONLY,
         };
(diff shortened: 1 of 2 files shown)


src/hyperlight_host/src/sandbox/file_mapping.rs
```diff
@@ -313,12 +314,36 @@ pub(crate) fn prepare_file_cow(
 
         let file_handle = HANDLE(file.as_raw_handle());
 
+        // Build a security descriptor with a NULL DACL (unrestricted
+        // access) so the surrogate process can map the section via
+        // MapViewOfFileNuma2. File-backed sections created with the
+        // default DACL fail with ERROR_ACCESS_DENIED when mapped
+        // cross-process on modern Windows.
+        let mut sd_bytes = [0u8; 40]; // SECURITY_DESCRIPTOR_MIN_LENGTH
```
andreiltd: This seems fragile, should we use https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Security/struct.SECURITY_DESCRIPTOR.html with default value instead?

src/hyperlight_host/src/sandbox/file_mapping.rs
```diff
@@ -313,12 +314,36 @@ pub(crate) fn prepare_file_cow(
...
+        unsafe {
+            let psd = PSECURITY_DESCRIPTOR(sd_bytes.as_mut_ptr() as *mut _);
+            windows::Win32::Security::InitializeSecurityDescriptor(
+                psd, 1, // SECURITY_DESCRIPTOR_REVISION
```
andreiltd: Use https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/SystemServices/constant.SECURITY_DESCRIPTOR_REVISION1.html

src/hyperlight_host/src/sandbox/file_mapping.rs
dblnz: If you modify this, please add the link to the docs for reference

src/hyperlight_host/src/sandbox/file_mapping.rs
dblnz: Same here

dblnz (commented):
I don't have anything else to add. Once the other comments are fixed, I'll approve

src/hyperlight_host/src/sandbox/file_mapping.rs
```diff
@@ -313,12 +317,39 @@ pub(crate) fn prepare_file_cow(
...
 
+        // Build a security descriptor with a NULL DACL (unrestricted
+        // access) so the surrogate process can map the section via
+        // MapViewOfFileNuma2. File-backed sections created with the
+        // default DACL fail with ERROR_ACCESS_DENIED when mapped
+        // cross-process on modern Windows.
+        // https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Security/struct.SECURITY_DESCRIPTOR.html
+        // https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/SystemServices/constant.SECURITY_DESCRIPTOR_REVISION1.html
+        let mut sd = SECURITY_DESCRIPTOR::default();
+        let psd = PSECURITY_DESCRIPTOR(std::ptr::addr_of_mut!(sd).cast());
```
jprendes, line 328: **nit:** you can use `&raw mut` instead of `addr_of_mut!` macro
```suggestion
        let psd = PSECURITY_DESCRIPTOR(&raw mut sd as _);
```

jprendes (approved):
LGTM

dblnz (approved):

PR #1387, opened 2026-04-17
fix(whp): respect IO port access size in IoPortAccess exit handler

danbugs:
## Summary

- The WHP `IoPortAccess` exit handler was returning the full 8 bytes of RAX regardless of the guest's operand width
- For `outb` (1-byte writes like serial console output), this produced 8 bytes per character, resulting in garbled multi-byte sequences in the host output
- Fix: extract `AccessSize` from `IoPortAccess.AccessInfo` (bits 1..3) and truncate the returned data vec to the actual operand width (1 for outb, 2 for outw, 4 for outl)


jprendes (dismissed):
LGTM

jprendes (approved):

PR #1391, opened 2026-04-20
feat(whp): copy-on-write file mapping for snapshot load

danbugs:
## Summary

Stacked follow-up to #1373. The Windows path in `ReadonlySharedMemory::from_file_windows` is currently created with `PAGE_READONLY` + `FILE_MAP_READ`. That matches the type name (`ReadonlySharedMemory`) but not the semantics the `from_snapshot` caller needs: the host-side view has to be **writable** so WHP (and MSHV) can service the copy-on-write faults the guest takes on first write. A read-only mapping triggers an access violation on the host thread the moment the guest touches any page.

Switch to `PAGE_WRITECOPY` + `FILE_MAP_COPY` — the Windows equivalent of Linux's `mmap(MAP_PRIVATE)` that the Linux `from_file` path already uses. Reads still come from the backing file; writes transparently allocate private copy-on-write pages.

## Relationship to #1373

- **Depends on #1373 landing first** — this branch targets `disk_snapshot`'s tip, so the diff here will reduce to just the CoW change once #1373 is merged.
- Until then, GitHub will show both #1373's commits and this one. Please review only the `shared_mem.rs` delta.

## Tested

On a Windows box with WHP enabled: `Snapshot::to_file` then `MultiUseSandbox::from_snapshot` → `restore` → `call` completes cleanly. Pre-fix, the same flow took an access violation on the first page fault the guest triggered after restore.

## Known follow-ups (not in this PR)

None for this change specifically. Separate PR will add `Registerable for MultiUseSandbox` so host functions can be wired on sandboxes acquired via `from_snapshot`.

### DIFF
--- a/docs/snapshot-file-implementation-plan.md
+++ b/docs/snapshot-file-implementation-plan.md
@@ -1,0 +1,322 @@
+# Snapshot File Format
+
+## Overview
+
+Save a `Snapshot` to disk and load it back via zero-copy file
+mapping, so that a `MultiUseSandbox` can be created directly from a
+file without re-parsing the guest ELF or re-running guest init code.
+
+- **Linux**: `mmap(MAP_PRIVATE)` at page-aligned offset - zero copy,
+  demand-paged by the kernel.
+- **Windows**: `CreateFileMappingA(PAGE_READONLY)` +
+  `MapViewOfFile(FILE_MAP_READ)` - zero copy, demand-paged by the OS.
+
+Cross-platform (Linux + Windows). Default feature flags only
+(`nanvix-unstable`, `crashdump`, `gdb` not handled).
+
+---
+
+## File Format
+
+The file uses a versioned header with two independent version checks:
+
+- **Format version** (`FormatVersion` enum): controls the byte layout
+  of the header itself. A format version mismatch may be convertible
+  by re-serializing the header.
+- **ABI version** (`SNAPSHOT_ABI_VERSION` constant): covers the
+  contents and interpretation of the memory blob. An ABI mismatch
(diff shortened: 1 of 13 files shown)


syntactically:
> The Windows path in ReadonlySharedMemory::from_file_windows is currently created with PAGE_READONLY + FILE_MAP_READ. That matches the type name (ReadonlySharedMemory) but not the semantics the from_snapshot caller needs: the host-side view has to be writable so WHP (and MSHV) can service the copy-on-write faults the guest takes on first write. A read-only mapping triggers an access violation on the host thread the moment the guest touches any page.

I didn't actually review any of the code in this PR yet, but I want to ask for some clarification from this line in the description. With FILE_MAP_COPY, the CoW will happen for the whole process when the host/guest writes into a page in the region, but that would mean the modification would be visible across all the sandboxes using the same `ReadonlySharedMemory` as their snapshot. So, it is important that this never happen. 

The reason for the `PROT_WRITE` on Linux is a basically a bug in the MSHV kernel driver, where, at the time of mapping the memory into the guest, even if the mapping is marked readonly and can never actually generate a write, the driver acts as though it was writing to every page in the mapping (without actually changing the contents of any page). Is this PR saying that an analogous bug exists in the WHP/windows kernel mshv root partition driver, and so there are some "fake" write faults that we need to make sure do not trap even though no writes ever actually happen, or is it saying that there are actual writes? The language "so WHP can service the copy-on-write faults the guest takes on first write" imply the latter, which I think would be a problem.

ludfjig:
I don't think this change is correct, I don't think we want cow in the host

PR #1397, opened 2026-04-21
Remove disallowed-macros lint

simongdavies:
Using the `disallowed-macros` lint to detect the use of `assert!` flags up the use of `debug-assert!` , use of assert in const expressions etc. This leads to more and more annotations in code to allow the valid use of asserts, rather than continually expand this this PR removes the lint in favour of detecting use of assert in release code via PR reviews


ludfjig (approved):

syntactically:
I think I agree that the lint was probably causing more trouble than it is worth, although it is kind of sad to lose a static analysis for this. It's a shame clippy doesn't have a more specific lint that would be easier to ensure was not false-positive on obvious cases like const code.

andreiltd (approved):

Issue #1406, opened 2026-04-23
Failure creating C artifacts for release


jprendes:
Creating the artifacts for C workflows fails in CI.
In particular, creating the tar with the libc headers fails since we switched to picolibc.
https://github.com/hyperlight-dev/hyperlight/actions/runs/24772551006/job/72486667291

ludfjig:
This is fixed

Issue #1399, opened 2026-04-21
Consider adding features to enable selected picolibc options


jprendes:
﻿The non-defaults serve the hyperlight guest model: single-threaded, no TLS, no semihosting, no libc malloc since we provide rust allocator, verbose asserts for better debuggability, and some I/O performance tweaks discussed above like io buffering.

There are some knobs like wchar support that were needed to allow building baremetal libc++ against picolibc.

We can expose build settings but I would rather do that trough cargo features not env variables. That being said I'm not sure if we should proactively expose all the build tweaks or wait and see if any of this turns out to be problematic?

_Originally posted by @andreiltd in https://github.com/hyperlight-dev/hyperlight/pull/831#discussion_r2980243499_

PR #1407, opened 2026-04-23
Use static picolibc.h file instead of generating it at buildtime

### LINKED ISSUE #1401: Avoid dirtying the source directory during libc build
I'm not sure how we fix this. Polluting source tree with generated files is obviously bad but we need a stable location for c guests that have no access to hashed rust target dir. 

One way of improving this would be to use the existing `HYPERLIGHT_GUEST_TOOLCHAIN_ROOT` mechanism for c guests. We could copy generated headers to ﻿`HYPERLIGHT_GUEST_TOOLCHAIN_ROOT/include` and set the env when building c guests. Then we can prepend the headers from that dir to clang calls.

I believe a proper fix would be to replace just script and cargo with polyglot build system.

_Originally posted by @andreiltd in https://github.com/hyperlight-dev/hyperlight/pull/831#discussion_r2980321976_

jprendes:
Fixes #1401 
It will also help fixing #1406


jprendes:
@copilot apply changes based on the comments in [this thread](https://github.com/hyperlight-dev/hyperlight/pull/1407#pullrequestreview-4164061030)

src/hyperlight_guest_bin/include/picolibc.h
```diff
@@ -0,0 +1,144 @@
...
+#undef __OBSOLETE_MATH_FLOAT
+#undef __PREFER_SIZE_OVER_SPEED // release build
+#undef __ATOMIC_UNGETC // single-thread
+#undef __IEEEFP_FUNCS
+#undef __INIT_FINI_FUNCS // using INIT_FINI_ARRAY instead
+#undef __HAVE_BITFIELDS_IN_PACKED_STRUCTS
+
+// compiler feature detection
+#ifndef __has_builtin
+#define picolibc_has_builtin(x) 0
```
andreiltd, line 75: Could this simply be `__has_builtin` ? And then you don't need else?

jprendes: I wanted to be able to `#undef` at the end, in case someone downstream does `#ifdef __has_builtin` and expects a fully functional macro.

andreiltd:
One thing that we were discussing was to expose `picolibc` configuration knobs via cargo features but IIUC that is no longer possible if we switch to entirely static header. Could we leave generated header but only improve builtin detection so it's done in the header rather than in the build script?

jprendes:
If we have configuration knobs and generated header, we need to figure out how that will work with published C artifacts.
A static file would let us fix the artifact publishing.

I am open to ideas :-)

jprendes:
Something I've been thinking is having a sibling project to cargo-hyperlight that provides clang-hyperlight (or hyperlight-clang), that prepares the sysroot just like cargo-hyperlight, but provides a clang wrapper instead of a cargo wrapper. But I there are some raough edges, like... which guest version do we use? with cargo-hyperlight we get that from the dependencies list, but for clang... how do we do that?

andreiltd:
> If we have configuration knobs and generated header, we need to figure out how that will work with published C artifacts. A static file would let us fix the artifact publishing.

I think the most straight forward approach to this is: offer libc build features to Rust users, C artifacts are build from default features -- we package `picolibc.h` generated for default features?

jprendes:
There are some questions regarding the generated header and race conditions (e.g., building a -sys crate that needs to build C code, and racing the build between the -sys dependency and building the hl-guest-bin dependency that generates the header).
There are also questions regarding how it would interact with cargo-hyperlight, as currently cargo-hyperlight avoid the race by copying the sysroot before invoking the actual cargo build, and also questions regarding how we can detect the features we need to enable/disable on the crate from cargo-hyperlight (AFAICT, cargo metadata doesn't tell you which features are in use, but I'm sure there's another way)

Because of this, we've decided that as a short term solution we would use a static file, and we will revisit generating the header when address #1399 .

andreiltd (approved):

PR #1422, opened 2026-04-28
Enable guest time

simongdavies:
This pull request introduces support for a paravirtualized guest clock in Hyperlight, allowing guest VMs to efficiently access wall-clock and monotonic time without incurring a VM exit. The implementation includes new documentation, shared host-guest structures, and updates to scratch memory layout and allocation to reserve and describe the clock page and associated metadata. This feature is off by default and is only enabled when the host is built with the `enable_guest_clock` Cargo feature.

**Paravirtualized guest clock feature:**

* Added a comprehensive documentation file (`docs/guest-time.md`) describing the design, usage, API, and memory layout of the paravirtualized guest clock, including how guests can read time and how the feature interacts with snapshots and restores.
* Updated the main documentation (`docs/README.md`) to link to the new guest clock documentation.

**Shared memory layout and structures:**

* Extended `src/hyperlight_common/src/layout.rs` with new constants and functions for the location and size of the clock page and associated metadata fields (`clock_type`, `boot_time_ns`) in the scratch region; added helpers for computing guest physical/virtual addresses of the clock page. 
* Added a new module `src/hyperlight_common/src/time.rs` containing host-guest shared structures for KVM pvclock and Hyper-V reference TSC, as well as a `ClockType` enum and related conversion logic.
* Registered the new `time` module in `src/hyperlight_common/src/lib.rs` for use by both host and guest.

**Guest-side integration:**

* Updated the guest physical page allocator (`src/hyperlight_guest/src/arch/amd64/prim_alloc.rs`) to always reserve three pages at the top of the scratch region (bookkeeping, shared-state, and clock page), ensuring a stable memory layout regardless of feature flags.
* Added a `time` module to the guest crate (`src/hyperlight_guest/src/lib.rs`) to support guest-side time reading.

### DIFF
--- a/docs/guest-time.md
+++ b/docs/guest-time.md
@@ -1,0 +1,135 @@
+# Paravirtualized Guest Clock
+
+Hyperlight's `enable_guest_clock` Cargo feature gives guests a cheap way to ask
+"what time is it?" without taking a VM exit. When the host is built with the
+feature, every sandbox exposes a paravirtualized clock that the guest can read
+using ordinary memory loads.
+
+## What the guest gets
+
+When the feature is enabled the host populates a single 4 KiB "clock page"
+inside the sandbox's scratch region. The page carries two pieces of
+information:
+
+- **A hypervisor-specific calibration block at offset `0x00`.** Written by
+  KVM (`kvm_clock`) or Hyper-V / MSHV (Reference TSC). Contains the TSC
+  frequency, scaling constants, and a sequence lock the guest uses to read it
+  atomically. The entire clock page is hypervisor-owned; Hyperlight does not
+  write to it.
+- **Hyperlight metadata in the scratch bookkeeping page** (separate from the
+  clock page): a `u64` [`ClockType`](../src/hyperlight_common/src/time.rs) tag
+  and `boot_time_ns`, the Unix-epoch origin of the monotonic clock computed
+  by the host as `wall_now - monotonic_now` (see below). These live at fixed
+  offsets from the top of scratch (`-0x28` and `-0x30`), NOT in the clock
+  page, so a future TLFS extension cannot clobber them.
+
+With those two pieces the guest can compute:
+
+- **Monotonic nanoseconds since boot** — read the TSC, apply the scaling
+  factors from the calibration block, giving you a `CLOCK_MONOTONIC`
+  equivalent.
+- **Wall-clock nanoseconds since the Unix epoch** — add `boot_time_ns` to the
+  monotonic value above, giving you a `CLOCK_REALTIME` / `gettimeofday`. `boot_time_ns` is computed by the host as
+  `SystemTime::now() - KVM_GET_CLOCK` (on KVM) or
+  `SystemTime::now() - TIME_REF_COUNT` (on Hyper-V) after sandbox
+  initialisation. Hyper-V has no equivalent to KVM's
+  `MSR_KVM_WALL_CLOCK_NEW`, so we use this uniform host-computed approach
+  on all backends.
+
+> **Note (KVM only):** Wall-clock time returns `None` during
+> `hyperlight_main` (guest init). On KVM, `KVM_GET_CLOCK` is unreliable
+> until the "master clock" is established at first vCPU entry, so
+> `boot_time_ns` is stamped after init completes. Monotonic time works
+> fine during init. Wall-clock time becomes available on the first
+> dispatch call.
+
+Both reads are lock-free (well, seqlock-protected for the calibration block)
+and never leave the guest.
+
+## Using it in a Rust guest
+
+The guest-side API lives in `hyperlight_guest::time` for the low-level
+readers and `hyperlight_guest_bin::time` for a `std::time`-flavoured
+wrapper:
+
+```rust
+// Low-level, no_std readers.
+use hyperlight_guest::time;
+
+if time::is_available() {
+    let mono_ns: u64 = time::monotonic_time_ns().unwrap();
+    let wall_ns: u64 = time::wall_clock_time_ns().unwrap();
+}
+
+// std::time-flavoured wrapper (hyperlight_guest_bin only).
+use hyperlight_guest_bin::time::{Instant, SystemTime, UNIX_EPOCH};
+
+let t0 = Instant::now()?;
+// ... do work ...
+let elapsed = t0.elapsed()?;
+
+let now = SystemTime::now()?;
+let unix_ns = now.duration_since(UNIX_EPOCH)?.as_nanos();
+```
+
+C guests that use picolibc get paravirt time for free: `hyperlight_guest_bin`
+wires `clock_gettime(CLOCK_MONOTONIC|CLOCK_REALTIME)` and `gettimeofday` into
+the same reader, so existing C code continues to work unchanged.
+
+## Snapshot / restore semantics
+
+Both `boot_time_ns` and the hypervisor calibration block live inside scratch
+memory, which is not included in snapshots. On every
+`MultiUseSandbox::restore`, the host re-arms the clock page: it re-installs
+the pvclock MSR / Hyper-V register against the fresh vCPU state and stamps a
+new `boot_time_ns` captured at the moment of restore. As a result a restored
+guest observes wall-clock time reflecting the restore moment, not the
+original boot — which is what wall clocks are supposed to do.
+
+## Enabling the feature
+
+Turn it on in the host's `Cargo.toml`:
+
+```toml
+[dependencies]
+hyperlight-host = { version = "...", features = ["enable_guest_clock"] }
+```
+
+The feature is x86_64 only; on aarch64 it has no effect. It is off by default
+so existing sandboxes don't pay for a facility they don't use. When off, the
+clock page is still reserved in the layout (so memory maps are stable) but
+left un-mapped against any hypervisor clock source; `hyperlight_guest::time`
+readers then report "unavailable" and fall back to whatever the guest wants
+to do about it (the picolibc wiring returns a synthetic 1-second-per-call
+counter, which is enough to stop `strftime` crashing and not much else).
+
+## Layout details
+
+The clock page sits 3 pages below the very top of the scratch region:
+
+| Offset from top | Size  | Contents                                       |
+|-----------------|-------|------------------------------------------------|
+| `-0x1000`       | 4 KiB | Bookkeeping (size, allocator counter, ...)     |
+| `-0x2000`       | 4 KiB | Reserved for shared-state counter              |
+| `-0x3000`       | 4 KiB | Paravirtualized clock page                     |
+
+Because the clock page is at the top of scratch, both the guest's main stack
+and its IST1 (exception) stack are configured to start one page below the
+clock page (at `MAX_GVA + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET`) so stack writes
+— including page-fault handlers running on IST1 — cannot clobber the trailer.
+The allocator reserves the top three pages unconditionally so the memory map
+stays identical whether or not the feature is enabled.
+
+## Non-goals
+
+- **Sub-microsecond accuracy.** `boot_time_ns` is computed from two
+  back-to-back host reads (`SystemTime::now()` and `KVM_GET_CLOCK` /
+  `TIME_REF_COUNT`). On KVM, residual disagreement between `KVM_GET_CLOCK`
+  and the pvclock page can add up to ~13ms of constant offset (observed on
+  WSL2; root cause uncertain). On Hyper-V the offset should be negligible.
+- **`CLOCK_PROCESS_CPUTIME_ID` and friends.** The clock page exposes only
+  monotonic and wall-clock time; per-thread / per-process CPU time is out of
+  scope.
+- **Timers or sleeps.** The guest can read the clock but has no way to ask
+  the hypervisor to wake it up later — that is still done through the
+  existing guest-function call model.
--- a/docs/guest-time.md
+++ b/docs/guest-time.md
@@ -101,7 +101,10 @@ clock page is still reserved in the layout (so memory maps are stable) but
 left un-mapped against any hypervisor clock source; `hyperlight_guest::time`
 readers then report "unavailable" and fall back to whatever the guest wants
 to do about it (the picolibc wiring returns a synthetic 1-second-per-call
-counter, which is enough to stop `strftime` crashing and not much else).
+counter).
+
+It is also a good stopgap for many other things that expect `gettimeofday` /
+`clock_gettime` to work (like StarlingMonkey and QuickJS).
 
 ## Layout details
 
--- a/docs/guest-time.md
+++ b/docs/guest-time.md
@@ -108,20 +108,21 @@ It is also a good stopgap for many other things that expect `gettimeofday` /
 
 ## Layout details
 
-The clock page sits 3 pages below the very top of the scratch region:
+The clock page is the second page from the very top of the scratch region.
+The top of scratch holds a fixed four-page reserved region:
 
 | Offset from top | Size  | Contents                                       |
 |-----------------|-------|------------------------------------------------|
-| `-0x1000`       | 4 KiB | Bookkeeping (size, allocator counter, ...)     |
-| `-0x2000`       | 4 KiB | Reserved for shared-state counter              |
-| `-0x3000`       | 4 KiB | Paravirtualized clock page                     |
+| `-0x1000`       | 4 KiB | Metadata / bookkeeping (size, allocator, ...)  |
+| `-0x2000`       | 4 KiB | Paravirtualized clock page                     |
+| `-0x4000`       | 8 KiB | Exception (IST1) stack (2 pages)               |
 
-Because the clock page is at the top of scratch, both the guest's main stack
-and its IST1 (exception) stack are configured to start one page below the
-clock page (at `MAX_GVA + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET`) so stack writes
-— including page-fault handlers running on IST1 — cannot clobber the trailer.
-The allocator reserves the top three pages unconditionally so the memory map
-stays identical whether or not the feature is enabled.
+The guest's IST1 (exception) stack starts at the clock-page base
+(`MAX_GVA + 1 - SCRATCH_TOP_EXN_STACK_OFFSET`) and grows downward through its
+two dedicated pages, so stack writes — including page-fault handlers running
+on IST1 — cannot clobber the clock page or the metadata page above. The
+allocator reserves the whole four-page region unconditionally so the memory
+map stays identical whether or not the feature is enabled.
 
 ## Non-goals
 
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -38,7 +38,31 @@ pub const SCRATCH_TOP_SIZE_OFFSET: u64 = 0x08;
 pub const SCRATCH_TOP_ALLOCATOR_OFFSET: u64 = 0x10;
 pub const SCRATCH_TOP_SNAPSHOT_PT_GPA_BASE_OFFSET: u64 = 0x18;
 pub const SCRATCH_TOP_SNAPSHOT_GENERATION_OFFSET: u64 = 0x20;
-pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x30;
+
+/// Offset from the top of scratch for the `clock_type` field (u64).
+///
+/// Identifies which paravirtualized clock the host configured
+/// ([`crate::time::ClockType`]). Lives in the bookkeeping page at the
+/// top of scratch — NOT in the clock page itself — so the hypervisor
+/// cannot clobber it if it extends the TLFS-reserved region.
+pub const SCRATCH_TOP_CLOCK_TYPE_OFFSET: u64 = 0x28;
+
+/// Offset from the top of scratch for the `boot_time_ns` field (u64).
+///
+/// The Unix-epoch origin of the monotonic clock, computed by the host
+/// as `SystemTime::now() - current_monotonic_ns()` and written in
+/// `arm_clock`. The guest recovers wall time as
+/// `boot_time_ns + monotonic_time_ns()`.
+///
+/// Hyper-V has no equivalent to KVM's `MSR_KVM_WALL_CLOCK_NEW`, so
+/// we use this uniform host-computed approach on all backends.
+pub const SCRATCH_TOP_BOOT_TIME_NS_OFFSET: u64 = 0x30;
+
+// ---- Next free offset in the bookkeeping page: 0x38 ----
+// When adding new host→guest shared fields, use the next multiple of
+// 8 after the last offset above. All fields in this page are u64,
+// little-endian, host-written and guest-read, and are excluded from
+// snapshots because they live in scratch memory.
 
 /// Offset from the top of scratch memory for a shared host-guest u64 counter.
 ///
@@ -48,6 +72,30 @@ pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x30;
 /// maximum representable frame number.
 #[cfg(feature = "guest-counter")]
 pub const SCRATCH_TOP_GUEST_COUNTER_OFFSET: u64 = 0x1008;
+
+/// Offset from the top of scratch memory for the start of the paravirtualized
+/// clock page.
+///
+/// The clock page is a single 4 KiB page occupying the scratch offsets
+/// `[0x3000, 0x2000)` from the top — i.e. one page lower than the
+/// guest-counter page, to avoid the i686 frame-number issue that forces the
+/// counter off the very last page (see [`SCRATCH_TOP_GUEST_COUNTER_OFFSET`]).
+///
+/// The constant is the *high* (exclusive) offset; the page base is one page
+/// below, at `top - SCRATCH_TOP_CLOCK_PAGE_OFFSET` + 1 byte — in other words,
+/// subtract this value from `MAX_GPA`/`MAX_GVA` + 1 to get the page base.
+///
+/// The page is only present when the host is built with the
+/// `enable_guest_clock` feature.
+pub const SCRATCH_TOP_CLOCK_PAGE_OFFSET: u64 = 0x3000;
+
+/// Size of the paravirtualized clock page in bytes (one 4 KiB page).
+/// The entire page is owned by the hypervisor (KVM pvclock or Hyper-V
+/// Reference TSC). Hyperlight's own metadata (`clock_type`,
+/// `boot_time_ns`) lives in the bookkeeping page at offsets
+/// `SCRATCH_TOP_CLOCK_TYPE_OFFSET` / `SCRATCH_TOP_BOOT_TIME_NS_OFFSET`,
+/// NOT in the clock page, so a future TLFS extension cannot clobber it.
+pub const CLOCK_PAGE_SIZE: u64 = 0x1000;
 
 pub fn scratch_base_gpa(size: usize) -> u64 {
     (MAX_GPA - size + 1) as u64
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -104,5 +104,26 @@ pub fn scratch_base_gva(size: usize) -> u64 {
     (MAX_GVA - size + 1) as u64
 }
 
+/// Guest physical address of the base of the paravirtualized clock page.
+///
+/// The clock page sits at a fixed offset from the top of the guest physical
+/// address space, independent of `scratch_size`: it is always
+/// `MAX_GPA + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET`.
+///
+/// Only meaningful when the host is built with the `enable_guest_clock`
+/// feature; otherwise the page is not populated.
+pub const fn clock_page_gpa() -> u64 {
+    (MAX_GPA as u64) + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET
+}
+
+/// Guest virtual address of the base of the paravirtualized clock page.
+///
+/// See [`clock_page_gpa`]. Scratch is mapped identity-style from
+/// `scratch_base_gva` to `scratch_base_gpa`, so the clock page sits at the
+/// equivalent offset in the guest virtual address space.
+pub const fn clock_page_gva() -> u64 {
+    (MAX_GVA as u64) + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET
+}
+
 /// Compute the minimum scratch region size needed for a sandbox.
 pub use arch::min_scratch_size;
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -85,8 +85,11 @@ pub const SCRATCH_TOP_GUEST_COUNTER_OFFSET: u64 = 0x1008;
 /// below, at `top - SCRATCH_TOP_CLOCK_PAGE_OFFSET` + 1 byte — in other words,
 /// subtract this value from `MAX_GPA`/`MAX_GVA` + 1 to get the page base.
 ///
-/// The page is only present when the host is built with the
-/// `enable_guest_clock` feature.
+/// The page is always reserved regardless of the `enable_guest_clock`
+/// feature so that the memory layout (and therefore stack positions)
+/// is stable across feature-flag builds. The host only populates it
+/// when the feature is enabled; otherwise it stays zero-filled and
+/// the guest sees `ClockType::None`.
 pub const SCRATCH_TOP_CLOCK_PAGE_OFFSET: u64 = 0x3000;
 
 /// Size of the paravirtualized clock page in bytes (one 4 KiB page).
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -63,8 +63,11 @@ pub const SCRATCH_TOP_BOOT_TIME_NS_OFFSET: u64 = 0x30;
 /// below, at `top - SCRATCH_TOP_CLOCK_PAGE_OFFSET` + 1 byte — in other words,
 /// subtract this value from `MAX_GPA`/`MAX_GVA` + 1 to get the page base.
 ///
-/// The page is only present when the host is built with the
-/// `enable_guest_clock` feature.
+/// The page is always reserved regardless of the `enable_guest_clock`
+/// feature so that the memory layout (and therefore stack positions)
+/// is stable across feature-flag builds. The host only populates it
+/// when the feature is enabled; otherwise it stays zero-filled and
+/// the guest sees `ClockType::None`.
 pub const SCRATCH_TOP_CLOCK_PAGE_OFFSET: u64 = 0x3000;
 
 /// Size of the paravirtualized clock page in bytes (one 4 KiB page).
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -82,5 +82,26 @@ pub fn scratch_base_gva(size: usize) -> u64 {
     (MAX_GVA - size + 1) as u64
 }
 
+/// Guest physical address of the base of the paravirtualized clock page.
+///
+/// The clock page sits at a fixed offset from the top of the guest physical
+/// address space, independent of `scratch_size`: it is always
+/// `MAX_GPA + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET`.
+///
+/// Only meaningful when the host is built with the `enable_guest_clock`
+/// feature; otherwise the page is not populated.
+pub const fn clock_page_gpa() -> u64 {
+    (MAX_GPA as u64) + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET
+}
+
+/// Guest virtual address of the base of the paravirtualized clock page.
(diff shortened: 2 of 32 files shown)


src/hyperlight_common/src/layout.rs
```diff
@@ -38,7 +38,31 @@ pub const SCRATCH_TOP_SIZE_OFFSET: u64 = 0x08;
 pub const SCRATCH_TOP_ALLOCATOR_OFFSET: u64 = 0x10;
 pub const SCRATCH_TOP_SNAPSHOT_PT_GPA_BASE_OFFSET: u64 = 0x18;
 pub const SCRATCH_TOP_SNAPSHOT_GENERATION_OFFSET: u64 = 0x20;
-pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x30;
+
```
andreiltd, line 31: What happened to `EXN_STACK_OFFSET`? I would assume it was moved to after clock pages but it looks like it's gone?

src/hyperlight_common/src/layout.rs
```diff
@@ -38,7 +38,31 @@ pub const SCRATCH_TOP_SIZE_OFFSET: u64 = 0x08;
...
+/// Offset from the top of scratch for the `clock_type` field (u64).
+///
+/// Identifies which paravirtualized clock the host configured
+/// ([`crate::time::ClockType`]). Lives in the bookkeeping page at the
+/// top of scratch — NOT in the clock page itself — so the hypervisor
+/// cannot clobber it if it extends the TLFS-reserved region.
+pub const SCRATCH_TOP_CLOCK_TYPE_OFFSET: u64 = 0x28;
```
andreiltd, line 38: I'm just curious, is top scratch now a goto mechanism to share configuration between host and guest replacing PEB?

src/hyperlight_common/src/time.rs
```diff
@@ -0,0 +1,165 @@
...
+}
+
+/// Hyper-V Reference TSC page structure (defined by Hyper-V ABI).
+///
+/// Used by both MSHV (Linux) and WHP (Windows). Time is in 100-nanosecond
+/// intervals.
+///
+/// Reference: Hyper-V TLFS §12.7.
+#[repr(C)]
+#[derive(Debug, Clone, Copy)]
```
andreiltd, line 87: Should this be `Copy`?

src/hyperlight_common/src/layout.rs
```diff
@@ -49,12 +73,60 @@ pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x30;
...
+/// The constant is the *high* (exclusive) offset; the page base is one page
+/// below, at `top - SCRATCH_TOP_CLOCK_PAGE_OFFSET` + 1 byte — in other words,
+/// subtract this value from `MAX_GPA`/`MAX_GVA` + 1 to get the page base.
+///
+/// The page is always reserved regardless of the `enable_guest_clock`
+/// feature so that the memory layout (and therefore stack positions)
+/// is stable across feature-flag builds. The host only populates it
+/// when the feature is enabled; otherwise it stays zero-filled and
+/// the guest sees `ClockType::None`.
+pub const SCRATCH_TOP_CLOCK_PAGE_OFFSET: u64 = 0x3000;
```
andreiltd: Could we add static assertions that the address is properly aligned to store `KvmPvclockVcpuTimeInfo` etc?

src/hyperlight_common/src/layout.rs
```diff
@@ -49,12 +73,60 @@ pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x30;
...
+///
+/// See [`clock_page_gpa`]. Scratch is mapped identity-style from
+/// `scratch_base_gva` to `scratch_base_gpa`, so the clock page sits at the
+/// equivalent offset in the guest virtual address space.
+pub const fn clock_page_gva() -> u64 {
+    (MAX_GVA as u64) + 1 - SCRATCH_TOP_CLOCK_PAGE_OFFSET
+}
+
 /// Compute the minimum scratch region size needed for a sandbox.
 pub use arch::min_scratch_size;
```
andreiltd, line 142: I don's see the arch specific `min_scratch_size` updates - should they account for new pages?

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -492,6 +495,124 @@ impl HyperlightVm {
...
+    ///   [`setup_clock`] before the first vCPU run), but wall-clock
+    ///   time returns `None` until this method stamps `boot_time_ns`.
+    ///
+    /// - **Snapshot restore**: called directly by the restore path.
+    ///   Re-stamps fresh `boot_time_ns` so the restored guest sees
+    ///   wall time reflecting the restore moment.
+    ///
+    /// Must be called while `scratch_memory` is `Some`.
+    #[cfg(all(feature = "enable_guest_clock", target_arch = "x86_64"))]
+    pub(crate) fn arm_clock(
```
andreiltd: I initially thought this was aarch64 only method 🙂

docs/guest-time.md
```diff
@@ -0,0 +1,135 @@
...
+hyperlight-host = { version = "...", features = ["enable_guest_clock"] }
+```
+
+The feature is x86_64 only; on aarch64 it has no effect. It is off by default
+so existing sandboxes don't pay for a facility they don't use. When off, the
+clock page is still reserved in the layout (so memory maps are stable) but
+left un-mapped against any hypervisor clock source; `hyperlight_guest::time`
+readers then report "unavailable" and fall back to whatever the guest wants
+to do about it (the picolibc wiring returns a synthetic 1-second-per-call
+counter, which is enough to stop `strftime` crashing and not much else).
```
jprendes:
```suggestion
counter).
```

It is also good stopgap for many other things that expect `gettimeofday` / `clock_gettime` to work (like StarlingMonkey and quickjs)

src/hyperlight_common/src/time.rs
jprendes: +1, specially with the padding this struct is huge.

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    None
+}
+
+/// Highest raw pvclock value ever returned. Lives in BSS so it
+/// survives snapshot/restore. Used to detect backward jumps when a
+/// snapshot is restored into a new partition whose monotonic clock
+/// starts from a lower value.
+static RAW_HIGH_WATER: AtomicU64 = AtomicU64::new(0);
```
jprendes: Can we record the high water mark on VM exit on the host side?
Then on VM-enter we can set the offset so that it continues where it left at?

src/hyperlight_guest_bin/src/time.rs
```diff
@@ -0,0 +1,222 @@
...
+///
+/// Represents wall-clock time, using the host's boot-time stamp combined
+/// with the paravirtualized monotonic clock. Snapshot-restore preserves
+/// the freshly re-stamped boot time, so `SystemTime::now()` will jump
+/// forward by real elapsed wall-clock time across a restore — exactly the
+/// behaviour a guest using `std::time::SystemTime` would expect.
+#[derive(Debug, Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SystemTime {
+    /// Nanoseconds since the Unix epoch.
+    ns: u64,
```
jprendes, line 165: This doesn't allow us to represent times before the epoch.
it might be ok, as we can't create times before the epoch anyway, right?

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -492,6 +495,124 @@ impl HyperlightVm {
...
+    /// The guest recovers wall time as
+    /// `boot_time_ns + monotonic_time_ns()`.
+    ///
+    /// # Call sites
+    ///
+    /// - **Initial sandbox creation**: called after `initialise()`
+    ///   returns (i.e. after the first vCPU run). On some backends
+    ///   (KVM) the monotonic clock source is unreliable until the
+    ///   first vCPU entry. Monotonic time is available during
+    ///   `hyperlight_main` via the pvclock page (set up by
```
jprendes, line 629: so, wall clock is not available during hyperlight_main?

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -492,6 +495,124 @@ impl HyperlightVm {
...
+
+        let gpa = clock_page_gpa();
+        let clock_type = self.vm.setup_pvclock(gpa)?;
+
+        let scratch_size = scratch.mem_size();
+
+        // Sample monotonic first, then wall clock. If preempted between
+        // the two reads, boot_time_ns shifts forward (guest wall clock
+        // runs slightly ahead of host) rather than backward — "slightly
+        // in the future" is more benign than "slightly in the past" for
```
jprendes, line 666: why is that?

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -492,6 +495,124 @@ impl HyperlightVm {
...
+        // most use cases. The gap is bounded by the 20ms test tolerance.
+        let mono_ns = self.vm.current_monotonic_ns()?;
+        let wall_ns = u64::try_from(
+            SystemTime::now()
+                .duration_since(UNIX_EPOCH)
+                .map_err(|e| crate::new_error!("system time before Unix epoch: {}", e))?
+                .as_nanos(),
+        )
+        .map_err(|_| crate::new_error!("wall_ns overflowed u64"))?;
+        let boot_time_ns = wall_ns.wrapping_sub(mono_ns);
```
jprendes, line 676: Can we store the wall?ns directly, and then before calling hyperlight_main we let the guest compute mono_ns and boot_time_ns? Then we could have wall clock on hyperlight_main

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -492,6 +495,124 @@ impl HyperlightVm {
...
+    ///   returns (i.e. after the first vCPU run). On some backends
+    ///   (KVM) the monotonic clock source is unreliable until the
+    ///   first vCPU entry. Monotonic time is available during
+    ///   `hyperlight_main` via the pvclock page (set up by
+    ///   [`setup_clock`] before the first vCPU run), but wall-clock
+    ///   time returns `None` until this method stamps `boot_time_ns`.
+    ///
+    /// - **Snapshot restore**: called directly by the restore path.
+    ///   Re-stamps fresh `boot_time_ns` so the restored guest sees
+    ///   wall time reflecting the restore moment.
```
jprendes, line 635: Is a standbox guarnteed to have ran by the time we call this function through the restore path?
If this is true, it's only right now, and would change with snapshot disk persistence, right?

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
```diff
@@ -468,6 +468,55 @@ impl VirtualMachine for KvmVm {
...
+            target: "hyperlight::pvclock",
+            clock_page_gpa,
+            "KVM pvclock armed"
+        );
+        Ok(hyperlight_common::time::ClockType::KvmPvclock)
+    }
+
+    #[cfg(feature = "enable_guest_clock")]
+    fn current_monotonic_ns(&self) -> crate::Result<u64> {
+        // KVM_GET_CLOCK returns kvmclock nanoseconds — the same time base
```
jprendes: This is the function that works only after the first run, right? might be worth documenting that here.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -183,6 +183,9 @@ pub enum InitializeError {
     SetupRegs(#[from] RegisterError),
     #[error("Guest initialised stack pointer to architecturally invalid value: {0}")]
     InvalidStackPointer(u64),
+    #[cfg(all(feature = "enable_guest_clock", target_arch = "x86_64"))]
+    #[error("Failed to arm paravirtualized guest clock: {0}")]
+    ArmClock(#[source] Box<HyperlightError>),
```
jprendes:
```suggestion
    #[error("Failed to setup paravirtualized guest clock: {0}")]
    ClockSetup(#[source] Box<HyperlightError>),
```

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -353,6 +353,32 @@ pub(crate) trait VirtualMachine: Debug + Send {
     #[cfg(not(feature = "i686-guest"))]
     fn set_xsave(&self, xsave: &[u32]) -> std::result::Result<(), RegisterError>;
 
+    /// Arm the hypervisor's paravirtualized clock for this vCPU, pointing it
```
jprendes:
```suggestion
    /// Setup the hypervisor's paravirtualized clock for this vCPU, pointing it
```

jprendes (commented):
It looks good, but I have a few questions:
* Can we make it so that the monotonic clock looks like it's ticking while the VM is running? We record the value when we enter the VM, and when we exit the VM, and compute an offset matching that. Then we can use that to measure "VM running time", regardless of restores, etc.
* I find the computation of the monotonic clock with the high watermark a bit difficult to parse (I struggle to check it's correctness)
* Instead of being a feature, cna this be a runtime option? When we contruct a sandbox, can we choose to enable or not the clock?

src/hyperlight_guest_bin/src/time.rs
simongdavies: Yes this is intentional, it follows `std:time::SystemTime` semantics and the host calcuates  `boot_time_ns` from `SystemTime::now()` so this cannot represent a time before the epoch.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
simongdavies: Wall time = `boot_time_ns + monotonic_ns`. To compute `boot_time_ns` we need `wall_now - mono_now` in the host using `KVM_GET_CLOCK` for the KVM monotonic epoch. Problem is `KVM_GET_CLOCK` is unreliable until KVM establishes its "master clock" — that only happens at first vCPU entry. So we defer stamping `boot_time_ns` until after initialise() returns (i.e. after the first vCPU run completes). On Hyper-V/MSHV this limitation doesn't exist — Reference TSC is available immediately — but we use the same deferred approach on all backends for uniformity.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
simongdavies: Correct. Monotonic time is available during hyperlight_main (the pvclock MSR is enabled before the first vCPU run), but wall-clock time returns None until after init completes. Wall time becomes available on the first dispatch call.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
simongdavies: Renamed

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
simongdavies: Yes — `stamp_pvclock_time_origin` is only reached after at least one vCPU run in all current paths:

1. **Normal evolve**: `initialise()` does the first vCPU run, then stamps the clock.
2. **`restore()`**: only callable on an already-initialised sandbox.
3. **`from_snapshot()`** (PR #1459): calls `initialise()` internally, which does the vCPU run + clock stamp.

For disk persistence (#1373/#1465) this remains safe because scratch is recreated fresh on load (not part of the persisted blob) — clock_type starts at 0 (None) and stamp_pvclock_time_origin populates it after the first vCPU run.

If scratch were ever persisted (which it might be) as a compile-time feature this would still be safe — the binary that snapshots and the binary that loads always agree on clock state. But with a runtime option, a mismatch becomes possible (snapshot-with-clock loaded into sandbox-without-clock), leaving stale clock_type != None pointing at an un-enabled clock page. That would need mitigation (zero clock region on load, or persist the flag in the snapshot header). Not a concern right now since scratch isn't persisted and clock is a feature not a runtime option.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
simongdavies: The idea is to stamp `wall_ns` (from `SystemTime::now()`) into scratch before the first vCPU run, so wall clock is available during `hyperlight_main`? 

The problem is that alone isn't enough — the guest still needs the monotonic value *at the moment wall_ns was captured* to derive a stable `boot_time_ns` offset. And the monotonic value comes from the pvclock page, which the hypervisor only updates with accurate calibration data on vCPU entry. Before that first entry, the clock page is either zeroed or contains stale data from MSR setup — the guest can't reliably read a monotonic timestamp from it yet.

So even if we stamped `wall_ns` beforehand, the guest couldn't compute a correct `boot_time_ns = wall_ns - mono_ns` during init because `mono_ns` isn't trustworthy until the vCPU is actually running. That's why we defer the whole thing to after the first vCPU run completes — at that point both the host's `KVM_GET_CLOCK` and the guest-visible pvclock page are consistent.

If wall clock during `hyperlight_main` is important, we could stamp a host-computed `wall_ns` directly and let the guest use it as-is (without deriving it from monotonic). But that gives you a point-in-time snapshot, not a live ticking wall clock — it drifts from reality the longer init takes.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
simongdavies: Fixed

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
simongdavies: Fixed

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -803,6 +803,56 @@ impl VirtualMachine for WhpVm {
...
+                WHvPartitionPropertyCodeReferenceTime,
+                &mut property as *mut WHV_PARTITION_PROPERTY as *mut c_void,
+                std::mem::size_of::<WHV_PARTITION_PROPERTY>() as u32,
+                Some(&mut written_size),
+            )
+            .map_err(|e| crate::new_error!("Failed to read WHP ReferenceTime: {}", e))?;
+        }
+        let ticks_100ns = unsafe { property.ReferenceTime };
+        Ok(ticks_100ns.wrapping_mul(100))
+    }
```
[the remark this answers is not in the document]
simongdavies: Fixed

src/hyperlight_guest_bin/src/libc_stubs.rs
```diff
@@ -112,7 +154,22 @@ extern "C" fn clock_gettime(clk_id: clockid_t, tp: *mut timespec) -> c_int {
...
+        return -1;
+    }
+
+    let (secs, nanos) = realtime();
+    unsafe {
+        (*tv).tv_sec = secs as c_long;
+        (*tv).tv_usec = (nanos / 1000) as c_long;
+    }
+    0
+}
```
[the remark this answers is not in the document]
simongdavies: Fixed

src/hyperlight_guest_bin/src/libc_stubs.rs
```diff
@@ -34,13 +35,46 @@ fn set_errno(val: u32) {
...
 
-/// Returns a synthetic monotonically-increasing time starting at Unix epoch
-/// increasing 1s each call.
-fn current_time() -> (u64, u64) {
+/// Matches picolibc `struct timespec` layout for x86_64 and aarch64.
+#[repr(C)]
+pub(crate) struct Timespec {
+    tv_sec: c_long,
+    tv_nsec: c_long,
+}
```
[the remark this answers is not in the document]
simongdavies: Fixed

src/hyperlight_common/src/layout.rs
simongdavies: I don't know, I feel like this is the right approach now for static memory allocations, do you agree?

src/hyperlight_common/src/layout.rs
simongdavies: Done

src/hyperlight_common/src/layout.rs
simongdavies: The 12 pages in `min_scratch_size` cover stacks, page tables, and IST metadata. The 3 reserved pages at the top (bookkeeping, shared-state counter, clock page) are accounted for separately — the host adds `CLOCK_PAGE_SIZE` in `SandboxMemoryLayout::new()`. The guest allocator unconditionally skips the top 3 pages so the layout is stable whether or not the clock feature is enabled. Added a doc comment to both arch implementations to clarify this split.

docs/guest-time.md
simongdavies: Fixed

src/hyperlight_common/src/layout.rs
andreiltd: Yeah, probably it's OK. If that is our approach we should probably think about some helpers to make the layout extending process less manual and safe,  for example validating if there is no overlaps etc. That is irrelevant to this PR though :)

andreiltd (dismissed):

src/hyperlight_common/src/layout.rs
simongdavies: This was a mess, the `guest-counter` feature introduced a page after the metadata page that contained a single u64, however it didn't take account of the fact that the exception stack started at 0x30 in the metadata page, I then removed the const and used the end of the clock page as the start point for the exception stack, this meant that the exception stack const was no longer required, however since the guest-feature is no longer needed I have updated things so that there is an explicit  SCRATCH_TOP_EXN_STACK_OFFSET again, the exception stack is still after the clock page but now is a calculated value ,  the metadata page now has free space for any additional data we might want to add there in the future.

src/hyperlight_guest/src/time.rs
simongdavies: I am not sure what you mean by on the host side? we would still need to record this in the snapshot to prevent the possibility of the clock going backwards, this is essentially what is happening since the `RAW_HIGH_WATER`and MONO_OFFSET are stored in the guests state (.bss) which is included in the snapshot.

I renamed the `RAW_HIGH_WATER` to `RAW_BASELINE` and added a function `monotonic_fixup` that calculates the `monotonic_time` to try and make this clearer (and testable)

simongdavies:
> It looks good, but I have a few questions:
> 
> * Can we make it so that the monotonic clock looks like it's ticking while the VM is running? We record the value when we enter the VM, and when we exit the VM, and compute an offset matching that. Then we can use that to measure "VM running time", regardless of restores, etc.

Do you mean in the guest? We could capture the `raw_monotonic_ns()?` at entry and exit and accumulate if that's what you mean? But if the thread running the vCPU gets pre-empted the clock will still keep going so it wouldn't be accurate. What do you want to measure here?


> * I find the computation of the monotonic clock with the high watermark a bit difficult to parse (I struggle to check it's correctness)

I have improved this with a function, variable renaming and more comments

> * Instead of being a feature, cna this be a runtime option? When we contruct a sandbox, can we choose to enable or not the clock?

Any particular reason why? In the previous I got feedback to make it an off by default feature?

src/hyperlight_host/src/hypervisor/virtual_machine/kvm/x86_64.rs
simongdavies: Done

andreiltd (dismissed):

jprendes:
> Any particular reason why? In the previous I got feedback to make it an off by default feature?

Features are resolved and coalesced for all uses of the library. That means if one use of hyperlight-host opts-in to the feature, then there's no way to opt-out for another use hyperlight-host in the same workspace. The per-sandbox runtime version would give you full control, and as long as it's not enabled by default (i.e., runtime behaviour desabled by default), then it's equivalent to an off by default feature (except for the slightly increased compile time).

I think the main point of the feedback (I might be putting words in someoelses mouth, sorry), is that we don't want this enabled by default, we want it to be a deliverate choice. Whether it's by feature or runtime config, it's not too relevant.

And I feel that a runtime toggle would provide a stronger guarantee in that sense.

simongdavies:
> And I feel that a runtime toggle would provide a stronger guarantee in that sense.

I think there is a difference, there may be scenarios where the developer never wants a user to be able to enable the feature, if its a config flag this is not possible, what if we add the config flag if the feature is enabled? And make it true by default in that case so that a user can turn it off dynamicallly?

jprendes:
> > And I feel that a runtime toggle would provide a stronger guarantee in that sense.
> 
> I think there is a difference, there may be scenarios where the developer never wants a user to be able to enable the feature, if its a config flag this is not possible, what if we add the config flag if the feature is enabled? And make it true by default in that case so that a user can turn it off dynamicallly?

That works for me. But who is the developer and who is the user? If the user is a rust developer they can always add hyperlight-host as a dependency and enable the feature. In any case, if we have a runtime option to switch it on/off, I'm fine with it.

However, I think that even with the feature enabled it should be off by default at runtime.

simongdavies:
> > > And I feel that a runtime toggle would provide a stronger guarantee in that sense.
> > 
> > 
> > I think there is a difference, there may be scenarios where the developer never wants a user to be able to enable the feature, if its a config flag this is not possible, what if we add the config flag if the feature is enabled? And make it true by default in that case so that a user can turn it off dynamicallly?
> 
> That works for me. But who is the developer and who is the user? If the user is a rust developer they can always add hyperlight-host as a dependency and enable the feature. In any case, if we have a runtime option to switch it on/off, I'm fine with it.
> 
> However, I think that even with the feature enabled it should be off by default at runtime.

OK I will do this, I think the challenge and where I was using the terms developer and user is that if a developer creates a solution on hyperlight that requires the clock (lets say for arguments sake hyperlight-js) then they don't want the user of their solution to be able to disable the feature via config (presumably it breaks or degrades in some way) this is relatively easy to deal with in todays model (since in the hyperlight-js case the developer could hide the runtime config and thus its always on) but I don't know how we will deal with this in a future unified API where we just load images (that may have dependencies on configs or features that may or may not be turned on) but that's a future problem

docs/guest-time.md
```diff
@@ -0,0 +1,139 @@
...
+
+let now = SystemTime::now()?;
+let unix_ns = now.duration_since(UNIX_EPOCH)?.as_nanos();
+```
+
+C guests that use picolibc get paravirt time for free: `hyperlight_guest_bin`
+wires `clock_gettime(CLOCK_MONOTONIC|CLOCK_REALTIME)` and `gettimeofday` into
+the same reader, so existing C code continues to work unchanged.
+
+## Snapshot / restore semantics
```
syntactically, line 79: If we're going to have this large design doc on this feature, can we include some more of the details, especially the mechanism used to to ensure that the monotonic clock is actually monotonic? From this description, it sounds as if it would not be monotonic, but I see that below there is indeed code doing the sensible thing.

src/hyperlight_common/src/layout.rs
```diff
@@ -34,5 +115,28 @@ pub fn scratch_base_gva(size: usize) -> u64 {
...
+    (MAX_GPA as u64) + 1 - SCRATCH_TOP_EXN_STACK_OFFSET
+}
+
+/// Guest virtual address of the base of the paravirtualized clock page.
+///
+/// See [`clock_page_gpa`]. Scratch is mapped identity-style from
+/// `scratch_base_gva` to `scratch_base_gpa`, so the clock page sits at the
+/// equivalent offset in the guest virtual address space.
+pub const fn clock_page_gva() -> u64 {
+    (MAX_GVA as u64) + 1 - SCRATCH_TOP_EXN_STACK_OFFSET
```
syntactically, line 138: Nit: the reason `MAX_GVA` is defined the way it is, is to avoid integer overflow problems if it's actually the largest representable integer. I think there is some other existing code that gets this wrong and needs to be cleaned up to make that actually work, but it would be nice not to add more... (same comment applies to some of the other uses in this PR)

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,526 @@
...
+//! Which one is active is decided by the host and advertised by the
+//! `clock_type` field in the scratch bookkeeping page. When the host is built
+//! without the `enable_guest_clock` feature the field reads back as
+//! [`ClockType::None`] and every function in this module returns `None`.
+//!
+//! # Concurrency invariant (current)
+//!
+//! In the current Hyperlight execution model the guest vCPU runs only
+//! while the host thread is blocked inside the vCPU run call: the host
+//! writes the clock page **before** entering the guest and cannot mutate
```
syntactically, line 45: That doesn't make sense to me. The hypervisor surely must update this page every time that it gets a timer interrupt, of which there will be many during a single guest execution.

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,526 @@
...
+/// Read time from the KVM pvclock structure.
+///
+/// Uses the seqlock-style protocol described in
+/// <https://docs.kernel.org/virt/kvm/x86/msr.html#pvclock>: the host sets
+/// `version` to an odd value before mutating and to a new even value
+/// afterwards; readers retry while `version` is odd or changes across the
+/// read. We cap retries with [`CLOCK_SEQLOCK_MAX_RETRIES`] so that a
+/// pathologically churning host can't make us spin forever.
+fn read_kvm_pvclock() -> Option<u64> {
+    // SAFETY: see `validate_clock` for the mapping invariant. Today the
```
syntactically, line 189: I'm not aware of any way to make seqlocks compatible with Rust's memory model right now.  Can you lay out the argument in a bit more detail?

If this is technically disallowed, but you think it's reasonable to use based on the same justification we are presently using for the use of u128 volatiles for reading/writing to the guest, please lay that out in a good bit more detail, and reference the justification we have for that.

 Or, perhaps we could rewrite the seqlock accesses in inline asm, which I'm increasingly thinking is what we should do for the read/write from guest memory as well...

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,526 @@
...
+        if version1 & 1 != 0 {
+            core::hint::spin_loop();
+            continue; // Update in progress.
+        }
+
+        // Pair with the hypervisor's write barrier between the version bump
+        // and the payload write. On x86_64 an Acquire fence is free (no
+        // instruction emitted), but we keep it for correctness under the
+        // memory model.
+        fence(Ordering::Acquire);
```
syntactically, line 212: There are no ordering guarantees between atomic fences and non-atomic instructions like the volatile reads surrounding this, so these fences will not really do anything.

docs/guest-time.md
```diff
@@ -0,0 +1,139 @@
...
+on IST1 — cannot clobber the clock page or the metadata page above. The
+allocator reserves the whole four-page region unconditionally so the memory
+map stays identical whether or not the feature is enabled.
+
+## Non-goals
+
+- **Sub-microsecond accuracy.** `boot_time_ns` is computed from two
+  back-to-back host reads (`SystemTime::now()` and `KVM_GET_CLOCK` /
+  `TIME_REF_COUNT`). On KVM, residual disagreement between `KVM_GET_CLOCK`
+  and the pvclock page can add up to ~13ms of constant offset (observed on
```
syntactically, line 132: Is that 13ms or us?

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,526 @@
...
+        let scaled = ((tsc_now as u128 * tsc_scale as u128) >> 64) as u64;
+        let time_100ns = scaled.checked_add_signed(tsc_offset)?;
+
+        return time_100ns.checked_mul(100);
+    }
+
+    None
+}
+
+/// The raw pvclock value from the previous [`monotonic_time_ns`] read: the
```
syntactically, line 296: Why not just have this in the host, either have the host set the appropriate mono offset in scratch or just adjust the actual clock to make it monotonic via `KVM_SET_CLOCK` and friends?  It's unnecessarily fragile to go about trying to detect the clock running backwards in the guest when the host knows precisely when snapshot/restore operations happen and could easily provide the offsets that it ought to know.

docs/guest-time.md
```diff
@@ -0,0 +1,139 @@
...
+  by the host as `wall_now - monotonic_now` (see below). These live at fixed
+  offsets from the top of scratch (`-0x28` and `-0x30`), NOT in the clock
+  page, so a future TLFS extension cannot clobber them.
+
+With those two pieces the guest can compute:
+
+- **Monotonic nanoseconds since boot** — read the TSC, apply the scaling
+  factors from the calibration block, giving you a `CLOCK_MONOTONIC`
+  equivalent.
+- **Wall-clock nanoseconds since the Unix epoch** — add `boot_time_ns` to the
```
syntactically, line 31: "Wall-clock nanoseconds since the Unix epoch" is extremely vague. There should at least be a timescale specified!

src/hyperlight_guest_bin/src/time.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// Panics if `rhs` is later than `self`. Mirrors the behaviour of
+    /// `std::time::Instant::sub`.
+    fn sub(self, rhs: Instant) -> Duration {
+        self.checked_duration_since(rhs)
+            .expect("supplied instant is later than self")
+    }
+}
+
+/// A measurement of the system clock, analogous to
+/// [`std::time::SystemTime`].
```
syntactically, line 155: I'm not convinced that an analogue to `SystemTime` is useful. One shouldn't use it for anything one can avoid because of its terribly broken behaviour around leap seconds (see e.g. https://www.ucolick.org/~sla/leapsecs/), so it's mostly useful for synchronising with other things on the same system that insist on using the broken POSIX "UTC but not quite (if they were UTC you would be able to differentiate files created at 23:59:59 and 23:59:60!)" timestamps. And of course there are more-or-less none of those in Hyperlight.

I generally think that you should use CLOCK_MONOTONIC for everything you possibly can, and only at the very edges where you encounter a human interface should you convert to human-friendly time. However, you cannot get time particularly useful for that out of this, since there's no tzinfo. 

I would much rather see this with just the monotonic clock, or ideally two variants of the monotonic clock (one for "time while the guest is running" and one for "time elapsed since the guest was first bootstrapped", as the current monotonic one is), and then see if we can do a proper job of providing support for converting the latter into human-interface time when necessary (probably by offering to embed tzinfo, or by having the host use its tzinfo to build some tables for the "near future")

src/hyperlight_guest/src/time.rs
```diff
@@ -0,0 +1,526 @@
...
+/// would shift wall time into the future by [`MONO_OFFSET`] after a
+/// cross-partition restore. This signature deliberately cannot see the offset.
+///
+/// Returns `None` when `boot_time_ns` is still 0 — the host hasn't stamped it
+/// yet (scratch is zero-initialised) — rather than a nonsense value.
+fn wall_time_from(boot_time_ns: u64, raw_ns: u64) -> Option<u64> {
+    if boot_time_ns == 0 {
+        return None;
+    }
+    Some(boot_time_ns.wrapping_add(raw_ns))
```
syntactically, line 430: If the goal is really to match broken posix `CLOCK_REALTIME` semantics, this does not achieve that since it does not account for leap seconds.

src/hyperlight_guest_bin/src/time.rs
```diff
@@ -0,0 +1,222 @@
...
+        }
+        let ns = time::wall_clock_time_ns().ok_or(TimeError::Retry)?;
+        Ok(Self { ns })
+    }
+
+    /// Returns the duration from `earlier` to `self`, or
+    /// `TimeError::NegativeDuration(d)` — where `d` is the magnitude of the
+    /// difference — if `earlier` is later than `self`. Mirrors
+    /// `std::time::SystemTime::duration_since`.
+    pub fn duration_since(&self, earlier: SystemTime) -> Result<Duration, TimeError> {
```
syntactically, line 186: It seems like with the rest of the system as it is now, whether or not this includes leap seconds will depend on whether or not they occurred in the interval between a snapshot and a restore, which seems untenable.

syntactically (changes requested):
I have serious concerns about the semantics of the wall-clock suggested in this PR; see the inline comments for further detail.

I will echo Jorge's comments that it would be nice if the monotonic clock only ran when the guest CPU was running. Personally, my ideal design (which this PR could become steps towards the first one or two of) would provide to the guest (with progressive levels of explicit opt-in on the host necessary):
- A monotonic clock that counts all the time spent executing by this guest and its snapshot predecessors
- A monotonic clock that includes also the time in between calls, but not between snapshots and restores
- A monotonic clock that includes also the time in between both calls and snapshot/restores ("wall clock")
- Enough information for the guest to correctly compute calendar time from these, which requires both calendar/leap day and leap second information

Separately, for the monotonic clock specifically (which I believe is intended to have semantics more-or-less matching the second of my three desired time information APIs?), I have concerns about why we are trying to handle the detection of snapshots/restores in the guest, which does not really have the information to do it "properly".

jprendes (approved):
This looks good to me, but we should address @syntactically 's comments.

simongdavies:
Closing this for now as I am going to be away for a while, will pick this up and address the concerns in a few months from now

PR #1438, opened 2026-05-06
Fix C headers archive for picolibc

jprendes:
This PR updates the tar-headers recipe in the Justfile to account for the picolibc changes

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -375,7 +375,7 @@ run-rust-examples-linux target=default-target features="": (run-rust-examples ta
 #########################
 
 tar-headers: (build-rust-capi) # build-rust-capi is a dependency because we need the hyperlight_guest.h to be built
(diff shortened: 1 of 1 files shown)


andreiltd (commented):
Could you remind me why do we copy headers from target dir?

jprendes:
We used to copy the files from different places in the src dir but we already do that when we build the sysroot, I thought it would be easier to copy them from the sysroot directly.
`cargo-hyperlight` builds the sysroot in the target dir.

jprendes:
As to why we do this, it's so that we can publish the sysroot in our release, so that people doing C development doesn't need to install Rust.

We publish 2 .tar.gz archives:
* **hyperlight-guest-c-api-linux.tar.gz**: contains the debug and the release version of `hyperlight-guest-capi` (and all its dependencies)
* **include.tar.gz**: contains the libc headers, plus 2 headers generated by cbindgen for `hyperlight-guest-capi`.

andreiltd (approved):

PR #1439, opened 2026-05-06
Add hyperlight-libc to CI publishing

jprendes:
This PR adds the `hyperlight-libc` to the CI publishing workflow.

### DIFF
--- a/.github/workflows/CargoPublish.yml
+++ b/.github/workflows/CargoPublish.yml
@@ -69,6 +69,7 @@ jobs:
             fi
           }
           
+          needs_publish hyperlight-libc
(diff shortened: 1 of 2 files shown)


andreiltd (approved):
Unrelated but we should really switch to trusted publishing: https://crates.io/docs/trusted-publishing

jprendes:
We are using trusted publishing: https://github.com/hyperlight-dev/hyperlight/pull/1109
That's this part in the workflow
```
      - name: Authenticate with crates.io
        uses: rust-lang/crates-io-auth-action@v1
        id: crates-io-auth
```

PR #1441, opened 2026-05-06
Use latest cargo-hyperlight in release workflow

jprendes:
To generate the header's tar and the lib tar, we use cargo-hyperlight.
With the libc changes we need to make sure we use the latest cargo-hyperlight.


ludfjig (approved):
What about `just ensure-cargo-hyperlight` or modifying it to update version?

jprendes:
I think we should have a better way of updating this in the base image, the base image currently ships with an ancient version of cargo-hyperlight.
I'll keep it as is now since we have the exact same pattern in a few different workflows, and will make it easier to find-and-replace.

PR #1446, opened 2026-05-08
Add snapshot golden tests

ludfjig:
This PR adds tests to help notice breaking changes in the guest<->host ABI. This matters because any change there breaks every snapshot already persisted to disk (bad!).

- persist a set of snapshots to GHCR, and on each PR pull them and run tests against them. Note that these snapshots are not expected to be regenerated often. I'm using GHCR instead of checking in raw binaries into repo to not clutter git repo (there might be a lot of them eventually), and also to exercise the "OCI-ness" of the snapshot format.
- add compile-time tripwires that pin every value defining the format (SNAPSHOT_ABI_VERSION, the media-type strings, the OCI layout version, HyperlightPEB field offsets and size, and the OutBAction ports), helping develpers catch changes that would be breaking
- add some additional regular tests for loading snapshots
- define and outline the development process for breaking snapshot compatibility (and regenerating new golden snapshtos)

When a change to the format is intentional, the author bumps the version and applies the `regen-goldens` label. A workflow then regenerates the goldens against the branch and pushes the new set to GHCR.

See docs/snapshot-versioning.md for details.


.github/workflows/ValidatePullRequest.yml
```diff
@@ -79,17 +79,37 @@ jobs:
...
     needs:
       - docs-pr
       - build-guests
+      - goldens-mode
     # Required because update-guest-locks is skipped on non-dependabot PRs,
     # and a skipped dependency transitively skips all downstream jobs.
     # See: https://github.com/actions/runner/issues/2205
     if: ${{ !cancelled() && !failure() }}
     strategy:
-      fail-fast: true
```
danbugs, line 94: Why? This applies to the entire matrix?

Justfile
```diff
@@ -249,8 +249,10 @@ test-integration target=default-target features="": (witguest-wit)
     @# run component-util integration tests that depend on generated WIT inputs
     {{ cargo-cmd }} test -p hyperlight-component-util --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test wasmtime_guest_codegen
 
-    @# run the rest of the integration tests
-    {{ cargo-cmd }} test -p hyperlight-host {{ if features =="" {''} else if features=="no-default-features" {"--no-default-features" } else {"--no-default-features -F " + features } }} --profile={{ if target == "debug" { "dev" } else { target } }} {{ target-triple-flag }} --test '*'
```
danbugs, line 254: This means that every test must be added here too or it won't run in CI, no?

danbugs (commented):
Mostly LGTM, just 2 minor comments--great work!! Let's just get 1 more person to look at this as it is a pretty big PR.

Justfile
ludfjig: yeah correct. Not ideal but I'm not sure how to fix it. My main concern is that I don't want `cargo test` to require users to pull OCI images in order to pass.

.github/workflows/ValidatePullRequest.yml
ludfjig: whoops good catch will fix!

src/hyperlight_host/tests/snapshot_goldens/platform.rs
```diff
@@ -0,0 +1,186 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Local platform detection and tag naming for snapshot goldens.
+//!
+//! A snapshot is not portable across `(hypervisor, cpu vendor,
+//! build profile)`. Each such triple gets its own tag, named
+//! `{GOLDENS_VERSION}-{hv}-{cpu}-{profile}`.
+
```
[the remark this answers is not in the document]
ludfjig: fixing...

Justfile
ludfjig: fixed! Found a better way to do it. The custom harness noop-passes without a flag

docs/snapshot-versioning.md
```diff
@@ -0,0 +1,353 @@
...
+* **Memory blob ABI**, `SNAPSHOT_ABI_VERSION` (a `u32` inside the
+  config blob, defined in
+  [src/hyperlight_host/src/sandbox/snapshot/file/media_types.rs](../src/hyperlight_host/src/sandbox/snapshot/file/media_types.rs)).
+  This is the host/guest runtime contract baked into the captured
+  memory: the `HyperlightPEB` layout (the struct host and guest share
+  to exchange state, field offsets and types), the `OutBAction` and
+  `VmAction` port numbers (the I/O ports the guest writes to for `Log`,
+  `CallFunction`, `Abort`, `DebugPrint`, and `Halt`), the layout of the
+  sandbox memory regions
+  (stack, heap, guest binary, input and output buffers, page tables),
```
syntactically: Some of these regions the host does not care about the layout of anymore (e.g. stack).

Also, this list should include the scratch bookkeeping area.

Justfile
```diff
@@ -573,3 +576,67 @@ install-vcpkg:
 
 install-flatbuffers-with-vcpkg: install-vcpkg
     cd ../vcpkg && ./vcpkg install flatbuffers || cd -
+
```
syntactically, line 595: Can we have `test-like-ci` run these like CI does please?

syntactically (commented):
I had time only for a pretty quick skim over this.

> add compile-time tripwires that pin every value defining the format (SNAPSHOT_ABI_VERSION, the media-type strings, the OCI layout version, HyperlightPEB field offsets and size, and the OutBAction ports), helping develpers catch 

Why are the PEB fields important? I think the are used only during creation of the original snapshot from a binary?  If they are not important I would like to be careful not to constrain on them for no reason.

> make sure snapshots taken on AMD are usable on Intel, and vice versa

Are we certain already that this is a goal? I thought we purposefully punted on this and compatibility across hypervisors until we had a better sense of whether it would be onerous.

ludfjig:
> Why are the PEB fields important? I think the are used only during creation of the original snapshot from a binary? If they are not important I would like to be careful not to constrain on them for no reason.

You're right, only the size of the PEB struct matters, the fields themselves are never read back by the host.

>> make sure snapshots taken on AMD are usable on Intel, and vice versa

>Are we certain already that this is a goal? I thought we purposefully punted on this and compatibility across hypervisors until we had a better sense of whether it would be onerous.

Addressed in #1587

Justfile
ludfjig: will add!

docs/snapshot-versioning.md
ludfjig: fixed

.github/workflows/dep_build_test.yml
```diff
@@ -174,3 +180,52 @@ jobs:
...
+      # branch. A missing tag fails the job and flags a format break.
+      # The host verifies only its own (arch, hypervisor, cpu vendor,
+      # profile) golden.
+      - name: Snapshot goldens (pull and verify)
+        if: ${{ inputs.regen_goldens != 'true' }}
+        env:
+          GHCR_USER: ${{ github.actor }}
+          GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "${GHCR_TOKEN}" | oras login ghcr.io -u "${GHCR_USER}" --password-stdin
```
simongdavies, line 200: nit: oras login should be in the Justfile?

.github/workflows/dep_build_test.yml
```diff
@@ -174,3 +180,52 @@ jobs:
...
+          just snapshot-goldens-pull ghcr.io/hyperlight-dev/hyperlight-snapshot-goldens ${{ inputs.config }}
+          just snapshot-goldens-verify ${{ inputs.config }}
+
+      # Label path: generate the goldens from the branch and load them
+      # back on this host. Used when no published tag set exists yet.
+      - name: Snapshot goldens (regenerate and verify)
+        if: ${{ inputs.regen_goldens == 'true' }}
```
simongdavies: It feels like this logic should all be in Justfile recipe (or script called by one) otherwise its hard to test locally?

.github/workflows/RegenSnapshotGoldens.yml
```diff
@@ -0,0 +1,267 @@
...
+    permissions:
+      contents: read
+      packages: read
+    outputs:
+      version: ${{ steps.decide.outputs.version }}
+      needs_publish: ${{ steps.decide.outputs.needs_publish }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+
+      - name: Install oras
```
simongdavies, line 71: Shall we add ORAS to our base images?

docs/github-labels.md
```diff
@@ -55,6 +55,12 @@ In addition to **kind/*** labels, we use optional **area/*** labels to specify t
 - **area/security** - Involves security-related changes or fixes.
 - **area/testing** - Related to tests or testing infrastructure.
 
+## Workflow labels
+
+Some labels change CI behaviour on a PR rather than categorizing it:
+
+- **regen-goldens** - Switches the snapshot golden verify job into regenerate mode. A PR that intentionally changes the snapshot format and bumps `GOLDENS_VERSION` carries this label so the verify job generates the goldens from the branch and runs them back through the branch loader, rather than pulling a published tag set that does not exist yet. See [snapshot-versioning.md](snapshot-versioning.md).
```
simongdavies, line 77: This might be in this PR and I may have missed it but does a PR that has this label validate that the ABI vesion changed?

simongdavies (changes requested):
I've not fully reviewed this but I have a couple of suggestions:

Can we move all the login that is embedded as script in the PR workflows into Justfile recipes and/or scripts so that they can easily beused/testd locally?

Do we want to run this on every PR? Can we run nightly and/or on release and then open a release blocker issue if it fails? We might also want running this as a gate to creating a new release?

.github/workflows/dep_build_test.yml
ludfjig: only CI needs to login to avoid rate potential limiting (I am guessing we might be rate limted). Users can just pull without authentication since it's a public package

.github/workflows/dep_build_test.yml
ludfjig: good idea, fixed

.github/workflows/RegenSnapshotGoldens.yml
ludfjig: that would probably be a good idea, filed an issue in the image repo

docs/github-labels.md
ludfjig: good idea, I will add more safety checks

ludfjig:
> Do we want to run this on every PR? Can we run nightly and/or on release and then open a release blocker issue if it fails? We might also want running this as a gate to creating a new release?

It should be very lightweight on each PR, it's just a couple of small tests. I think it's worth it to catch early rather than to discover after the fact

ludfjig:
Pushed some updates. I'd like feedback on the following

- `just test-like-ci` runs the snapshot golden test without first pulling. You will need to manually pull images first using `just snapshot-goldens-verify` to pass the tests. Should we include pulling in this recipe (or pull-if-absent logic maybe).
- there is duplicated logic in bash/rust for detectign host machine details. This is on purpose so users don't need to compile any rust in order to pull images
- `host_cpu_vendor_golden_tag` is a new public method in hyperlight-host, needed for the new integration test harness. Ideally I want to avoid exposing this publicly, but since integration tests ignore cfg(test), i'm sure if there's a better way (adding a cargo feature just for this seems a little extreme as well?)

docs/github-labels.md
ludfjig: added

.github/workflows/dep_build_test.yml
simongdavies: > Users can just pull without authentication since it's a public package

I am not sure this is true for GHCR, IIRC even if you make a package public you have to be authenticated to pull it, i.e. public == authenticated users not anonymous users

simongdavies:
> > Do we want to run this on every PR? Can we run nightly and/or on release and then open a release blocker issue if it fails? We might also want running this as a gate to creating a new release?
> 
> It should be very lightweight on each PR, it's just a couple of small tests. I think it's worth it to catch early rather than to discover after the fact

OK maybe we should add cache entries int the job for the golden images? or perhaps if add the oras cli install to the image then we should pull all the OCIs onto the CI image that way we dont waste time each run waiting for the pulls? Also means that if they change the changes will still get pulled and everything will work.

simongdavies:
> * `just test-like-ci` runs the snapshot golden test without first pulling. You will need to manually pull images first using `just snapshot-goldens-verify` to pass the tests. Should we include pulling in this recipe (or pull-if-absent logic maybe).

Maybe but perhaps we should add the tools as pre-reqs in the docs and then list the pull command/script needed in the docs also?

simongdavies:
> * `host_cpu_vendor_golden_tag` is a new public method in hyperlight-host, needed for the new integration test harness. Ideally I want to avoid exposing this publicly, but since integration tests ignore cfg(test), i'm sure if there's a better way (adding a cargo feature just for this seems a little extreme as well?)

Urghhh , maybe we can make this debug only and no-op the golden tests for release profile??

.github/workflows/dep_build_test.yml
ludfjig: I tried it (installed oras on a vm, didn't login, then pulled) and didn't need any authentication

ludfjig:
> OK maybe we should add cache entries int the job for the golden images? or perhaps if add the oras cli install to the image then we should pull all the OCIs onto the CI image that way we dont waste time each run waiting for the pulls? Also means that if they change the changes will still get pulled and everything will work.

I am not worried about this, the images are really very small and should pull super quick. If we really want to, we could do this in a follow-up. Installing oras into the image is a good idea though.

>>just test-like-ci runs the snapshot golden test without first pulling. You will need to manually pull images first using just snapshot-goldens-verify to pass the tests. Should we include pulling in this recipe (or pull-if-absent logic maybe).

>Maybe but perhaps we should add the tools as pre-reqs in the docs and then list the pull command/script needed in the docs also?

Added to docs in snapshot-versioning.md

>> host_cpu_vendor_golden_tag is a new public method in hyperlight-host, needed for the new integration test harness. Ideally I want to avoid exposing this publicly, but since integration tests ignore cfg(test), i'm sure if there's a better way (adding a cargo feature just for this seems a little extreme as well?)

> Urghhh , maybe we can make this debug only and no-op the golden tests for release profile??

Moved to `hyperlight_host::__private::host_cpu_vendor_golden_tag()`, __private module should signify nobody should use this but us. I do really want to test release as well

simongdavies (approved):

PR #1456, opened 2026-05-14
Move snapshot file + small reset_vcpu/apply_sregs refactor

ludfjig:
Two pre-OCI refactors with no behavior change. Move sandbox/snapshot.rs into a snapshot/ module directory (more files coming there later), and extract HyperlightVm::apply_sregs from reset_vcpu so it can be reused by upcoming snapshot-restore code.


danbugs (approved):
LGTM

PR #1459, opened 2026-05-15
Create sandboxes directly from snapshots

ludfjig:
Builds a ready-to-use MultiUseSandbox directly from an `Arc<Snapshot>` without going through `UninitializedSandbox` + `evolve()`. Building block for OCI-backed snapshot loading on a follow-up branch.

Adds:

- `HostFunctions` newtype around the internal FunctionRegistry. default() pre-registers HostPrint, empty() starts empty.
- `MultiUseSandbox::from_snapshot(snap, host_funcs, cfg)`. Snapshot is the source of truth for layout. cfg is honored for non-layout fields (timeouts, debug info, interrupt retry delay).
- Snapshot records host function metadata. from_snapshot rejects registries missing functions or with mismatched signatures. Pre-init snapshots record nothing and accept any registry, matching the UninitializedSandbox contract.
- gdb support extended to sandboxes built via from_snapshot.

Recommend to review commit-by-commit

### DIFF
--- a/src/hyperlight_host/examples/guest-debugging/main.rs
+++ b/src/hyperlight_host/examples/guest-debugging/main.rs
@@ -115,6 +115,76 @@ mod tests {
     #[cfg(windows)]
     const GDB_COMMAND: &str = "gdb";
 
+    /// Construct the (out_file_path, cmd_file_path, manifest_dir)
+    /// triple every gdb test needs.
+    fn gdb_test_paths(name: &str) -> (String, String, String) {
+        let out_dir = std::env::var("OUT_DIR").expect("Failed to get out dir");
+        let manifest_dir = std::env::var("CARGO_MANIFEST_DIR")
+            .expect("Failed to get manifest dir")
+            .replace('\\', "/");
+        let out_file_path = format!("{out_dir}/{name}.output");
+        let cmd_file_path = format!("{out_dir}/{name}-commands.txt");
+        (out_file_path, cmd_file_path, manifest_dir)
+    }
+
+    /// Build a gdb script that connects to `port`, sets a single
+    /// breakpoint at `breakpoint`, prints `echo_msg` when hit, and
+    /// detaches before quitting.
+    ///
+    /// The breakpoint commands end with `detach` + `quit` instead of
+    /// `continue`. The previous "inner continue, outer continue, quit"
+    /// shape races with the inferior exit. After the breakpoint hits
+    /// and the inner `continue` resumes the guest, the guest may run
+    /// to completion and the gdb stub may close the remote before gdb
+    /// has dispatched the outer `continue`, producing a non-zero exit
+    /// with `Remote connection closed`. Detaching from the breakpoint
+    /// commands removes that window. The host process keeps running
+    /// the guest call to completion on its own after detach.
+    fn single_breakpoint_script(
+        manifest_dir: &str,
+        port: u16,
+        out_file_path: &str,
+        breakpoint: &str,
+        echo_msg: &str,
+    ) -> String {
+        let cmd = format!(
+            "file {manifest_dir}/../tests/rust_guests/bin/debug/simpleguest
+                target remote :{port}
+
+                set pagination off
+                set logging file {out_file_path}
+                set logging enabled on
+
+                break {breakpoint}
+                    commands
+                    echo \"{echo_msg}\\n\"
+                    backtrace
+
+                    set logging enabled off
+                    detach
+                    quit
+                end
+
+                continue
+            "
+        );
+        #[cfg(windows)]
+        let cmd = format!("set osabi none\n{cmd}");
+        cmd
+    }
+
+    /// Spawn the gdb client to execute the script in `cmd_file_path`.
+    fn spawn_gdb_client(cmd_file_path: &str) -> std::process::Child {
+        Command::new(GDB_COMMAND)
+            .arg("-nx")
+            .arg("--nw")
+            .arg("--batch")
+            .arg("-x")
+            .arg(cmd_file_path)
+            .spawn()
+            .expect("Failed to start gdb")
+    }
+
     fn write_cmds_file(cmd_file_path: &str, cmd: &str) -> io::Result<()> {
         let file = File::create(cmd_file_path)?;
         let mut writer = BufWriter::new(file);
@@ -163,14 +233,7 @@ mod tests {
         // wait 3 seconds for the gdb to connect
         thread::sleep(Duration::from_secs(3));
 
-        let mut gdb = Command::new(GDB_COMMAND)
-            .arg("-nx") // Don't load any .gdbinit files
-            .arg("--nw")
-            .arg("--batch")
-            .arg("-x")
-            .arg(cmd_file_path)
-            .spawn()
-            .map_err(|e| new_error!("Failed to start gdb process: {}", e))?;
+        let mut gdb = spawn_gdb_client(cmd_file_path);
 
         // wait 3 seconds for the gdb to connect
         thread::sleep(Duration::from_secs(10));
@@ -245,37 +308,15 @@ mod tests {
     #[test]
     #[serial]
     fn test_gdb_end_to_end() {
-        let out_dir = std::env::var("OUT_DIR").expect("Failed to get out dir");
-        let manifest_dir = std::env::var("CARGO_MANIFEST_DIR")
-            .expect("Failed to get manifest dir")
-            .replace('\\', "/");
-        let out_file_path = format!("{out_dir}/gdb.output");
-        let cmd_file_path = format!("{out_dir}/gdb-commands.txt");
+        let (out_file_path, cmd_file_path, manifest_dir) = gdb_test_paths("gdb");
 
-        let cmd = format!(
-            "file {manifest_dir}/../tests/rust_guests/bin/debug/simpleguest
-                target remote :8080
-
-                set pagination off
-                set logging file {out_file_path}
-                set logging enabled on
-
-                break hyperlight_main
-                    commands
-                    echo \"Stopped at hyperlight_main breakpoint\\n\"
-                    backtrace
-
-                    set logging enabled off
-                    detach
-                    quit
-                end
-
-                continue
-            "
+        let cmd = single_breakpoint_script(
+            &manifest_dir,
+            8080,
+            &out_file_path,
+            "hyperlight_main",
+            "Stopped at hyperlight_main breakpoint",
         );
-
-        #[cfg(windows)]
-        let cmd = format!("set osabi none\n{}", cmd);
 
         let checker = |contents: String| contents.contains("Stopped at hyperlight_main breakpoint");
 
@@ -288,13 +329,8 @@ mod tests {
     #[test]
     #[serial]
     fn test_gdb_sse_check() {
-        let out_dir = std::env::var("OUT_DIR").expect("Failed to get out dir");
-        let manifest_dir = std::env::var("CARGO_MANIFEST_DIR")
-            .expect("Failed to get manifest dir")
-            .replace('\\', "/");
+        let (out_file_path, cmd_file_path, manifest_dir) = gdb_test_paths("gdb-sse");
         println!("manifest dir {manifest_dir}");
-        let out_file_path = format!("{out_dir}/gdb-sse.output");
-        let cmd_file_path = format!("{out_dir}/gdb-sse--commands.txt");
 
         let cmd = format!(
             "file {manifest_dir}/../tests/rust_guests/bin/debug/simpleguest
@@ -329,5 +365,75 @@ mod tests {
 
         cleanup(&out_file_path, &cmd_file_path);
         assert!(result.is_ok(), "{}", result.unwrap_err());
+    }
+
+    #[test]
+    #[serial]
+    fn test_gdb_from_snapshot() {
+        use hyperlight_host::HostFunctions;
+
+        const PORT: u16 = 8081;
+
+        let (out_file_path, cmd_file_path, manifest_dir) = gdb_test_paths("gdb-from-snapshot");
+
+        // Build a sandbox the normal way and snapshot it in-memory.
+        let mut producer: MultiUseSandbox = UninitializedSandbox::new(
+            hyperlight_host::GuestBinary::FilePath(
+                hyperlight_testing::simple_guest_as_string().unwrap(),
+            ),
+            None,
+        )
+        .unwrap()
+        .evolve()
+        .unwrap();
+        let snap = producer.snapshot().unwrap();
+
+        // Order matters. The gdb stub event loop must enter (i.e.
+        // `VcpuStopped` must be sent on the channel) before the gdb
+        // client connects, otherwise the wire protocol desyncs. The
+        // evolve case gets this for free because `evolve()` runs
+        // `vm.initialise()` which trips the entry breakpoint
+        // immediately. For a `Call` snapshot `vm.initialise` is a
+        // no-op, so we trigger the breakpoint by running `sbox.call`
+        // here before the client is launched below.
+        let snap_thread = snap.clone();
+        let sandbox_thread = thread::spawn(move || -> Result<()> {
+            let mut cfg = SandboxConfiguration::default();
+            cfg.set_guest_debug_info(DebugInfo { port: PORT });
+
+            let mut sbox =
+                MultiUseSandbox::from_snapshot(snap_thread, HostFunctions::default(), Some(cfg))?;
+            sbox.call::<i32>(
+                "PrintOutput",
+                "Hello from a from_snapshot sandbox\n".to_string(),
+            )?;
+            Ok(())
+        });
+
+        // Wait for the sandbox thread to bind the listener, install
+        // the one-shot breakpoint, and trip it.
+        thread::sleep(Duration::from_secs(3));
+
+        let cmd = single_breakpoint_script(
+            &manifest_dir,
+            PORT,
+            &out_file_path,
+            "main.rs:simpleguest::print_output",
+            "Stopped at print_output breakpoint",
+        );
+        write_cmds_file(&cmd_file_path, &cmd).expect("Failed to write gdb commands");
+
+        let mut gdb = spawn_gdb_client(&cmd_file_path);
+        let _ = gdb.wait();
+        let sandbox_result = sandbox_thread
+            .join()
+            .expect("from_snapshot sandbox thread panicked");
+
+        let checker = |contents: String| contents.contains("Stopped at print_output breakpoint");
+        let result = check_output(&out_file_path, checker);
+
+        cleanup(&out_file_path, &cmd_file_path);
+        sandbox_result.expect("from_snapshot sandbox returned error");
+        result.expect("gdb output missing expected breakpoint hit");
     }
 }
--- a/src/hyperlight_host/src/mem/layout.rs
+++ b/src/hyperlight_host/src/mem/layout.rs
@@ -226,16 +226,16 @@ pub(crate) struct SandboxMemoryLayout {
     /// The size of the guest code section.
     pub(crate) code_size: usize,
     /// The size of the init data section (guest blob).
-    init_data_size: usize,
+    pub(crate) init_data_size: usize,
     /// Permission flags for the init data region.
     #[cfg_attr(feature = "i686-guest", allow(unused))]
-    init_data_permissions: Option<MemoryRegionFlags>,
+    pub(crate) init_data_permissions: Option<MemoryRegionFlags>,
     /// The size of the scratch region in physical memory.
-    scratch_size: usize,
+    pub(crate) scratch_size: usize,
     /// The size of the snapshot region in physical memory.
-    snapshot_size: usize,
+    pub(crate) snapshot_size: usize,
     /// The size of the page tables (None if not yet set).
-    pt_size: Option<usize>,
+    pub(crate) pt_size: Option<usize>,
 }
 
 impl Debug for SandboxMemoryLayout {
@@ -295,7 +295,7 @@ impl SandboxMemoryLayout {
     /// Both the scratch region and the snapshot region are bounded by
     /// this size. The value is arbitrary but chosen to be large enough
     /// for most workloads while preventing accidental resource exhaustion.
-    const MAX_MEMORY_SIZE: usize = (16 * 1024 * 1024 * 1024) - Self::BASE_ADDRESS; // 16 GiB - BASE_ADDRESS
+    pub(crate) const MAX_MEMORY_SIZE: usize = (16 * 1024 * 1024 * 1024) - Self::BASE_ADDRESS; // 16 GiB - BASE_ADDRESS
 
     /// The base address of the sandbox's memory.
     pub(crate) const BASE_ADDRESS: usize = 0x1000;
--- a/src/hyperlight_host/src/mem/mgr.rs
+++ b/src/hyperlight_host/src/mem/mgr.rs
@@ -22,6 +22,7 @@ use hyperlight_common::flatbuffer_wrappers::function_call::{
 };
 use hyperlight_common::flatbuffer_wrappers::function_types::FunctionCallResult;
 use hyperlight_common::flatbuffer_wrappers::guest_log_data::GuestLogData;
+use hyperlight_common::flatbuffer_wrappers::host_function_details::HostFunctionDetails;
 use hyperlight_common::vmem::{self, PAGE_TABLE_SIZE};
 #[cfg(all(feature = "crashdump", not(feature = "i686-guest")))]
 use hyperlight_common::vmem::{BasicMapping, MappingKind};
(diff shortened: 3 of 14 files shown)


src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,179 @@ impl MultiUseSandbox {
...
+            peb_addr,
+            seed,
+            page_size as u32,
+            &mut hshm,
+            &host_funcs,
+            None,
+            #[cfg(gdb)]
+            dbg_mem_access_hdl,
+        )
+        .map_err(crate::hypervisor::hyperlight_vm::HyperlightVmError::Initialize)?;
```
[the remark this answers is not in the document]
ludfjig: this is known issue, currently specified in https://github.com/hyperlight-dev/hyperlight/pull/1373 I will create issues later

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -115,6 +116,13 @@ pub struct Snapshot {
     /// restored sandbox's guest-visible counter so the guest can tell
     /// which snapshot it is currently a clone of.
     snapshot_generation: u64,
+
+    /// Names and signatures of host functions registered on the
+    /// sandbox at the time this snapshot was taken. Used by
+    /// [`crate::MultiUseSandbox::from_snapshot`] to reject a
+    /// `HostFunctions` set that is missing required functions or
+    /// has mismatched signatures.
+    host_functions: HostFunctionDetails,
```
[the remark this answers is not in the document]
ludfjig: Hash is not used currently. I am planning to remove it in upcoming PR completely

src/hyperlight_host/examples/guest-debugging/main.rs
```diff
@@ -330,4 +366,74 @@ mod tests {
...
+        let mut gdb = spawn_gdb_client(&cmd_file_path);
+        let _ = gdb.wait();
+        let sandbox_result = sandbox_thread
+            .join()
+            .expect("from_snapshot sandbox thread panicked");
+
+        let checker = |contents: String| contents.contains("Stopped at print_output breakpoint");
+        let result = check_output(&out_file_path, checker);
+
+        cleanup(&out_file_path, &cmd_file_path);
```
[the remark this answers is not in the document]
ludfjig: matches existing pattern

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,179 @@ impl MultiUseSandbox {
...
+        #[cfg(gdb)]
+        let dbg_mem_wrapper = Arc::new(Mutex::new(hshm.clone()));
+
+        let mut sbox = MultiUseSandbox::from_uninit(
+            host_funcs,
+            hshm,
+            vm,
+            #[cfg(gdb)]
+            dbg_mem_wrapper,
+        );
```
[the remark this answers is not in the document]
ludfjig: this matches existing evolve code

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -295,7 +295,7 @@ impl SandboxMemoryLayout {
     /// Both the scratch region and the snapshot region are bounded by
     /// this size. The value is arbitrary but chosen to be large enough
     /// for most workloads while preventing accidental resource exhaustion.
-    const MAX_MEMORY_SIZE: usize = (16 * 1024 * 1024 * 1024) - Self::BASE_ADDRESS; // 16 GiB - BASE_ADDRESS
+    pub(crate) const MAX_MEMORY_SIZE: usize = (16 * 1024 * 1024 * 1024) - Self::BASE_ADDRESS; // 16 GiB - BASE_ADDRESS
```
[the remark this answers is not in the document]
ludfjig: I'm using this in the next PR

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -35,8 +35,76 @@ pub struct FunctionRegistry {
...
+    /// `UninitializedSandbox::new()`, so a snapshot taken from a
+    /// regular sandbox can be loaded with
+    /// `MultiUseSandbox::from_snapshot(snap, HostFunctions::default(), None)`
+    /// without registering anything else.
+    ///
+    /// Use [`HostFunctions::new`] for an empty registry.
+    fn default() -> Self {
+        Self(FunctionRegistry::with_default_host_print())
+    }
+}
```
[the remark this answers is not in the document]
ludfjig: fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,179 @@ impl MultiUseSandbox {
...
+        // the override is at least visible.
+        let caller_supplied_config = config.is_some();
+        let mut config = config.unwrap_or_default();
+        if caller_supplied_config {
+            warn_on_layout_override(&config, snapshot.layout());
+        }
+        config.set_input_data_size(snapshot.layout().input_data_size);
+        config.set_output_data_size(snapshot.layout().output_data_size);
+        config.set_heap_size(snapshot.layout().heap_size as u64);
+        config.set_scratch_size(snapshot.layout().get_scratch_size());
```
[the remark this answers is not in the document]
ludfjig: This is fine

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,179 @@ impl MultiUseSandbox {
...
+        config: Option<crate::sandbox::SandboxConfiguration>,
+    ) -> Result<Self> {
+        use rand::RngExt;
+
+        use crate::mem::ptr::RawPtr;
+        use crate::sandbox::uninitialized_evolve::set_up_hypervisor_partition;
+
+        // Validate that the provided host functions are a superset of
+        // those required by the snapshot.
+        snapshot.validate_host_functions(&host_funcs)?;
```
jsturtevant, line 204: I assume the snapshot manager does other validation?  Should this validation live with that validation? if not are we missing validation? Can a maliciously crafted snapshot bypass any assumptions we have about the memory layout?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -2588,4 +2809,222 @@ mod tests {
...
+            let mut sbox = make_sandbox_with_add();
+            let snap = sbox.snapshot().unwrap();
+            let mut hf = HostFunctions::default();
+            hf.register_host_function("Add", |a: String, b: String| Ok(format!("{a}{b}")))
+                .unwrap();
+            let err = MultiUseSandbox::from_snapshot(snap, hf, None)
+                .expect_err("signature mismatch on `Add` must be rejected");
+            let msg = format!("{}", err);
+            assert!(msg.contains("Add"), "got: {}", msg);
+        }
```
jsturtevant, line 3038: we are missing a test that catches the case: `snapshot, register a new host function, snapshot again`

A test that demonstrates this is missing and fails:

```
 let mut sbox = make_sandbox();
let _ = sbox.snapshot().unwrap();

sbox.register_host_function("Echo42", || Ok(42i64)).unwrap();

let snap = sbox.snapshot().unwrap();
let err = MultiUseSandbox::from_snapshot(snap, HostFunctions::default(), None)
    .expect_err("late-registered `Echo42` must be required by the new snapshot");
let msg = format!("{}", err);
assert!(msg.contains("Echo42"), "got: {}", msg);
```

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -943,6 +1122,48 @@ impl std::fmt::Debug for MultiUseSandbox {
     }
 }
 
+/// Emit a warning for each memory-layout field in `caller` that
+/// disagrees with `snapshot`. Used by [`MultiUseSandbox::from_snapshot`]
+/// to surface ignored caller-supplied layout values, since those
+/// fields are always taken from the snapshot.
+fn warn_on_layout_override(
```
jsturtevant, line 1120: what's the consequences of lay override?

jsturtevant (commented):
LGTM, I've left a few comments and one missing testing that needs a fix

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Nothing but it might be confusing, as from the args to the function it looks like any fields on the config will be respected. But instead we ignore some of the passed in ones.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: There's no way to create malicious snapshots (yet). Right now they are completely opaque structs to the user. This will change when we load them from disk though (future PR, where I add a lot of validation)

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Whoops good call let me fix this...

ludfjig: fixed by invalidating snapshot cache when registering host func

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: > (future PR, where I add a lot of validation)

Glad to hear we are planning this.  What if a malicious guest does something (writes some handcrafted assembly) and we take a snapshot with that in there?

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -298,6 +299,7 @@ where
     }
 
     /// Create a snapshot with the given mapped regions
+    #[allow(clippy::too_many_arguments)]
```
danbugs, line 302: Should we consider a snapshot builder to avoid this?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,179 @@ impl MultiUseSandbox {
...
+        config.set_input_data_size(snapshot.layout().input_data_size);
+        config.set_output_data_size(snapshot.layout().output_data_size);
+        config.set_heap_size(snapshot.layout().heap_size as u64);
+        config.set_scratch_size(snapshot.layout().get_scratch_size());
+        let load_info = snapshot.load_info();
+
+        let mgr = crate::mem::mgr::SandboxMemoryManager::from_snapshot(&snapshot)?;
+        let (mut hshm, gshm) = mgr.build()?;
+
+        let page_size = u32::try_from(page_size::get())? as usize;
```
danbugs, line 230: Hm--just noticed set `set_up_hypervisor_partition` takes a `usize` `page_size` but `vm.initialise` takes a u32. Can we improve this to avoid the casts?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -676,6 +676,65 @@ impl Snapshot {
...
+    /// Validate that `provided` is a superset of the host functions
+    /// recorded in this snapshot: every function that was registered
+    /// at snapshot time must also be present in `provided` with a
+    /// matching signature. Extras in `provided` are allowed.
+    ///
+    /// A snapshot with no recorded host functions (e.g. one
+    /// produced by a test-only constructor) accepts any `provided`
+    /// set.
+    pub(crate) fn validate_host_functions(&self, provided: &crate::HostFunctions) -> Result<()> {
+        let required = match &self.host_functions.host_functions {
```
danbugs, line 689: nit: `host_functions.host_functions`

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -676,6 +676,65 @@ impl Snapshot {
...
+        let provided_details: HostFunctionDetails = provided.inner().into();
+        let provided_funcs = provided_details.host_functions.unwrap_or_default();
+
+        let mut missing: Vec<String> = Vec::new();
+        let mut signature_mismatches: Vec<String> = Vec::new();
+
+        for req in required {
+            match provided_funcs
+                .iter()
+                .find(|f| f.function_name == req.function_name)
```
danbugs: Can we improve this lookup? For every req function, we look at all provided functions (i.e., O(req * provided)). What if we built a `HashMap` for the provided functions? This way we'd have a O(provided + req)?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,179 @@ impl MultiUseSandbox {
...
+
+        // If the snapshot was taken from an already-initialized guest
+        // (NextAction::Call), apply the captured special registers so
+        // the guest resumes in the correct CPU state.
+        #[cfg(not(feature = "i686-guest"))]
+        if matches!(snapshot.entrypoint(), super::snapshot::NextAction::Call(_)) {
+            let sregs = snapshot.sregs().ok_or_else(|| {
+                crate::new_error!("snapshot with NextAction::Call must have captured sregs")
+            })?;
+            vm.apply_sregs(hshm.layout.get_pt_base_gpa(), sregs)
```
danbugs, line 292: Might be worth adding a comment that this is overwriting the `apply_sregs` from `vm.initialise`?

src/hyperlight_host/examples/guest-debugging/main.rs
```diff
@@ -115,6 +115,76 @@ mod tests {
     #[cfg(windows)]
     const GDB_COMMAND: &str = "gdb";
 
+    /// Construct the (out_file_path, cmd_file_path, manifest_dir)
+    /// triple every gdb test needs.
+    fn gdb_test_paths(name: &str) -> (String, String, String) {
```
danbugs, line 120: GDB changes look good to me, but it might be worth getting @dblnz to have a look :)

danbugs (commented):
Mostly LGTM!

src/hyperlight_host/src/mem/mgr.rs
ludfjig: Hmm I don't think so, none of these are really optional in that sense

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: good point, but I'd prefer to fix this in follow up. I don't think we should require both methods to take in page size as well, that seems strange

src/hyperlight_host/src/sandbox/snapshot/mod.rs
ludfjig: good point let me investigate. HostFunc is already based on HashMap

src/hyperlight_host/examples/guest-debugging/main.rs
ludfjig: agree I would be more confident if @dblnz took a look as well :)

src/hyperlight_host/src/sandbox/snapshot/mod.rs
ludfjig: fixed

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: it's not, vm.initialize is a noop

src/hyperlight_host/src/sandbox/snapshot/mod.rs
ludfjig: Agree. I can fix this in follow up as well

simongdavies:
> Snapshot is the source of truth for layout. cfg is honored for non-layout fields (timeouts, debug info, interrupt retry delay).

Can we split the config out so we have a runtime config or something similar? Or maybe this is a part of a builder in future API changes?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -226,16 +226,16 @@ pub(crate) struct SandboxMemoryLayout {
     /// The size of the guest code section.
     pub(crate) code_size: usize,
     /// The size of the init data section (guest blob).
-    init_data_size: usize,
+    pub(crate) init_data_size: usize,
```
simongdavies, line 229: Why are these changed? Are they used outside current scope anywhere?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,180 @@ impl MultiUseSandbox {
...
+        config: Option<crate::sandbox::SandboxConfiguration>,
+    ) -> Result<Self> {
+        use rand::RngExt;
+
+        use crate::mem::ptr::RawPtr;
+        use crate::sandbox::uninitialized_evolve::set_up_hypervisor_partition;
+
+        // Validate that the provided host functions are a superset of
+        // those required by the snapshot.
+        snapshot.validate_host_functions(&host_funcs)?;
```
simongdavies, line 204: I notice restore does not validate host funcs, consider this scenario:

```rust
let sa = from_snapshot(s0, {HostPrint, Add});
let sb = from_snapshot(s0, {HostPrint, Add, Mul});
let s1 = sb.snapshot();
sa.restore(s1); // works because id is equal but hf are different
```

src/hyperlight_host/src/mem/layout.rs
ludfjig: They are used in the next PR, just happened to go in with this one 😅

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: This is true and not something I considered. However I am not sure this is an issue. It might actually be beneficial. Maybe it's sufficient to documen that restoring to a snapshot does not modify the sandbox's registered host functions?

jsturtevant:
It sounds like this is a good step forward. Could we make sure we track any remaining work items in https://github.com/hyperlight-dev/hyperlight/issues/1268 so we can make sure they are closed before the next release?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: > any assumptions we have about the memory layout?

Our security model says that the host needs to secure against a fully compromised, running arbitrary code, guest, so at least in theory, there should not be any security-sensitive assumptions---rather, things will simple Not Work if the sandbox isn't cooperating, which is fine.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: Ideally we would split the SandboxConfiguration type into _Sandbox_Configuration and InitialSnapshotConfiguration or something like that.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
syntactically: > I don't think we should require both methods to take in page size as well, that seems strange

@ludfjig see https://github.com/hyperlight-dev/hyperlight/pull/1474/changes/46d81e63af9389e8262ace5889bbf6c76255b3c5, which removes the argument from `HyperlightVm::initialise`.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -676,6 +676,65 @@ impl Snapshot {
     pub(crate) fn entrypoint(&self) -> NextAction {
         self.entrypoint
     }
+
+    /// Validate that `provided` is a superset of the host functions
+    /// recorded in this snapshot: every function that was registered
+    /// at snapshot time must also be present in `provided` with a
+    /// matching signature. Extras in `provided` are allowed.
```
syntactically, line 683: Are you thinking of a particular use case for allowing extras?

I can see the potential benefit to this because it makes it very easy to make a base snapshot depending on a minimal set of functions and then make treelike layers that use more. However, I want to call it out explicitly, since it seems like a significant decision about the semantic model of snapshotting (as far as I can recall, it _used_ to be the case that if you tried, in a single snapshot, to call a host function and had not-found failure, that could never change---but now it can, _and_ only when a snapshot/restore cycle is involved). 

If we do want this, I might be in favour of going all-in and also allowing registration of extra functions after the creation of the `Sandbox`, which I believe was previously impossible.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -145,6 +145,180 @@ impl MultiUseSandbox {
...
+            vm,
+            #[cfg(gdb)]
+            dbg_mem_wrapper,
+        );
+        // Use the snapshot's sandbox_id so that restore() back to this
+        // snapshot is permitted. The id is process-local and never
+        // persisted to disk: `Snapshot::from_oci` assigns a fresh id
+        // on every load, so two `from_file` calls of the same path
+        // yield restore-incompatible sandboxes (which is the intended
+        // safer default). Sandboxes built from clones of the same
```
syntactically, line 315: > so two `from_file` calls of the same path yield restore-incompatible sandboxes (which is the intended safer default).

This seems unusable for a lot of use cases, where we do in fact need to switch one sandbox between multiple guests. I think the sandbox ID was intended mostly as a stopgap for the fact that we could have layout and host-function mismatches, both of which this avoids. Can we replace the sandbox ID with a deterministic hash of the layout and host function, or even remove it entirely and do the appropriate checks in reset() as well as here?

src/hyperlight_host/examples/guest-debugging/main.rs
```diff
@@ -330,4 +366,74 @@ mod tests {
...
+                hyperlight_testing::simple_guest_as_string().unwrap(),
+            ),
+            None,
+        )
+        .unwrap()
+        .evolve()
+        .unwrap();
+        let snap = producer.snapshot().unwrap();
+
+        // Order matters. The gdb stub event loop must enter (i.e.
```
syntactically, line 391: This should probably be documented somewhere obvious, as I imagine the fact that the wire protocol won't connect properly until the call happens may be non-intuitive.

syntactically (dismissed):
Happy to see this merge. I have a couple of purely informative comments/questions that I'm curious about the answers to, but these are not at all blockers to merging.

As well as the inline comments, I will remark that the lifecycle of (un)initialised snapshots and (un)initialised sandboxes has become somewhat complex. I assume that there is an intention to remove the `UninitialisedSandbox` entirely in a follow-up PR, which will help. However, I also wonder a little whether the uninitialised/initialised Snapshot dichotomy makes sense at all---especially since it leads to the question of whether we need to keep the distinction on disk. I am curious about opinions on running the initialisation in `Snapshot::from_env`, so that more-or-less all the other code assumes initialisation is complete.

Also not blocking for this, but if we don't have one already, please make a follow-up issue about the need for a reseed-random  path, probably in both restore and this creation function.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
syntactically: It was brought to my attention that we actually added the "all-in" feature without really discussing it when a lot of high-priority things were going in quickly a few weeks ago. I would still like to draw attention to the fact that it's there and see if there are any objections, and document it better if not.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Agree, this will be addressed https://github.com/hyperlight-dev/hyperlight/issues/1268?issue=hyperlight-dev%7Chyperlight%7C1479

src/hyperlight_host/examples/guest-debugging/main.rs
ludfjig: agree, will fix https://github.com/hyperlight-dev/hyperlight/issues/1268?issue=hyperlight-dev%7Chyperlight%7C1479

src/hyperlight_host/src/sandbox/snapshot/mod.rs
ludfjig: It's a good point, and important semantically. I think that given we already support adding host functions after sandbox is created, this is equivalent. Listed here as well https://github.com/hyperlight-dev/hyperlight/issues/1268?issue=hyperlight-dev%7Chyperlight%7C1479

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Added documentation to `restore()`

jsturtevant (approved):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: Fixing in #1499

PR #1464, opened 2026-05-19
Validate Guest Address Ranges for Overlapping Regions in map_region

### LINKED ISSUE #1289: Validate guest address ranges for overlapping regions in `map_region` / `map_file_cow`
## Summary

`HyperlightVm::map_region` does not check whether a new guest address range overlaps with an already-mapped region. The `VirtualMachine::map_memory` safety contract explicitly requires non-overlapping regions, but `map_region` does not enforce this precondition. On WHP, overlapping mappings return a cryptic HRESULT (confirmed — `WHvMapGpaRange2` rejects the call and the error is wrapped as a generic `WindowsError`). On KVM and MSHV, the behaviour is kernel/hypervisor-version-dependent and may range from an explicit error to silent shadowing of the previous mapping, risking data corruption.

## Current Behaviour

`HyperlightVm::map_region` (in `src/hyperlight_host/src/hypervisor/hyperlight_vm.rs`) validates page alignment of guest and host addresses, assigns a unique slot, and calls the platform-specific `map_memory`. It does **not** check the new `guest_region` against existing entries in `self.mmap_regions`.

The existing mapped regions are readily available via `self.mmap_regions` (a `Vec<(u32, MemoryRegion)>`), so adding the check is straightforward.

Richard-Durkee:
Add overlap validation to `HyperlightVm::map_region` to enforce the safety contract documented on `VirtualMachine::map_memory`, which requires non-overlapping regions.

Checks the new region against existing dynamically mapped regions (`mmap_regions`), the snapshot region (starting at `BASE_ADDRESS`), and the scratch region (at the top of the guest physical address space).

Adds `Overlapping` variant to `MapRegionError` with a descriptive message showing both the new and conflicting ranges.

Also adds early validation at the `MultiUseSandbox::map_region` level to reject invalid input before side effects (snapshot reset) occur.

Tested on KVM (`c8i.2xlarge` with nested virtualization enabled).

Closes #1289

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
@@ -243,6 +243,15 @@ pub enum MapRegionError {
     MapMemory(#[from] MapMemoryError),
     #[error("Region is not page-aligned (page size: {0:#x})")]
     NotPageAligned(usize),
+    #[error(
+        "Region [{new_start:#x}..{new_end:#x}) overlaps existing region [{existing_start:#x}..{existing_end:#x})"
+    )]
+    Overlapping {
+        new_start: usize,
+        new_end: usize,
+        existing_start: usize,
+        existing_end: usize,
+    },
 }
 
 /// Errors that can occur when unmapping a memory region
@@ -418,6 +427,53 @@ impl HyperlightVm {
         .any(|x| x % self.page_size != 0)
         {
             return Err(MapRegionError::NotPageAligned(self.page_size));
+        }
+
+        let new_start = region.guest_region.start;
+        let new_end = region.guest_region.end;
+
+        // Check against existing dynamically mapped regions
+        for (_, existing) in &self.mmap_regions {
+            if new_start < existing.guest_region.end && new_end > existing.guest_region.start {
+                return Err(MapRegionError::Overlapping {
+                    new_start,
+                    new_end,
+                    existing_start: existing.guest_region.start,
+                    existing_end: existing.guest_region.end,
+                });
+            }
+        }
+
+        // Check against the snapshot region
+        if let Some(ref snapshot) = self.snapshot_memory {
+            let snap_start = crate::mem::layout::SandboxMemoryLayout::BASE_ADDRESS;
+            #[cfg(not(unshared_snapshot_mem))]
+            let snap_end = snap_start + snapshot.guest_mapped_size();
+            #[cfg(unshared_snapshot_mem)]
+            let snap_end = snap_start + snapshot.mem_size();
+            if new_start < snap_end && new_end > snap_start {
+                return Err(MapRegionError::Overlapping {
+                    new_start,
+                    new_end,
+                    existing_start: snap_start,
+                    existing_end: snap_end,
+                });
+            }
+        }
+
+        // Check against the scratch region
+        if let Some(ref scratch) = self.scratch_memory {
+            let scratch_start =
+                hyperlight_common::layout::scratch_base_gpa(scratch.mem_size()) as usize;
+            let scratch_end = scratch_start + scratch.mem_size();
+            if new_start < scratch_end && new_end > scratch_start {
+                return Err(MapRegionError::Overlapping {
+                    new_start,
+                    new_end,
+                    existing_start: scratch_start,
+                    existing_end: scratch_end,
+                });
+            }
         }
 
         // Try to reuse a freed slot first, otherwise use next_slot
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -568,9 +568,10 @@ impl MultiUseSandbox {
             // writes can be rolled back when necessary.
             log_then_return!("TODO: Writable mappings not yet supported");
         }
-        // Reset snapshot since we are mutating the sandbox state
-        self.snapshot = None;
+
+        // Map first so overlaps are rejected before resetting the snapshot
         unsafe { self.vm.map_region(rgn) }.map_err(HyperlightVmError::MapRegion)?;
+        self.snapshot = None;
         self.mem_mgr.mapped_rgns += 1;
         Ok(())
     }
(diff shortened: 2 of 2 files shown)


ludfjig (commented):
thanks for your contribution!

I think we can drop the validation in `MultiUseSandbox::map_region` entirely and rely on the check inside `HyperlightVm::map_region`, then reorder so the call happens before the snapshot reset:

```rust
unsafe { self.vm.map_region(rgn) }.map_err(HyperlightVmError::MapRegion)?;
self.snapshot = None;
self.mem_mgr.mapped_rgns += 1;
```

The same pattern can probably applies to map_file_cow too I think. What do you think about this?

Richard-Durkee:
> thanks for your contribution!
> 
> I think we can drop the validation in `MultiUseSandbox::map_region` entirely and rely on the check inside `HyperlightVm::map_region`, then reorder so the call happens before the snapshot reset:
> 
> ```rust
> unsafe { self.vm.map_region(rgn) }.map_err(HyperlightVmError::MapRegion)?;
> self.snapshot = None;
> self.mem_mgr.mapped_rgns += 1;
> ```
> 
> The same pattern can probably applies to map_file_cow too I think. What do you think about this?

Thanks for the feedback! 

This makes sense to me. I just pushed again and included these suggestions. Please let me know if there are any other changes you'd like.

ludfjig:
> > thanks for your contribution!
> > I think we can drop the validation in `MultiUseSandbox::map_region` entirely and rely on the check inside `HyperlightVm::map_region`, then reorder so the call happens before the snapshot reset:
> > ```rust
> > unsafe { self.vm.map_region(rgn) }.map_err(HyperlightVmError::MapRegion)?;
> > self.snapshot = None;
> > self.mem_mgr.mapped_rgns += 1;
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > The same pattern can probably applies to map_file_cow too I think. What do you think about this?
> 
> Thanks for the feedback!
> 
> This makes sense to me. I just pushed again and included these suggestions. Please let me know if there are any other changes you'd like.

Looks great! Just one minor thing I just realized the page-table "tail" of compacted snapshots are not mapped into the vm, so the could be less than `mem_size()`. I think switching to

```rust
#[cfg(not(unshared_snapshot_mem))]
let snap_end = snap_start + snapshot.guest_mapped_size();
#[cfg(unshared_snapshot_mem)]
let snap_end = snap_start + snapshot.mem_size();
```

should do it. Thanks!

Richard-Durkee:
Ah, okay -- I just learned about compacted snapshots. Pushed the change. Thanks again for the help.

ludfjig (dismissed):
LGTM. Would you mind gpg-sign your commit. Thanks alot!

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -243,6 +243,8 @@ pub enum MapRegionError {
     MapMemory(#[from] MapMemoryError),
     #[error("Region is not page-aligned (page size: {0:#x})")]
     NotPageAligned(usize),
+    #[error("Region [{0:#x}..{1:#x}) overlaps existing region [{2:#x}..{3:#x})")]
+    Overlapping(usize, usize, usize, usize),
```
danbugs: nit: maybe make these named fields so you don't have `usize, usize, usize, usize`?

```
  Overlapping {
      new_start: usize,
      new_end: usize,
      existing_start: usize,
      existing_end: usize,
  }
```

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -420,6 +422,50 @@ impl HyperlightVm {
...
+        for (_, existing) in &self.mmap_regions {
+            if new_start < existing.guest_region.end && new_end > existing.guest_region.start {
+                return Err(MapRegionError::Overlapping(
+                    new_start,
+                    new_end,
+                    existing.guest_region.start,
+                    existing.guest_region.end,
+                ));
+            }
+        }
```
danbugs, line 445: nit: I think w/ this we can now remove: https://github.com/hyperlight-dev/hyperlight/blob/29bf1809e20505c3235fff89ce9e19598027287e/src/hyperlight_host/src/sandbox/initialized_multi_use.rs#L645-L655?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -2588,4 +2588,99 @@ mod tests {
...
+            &mem,
+            crate::mem::layout::SandboxMemoryLayout::BASE_ADDRESS,
+            MemoryRegionFlags::READ,
+        );
+        let err = unsafe { sbox.map_region(&region) }.unwrap_err();
+        assert!(
+            format!("{err:?}").contains("Overlapping"),
+            "Expected Overlapping error for snapshot overlap, got: {err:?}"
+        );
+    }
```
danbugs, line 2673: iiuc, we have tests covering overlaps on the mmap_regions as well as snapshots--yes? Would be nice to add a test checking for overlap on scratch too.

danbugs (commented):
Mostly nits, nothing blocking. LGTM.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
Richard-Durkee: Yeah, good idea. Added.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
Richard-Durkee: Good catch. Removed in the updated commit.

ludfjig (changes requested):
I'm back sorry for the delay, I was OOO for a couple of days. Looks good but it looks like you accidentally pushed some unrelated testing files

Richard-Durkee:
> I'm back sorry for the delay, I was OOO for a couple of days. Looks good but it looks like you accidentally pushed some unrelated testing files

Oops -- sorry about that. Should be fixed now.

ludfjig (dismissed):

ludfjig:
looks like the signature got lost, mind signing it again? :P

Richard-Durkee:
> looks like the signature got lost, mind signing it again? :P

Done! Thanks for your patience.

ludfjig (approved):

ludfjig:
thank you for your contribution

PR #1465, opened 2026-05-20
Save/load snapshot from disk (oci format)

ludfjig:
Adds save/load of `Snapshot` to disk as an OCI Image Layout. Public API:

* `Snapshot::save(path, tag)` (returns the manifest's sha256 digest)
* `Snapshot::checked_load(path, reference)` where `reference` is a tag or a sha256 digest (verifies sha256 on every blob)
* `Snapshot::load(path, reference)` (skips the sha256 digest checks for speed)

## Known limitations

* Core dumps from a snapshot-loaded sandbox lack `binary_path` and AT_ENTRY for
  `Call` snapshots. `mem_profile` lacks accurate traces. Fixing either requires
  extending the on-disk format.
* `max_guest_log_level` is not plumbed through snapshot load. It is also
  intrinsically ineffective for `Call` snapshots and should be addressed
  separately.
* The backing OCI directory must not be modified, truncated, renamed over, or
  deleted for the lifetime of a loaded `Snapshot` or any `MultiUseSandbox`
  built from it. On Linux this is unenforced. On Windows the OS refuses the
  operation with `ERROR_USER_MAPPED_FILE` (1224). Firecracker has the same
  constraint:
  > The memory file (pointed by `backend_path` when using `File` backend type,
  > or pointed by `mem_file_path`) must be considered immutable from
  > Firecracker and host point of view. It backs the guest OS memory for read
  > access through the page cache. External modification to this file corrupts
  > the guest memory and leads to undefined behavior.

  [firecracker docs](https://github.com/firecracker-microvm/firecracker/blob/main/docs/snapshotting/snapshot-support.md#loading-snapshots
)

## Future work

Typed error variants. TSC and rand seeding capture. Fuzz target for `checked_load`.
CoW overlay layers. Cross-hypervisor portability via sregs normalisation. Huge
page support (`MAP_HUGETLB`)

### DIFF
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -48,9 +48,14 @@ thiserror = "2.0.18"
 chrono = { version = "0.4", optional = true }
 anyhow = "1.0"
 metrics = "0.24.6"
+serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 elfcore = { version = "2.0", optional = true }
 uuid = { version = "1.23.3", features = ["v4"] }
+oci-spec = { version = "0.8", default-features = false, features = ["image"] }
+sha2 = "0.10"
+hex = "0.4"
+tempfile = "3.27.0"
 
 [target.'cfg(windows)'.dependencies]
 windows = { version = "0.62", features = [
@@ -82,10 +87,8 @@ mshv-ioctls = { version = "0.6", optional = true}
 [dev-dependencies]
 uuid = { version = "1.23.3", features = ["v4"] }
 signal-hook-registry = "1.4.8"
-serde = "1.0"
 iced-x86 = { version = "1.21", default-features = false, features = ["std", "code_asm"] }
 proptest = "1.11.0"
-tempfile = "3.27.0"
 crossbeam-queue = "0.3.12"
 tracing-serde = "0.2.0"
 serial_test = "3.5.0"
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -1568,6 +1568,8 @@ impl ReadonlySharedMemory {
     fn map_file(file: &std::fs::File, len: usize) -> Result<Arc<HostMapping>> {
         use std::os::unix::io::AsRawFd;
 
+        #[cfg(mshv3)]
+        use libc::PROT_WRITE;
         use libc::{
             MAP_ANONYMOUS, MAP_FAILED, MAP_FIXED, MAP_NORESERVE, MAP_PRIVATE, PROT_NONE, PROT_READ,
             mmap, off_t, size_t,
@@ -1605,9 +1607,17 @@ impl ReadonlySharedMemory {
         };
 
         // 2. Overlay the file content on the middle slot with
-        //    `MAP_FIXED`. The guest maps these pages READ|EXECUTE,
-        //    so the host VMA is read-only. `MAP_PRIVATE` keeps the
-        //    mapping detached from the underlying file.
+        //    `MAP_FIXED`. `MAP_PRIVATE` keeps the mapping detached
+        //    from the underlying file.
+        //
+        //    MSHV's map_user_memory requires host-writable pages
+        //    (the kernel module calls `pin_user_pages(FOLL_PIN|FOLL_WRITE)`
+        //    on the region when it is mapped into the partition).
+        //    KVM accepts read-only host pages for read-only guest slots.
+        #[cfg(mshv3)]
+        let file_prot = PROT_READ | PROT_WRITE;
+        #[cfg(not(mshv3))]
+        let file_prot = PROT_READ;
         // SAFETY: `total_size = len + 2 * PAGE_SIZE_USIZE`, so
         // `base + PAGE_SIZE_USIZE` is in-bounds of the reservation.
         let usable_ptr = unsafe { (base as *mut u8).add(PAGE_SIZE_USIZE) };
@@ -1620,7 +1630,7 @@ impl ReadonlySharedMemory {
             mmap(
                 usable_ptr as *mut c_void,
                 len as size_t,
-                PROT_READ,
+                file_prot,
                 MAP_PRIVATE | MAP_FIXED | MAP_NORESERVE,
                 fd,
                 0 as off_t,
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -1521,7 +1521,6 @@ impl ReadonlySharedMemory {
     /// The file's length must be a non-zero multiple of `PAGE_SIZE`.
     /// `guest_mapped_size` must be a non-zero multiple of `PAGE_SIZE`
     /// no greater than the file's length.
-    #[cfg_attr(not(test), expect(dead_code))]
     pub(crate) fn from_file(file: &std::fs::File, guest_mapped_size: usize) -> Result<Self> {
         let len: usize = file
             .metadata()
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -160,6 +160,21 @@ impl MultiUseSandbox {
     /// # Ok(())
     /// # }
     /// ```
+    ///
+    /// From a snapshot loaded from disk:
+    ///
+    /// ```no_run
+    /// # use std::sync::Arc;
+    /// # use hyperlight_host::{HostFunctions, MultiUseSandbox};
+    /// # use hyperlight_host::sandbox::snapshot::{OciTag, Snapshot};
+    /// # fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let tag = OciTag::new("latest")?;
+    /// let snapshot = Arc::new(Snapshot::load("./guest_snapshot", tag)?);
+    /// let mut sandbox = MultiUseSandbox::from_snapshot(snapshot, HostFunctions::default(), None)?;
+    /// let result: String = sandbox.call("Echo", "hello".to_string())?;
+    /// # Ok(())
+    /// # }
+    /// ```
     #[instrument(err(Debug), skip_all, parent = Span::current(), level = "Trace")]
     pub fn from_snapshot(
         snapshot: Arc<Snapshot>,
--- a/src/hyperlight_host/src/sandbox/snapshot/file/config.rs
+++ b/src/hyperlight_host/src/sandbox/snapshot/file/config.rs
@@ -1,0 +1,581 @@
+/*
+Copyright 2025 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use hyperlight_common::flatbuffer_wrappers::function_types::{ParameterType, ReturnType};
+use hyperlight_common::flatbuffer_wrappers::host_function_definition::HostFunctionDefinition;
+use hyperlight_common::vmem::PAGE_SIZE;
+use serde::{Deserialize, Serialize};
+
+use super::media_types::SNAPSHOT_ABI_VERSION;
+use crate::hypervisor::regs::{CommonSegmentRegister, CommonSpecialRegisters, CommonTableRegister};
+use crate::mem::layout::SandboxMemoryLayout;
+
+// --- Arch and hypervisor identifiers --------------------------------
+
+/// Guest architecture the snapshot was captured for. Checked on load
+/// against the running host.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub(super) enum Arch {
+    X86_64,
+    Aarch64,
+}
+
+impl Arch {
+    pub(super) fn current() -> Self {
+        #[cfg(target_arch = "x86_64")]
+        {
+            Self::X86_64
+        }
+        #[cfg(target_arch = "aarch64")]
+        {
+            Self::Aarch64
+        }
+    }
+
+    /// Lowercase token matching the config JSON serialization, used
+    /// for the advisory arch annotation on the manifest descriptor.
+    pub(super) fn as_str(&self) -> &'static str {
+        match self {
+            Self::X86_64 => "x86_64",
+            Self::Aarch64 => "aarch64",
+        }
+    }
+}
+
+/// Hypervisor backend the snapshot was captured under. Checked on
+/// load because vCPU register state is backend-specific.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub(super) enum Hypervisor {
+    Kvm,
+    Mshv,
+    Whp,
+}
+
+impl Hypervisor {
+    pub(super) fn current() -> Option<Self> {
+        #[allow(unused_imports)]
+        use crate::hypervisor::virtual_machine::HypervisorType;
+        use crate::hypervisor::virtual_machine::get_available_hypervisor;
+
+        match get_available_hypervisor() {
+            #[cfg(kvm)]
+            Some(HypervisorType::Kvm) => Some(Self::Kvm),
+            #[cfg(mshv3)]
+            Some(HypervisorType::Mshv) => Some(Self::Mshv),
+            #[cfg(target_os = "windows")]
+            Some(HypervisorType::Whp) => Some(Self::Whp),
+            None => None,
+        }
+    }
+
+    fn name(&self) -> &'static str {
+        match self {
+            Self::Kvm => "KVM",
+            Self::Mshv => "MSHV",
+            Self::Whp => "WHP",
+        }
+    }
+
+    /// Lowercase token matching the config JSON serialization, used
+    /// for the advisory hypervisor annotation on the manifest
+    /// descriptor.
+    pub(super) fn as_str(&self) -> &'static str {
+        match self {
+            Self::Kvm => "kvm",
+            Self::Mshv => "mshv",
+            Self::Whp => "whp",
+        }
+    }
+}
+
+// --- Config JSON shape ----------------------------------------------
+
+/// Top-level Hyperlight snapshot config JSON. Lives at
+/// `blobs/sha256/<config-digest>` with media type
+/// `application/vnd.hyperlight.snapshot.config.v1+json`.
+///
+/// In OCI terms this is the "image config" blob that the manifest's
+/// `config` descriptor points to. It describes the accompanying
+/// memory layer (the snapshot bytes) and everything the loader needs
+/// to reconstruct a runnable `Snapshot`.
+#[derive(Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub(super) struct OciSnapshotConfig {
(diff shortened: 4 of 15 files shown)


src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,693 @@
...
+use self::digest::{Digest256, oci_digest, parse_oci_digest, verify_blob_bytes, verify_blob_file};
+use self::fsutil::{read_bounded, write_blob_if_absent, write_file_atomic};
+use self::media_types::{
+    MT_CONFIG_CURRENT, MT_CONFIG_V1, MT_SNAPSHOT_CURRENT, MT_SNAPSHOT_V1, SNAPSHOT_ABI_VERSION,
+};
+use super::{NextAction, Snapshot};
+use crate::hypervisor::regs::CommonSpecialRegisters;
+use crate::mem::layout::SandboxMemoryLayout;
+use crate::mem::memory_region::MemoryRegionFlags;
+use crate::mem::shared_mem::{ReadonlySharedMemory, SharedMemory};
```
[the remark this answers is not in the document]
ludfjig: this is bull

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1655,6 +1656,15 @@ impl ReadonlySharedMemory {
         //    `MAP_FIXED`. The guest maps these pages READ|EXECUTE,
         //    so the host VMA is read-only. `MAP_PRIVATE` keeps the
         //    mapping detached from the underlying file.
```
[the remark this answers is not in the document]
ludfjig: There's a comment right there

src/hyperlight_host/Cargo.toml
```diff
@@ -48,9 +48,13 @@ thiserror = "2.0.18"
 chrono = { version = "0.4", optional = true }
 anyhow = "1.0"
 metrics = "0.24.6"
+serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 elfcore = { version = "2.0", optional = true }
 uuid = { version = "1.23.2", features = ["v4"] }
+oci-spec = { version = "0.8", default-features = false, features = ["image"] }
```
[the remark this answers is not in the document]
ludfjig: image

src/hyperlight_host/src/sandbox/snapshot/file/fsutil.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Intended for the large snapshot blob, where the cost of one full
+/// re-hash of the existing file is far less than the cost of an
+/// unconditional rewrite.
+pub(super) fn put_blob_if_absent(
+    blobs_dir: &Path,
+    digest: &Digest256,
+    bytes: &[u8],
+) -> crate::Result<()> {
+    let target = blobs_dir.join(&digest.hex);
+    if let Ok(meta) = std::fs::metadata(&target)
```
andreiltd: Do we want to check for symlinks as well? I think if blob is symlink it should be rejected.

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -0,0 +1,555 @@
...
+
+// --- Arch and hypervisor identifiers --------------------------------
+
+/// Guest architecture the snapshot was captured for. Checked on load
+/// against the running host.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+pub(super) enum Arch {
+    X86_64,
+    Aarch64,
```
andreiltd, line 34: Does this require synchronization with `aarch64` support PR? It looks like registers for arm are placeholders only now.

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+        //    the raw memory image. `ReadonlySharedMemory::from_file`
+        //    surrounds it with host guard pages. The guest mapping
+        //    of the snapshot region covers only the data prefix
+        //    (`snapshot_size`). The PT tail sits past that prefix
+        //    in the host mapping and is copied into the scratch
+        //    region on restore. Keeping it out of the guest mapping
+        //    of the snapshot region avoids overlap with
+        //    `map_file_cow` regions installed immediately after the
+        //    snapshot in guest PA space.
+        let memory = ReadonlySharedMemory::from_file(&snap_file, layout.snapshot_size)?;
```
andreiltd, line 816: Should we check memory size here again against `cfg.memory_size`? 

```rust
 if memory.mem_size() as u64 != cfg.memory_size {
     return Err(...);
 }
```

Is it possible that file will be modified between verification and mapping or is file immutability enforced somehow? If it can be modified, I think that without this check later restore can explode, no?

```rust
 let snapshot_pt_end = self.shared_mem.mem_size();
 let snapshot_pt_start = snapshot_pt_end - snapshot_pt_siz
```

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -0,0 +1,555 @@
...
+        Self {
+            function_name: r.function_name,
+            parameter_types: Some(r.parameter_types.into_iter().map(Into::into).collect()),
+            return_type: r.return_type.into(),
+        }
+    }
+}
+
+impl OciSnapshotConfig {
+    pub(super) fn validate_for_load(&self) -> crate::Result<()> {
```
andreiltd, line 447: I think we are missing `stack_top_gva` validation. It looks like it's loaded but not validated.

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+                snap_file_len,
+                snap_desc.size()
+            ));
+        }
+        if verify_blobs {
+            verify_blob_file("snapshot", &mut snap_file, &snap_hex)?;
+        }
+
+        // 6. Reconstruct layout.
+        let mut sbox_cfg = crate::sandbox::SandboxConfiguration::default();
```
andreiltd, line 764: I think all those config fields that are used to calculate layout arithmetic should be verified/bounded before creating layout?

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -0,0 +1,555 @@
...
+    pub(super) input_data_size: usize,
+    pub(super) output_data_size: usize,
+    pub(super) heap_size: usize,
+    pub(super) code_size: usize,
+    pub(super) init_data_size: usize,
+    /// Memory region flag bits. `None` means default permissions.
+    pub(super) init_data_permissions: Option<u32>,
+    pub(super) scratch_size: usize,
+    pub(super) snapshot_size: usize,
+    pub(super) pt_size: Option<usize>,
```
andreiltd, line 161: Why this is optional? Can we populate 0 size pt?

https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_host/src/mem/mgr.rs#L624

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -184,6 +184,20 @@ impl MultiUseSandbox {
...
     /// ```
+    ///
+    /// From a snapshot loaded from disk:
+    ///
+    /// ```no_run
+    /// # use std::sync::Arc;
+    /// # use hyperlight_host::{HostFunctions, MultiUseSandbox};
+    /// # use hyperlight_host::sandbox::snapshot::Snapshot;
+    /// # fn example() -> Result<(), Box<dyn std::error::Error>> {
+    /// let snapshot = Arc::new(Snapshot::from_oci("./guest_snapshot", "latest")?);
```
simongdavies: Is there a reason this is two step? can we not just have `MultiUseSandbox::from_oci`

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+            .map_err(|e| crate::new_error!("from_oci failed to stat {:?}: {}", path, e))?;
+        if !meta.is_dir() {
+            return Err(crate::new_error!(
+                "from_oci path {:?} is not a directory",
+                path
+            ));
+        }
+
+        // 1. oci-layout
+        let layout_bytes = read_bounded(&path.join("oci-layout"), MAX_JSON_BLOB_SIZE)
```
simongdavies: The use of `MAX_JSON_BLOB_SIZE` seems to be asymmetric, in that its not used to validate size in `write_blobs_and _build_descriptor`

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+    /// Use [`Snapshot::from_oci_unchecked`] to skip digest verification
+    /// in trusted paths.
+    ///
+    /// Returns an error for arch, hypervisor, and ABI mismatches.
+    ///
+    /// # File-mutation hazard
+    ///
+    /// Do not modify or replace files in `path` while the returned
+    /// `Snapshot` (or sandboxes built from it) is still alive.
+    pub fn from_oci(path: impl AsRef<Path>, tag: &str) -> crate::Result<Self> {
```
simongdavies: We should consider making this an `enum` that accepts a tag or a digest or introducing `from_oci_by_digest`, its valid to not have a tag and a user would expect it to work , even though we write a tag when we create a snapshot , a user could remove this (or alter it).

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+    /// `imageLayoutVersion`. Otherwise the call errors without
+    /// touching the directory.
+    ///
+    /// # Portability
+    ///
+    /// Snapshot images are bound to a specific CPU architecture and
+    /// hypervisor. Both are recorded in the config blob and checked
+    /// at load time, with mismatches rejected with a clear error.
+    /// The hypervisor tag (kvm/mshv/whp) constrains the host OS.
+    pub fn to_oci(&self, path: impl AsRef<Path>, tag: &str) -> crate::Result<()> {
```
simongdavies: Consider returning the digest here? I think in order to support a user signing the digest is required so this will help in that scenarios

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+
+        // Config blob.
+        let cfg = self.build_config()?;
+        let cfg_bytes = serde_json::to_vec_pretty(&cfg)
+            .map_err(|e| crate::new_error!("failed to serialise config JSON: {}", e))?;
+        let cfg_digest = Digest256::from_bytes(&cfg_bytes);
+        put_blob(&blobs_dir, &cfg_digest, &cfg_bytes)?;
+
+        // Manifest blob.
+        let config_descriptor = DescriptorBuilder::default()
```
simongdavies, line 490: We should add a platform property so that tools can inspect the image and see the architecture/os , we should also add the hypervisor as an annotation so its visible to tools as well

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+            existing.manifests().to_vec()
+        } else {
+            Vec::new()
+        };
+
+        let new_desc = self.write_blobs_and_build_descriptor(path, tag)?;
+
+        // Replacement is by tag, not by digest: a new snapshot may
+        // hash to a different value but still claim the same logical
+        // ref. Blobs from the replaced manifest become orphans.
```
simongdavies, line 423: This is another reason why returning/accepting a digest is important since the image still exists when the tag is reused, if someone accidentally overwrites a tag they can still load the snapshot using a digest

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
simongdavies: Should we require a parameter (overwrite,clobber etc.) to control behaviour of re-using an existing tag?

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -0,0 +1,697 @@
...
+    /// Like [`Snapshot::from_oci`] but skips the sha256 digest check
+    /// on the three content-addressed blobs under
+    /// `path/blobs/sha256/`: the OCI image manifest, the Hyperlight
+    /// config JSON, and the raw snapshot memory image.
+    ///
+    /// # Safety
+    ///
+    /// The caller must guarantee that every blob under `path` is
+    /// byte-for-byte the output of a trusted `Snapshot::to_oci`
+    /// call.
```
simongdavies: Might want to update the comments to advise a user that if they were using a signed image then using this function (even if they check the hashes of the blob(s) themselves before calling this function there is a TOUTOC issue)

simongdavies (changes requested):
A few comments but other than those it looks good, I think we should add some docs/examples showing how to use this with Oras etc. (push/pull/tag etc), sign with notation, etc I think this is useful not just for users but also to validate that what we are doing here works as we intended

simongdavies:
can we get the limitations/future work into a new issue with sub-issues so that we can track all the work that's needed to complete this feature?

ludfjig:
> can we get the limitations/future work into a new issue with sub-issues so that we can track all the work that's needed to complete this feature?

Will put all those things combined in this issue https://github.com/hyperlight-dev/hyperlight/issues/1479

src/hyperlight_host/src/sandbox/snapshot/file/fsutil.rs
ludfjig: Agree. Rejects reading symlinks now

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
ludfjig: Yes there will be some conflicts for sure

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
ludfjig: Oops, fixed!

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: Done, each region is capped at MAX_MEMORY_SIZE now

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: There's nothing preventing the underlying file from being mutated after the verification has passed but before it's mapped. I'm not sure if there's a proper way to fix this. Maybe memfd + F_SEAL_WRITE but not sure the perf implications or if it's overkill for what we want. Any suggestions?

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
ludfjig: pre-init snapshots (snapshots taken of an a guest binary before vm is ran) have 0 here.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: It's technically a snapshot, so I think this API makes sense. For one it kinda prevents `MultiUseSandbox::from_oci` from being called at the same path twice. With the current api it's more clear that you load the snapshot once, and then can create 2 sandboxes from the same 1 snapshot.

If we want, we can add a convenience method later, but I feel like we should think more about it. Tracking in #1479

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: It's there to cap the untrusted read allocation, I don't think we need it on the write site?

simongdavies: But a write could create a file that is too big (1000s of functions) and then the creation of the snapshot would work but the restore would fail with a weird error, seems odd to me that its possible to create a snapshot that cannot be used for restore

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
andreiltd: I think what I suggested above is just enough: comparing size against config and error out on mismatch?

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: This is true. I'm working on redesigning the API a bit

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: good idea

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: agree

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: it's true. But the cap is already big enough for 1000s of functions

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: fixed!

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: fixed!

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: fixed!

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: I thought about it, but I don't think we need it given the current semantics. Nothing is ever overwritten. A tag can be moved to point to a new digest, but that's it

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: thanks, added

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: would you prefer we error out if we notice the written file is too large?

andreiltd (dismissed):
LGTM! 🚀

andreiltd (dismissed):

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: Change api and docs, but I'm not saying anything about signatures or signing here at all

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
simongdavies: Yes, I think that its better to have a clear error when writing than an ambiguous one when reading.

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
simongdavies: OK , I think if we have a clear set of examples and docs about how to use the APIs this is ine.

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: fixed + added tests

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
ludfjig: sounds good will add this later, it's in the tracking issue

simongdavies (approved):
LGTM, nice work

PR #1476, opened 2026-05-27
perf(mshv): skip RIP advance on VmAction::Halt fast path

ludfjig:
Fixes an issue where we did an extra IOCTL when it wasn't needed (regs gets set anyway on next guest call). 

Note that the perf regressions were mostly fixed by https://github.com/hyperlight-dev/hyperlight/pull/1366 due to using VP page for setting registers instead of IOCTL, so the impact of this PR is minor. But it's still good to have


jsturtevant (approved):

dblnz (approved):
LGTM!

PR #1477, opened 2026-05-27
Fix broken rust cache

ludfjig:
Our existing caching is broken because the cache is written from pr branches, but cannot be consumed by other branches 
https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache.

This PR adds a workflow on default branch (main) to prime the caches (to make it available to all branches), and makes PRs branches read-only for the caches. Also splits cache on kvm (ubuntu) and mshv (azl3) since I believe this is necessary.

### DIFF
--- a/.github/workflows/dep_benchmarks.yml
+++ b/.github/workflows/dep_benchmarks.yml
@@ -97,7 +97,7 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          shared-key: "${{ runner.os }}-release"
+          shared-key: "${{ runner.os }}-${{ inputs.hypervisor }}-release"
           save-if: "false"
 
       - name: Download Rust guests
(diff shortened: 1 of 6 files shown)


.github/workflows/dep_benchmarks.yml
```diff
@@ -97,7 +97,7 @@ jobs:
       - name: Rust cache
         uses: Swatinem/rust-cache@v2
         with:
-          shared-key: "${{ runner.os }}-release"
+          shared-key: "${{ runner.os }}-${{ inputs.hypervisor }}-release"
           save-if: "false"
```
jsturtevant, line 101: why do we have this here if its always false?

jsturtevant: found my answer `# If `false`, the cache is only restored.`

jsturtevant (approved):

PR #1500, opened 2026-06-02
refactor: replace tracing-log with native tracing macros for guest log forwarding

### LINKED ISSUE #1028: Use `tracing` crate instead of `log` to create logs

cshung:
## Summary

Remove the `tracing-log` dependency from hyperlight-host and replace the `format_trace`/`log::Record` approach in `outb_log` with direct tracing event macros (`error!`, `warn!`, `info!`, `debug!`, `trace!`).

This consolidates guest log emission on the `tracing` crate, while preserving backward compatibility for consumers using only the `log` crate (via tracing's built-in `log` feature).

## Changes

- **src/hyperlight_host/Cargo.toml**: Removed `tracing-log` dependency
- **src/hyperlight_host/src/sandbox/outb.rs**: Replaced `format_trace`/`log::Record` with native `tracing` macros matched on `LogLevel`
- **src/hyperlight_host/tests/integration_test.rs**: Switched `use log::{error, trace}` to `use tracing::{error, trace}`
- **src/tests/rust_guests/simpleguest/src/main.rs**: Switched `log::info!` to `tracing::info!` (kept `log` import for backward compat test)

## Testing

- All unit tests pass (debug + release)
- All integration tests pass (debug + release)
- Clippy clean (debug + release)
- `test_log_outb_log` verifies backward compatibility with `log`-only consumers
- `test_trace_outb_log` verifies tracing subscriber path

Closes #1028

### DIFF
--- a/src/hyperlight_host/src/sandbox/outb.rs
+++ b/src/hyperlight_host/src/sandbox/outb.rs
@@ -86,7 +86,7 @@ pub(super) fn outb_log(
         LogLevel::Error | LogLevel::Critical => {
             tracing::error!(
                 target: "hyperlight_guest",
-                guest_source_file = source_file,
+                guest_file = source_file,
                 guest_line = line,
                 guest_module = source,
                 "{}",
@@ -96,7 +96,7 @@ pub(super) fn outb_log(
         LogLevel::Warning => {
             tracing::warn!(
                 target: "hyperlight_guest",
-                guest_source_file = source_file,
+                guest_file = source_file,
                 guest_line = line,
                 guest_module = source,
                 "{}",
@@ -106,7 +106,7 @@ pub(super) fn outb_log(
         LogLevel::Information => {
             tracing::info!(
                 target: "hyperlight_guest",
-                guest_source_file = source_file,
+                guest_file = source_file,
                 guest_line = line,
                 guest_module = source,
                 "{}",
@@ -116,7 +116,7 @@ pub(super) fn outb_log(
         LogLevel::Debug => {
             tracing::debug!(
                 target: "hyperlight_guest",
-                guest_source_file = source_file,
+                guest_file = source_file,
                 guest_line = line,
                 guest_module = source,
                 "{}",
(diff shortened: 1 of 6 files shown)


src/hyperlight_host/src/sandbox/outb.rs
```diff
@@ -65,64 +62,77 @@ pub enum HandleOutbError {
...
+    // Consumers using a `log` logger (without a tracing subscriber) still
+    // receive these events thanks to the `tracing` crate's `log` feature,
+    // which forwards tracing events to the `log` facade when no subscriber
+    // is set.
+    let source_file = log_data.source_file.as_str();
+    let line = log_data.line;
+    let source = log_data.source.as_str();
+    let message = log_data.message.as_str();
+
+    match &log_data.level {
```
ludfjig, line 85: Can we rename the fields to `guest.file`, `guest.line`, and `guest.module` instead

ludfjig (changes requested):
Thanks for contribution! Would you mind also updating `docs/hyperlight-metrics-logs-and-traces.md` (remove any mention of tracing-log crate for exapmle). Thanks!

Eventually I want to get rid of all logging (non-tracing) code in both guest and host, so if you're up for doing something more relating to this topic we would greatly appreciate it!

src/hyperlight_host/src/sandbox/outb.rs
ludfjig: Are only underscores supported (guest_file, guest_line, guest_module), or would periods (guest.file, guest.line, guest.module) work too?

cshung: Only underscores are supported. Tracing macros use Rust identifier syntax for field names, and dots are not valid identifier characters. If dotted names are needed in the output (e.g., for OpenTelemetry semantic conventions), a custom subscriber layer could rename them at emit time.

ludfjig (approved):
Thank you!

Issue #1483, opened 2026-05-29
Rust guests uses picolibc


ludfjig:
This is unintended, and is because simpleguest depends on hyperlight-guest-bin which has default libc feature. We should turn it off default-features for simpleguest rust to make sure we get llvm memcpy etc, and not the libc stuff

PR #1501, opened 2026-06-03
Create inner workspace for rust guests to reduce compile time

ludfjig:
Reduces time to build rust guests by putting them in a workspace together. This avoids rebuilding hyperlight-libc from scratch and hyperlight-guest-bin from scratch for each guest which is very slow. On my machine `just guests` time is reduced by 6 minutes.

Aside: rust guests shouldn't need hyperlight-libc (https://github.com/hyperlight-dev/hyperlight/issues/1483), but this is a good change regardless since hyperlight-guest-bin will be shared even if the libc issue is fixed in the future


simongdavies (approved):
Nice!! LGTM

PR #1505, opened 2026-06-04
Build rust guests without hyperlight-libc

### LINKED ISSUE #1483: Rust guests uses picolibc

ludfjig:
Rust guests transitively depend on hyperlight-libc (picolibc) through hyperlight-guest-bin's default libc feature. The only reason the dependency was needed was an unconditional call to picolibc's srand in generic_init. The Rust guests themselves never use it.

This PR: 
- Gates the srand extern and the call in generic_init behind #[cfg(feature = "libc")].
- Sets default-features = false on hyperlight-guest-bin in simpleguest, dummyguest, and witguest

I am also strongly considering making cargo feature libc a non-default cargo feature of hyperlight-guest-bin, since I am guessing most people who write rust guests do not need hyperlight-libc, but would appreciate more thoughts/feedback on that.

Closes #1483

### DIFF
--- a/src/hyperlight_guest_bin/src/lib.rs
+++ b/src/hyperlight_guest_bin/src/lib.rs
@@ -203,6 +203,8 @@ fn _panic_handler(info: &core::panic::PanicInfo) -> ! {
 
 unsafe extern "C" {
     fn hyperlight_main();
+
+    #[cfg(feature = "libc")]
     fn srand(seed: u32);
 }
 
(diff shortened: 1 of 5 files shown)


andreiltd (approved):
LGTM!

PR #1514, opened 2026-06-05
feat(virtq): add high-level packed virtqueue producer/consumer api

andreiltd:
This patch adds a higher level api on top of the packed ring primitives in hyperlight_common so callers don't manage raw descriptors directly.

- Add VirtqProducer/VirtqConsumer in producer.rs and consumer.rs that handle buffer allocation, chain lifecycle, and event suppression/notification decisions; expand virtq/mod.rs with the public API, builders, and docs.
- Add buffer management: the BufferProvider trait and shared types in buffer.rs plus two concrete allocators - a two-tier variable-size BufferPool and a fixed-slot RecyclePool.
- Add an 8-byte wire message header in msg.rs  for message type discrimination and request/response correlation.
- Add loom-based concurrency tests in concurrency.rs using shadow atomics
- Add criterion benchmarks for the buffer pool and virtq API.

### DIFF
--- a/src/hyperlight_common/Cargo.toml
+++ b/src/hyperlight_common/Cargo.toml
@@ -19,6 +19,8 @@ arbitrary = {version = "1.4.2", optional = true, features = ["derive"]}
 anyhow = { version = "1.0.102", default-features = false }
 bitflags = "2.13.0"
 bytemuck = { version = "1.24", features = ["derive"] }
+bytes = { version = "1", default-features = false }
+fixedbitset = { version = "0.5.7", default-features = false }
 flatbuffers = { version = "25.12.19", default-features = false }
 log = "0.4.32"
 smallvec = "1.15.1"
@@ -39,9 +41,22 @@ nanvix-unstable = ["i686-guest"]
 guest-counter = []
 
 [dev-dependencies]
+criterion = "0.8.1"
+hyperlight-testing = { workspace = true }
 quickcheck = "1.0.3"
 rand = "0.10.1"
+
+[target.'cfg(loom)'.dev-dependencies]
+loom = "0.7"
 
 [lib]
 bench = false # see https://bheisler.github.io/criterion.rs/book/faq.html#cargo-bench-gives-unrecognized-option-errors-for-valid-command-line-options
 doctest = false # reduce noise in test output
+
+[[bench]]
+name = "buffer_pool"
+harness = false
+
+[[bench]]
+name = "virtq_api"
+harness = false
--- a/src/hyperlight_common/src/virtq/concurrency.rs
+++ b/src/hyperlight_common/src/virtq/concurrency.rs
@@ -1,0 +1,540 @@
+/*
+Copyright 2026  The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Loom-based concurrency testing for the virtqueue implementation.
+//!
+//! Loom will explores all possible thread interleavings to find data races
+//! and other concurrency bugs. However, it has specific requirements that
+//! make our memory model more involved:
+//!
+//! ## Flag-Based Synchronization
+//!
+//! The virtqueue protocol uses flag-based synchronization:
+//! 1. Producer writes descriptor fields (addr, len, id), then writes flags with release semantics
+//! 2. Consumer reads flags with acquire semantics, then reads descriptor fields
+//!
+//! Loom  would see this as concurrent access to the same memory and report a race, even though
+//! acquire/release on flags provides proper synchronization.
+//!
+//! ## Shadow Atomics for Flags
+//!
+//! We maintain shadow atomics that loom tracks for synchronization:
+//!
+//! - `desc_flags`: One `AtomicU16` per descriptor for flags field
+//! - `drv_flags`: `AtomicU16` for driver event suppression flags
+//! - `dev_flags`: `AtomicU16` for device event suppression flags
+//!
+//! The `load_acquire`/`store_release` operations use these loom atomics, while
+//! `read`/`write` access the underlying data. All shared regions (descriptors,
+//! the event-suppression structs, and the pool) are held in
+//! `loom::cell::UnsafeCell`s, so loom also tracks every data access and verifies
+//! the flag-based synchronization actually orders them race-free.
+//!
+//! ## Memory Regions
+//!
+//! We use a `BTreeMap` to map addresses to memory regions:
+//! - `Desc(idx)`: Individual descriptors in the ring
+//! - `DrvEvt`: Driver event suppression structure
+//! - `DevEvt`: Device event suppression structure
+//! - `Pool`: Buffer pool for chain payload data
+
+use alloc::collections::BTreeMap;
+use alloc::sync::Arc;
+use alloc::vec;
+use core::num::NonZeroU16;
+
+use bytemuck::Zeroable;
+use loom::sync::atomic::{AtomicU16, AtomicUsize, Ordering};
+use loom::thread;
+
+use super::*;
+use crate::virtq::desc::Descriptor;
+use crate::virtq::pool::BufferPoolSync;
+
+#[derive(Debug)]
+pub struct MemErr;
+
+#[derive(Debug, Clone, Copy)]
+enum RegionKind {
+    Desc(usize),
+    DrvEvt,
+    DevEvt,
+    Pool,
+}
+
+#[derive(Debug, Clone, Copy)]
+struct RegionInfo {
+    kind: RegionKind,
+    size: usize,
+}
+
+#[derive(Debug)]
+pub struct LoomMem {
+    // All shared regions live in `loom::cell::UnsafeCell` so loom tracks every
+    // data access and verifies it is race-free.
+    descs: Vec<loom::cell::UnsafeCell<Descriptor>>,
+    drv: loom::cell::UnsafeCell<EventSuppression>,
+    dev: loom::cell::UnsafeCell<EventSuppression>,
+    pool: loom::cell::UnsafeCell<Vec<u8>>,
+
+    desc_flags: Vec<AtomicU16>,
+    drv_flags: AtomicU16,
+    dev_flags: AtomicU16,
+
+    regions: BTreeMap<u64, RegionInfo>,
+    layout: Layout,
+}
+
+unsafe impl Sync for LoomMem {}
+unsafe impl Send for LoomMem {}
+
+impl LoomMem {
+    pub fn new(ring_base: u64, num_descs: usize, pool_base: u64, pool_size: usize) -> Self {
+        let descs_nz = NonZeroU16::new(num_descs as u16).unwrap();
+        let layout = unsafe { Layout::from_base(ring_base, descs_nz).unwrap() };
+
+        let descs: Vec<_> = (0..num_descs)
+            .map(|_| loom::cell::UnsafeCell::new(Descriptor::zeroed()))
+            .collect();
+        let desc_flags: Vec<_> = (0..num_descs).map(|_| AtomicU16::new(0)).collect();
+
+        let mut regions = BTreeMap::new();
+
+        // Register each descriptor as a separate region
+        for i in 0..num_descs {
+            let addr = layout.desc_table_addr() + (i * Descriptor::SIZE) as u64;
+            regions.insert(
+                addr,
+                RegionInfo {
+                    kind: RegionKind::Desc(i),
+                    size: Descriptor::SIZE,
+                },
+            );
+        }
+
+        regions.insert(
+            layout.drv_evt_addr(),
+            RegionInfo {
+                kind: RegionKind::DrvEvt,
+                size: EventSuppression::SIZE,
+            },
+        );
+
+        regions.insert(
+            layout.dev_evt_addr(),
+            RegionInfo {
+                kind: RegionKind::DevEvt,
+                size: EventSuppression::SIZE,
+            },
+        );
+
+        regions.insert(
+            pool_base,
+            RegionInfo {
+                kind: RegionKind::Pool,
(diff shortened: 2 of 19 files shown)


src/hyperlight_common/src/virtq/concurrency.rs
```diff
@@ -0,0 +1,482 @@
...
+struct RegionInfo {
+    kind: RegionKind,
+    size: usize,
+}
+
+#[derive(Debug)]
+pub struct LoomMem {
+    descs: Vec<Descriptor>,
+    drv: core::cell::UnsafeCell<EventSuppression>,
+    dev: core::cell::UnsafeCell<EventSuppression>,
```
ludfjig: any point in making these `loom::cell::UnsafeCell` ?

ludfjig (dismissed):
LGTM, very clean!

src/hyperlight_common/src/virtq/producer.rs
```diff
@@ -0,0 +1,1761 @@
...
+        &self,
+        elems: impl IntoIterator<Item = BufferElement>,
+    ) -> Result<(), VirtqError> {
+        let mut first_err = None;
+        for elem in elems {
+            if let Err(err) = self.pool.dealloc(elem.addr)
+                && first_err.is_none()
+            {
+                first_err = Some(VirtqError::Alloc(err));
+            }
```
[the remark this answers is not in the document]
andreiltd: Let chains are valid rust syntax.

src/hyperlight_common/src/virtq/producer.rs
```diff
@@ -0,0 +1,1761 @@
...
+        Ok(())
+    }
+
+    /// Begin building a descriptor chain for submission.
+    ///
+    /// Returns a [`ChainBuilder`] that allocates buffers from the pool.
+    /// ```
+    pub fn chain(&self) -> ChainBuilder<M, P> {
+        ChainBuilder::new(self.inner.mem().clone(), self.pool.clone())
+    }
```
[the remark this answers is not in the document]
andreiltd: Fixed.

src/hyperlight_common/src/virtq/consumer.rs
```diff
@@ -0,0 +1,804 @@
...
+            Ok(x) => x,
+            Err(RingError::WouldBlock) => return Ok(None),
+            Err(e) => return Err(e.into()),
+        };
+
+        let readables = chain.readables();
+        let writables = chain.writables();
+        if readables.is_empty() && writables.is_empty() {
+            return Err(VirtqError::BadChain);
+        }
```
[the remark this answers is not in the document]
andreiltd: Added `abort_chain`.

src/hyperlight_common/src/virtq/pool.rs
```diff
@@ -0,0 +1,1311 @@
...
+        // advance and align upper base to N
+        aligned = aligned
+            .checked_add(lower.capacity() as u64)
+            .ok_or(AllocError::Overflow)?;
+
+        aligned = align_up(aligned as usize, U) as u64;
+        let upper = Slab::<U>::new(aligned, upper_region)?;
+
+        Ok(Self { lower, upper })
+    }
```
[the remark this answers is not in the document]
andreiltd: Fixed

src/hyperlight_common/src/virtq/concurrency.rs
```diff
@@ -0,0 +1,482 @@
...
+        let (info, offset) = self.region(addr).ok_or(MemErr)?;
+
+        match info.kind {
+            RegionKind::Pool => Ok(self
+                .pool
+                .get_mut()
+                .with(|buf| unsafe { &mut (&mut *buf)[offset..offset + len] })),
+            _ => Err(MemErr),
+        }
+    }
```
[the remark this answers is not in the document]
andreiltd: This is compiling and the shortcuts like this are fine in tests, which concurrency module is - loom model tests.

src/hyperlight_common/Cargo.toml
```diff
@@ -19,6 +19,8 @@ arbitrary = {version = "1.4.2", optional = true, features = ["derive"]}
 anyhow = { version = "1.0.102", default-features = false }
 bitflags = "2.10.0"
 bytemuck = { version = "1.24", features = ["derive"] }
+bytes = { version = "1", default-features = false }
```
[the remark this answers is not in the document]
andreiltd: Bytes has no alloc feature.

src/hyperlight_common/src/virtq/consumer.rs
```diff
@@ -0,0 +1,804 @@
...
+
+        // Copy chain payload from shared memory
+        let data = match self.read_elements(readables) {
+            Ok(d) => d,
+            Err(e) => {
+                // Read failed - clear inflight before propagating
+                self.inflight.set(id_idx, false);
+                return Err(e);
+            }
+        };
```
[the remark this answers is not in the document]
andreiltd: Added abort_chain to account for that.

src/hyperlight_common/src/virtq/concurrency.rs
```diff
@@ -0,0 +1,482 @@
...
+                // Safety: pool memory is a contiguous Vec<u8>; caller ensures
+                // no concurrent writes for the lifetime of the returned slice.
+                Ok(self
+                    .pool
+                    .get()
+                    .with(|buf| unsafe { &(&*buf)[offset..offset + len] }))
+            }
+            _ => Err(MemErr),
+        }
+    }
```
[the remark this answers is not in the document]
andreiltd: It is OK in tests.

src/hyperlight_common/src/virtq/pool.rs
```diff
@@ -0,0 +1,1311 @@
...
+///
+/// The wrapped value must only be accessed from a single thread.
+#[derive(Debug)]
+pub(super) struct SyncWrap<T>(pub(super) T);
+
+// SAFETY: The wrapped value must only be accessed from a single thread.
+unsafe impl<T> Send for SyncWrap<T> {}
+// SAFETY: The wrapped value must only be accessed from a single thread.
+unsafe impl<T> Sync for SyncWrap<T> {}
+
```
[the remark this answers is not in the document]
andreiltd: This code is instantiated on guest only, which currently is single threaded.

src/hyperlight_common/src/virtq/concurrency.rs
andreiltd: I switched this to model the descriptor and event-suppression memory with loom `UnsafeCell` too. An interesting outcome of that switch was that loom started reporting data races when reading `Descriptors` and `EventSuppressions`.
 
The cause was our single stage read api, we read the whole struct in one call: acquire-loading the flags and then reading the body. If the flags indicated the entry was ours, the body had already been read and if not, we discarded the body and kept waiting for valid flags. That's sound as long as the discarded body is never used, but loom (correctly) saw the unconditional body read racing the concurrent write to the same cell.
 
 So I split the read api for both `Descriptor` and `EventSuppression` into two stages: `read_flags_acquire` and `read_body`. It's a bit more verbose, but it keeps loom happy and makes the concurrency tests more meaningful.

Good catch!

src/hyperlight_common/src/virtq/producer.rs
```diff
@@ -0,0 +1,1761 @@
...
+    /// No outstanding [`UsedChain::Data`] buffers, borrowed segment views, or
+    /// peer accesses to previously submitted descriptors may exist. The new pool
+    /// may manage the same shared-memory addresses as the old pool, so old
+    /// zero-copy buffers must not outlive this transition.
+    pub unsafe fn reset_with_pool(&mut self, pool: P) {
+        self.pending.clear();
+        self.inflight.iter_mut().for_each(|slot| *slot = None);
+        self.inner.reset();
+        self.pool = pool;
+    }
```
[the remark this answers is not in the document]
andreiltd: Fixed.

src/hyperlight_common/src/virtq/consumer.rs
```diff
@@ -0,0 +1,804 @@
...
+            id,
+        };
+        self.next_token = self.next_token.wrapping_add(1);
+
+        if recv_len > max_recv_len {
+            return Err(VirtqError::PayloadTooLarge {
+                recv: recv_len,
+                limit: max_recv_len,
+            });
+        }
```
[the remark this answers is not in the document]
andreiltd: Same, it is now handled by `abort_chain`

ludfjig (dismissed):
ship it!

src/hyperlight_common/src/virtq/pool.rs
```diff
@@ -0,0 +1,1332 @@
...
+///
+/// # Safety
+///
+/// The `Send`/`Sync` assertions are only sound while the wrapped value - and
+/// every `Bytes` handed out from it - stays on a single thread. Hyperlight
+/// guests are single-threaded, so this holds for guest-side use. It is unsound
+/// to move a pool (or a reply `Bytes`) to another thread, e.g. by using these
+/// pools with a producer/consumer on the multi-threaded host.
+#[derive(Debug)]
+pub(super) struct SyncWrap<T>(pub(super) T);
```
simongdavies: If this is only intended to be used for the BufferPool and RecyclePool  is there any reason to have the unconstrainted T?

Could we either constraint this to these types:

```rust
unsafe impl<const L: usize, const U: usize> Send for SyncWrap<Rc<RefCell<Inner<L, U>>>> {}
unsafe impl<const L: usize, const U: usize> Sync for SyncWrap<Rc<RefCell<Inner<L, U>>>> {}
unsafe impl Send for SyncWrap<Rc<RefCell<RecycleList>>> {}
unsafe impl Sync for SyncWrap<Rc<RefCell<RecycleList>>> {}
```

Or remove it entirely?

```rust

pub struct BufferPool<const L: usize = 256, const U: usize = 4096> {
    inner: Rc<RefCell<Inner<L, U>>>,
}
pub struct RecyclePool {
    inner: Rc<RefCell<RecycleList>>,
}

// SAFETY: these pools hold their state in a non-atomic `Rc<RefCell<..>>`. The
// Send/Sync assertions are sound only for single-threaded (guest-side) use,
// where the pool and every reply `Bytes` derived from it stay on one thread.
// Hyperlight guests are single-threaded. Moving a pool (or a reply `Bytes`) to
// another thread — e.g. a producer/consumer on the multi-threaded host — is unsound.
unsafe impl<const L: usize, const U: usize> Send for BufferPool<L, U> {}
unsafe impl<const L: usize, const U: usize> Sync for BufferPool<L, U> {}
unsafe impl Send for RecyclePool {}
unsafe impl Sync for RecyclePool {}
```

src/hyperlight_common/src/virtq/desc.rs
```diff
@@ -186,6 +180,28 @@ impl Descriptor {
...
+
+    /// Acquire-load only the flags word - the packed-ring publish point -
+    /// without reading the descriptor body.
+    pub fn read_flags_acquire<M: MemOps>(mem: &M, addr: u64) -> Result<DescFlags, M::Error> {
+        let flags = mem.load_acquire(addr + Self::FLAGS_OFFSET as u64)?;
+        Ok(DescFlags::from_bits_truncate(flags))
+    }
+
+    /// Read the descriptor body (`addr`/`len`/`id`) and combine it with flags
+    /// already obtained from [`load_flags_acquire`](Self::load_flags_acquire).
```
simongdavies: this link is broken should be `read_flags_acquire`

src/hyperlight_common/src/virtq/consumer.rs
```diff
@@ -0,0 +1,857 @@
...
+    /// Submit a reply/ack for a received chain back to the ring.
+    ///
+    /// Accepts both [`WritableChain`] (with written byte count) and
+    /// [`AckChain`] (zero-length) via the [`ReplyChain`] enum.
+    /// Clears the inflight slot and notifies the producer if event
+    /// suppression allows.
+    pub fn complete(&mut self, reply: impl Into<ReplyChain<M>>) -> Result<(), VirtqError> {
+        let reply = reply.into();
+        let id = reply.token().id;
+        let written = reply.written() as u32;
```
simongdavies: should this use `u32::try_from(reply.written()).map_err(|_| VirtqError::ReplyTooLarge)`?

src/hyperlight_common/src/virtq/pool.rs
```diff
@@ -0,0 +1,1332 @@
...
+#[cfg(test)]
+impl<const N: usize> Slab<N> {
+    fn free_bytes(&self) -> usize {
+        (self.used_slots.len() - self.used_slots.count_ones(..)) * N
+    }
+}
+
+#[inline]
+fn align_up(val: usize, align: usize) -> usize {
+    assert!(align > 0);
```
simongdavies: Could we make this return a result? 
```rust
fn align_up(val: usize, align: usize) -> Result<usize, AllocError> {
    if align == 0 { return Err(AllocError::InvalidArg); }
    Ok(val.next_multiple_of(align))
}
```

or make the assert a const? 

Also there is an align_up in mod.rs (which also panics on align == 0)

simongdavies (changes requested):
A couple of minor issues, but otherwise LGTM

src/hyperlight_common/src/virtq/pool.rs
ludfjig: or use https://doc.rust-lang.org/std/primitive.u32.html#method.next_multiple_of. Result might be a bit restrictive

src/hyperlight_common/src/virtq/consumer.rs
andreiltd: Fixed!

src/hyperlight_common/src/virtq/desc.rs
andreiltd: Fixed!

src/hyperlight_common/src/virtq/pool.rs
andreiltd: That makes sense, I kept `SyncWrap` to make it a bit more explicit. I also narrowed this to Send only.

andreiltd: BTW, the whole Send/Sync dance here is required to satisfy this trait: https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner

This way we can map shared memory on a guest as `Bytes` and pass cheap reference around.

src/hyperlight_common/src/virtq/pool.rs
andreiltd: Fixed!

ludfjig (approved):

simongdavies (approved):

PR #1385, opened 2026-04-16
feat: i686 page tables, snapshot compaction, and CoW (standalone)

danbugs:
## Summary

The cleaned-up version of #1381: drops the dependencies on #1379 and #1380 so this PR can land on its own.

## Commits
- `refactor: replace nanvix-unstable with i686-guest and guest-counter features`
- `feat: i686 protected-mode boot and unified restore path`
- `feat: i686 page tables, snapshot compaction, and CoW support`
- `fixup: address PR #1381 review comments`

## What this adds

- **i686 guest support** on the x86_64 host: 32-bit protected-mode boot, unified restore path, 2-level page-table walking and snapshot compaction with CoW-resolved PTE reads.
- **Feature rename** — the prior `nanvix-unstable` gates split into `i686-guest` (PT walker, protected-mode boot, compaction) and `guest-counter` (scratch counter plumbing). No behavior change for consumers using the old feature flag; Cargo.toml / Justfile / build.rs updated to match.

## What it does *not* include (vs #1381)

- The NOBITS-section ELF-loader zero-fill (#1379).
- The PEB `file_mappings` removal (#1380).

Both can land separately; #1381 happened to stack on top of them.

## Review items from #1381 addressed here

- PTE decode uses `u32::from_le_bytes` (previously `from_ne_bytes`) — PTEs are little-endian by arch spec.
- `i686-guest` feature guarded with a `compile_error!` on targets that are neither `x86` (guest) nor `x86_64` (host).
- `restore_snapshot` rewrites the scratch PD-roots bookkeeping (`SCRATCH_TOP_PD_ROOTS_{COUNT,ARRAY}_OFFSET`) so a subsequent `snapshot()` doesn't see a stale zero count. `Snapshot` persists the root count; root GPAs are deterministic (`layout.get_pt_base_gpa() + i * PAGE_SIZE`).

## Verification

Built with `cargo check -p hyperlight-{common,host,guest} --features kvm,mshv3,executable_heap,i686-guest,guest-counter` — clean. End-to-end tested against [nanvix@e61306676](https://github.com/nanvix/nanvix/tree/hyperlight-end-to-end-dev): **Hello 5/5** with `NANVIX_REPEAT=4` through the restore+call loop.

PR #1525, opened 2026-06-10
Remove i686-guest, nanvix-unstable and guest-counter features

simongdavies:
Removes the experimental 32-bit i686 guest support along with the nanvix-unstable and guest-counter cargo features that depended on or accompanied it. This deletes the i686 arch modules, the PEB file-mapping plumbing (labels, preallocated FileMappingInfo array), the GuestCounter API, and the associated cfg gating, tests, Justfile recipes, CI matrix entries and docs.

As part of the cleanup, the ReadableSharedMemory trait machinery in mem::layout is now gated on a precise 'readable_shared_mem' cfg alias (gdb debug path or shared-snapshot mem_profile path) instead of a blanket

#[allow(unused)], so it is compiled out entirely when unused.


ludfjig (changes requested):
LGTM. 
- Can we remove SpaceAwareMapping::AnotherSpace, space_aware_map, SpaceReferenceMapping, SpaceId, built_roots, PtRootFinder/set_pt_root_finder, root_pt_gpas: &[u64], since they are not used anymore. I think we only need single-root walker and not multi-root walkers. They were introduced in https://github.com/hyperlight-dev/hyperlight/pull/1385
- Add breaking change to map_file_cow signature to the changelog

If you prefer the top bullet can be separate pr

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -161,10 +158,8 @@ pub(crate) struct SandboxMemoryManager<S: SharedMemory> {
 pub(crate) struct GuestPageTableBuffer {
     buffer: std::cell::RefCell<Vec<u8>>,
     phys_base: usize,
-    /// Absolute GPA of the currently-active root table. For
-    /// multi-root guests, `set_root` switches which root subsequent
-    /// `vmem::map` / `vmem::space_aware_map` calls target — typically
-    /// to an address previously returned by `alloc_table`.
+    /// Absolute GPA of the root table used while rebuilding snapshot
+    /// mappings.
     root: std::cell::Cell<u64>,
```
ludfjig: I think this `root` field can be deleted now

src/hyperlight_common/src/vmem.rs
ludfjig: I think we can remove this as well

ludfjig (changes requested):
LGTM with some more cleanup

src/hyperlight_host/src/mem/mgr.rs
simongdavies: This is required by the trait `TableReadOps` so it seems like this would require lots more changes, if it is truly unused then maybe we should do it as a separate PR.

ludfjig: Isn't it just a 3-line-change? root is a duplicate of phys_base

```rust
fn root_table(&self) -> u64 {
  self.phys_base as u64
}
```

simongdavies: OK let me take a look

ludfjig (approved):

ludfjig (approved):
LGTM, nit: commit 2f5c85264e606f9f52b59bceed1dce10f8c60aa5 commit message has a weird trailer in the commit message but should be fine if you squash

PR #1527, opened 2026-06-11
Updates rust version to 1.94

simongdavies:
Updates rust version to 1.94 , MSRV remains at 1.89.

Latest rust version is 1.96 but this requires a change to cargo-hyperlight. I will open a PR for that shortly along with a PR to update the common-workflow and dev-images

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -48,7 +48,7 @@ build target=default-target:
 guests: build-and-move-rust-guests build-and-move-c-guests
 
 ensure-cargo-hyperlight:
-    cargo install --locked cargo-hyperlight
+    cargo install --locked --version 0.1.11 cargo-hyperlight
 
 witguest-wit:
     cargo install --locked wasm-tools
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -49,7 +49,7 @@ RUN SNIPPET="export PROMPT_COMMAND='history -a' && export HISTFILE=/commandhisto
 
 USER $USER
 
-ARG RUST_TOOLCHAIN=1.89
+ARG RUST_TOOLCHAIN=1.94
 
 # Install rust
 RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y \
--- a/.github/workflows/CargoAudit.yml
+++ b/.github/workflows/CargoAudit.yml
@@ -19,7 +19,7 @@ jobs:
       # TODO: Once the runner image is updated to include the necessary tools (without downloading), we can switch to the common workflow.
       - uses: actions-rust-lang/setup-rust-toolchain@46268bd060767258de96ed93c1251119784f2ab6  # v1.16.1
         with:
-          toolchain: "1.89"
+          toolchain: "1.94"
 
       - uses: rustsec/audit-check@69366f33c96575abad1ee0dba8212993eecbe998  # v2.0.0
         with:
--- a/.github/workflows/CargoPublish.yml
+++ b/.github/workflows/CargoPublish.yml
@@ -35,7 +35,7 @@ jobs:
 
       - uses: hyperlight-dev/ci-setup-workflow@f6bd9cc86d0737976d2128c8b8ced8edc017cbb4  # v1.9.0
         with:
-          rust-toolchain: "1.89"
+          rust-toolchain: "1.94"
 
       - name: Check crate versions
         shell: bash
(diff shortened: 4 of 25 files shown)


Justfile
```diff
@@ -48,7 +48,10 @@ build target=default-target:
 guests: build-and-move-rust-guests build-and-move-c-guests
 
 ensure-cargo-hyperlight:
-    cargo install --locked cargo-hyperlight
+    # Force reinstall so self-hosted runners don't keep a preinstalled binary with the same semver.
+    cargo install --locked --version 0.1.11 --force cargo-hyperlight
+    cargo hyperlight --version
+    {{ if os() == "windows" { "Get-Command cargo-hyperlight" } else { "command -v cargo-hyperlight" } }}
```
ludfjig: This is ran as part of `just guest`, which will now reinstall cargo-hyperlight evertime which seems wrong

simongdavies: I dont know what went wrong with the runners but this was the fix, I'll look at it some more

simongdavies: fixed

simongdavies:
# Test `reset_vcpu_simple` failing after upgrade.

This PR uncovered a failure in the `reset_vcpu_simple` test:

image

The fix to this is to update the test only compare 2432 bytes as detailed below.

## Test Fix Justification: `reset_vcpu_simple` XSAVE assertion

### What's failing

The test `hypervisor::hyperlight_vm::x86_64::tests::reset_vcpu_simple` fails on MSHV3 when compiled with Rust ≥1.92. It asserts that all 4096 bytes of the XSAVE buffer returned by `get_xsave()` should be zero (except known fields). Bytes past offset 2432 contain non-zero values.

### Root cause (proved on Azure Linux 3 + MSHV3)

**`mshv-ioctls` 0.6.7 returns uninitialised heap memory in the XSAVE buffer.**

The call chain:
1. `mshv-ioctls::get_xsave()` allocates a 4096-byte buffer via `Buffer::new()`, which calls `std::alloc::alloc` — **uninitialised memory**
2. The MSHV kernel ioctl fills **exactly 2432 bytes** (confirmed: `vp_state.buf_sz == 2432` after ioctl returns)
3. `XSave::try_from(buffer)` copies all 4096 bytes into the returned struct — including 1664 bytes of heap garbage

Bytes 2432-4095 are whatever happened to be at that heap address. With Rust 1.91 the allocator happened to give a fresh (zeroed) page. With Rust 1.92+, it gives a recycled page with non-zero residue.

### Evidence

| Test | Result | Proves |
|------|--------|--------|
| Same code, Rust 1.91 | ✅ PASS | Allocator gives zeroed page by luck |
| Same code, Rust 1.92 | ❌ FAIL | Allocator gives dirty page |
| Patched `Buffer::new` to fill `0xAA` before ioctl, Rust 1.91 | ❌ FAIL | Proves kernel doesn't zero the tail regardless of compiler — test only passed by luck |
| `vp_state.buf_sz` after ioctl | Always **2432** | Kernel reports it only wrote 2432 bytes |

### Why KVM passes

KVM's `get_xsave` uses `kvm_xsave::default()` (stack-allocated, zeroed struct). The Linux `KVM_GET_XSAVE` ioctl uses standard (non-compacted) XSAVE format and the kernel zeroes the output buffer. No uninitialised memory is exposed.

### Why WHP passes

WHP's `WHvGetVirtualProcessorXsaveState` returns a `written_bytes` count. Hyperlight allocates the buffer with `vec![0u8; size]` (pre-zeroed) and the API only writes `written_bytes` into it. Tail stays zero.

### Why the test is wrong regardless of the mshv bug

The test asserts zeros at bytes that have no defined content. The Intel XSAVE compacted format (SDM Vol 1 §13.4.3) only defines the layout of bytes within declared components. Bytes past the last component (offset 2432 on this hardware) are not part of any XSAVE specification — their contents should not be asserted by the test.

### Test Fix

Only compare bytes within valid XSAVE regions:
- Legacy region (0-511)
- XSAVE header (512-575)
- Extended components (576-2432, based on CPUID-reported component sizes in compacted layout)

Ignore bytes past the end of the last declared component.

### Upstream fix 

Change `Buffer::new` in `mshv-bindings/src/x86_64/regs.rs` from `std::alloc::alloc` to `std::alloc::alloc_zeroed`. This eliminates the uninitialised memory exposure for all callers (`get_xsave`, `get_lapic`, and any future users).

andreiltd (approved):
LGTM! 🚀

jsturtevant (approved):

Issue #1408, opened 2026-04-23
Make elf loading respect phdr VAs


syntactically:
Currently, we load the elf binary in an odd way that does not really respect the VAs on the segments in the file, because, in the past, the virtual memory system depended on identity-mapping for code segments, and we were trying not to use too much VA space. The logic assumes the binary is PIE, although it currently works somewhat by happenstance for at least one non-PIE binary (https://github.com/hyperlight-dev/hyperlight/pull/1385#discussion_r3132233017).

Now that we have much less dependence on identity-mapped code segments, we should change this to a more normal loader that actually creates the VAs in the header (basing PIE executables at some reasonable base address).

PR #1530, opened 2026-06-14
Make ELF loading respect program header virtual addresses for non-PIE binaries

cshung:
## Summary

For non-PIE (ET_EXEC) ELF binaries, the guest page table now maps the code region at the ELF's declared virtual address rather than identity-mapping it at the GPA. This allows statically-linked binaries with a fixed load address (e.g., `--image-base=0x200000`) to execute correctly.

## Problem

Previously, Hyperlight assumed code GVA == code GPA (identity mapping). Non-PIE binaries that declare a non-zero base virtual address (via program header `p_vaddr`) would triple-fault because the guest CPU jumped to the ELF's declared entrypoint VA, which wasn't mapped in the page tables.

## Solution

- Compute `code_virt_base` from the ELF's lowest LOAD segment `p_vaddr`
- For non-PIE (`base_va > 0`): map code at the declared VA in the guest page tables
- For PIE (`base_va == 0`): preserve existing identity mapping behavior (with assertion to guard the invariant)
- Compute entrypoint as `code_virt_base + (entrypoint_va - base_va)`

The fix leverages the existing `Mapping` struct's support for `phys_base != virt_base` — no changes to the page table code itself.

## Testing

- New `non_pie_guest_hello_world` integration test exercises full guest lifecycle (init, COW, function call, return value) with a non-PIE simpleguest built at `--image-base=0x200000`
- All existing PIE guest tests continue to pass (identity mapping preserved)
- Tested on Windows/WHP and Linux/KVM

## Build infrastructure

- Added `build-rust-guests-non-pie` Justfile targets
- Non-PIE build runs first in `guests` recipe to avoid clobbering normal guest binaries
- Added `simple_guest_non_pie_as_string()` test helper

Contributes to: #1408


cshung:
Addressed Copilot review feedback:

1. **Justfile (RUSTFLAGS on Linux)** — Fixed. \RUSTFLAGS\ now prefixes \cargo\ directly after \cd &&\ so the env var applies to the build command.

2. **snapshot/mod.rs (virt_base mapping)** — No change needed. The code region is a single contiguous block where \
gn.guest_region.start == load_addr\ — there is no offset to lose. The \ssert_eq!\ is deliberate per codebase convention.

3. **snapshot/mod.rs (entrypoint underflow)** — Fixed. Now uses \checked_sub\ with a proper error for malformed ELFs.

4. **lib.rs (doc says 'elf binary')** — No change needed. Hyperlight guests are always ELF regardless of host OS.

Justfile
```diff
@@ -69,6 +69,16 @@ build-rust-guests target=default-target features="": (witguest-wit) (ensure-carg
 build-and-move-rust-guests: (build-rust-guests "debug") (move-rust-guests "debug") (build-rust-guests "release") (move-rust-guests "release")
 build-and-move-c-guests: (build-c-guests "debug") (move-c-guests "debug") (build-c-guests "release") (move-c-guests "release")
 
+# Build non-PIE variants of rust guests for testing ELF VA mapping
+build-rust-guests-non-pie target=default-target: (ensure-cargo-hyperlight)
+    {{ if os() == "windows" { "$env:RUSTFLAGS='-C relocation-model=static -C link-args=--no-pie -C link-args=--image-base=0x200000';" } else { "RUSTFLAGS='-C relocation-model=static -C link-args=--no-pie -C link-args=--image-base=0x200000'" } }} cd src/tests/rust_guests/simpleguest && cargo hyperlight build --profile={{ if target == "debug" { "dev" } else { target } }}
```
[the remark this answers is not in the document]
cshung: Fixed — RUSTFLAGS now prefixes cargo directly after cd on Linux.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -340,9 +350,26 @@ impl Snapshot {
...
+                    assert_eq!(
+                        code_virt_base,
+                        rgn.guest_region.start as u64,
+                        "PIE code region should be identity-mapped"
+                    );
+                }
+                code_virt_base
+            } else {
+                rgn.guest_region.start as u64
+            };
```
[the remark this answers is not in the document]
cshung: The code region is a single contiguous block where rgn.guest_region.start == load_addr — there is no offset to lose. The assert_eq! is deliberate per codebase convention (566 uses of assert vs 8 of debug_assert).

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -367,7 +394,7 @@ impl Snapshot {
             load_info,
             stack_top_gva: exn_stack_top_gva,
             sregs: None,
-            entrypoint: NextAction::Initialise(load_addr + entrypoint_va - base_va),
+            entrypoint: NextAction::Initialise(code_virt_base + (entrypoint_va - base_va)),
```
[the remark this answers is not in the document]
cshung: Fixed — now uses checked_sub with a proper error message for malformed ELFs.

src/hyperlight_testing/src/lib.rs
```diff
@@ -88,6 +88,37 @@ pub fn dummy_guest_as_string() -> Result<String> {
         .ok_or_else(|| anyhow!("couldn't convert dummy guest PathBuf to string"))
 }
 
+/// Get a fully qualified OS-specific path to the non-PIE simpleguest elf binary
+pub fn simple_guest_non_pie_as_string() -> Result<String> {
```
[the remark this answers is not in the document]
cshung: Hyperlight guests are always ELF regardless of host OS — the VMM loads ELF directly into the micro-VM.

ludfjig (commented):
LGTM. @syntactically could you have a look too

Justfile
```diff
@@ -69,6 +69,22 @@ build-rust-guests target=default-target features="": (witguest-wit) (ensure-carg
...
...-hyperlight-none/target.json " + root + "/src/tests/rust_guests/target-non-pie/x86_64-hyperlight-none.json -Force; $env:RUST_TARGET_PATH='" + root + "/src/tests/rust_guests/target-non-pie'; $env:RUSTFLAGS='--sysroot " + root + "/src/tests/rust_guests/target/sysroot --cfg=hyperlight --check-cfg=cfg(hyperlight) -C link-args=-eentrypoint -C relocation-model=static -C link-args=--no-pie -C link-args=--image-base=0x200000'; $env:RUSTC_BOOTSTRAP='1';" } else { "mkdir -p " + root + "/src/tests/rust_guests/target-non-pie && cp " + root + "/src/tests/rust_guests/target/sysroot/lib/rustlib/x86_64-hyperlight-none/target.json " + root + "/src/tests/rust_guests/target-non-pie/x86_64-hyperlight-none.json && RUST_TARGET_PATH='" + root + "/src/tests/rust_guests/target-non-pie' RUSTFLAGS='--sysroot " + root + "/src/tests/rust_guests/target/sysroot --cfg=hyperlight --check-cfg=cfg(hyperlight) -C link-args=-eentrypoint -C relocation-model=static -C link-args=--no-pie -C link-args=--image-base=0x200000' RUSTC_BOOTSTRAP=1" } }} cd src/tests/rust_guests/simpleguest && cargo build --target x86_64-hyperlight-none --target-dir ../target-non-pie --profile={{ if target == "debug" { "dev" } else { target } }}
```
syntactically: On first glance, linking the standard library built with different codegen opts seemed a little sketchy, but I guess it should probably be OK here, at least as long as we don't have something like a GOT-generating relocation show up for some reason?

Do you need the `-C link-args=--no-pie`? My read of the docs / quick test on aarch64 made me think that that should be more-or-less implied by the `-C relocation-model=static` on the rust command that invokes the linker.

More seriously, I would prefer to avoid copying so much of the RUSTFLAGS logic from cargo-hyperlight. Is the sysroot rebuild a serious problem for CI performance? If so, maybe there is a way to address it in cargo-hyperlight by e.g. adding a flag to prevent the sysroot being rebuilt (or stripping some of these things like `link-args` when building the sysroot).

Also, if you do pursue the approach here, why copy the target json to a new path when the path has to be passed in via an environment variable anyway?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -340,9 +348,25 @@ impl Snapshot {
                     executable,
                 })
             };
+
+            // For the code region, use code_virt_base as the GVA.
+            // For non-PIE this is the ELF's declared base VA (non-identity mapping).
+            // For PIE this should equal the GPA (identity mapping).
+            let virt_base = if rgn.region_type == MemoryRegionType::Code {
```
syntactically: Probably we should have some check that the executable mapping here does not conflict with any other mappings.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -308,6 +310,12 @@ impl Snapshot {
         let base_va = exe_info.base_va();
         let entrypoint_va: u64 = exe_info.entrypoint().into();
 
+        // Determine the virtual base address for the code region.
+        // For non-PIE binaries (base_va > 0), the code should appear at the
+        // ELF's declared virtual address. For PIE binaries (base_va == 0),
+        // we use the physical load address (identity mapping).
+        let code_virt_base = if base_va > 0 { base_va } else { load_addr };
```
syntactically: Should this choice be made based on whether the executable has `.et_type = ET_DYN` vs `ET_EXEC`, rather than the first phdr VA?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -340,9 +348,25 @@ impl Snapshot {
...
+                }
+                code_virt_base
+            } else {
+                rgn.guest_region.start as u64
+            };
+
             let mapping = Mapping {
                 phys_base: rgn.guest_region.start as u64,
-                virt_base: rgn.guest_region.start as u64,
+                virt_base,
```
syntactically: This approach probably works for now, especially with executables where all the segments are contiguous. One other thing that we could do while we are already touching this code is switch to mapping the ELF a bit more "properly", by copying every section copied by a segment into the snapshot contiguously (similarly to how it is being done now, but without using `phdr.p_vaddr`, so that there is no risk of large gaps appearing for no reason) and then making one of these `vmem::map` calls for each segment (~= `PT_LOAD` program header). In addition to being more efficient if we encounter a file with gaps between segments (at least a few kilobytes of gaps often exist due to page-aligning the phdr VAs), this would allow us to have better default permissions in the guest page tables by using the permission information in `phdr.p_flags`. Probably the way to do that would be to remove this mapping of this region entirely and change the exe interface in `exe.rs`/implementation in `elf.rs` to replace `load`/`load_at` with functions that take a mapping operations and do the mapping as they walk over the phdrs. That would be my slight preference/how I would do it, but I will leave it up to you whether or not to pursue it, since it is a nontrivially-larger change.

(Unrelated, not-really-review note: the need to fix up the physical vs virtual discontinuity is continuing a trend of several things that makes me slightly question the actual utility of `SandboxMemoryLayout::get_memory_regions`.   It might be worth thinking about whether it makes sense to get rid of that / some more bits of `SandboxMemoryLayout` at some point, but that's definitely not a discussion for this PR).

syntactically (commented):
Thanks for doing this! It looks like it is moving in a good direction. I have a couple of minor suggestions/comments, as well as a bit of an alternative design that you could take (but totally up to you on that one!)

Justfile
cshung: Thanks for the feedback. I've switched to using cargo hyperlight build directly with just the non-PIE link flags and a separate --target-dir. This avoids duplicating any cargo-hyperlight internals.

Regarding --no-pie - you're probably right that relocation-model=static implies it, but I kept it explicit for clarity. Happy to remove if you prefer.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
cshung: Good call - added a conflict check that verifies the non-PIE code VA range doesn't overlap with any other memory region. This actually caught a real bug: the original --image-base=0x200000 produced a binary that extended into the PEB region. Bumped the base to 0x1000000 (16MB) to avoid this.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
cshung: Agreed - now using elf.header.e_type == ET_DYN via an is_pie() method on ExeInfo/ElfInfo instead of checking base_va > 0. Much cleaner semantically.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
cshung: Agreed this would be the more proper approach. I'll leave it for a follow-up since it's a larger change, but it's on the radar.

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -1872,3 +1872,17 @@ fn hw_timer_interrupts() {
         );
     });
 }
+
+#[test]
+#[serial]
```
ludfjig: does this need the `#[serial]` attribute?

ludfjig (commented):
LGTM but will once again defer to @syntactically for final review

src/hyperlight_host/tests/integration_test.rs
cshung: Fixed

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -340,9 +373,25 @@ impl Snapshot {
                     executable,
                 })
             };
+
+            // For the code region, use code_virt_base as the GVA.
+            // For non-PIE this is the ELF's declared base VA (non-identity mapping).
+            // For PIE this should equal the GPA (identity mapping).
+            let virt_base = if rgn.region_type == MemoryRegionType::Code {
```
syntactically: One other place this logic (and the above non-overlapping check) could go is in `hyperlight_host/src/mem/layout.rs` where the region list is being built. I think it's maybe a bit better for the existing separation of concerns to have the snapshot code dumbly process the list that that function creates, and have all of the logic around guest layout there?

syntactically (dismissed):
This looks good to me, thank you and sorry for the delay reviewing!

I do think that there is one minor move that would make the code a little bit cleaner / better fitting with existing conventions, which I've commented on inline. Feel free to push back with a reason why this code belongs here, though!

cshung:
All review comments have been addressed. The suggestion to move conflict-check logic into \layout.rs\ (from the Jul 15 review) is implemented in the follow-up PR #1655 via the \code_virt_base()\ method which handles both VA selection and overlap validation.

@syntactically — ready for re-review when you get a chance!

syntactically (dismissed):
If you've already done that, I think it would be slightly nice to get it cherry-picked here without the ASLR changes, since I'm not totally sure we are decided on doing ASLR---it has pretty limited benefits for the modal hyperlight use case because one can't (generally) re-slide an image after a snapshot is taken, and in practice almost all images of a given binary are expected to descend from one snapshot.

As I said before I don't feel incredibly strongly though---I will leave the decision up to you.

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -874,6 +874,7 @@ impl Snapshot {
             sregs: Some(cfg.sregs),
             next_action,
             original_entrypoint: cfg.original_entrypoint_addr,
+            code_virt_base: 0, // TODO: persist code_virt_base in snapshot file format
```
jsturtevant: do we have a tracking item for this?  I would assume we would want to get this into the snapshot format before 1.0

jsturtevant (dismissed):

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -555,6 +555,64 @@ impl SandboxMemoryLayout {
...
+        elf_base_va: u64,
+        loaded_size: u64,
+    ) -> Result<(u64, Vec<MemoryRegion_<GuestMemoryRegion>>)> {
+        let load_addr = self.get_guest_code_address() as u64;
+        let code_virt_base = if is_pie { load_addr } else { elf_base_va };
+
+        let mut regions = self.get_memory_regions_::<GuestMemoryRegion>(())?;
+
+        if !is_pie {
+            let code_virt_end = code_virt_base + loaded_size;
```
jsturtevant: This will wrap and panic if the elf file is mal formed

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
jsturtevant: nice, thanks for adding a test too

syntactically:
Thanks for all the great work here & putting up with so many rounds of iteration @cshung!

The addition of `guest_virt_addr` made me a bit uncomfortable, because I knew that we were using memory regions that had guest virtual addresses in them already in other places.  I took a little bit more of a look at it, and I realised that this PR has exposed an existing semantic inconsistency in the way that we use `MemoryRegion`s.

As you probably noticed, we have several different variants of `MemoryRegion`, defined as a base `MemoryRegion_<T>` struct that is parameterised by a marker type.  There are IIRC three of those marker types that we are using right now, and it turns out that all of them have a different idea of what "host" and "guest" mean:
- `HostGuestMemoryRegion` maps host virtual addresses to guest physical addresses
- `GuestMemoryRegion`, which presently is not a mapping at all---just legitimately a region chunk of a single address space
- `CrashDumpMemoryRegion`, which maps host virtual addresses to guest virtual addresses.

Adding another field for GVA to the base struct is not ideal, then, because it's semantically confusing for the other variants (one of which even already contains GVAs!).  I think what this PR really needs is for `GuestMemoryRegion` to be a mapping, the same as the other memory regions are, which is a nice increase in consistency.  We would then switch to using the two sides of the mapping as the guest phys/virt addresses.

The obvious way to fix this is just by changing `<GuestMemoryRegion as MemoryRegionKind>::HostBaseType` to `u64` and using it as the guest physical address. The big downside is of course that the naming becomes even more misleading: whilst "host_region" would be "towards the host side" (on a semantic line of 'host physical address <- host virtual address <- guest physical address <- guest virtual address'), it's not actually located on the host.

If you wanted to make another change to rename that type and fields to something more sensible (either as another PR using stacked PRs to keep this on top, or as an early commit in this PR _which we must keep without squashing when merging, since it's a noisy, large, and semantically distinct change_), that would be great, but is by no means required/expected.  If you did do that, taking the opportunity to (again, as distinct commits that _we must not squash on merge_) to also remove one of the redundant size fields in the structure (so changing from host range/guest range to host base/guest base/size) would be great.

A couple more minor nits:
- I believe that there are no longer any users of `SandboxMemoryLayout::get_memory_regions_` using a region kind other than `GuestMemoryRegion`. So, we could rename it to `get_memory_regions` and concretise its type. Since it would then always be returning things with the correct host-side address type, we could avoid adding the new `_with_code_va` wrapper method, and just make the changes directly to `get_memory_regions`
- Why is `code_virt_base` pushed through snapshots / into the mgr? I see a comment about it being needed for gdb and tracing to resolve the offsets to the binary, but I don't see that happening
- If other regions intersect the code VA, should they be pushed around to make room, instead of signalling an error? I don't recall off the top of my head whether there is anything making assumptions about the layout of those regions, but I would rather hope not.  There aren't many of them, and I hope they will all go away quite soon, so it does not matter too much, however.

cshung:
Thanks for the thorough review @syntactically! The two refactor commits address the main feedback:

- **GuestMemoryRegion is now a proper GPA-to-GVA mapping** — \guest_virt_addr\ is removed. \host_region\ carries the GPA, \guest_region\ carries the GVA. For identity-mapped regions both are the same.
- **\get_memory_regions\ concretised** — renamed from \get_memory_regions_\, generic removed since all callers use \GuestMemoryRegion\.
- **\checked_add\** added for the overlap check (@jsturtevant's nit).

Responding to the remaining questions:

> Why is \code_virt_base\ pushed through snapshots / into the mgr?

It is used by the GDB debug path (\GetCodeSectionOffset\ in \x86_64.rs\) to report where the code section lives in guest virtual space. Without it, GDB cannot resolve symbols after snapshot restore. It is also needed for \
ead_guest_memory_by_gva\ (tracing feature) to translate GVAs back to physical offsets.

> If other regions intersect the code VA, should they be pushed around to make room, instead of signalling an error?

For non-PIE, the code VA is fixed by the ELF linker, so the user controls it. If it conflicts, that is a build-time problem the user should fix (e.g., pick a different \--image-base\). Silently relocating other regions would make behavior hard to reason about, and those regions have fixed GPAs that the PEB and guest runtime depend on. I think erroring is the right call here.

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -470,11 +470,8 @@ impl SandboxMemoryLayout {
 
     /// Returns the memory regions associated with this memory layout,
     /// suitable for passing to a hypervisor for mapping into memory
-    pub(crate) fn get_memory_regions_<K: MemoryRegionKind>(
-        &self,
-        host_base: K::HostBaseType,
-    ) -> Result<Vec<MemoryRegion_<K>>> {
-        let mut builder = MemoryRegionVecBuilder::new(Self::BASE_ADDRESS, host_base);
+    pub(crate) fn get_memory_regions(&self) -> Result<Vec<MemoryRegion_<GuestMemoryRegion>>> {
```
syntactically, line 430: Is there ever a good reason to call this instead of `get_guest_regions_with_code_va` below? I see that there is a caller in hyperlight_vm/x86_64.rs, but it looks like that is in a test where the distinction does not matter.

If not, can we either inline the adjustment from get_guest_regions_with_code_va, or make this function private so that nobody calls it by accident?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -470,11 +470,8 @@ impl SandboxMemoryLayout {
...
+        let mut builder = MemoryRegionVecBuilder::new(Self::BASE_ADDRESS, Self::BASE_ADDRESS);
 
         // code
         let peb_offset = builder.push_page_aligned(
```
syntactically, line 434: There's no reason for this to exist in the non-PIE binary case?

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -280,20 +280,26 @@ impl MemoryRegionKind for HostGuestMemoryRegion {
...
-    fn add(_base: Self::HostBaseType, _size: usize) -> Self::HostBaseType {}
+    fn add(base: Self::HostBaseType, size: usize) -> Self::HostBaseType {
+        base + size
+    }
 }
 
 /// represents a single memory region inside the guest. All memory within a region has
 /// the same memory permissions
 #[derive(Debug, Clone, PartialEq, Eq, Hash)]
 pub struct MemoryRegion_<K: MemoryRegionKind> {
```
syntactically, line 281: I think it would be nice to be more careful about naming+documentation here, but that can wait if you would prefer.

src/hyperlight_host/src/mem/memory_region.rs
```diff
@@ -365,8 +371,9 @@ impl<K: MemoryRegionKind> MemoryRegionVecBuilder<K> {
         // we know this is safe because we check if the regions are empty above
         let last_region = self.regions.last().unwrap();
         let host_end = <K as MemoryRegionKind>::add(last_region.host_region.end, size);
+        let guest_start = last_region.guest_region.end;
```
syntactically: Does this actually change any behaviour, or is it just a functionally identical incidental formatting cleanup?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -337,6 +342,14 @@ impl Snapshot {
         let load_addr = layout.get_guest_code_address() as u64;
```
syntactically: Is layout.get_guest_code_address() meaningful at all anymore? I guess it still points to the GPA for the code region---is that actually used anywhere?

syntactically: (I am imagining an approach here where you, early on, do `layout.set_code_va(Some(the appropriate thing))` and then `get_guest_code_address` and `get_memory_regions` both are made to incidentally do the correct thing so that further on code does not have to fix up the layout with an extra value that is threaded around next to it).

syntactically (commented):
> It is used by the GDB debug path (\GetCodeSectionOffset\ in \x86_64.rs) to report where the code section lives in guest virtual space. Without it, GDB cannot resolve symbols after snapshot restore. It is also needed for
ead_guest_memory_by_gva\ (tracing feature) to translate GVAs back to physical offsets.

Ah, of course. I think the `SandboxMemoryLayout` is already accessible in the relevant places and a lot of the information about the layout is already centralised there; could we just move this into that?

> Silently relocating other regions would make behavior hard to reason about, and those regions have fixed GPAs that the PEB and guest runtime depend on. I think erroring is the right call here

Well, in the non-PIE mode, those gpas are not independent of the executable, because the code segment is near the bottom of the initial setup snapshot region. So, it's kind of being consistent to move them. However, I don't feel strongly.

src/hyperlight_host/src/sandbox/snapshot/file/mod.rs
```diff
@@ -610,7 +610,7 @@ impl Snapshot {
             stack_top_gva: self.stack_top_gva,
             entrypoint_addr,
             original_entrypoint_addr: self.original_entrypoint,
-            code_virt_base: self.code_virt_base,
+            code_virt_base: self.layout.get_guest_code_gva() as u64,
```
syntactically: This should probably go in the `MemoryLayout` structure below, which is more-or-less the serialisation proxy for `SandboxMemoryLayout`

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -339,17 +334,14 @@ impl Snapshot {
             guest_blob_mem_flags,
         )?;
 
-        let load_addr = layout.get_guest_code_address() as u64;
+        let load_addr = layout.get_guest_code_gpa() as u64;
```
syntactically, line 324: This load_addr is used in relocation processing, so it seems wrong for it to be the gpa? I think we only support `R_<cls>_RELATIVE` right now, which is probably not used in the non-PIC binaries that are relevant, but we should probably still do the correct math if we did encounter one.

Actually, it looks like there is a latent bug here even in the PIC code (since a PIC executable doesn't actually have to have a link address of 0): I think probably `load_at` needs to be adjusted to properly take into account the offset between `load_addr` and `get_base_va`, since for e.g. `R_AARCH64_RELATIVE`, the ABI notes that it "[...] represents a relative adjustment to the place based on _the load address of the object relative to its original link address_" (emphasis added).

So, probably `load_at` should take the gva, but also change its relocation process to use a (signed) `load_addr - get_base_va()` delta as the adjustment to the addend.

dblnz:
Have you by any chance tried debugging a guest using gdb/lldb to ensure PIE binaries behave correctly when debugged? Also, the same question applies to crashdump.

cshung:
Yes. The branch has end-to-end tests using the PIE `simpleguest` binary.

* `test_gdb_end_to_end` verifies source breakpoint and symbol resolution.
* `test_gdb_from_snapshot` verifies the same after snapshot restore.
* `test_crashdump_gdb_symbols` verifies that GDB resolves the saved PC using the core file's `AT_ENTRY` load bias.
* `test_crashdump_gdb_symbols_from_snapshot` verifies that path after snapshot restore.

The latest x64 KVM run executed and passed all four tests. The GDB tests also passed on x64 Hyper-V. These were real GDB sessions, not compile-only checks or skipped tests: https://github.com/hyperlight-dev/hyperlight/actions/runs/35659787035

I have not separately exercised LLDB. The GDB and crashdump tests cover the PIE load-address behavior this PR changes.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -377,7 +383,15 @@ impl Snapshot {
...
-        let entrypoint_gva = load_addr + entrypoint_va - base_va;
+        let entrypoint_offset = entrypoint_va.checked_sub(base_va).ok_or_else(|| {
+            crate::new_error!(
+                "ELF entrypoint VA ({:#x}) is below base VA ({:#x})",
+                entrypoint_va,
+                base_va
+            )
+        })?;
+
+        let entrypoint_gva = layout.get_guest_code_gva() as u64 + entrypoint_offset;
```
simongdavies: Should we validate that entrypoint_va lies within an executable PT_LOAD segment before calculating the relocated entrypoint? Checking only entrypoint_va >= base_va permits entrypoints in segment gaps or non-executable data.

The final addition should also use checked_add:

```Rust
let entrypoint_gva = layout
    .get_guest_code_gva()
    .checked_add(entrypoint_offset)
    .ok_or_else(|| new_error!("ELF entrypoint GVA overflows"))?;
```

Ideally, ElfInfo::new validates e_entry against a PT_LOAD range allowing this calculation to rely on that invariant.

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -515,10 +525,10 @@ impl OciSnapshotConfig {
             })?;
         if self.original_entrypoint_addr < code_lo || self.original_entrypoint_addr >= snapshot_hi {
             return Err(crate::new_error!(
-                "snapshot original entrypoint addr {:#x} is outside the snapshot region [{:#x}, {:#x})",
+                "snapshot original entrypoint addr {:#x} is outside the code region [{:#x}, {:#x})",
```
simongdavies: The error now describes [code_lo, code_hi) as the valid code range, but the condition above still accepts values up to snapshot_hi. For non-PIE guests, code_lo + snapshot_size is not a contiguous mapped snapshot range because only code moves to code_virt_base. Should the condition use code_hi too?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
cshung: Addressed in 55680995. `ElfInfo::new` now requires `e_entry` to be inside an executable `PT_LOAD` segment, with tests for segment gaps and non-executable segments. The relocated GVA also uses `checked_add`.

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
cshung: Addressed in 55680995. `original_entrypoint_addr` is now validated against `code_hi`, with a focused test for an address at the end of the code range.

simongdavies (approved):

PR #1533, opened 2026-06-15
Update surrogate process to nostd, avoid overhead of loading unecessary libraries

simongdavies:
Creating a surrogate process on windows unnecessarily loads the Rust std library (which it doesn't need since it starts suspended and sleeps forever), the overhead of this on creating a process appears to be  large on AARCH64 (200ms) , this PR makes the surrogate `no_std` which removes this overhead.

### DIFF
--- a/src/hyperlight_host/src/hyperlight_surrogate/src/main.rs
+++ b/src/hyperlight_host/src/hyperlight_surrogate/src/main.rs
@@ -14,6 +14,40 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-fn main() {
-    std::thread::sleep(std::time::Duration::MAX)
+// This binary is used solely as an address-space host for WHvMapGpaRange2.
+// It is created with CREATE_SUSPENDED and never executes user code.
+//
+// We use #![no_std] to avoid linking dbghelp.dll and VCRUNTIME140.dll,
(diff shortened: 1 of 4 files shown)


src/hyperlight_host/src/hyperlight_surrogate/src/main.rs
```diff
@@ -14,6 +14,40 @@ See the License for the specific language governing permissions and
...
+// which add ~200ms to CreateProcess on ARM64 Windows. The process never
+// runs, so the Rust stdlib is pure dead weight here.
+
+#![no_std]
+#![no_main]
+
+#[link(name = "kernel32")]
+unsafe extern "system" {
+    fn Sleep(dwMilliseconds: u32);
+}
```
ludfjig, line 30: Why do we need this if process is created with CREATE_SUSPEND? Defense in depth?

ludfjig (approved):

src/hyperlight_host/src/hyperlight_surrogate/src/main.rs
simongdavies: I think you are correct, we don't really need it, but I guess first its protection against the process not getting started suspended , or somehow being unsuspended and second it sort of self documents what is going on

PR #1561, opened 2026-06-18
Fix just-clippyw recipe

### LINKED ISSUE #1559: `just clippyw` recipe broken
`just clippyw` is useful for checking clippy targeting windows from a linux machine. https://github.com/hyperlight-dev/hyperlight/pull/1533 broke it due adding MSVC specific args to surrogate build.rs. We should be able to fix it by checking for CARGO_CFG_TARGET_ENV=="gnu" and using the gnu specific flags

ludfjig:
Fixes #1559

I use this all the time on linux machine to verify `just clippy` would pass on windows target

### DIFF
--- a/src/hyperlight_host/src/hyperlight_surrogate/build.rs
+++ b/src/hyperlight_host/src/hyperlight_surrogate/build.rs
@@ -15,8 +15,20 @@ limitations under the License.
 */
 
 fn main() {
+    if std::env::var("CARGO_CFG_TARGET_OS").as_deref() != Ok("windows") {
+        panic!("hyperlight_surrogate can only be built for Windows targets");
+    }
+
+    let target_env = std::env::var("CARGO_CFG_TARGET_ENV")
+        .expect("CARGO_CFG_TARGET_ENV should be set for Windows targets");
     // The surrogate is a #![no_std] binary with a custom entry point.
-    // We need to tell the MSVC linker the entry point and subsystem.
-    println!("cargo:rustc-link-arg=/ENTRY:mainCRTStartup");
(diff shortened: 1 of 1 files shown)


jsturtevant (approved):
Seems reasonable to me

andreiltd (approved):

src/hyperlight_host/src/hyperlight_surrogate/build.rs
```diff
@@ -15,8 +15,20 @@ limitations under the License.
...
+    }
+
+    let target_env = std::env::var("CARGO_CFG_TARGET_ENV")
+        .expect("CARGO_CFG_TARGET_ENV should be set for Windows targets");
     // The surrogate is a #![no_std] binary with a custom entry point.
-    // We need to tell the MSVC linker the entry point and subsystem.
-    println!("cargo:rustc-link-arg=/ENTRY:mainCRTStartup");
-    println!("cargo:rustc-link-arg=/SUBSYSTEM:CONSOLE");
+    if target_env == "gnu" {
+        // useful for `just clippyw` (targeting x86_64-pc-windows-gnu on linux host)
```
jprendes, line 26: shouldn't we use x86_64-pc-windows-msvc for clippy on linux?

jprendes (approved):

src/hyperlight_host/src/hyperlight_surrogate/build.rs
ludfjig: I don't think that's possible

jprendes: Certainly possible, you need to install the `x86_46-pc-windows-msvc` target using rustup, here's for a different project on my Linux machine:
image

PR #1568, opened 2026-06-22
Automate removal of the ready-for-review label

simongdavies:
## What

Automates removal of the `ready-for-review` label from a pull request once it is
no longer awaiting review.

The label is used to automate requests for review.

The label is removed when **any** of the following become true:

- the PR is closed or merged,
- the PR is converted to a draft,
- the PR has **two or more** approvals, or
- the PR has **two or more** change requests.

Adding the label stays a manual / author action (now documented in
`docs/github-labels.md`); only removal is automated.

## How

The reusable workflow lives in the organisation's `.github` repository
(`hyperlight-dev/.github`) so every repository in the org can share one implementation.

`pull_request_target` is used so the job has a write-scoped token for pull
requests from forks. This is safe because the engine never checks out or
executes PR code — it only reads PR metadata and removes a label.

### DIFF
--- a/.github/workflows/ready-for-review-label.yml
+++ b/.github/workflows/ready-for-review-label.yml
@@ -32,6 +32,6 @@ jobs:
   manage-label:
     # Shared engine in the org-wide `.github` repository, pinned to a commit SHA
     # (matching this repository's action-pinning convention).
-    uses: hyperlight-dev/.github/.github/workflows/manage-ready-for-review.yml@12c76e8040b3a64d36cc8e95b129996dba6a8c54
+    uses: hyperlight-dev/.github/.github/workflows/manage-ready-for-review.yml@8eaae4229db908ea64fff06c31dbf5677611ce5e
     with:
       pr-number: ${{ github.event.pull_request.number }}
--- a/.github/workflows/ready-for-review-label.yml
+++ b/.github/workflows/ready-for-review-label.yml
@@ -26,8 +26,8 @@ concurrency:
 
 permissions:
   contents: read
-  pull-requests: write
-
+  pull-requests: read
+  issues: write
 jobs:
   manage-label:
     # Shared engine in the org-wide `.github` repository, pinned to a commit SHA
--- a/.github/workflows/ready-for-review-label.yml
+++ b/.github/workflows/ready-for-review-label.yml
@@ -1,0 +1,37 @@
+name: Ready-for-review label
+
+# Caller workflow that wires this repository's pull-request events up to the
+# reusable `manage-ready-for-review.yml` workflow, which removes the
+# "ready-for-review" label once a PR is no longer awaiting review.
+#
+# `pull_request_target` (rather than `pull_request`) is required so that the job
+# has a read/write token for pull requests opened from forks. This is safe here
+# because the reusable workflow never checks out or executes pull-request code.
+#
+# The label logic lives in the organisation's `.github` repository so that every
+# repository consumes the same engine. To roll this out to another repository,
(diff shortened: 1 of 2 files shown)


.github/workflows/ready-for-review-label.yml
```diff
@@ -0,0 +1,37 @@
...
+
+permissions:
+  contents: read
+  pull-requests: read
+  issues: write
+jobs:
+  manage-label:
+    # Shared engine in the org-wide `.github` repository, pinned to a commit SHA
+    # (matching this repository's action-pinning convention).
+    uses: hyperlight-dev/.github/.github/workflows/manage-ready-for-review.yml@8eaae4229db908ea64fff06c31dbf5677611ce5e
```
jsturtevant: TIL

.github/workflows/ready-for-review-label.yml
```diff
@@ -0,0 +1,37 @@
+name: Ready-for-review label
+
+# Caller workflow that wires this repository's pull-request events up to the
+# reusable `manage-ready-for-review.yml` workflow, which removes the
+# "ready-for-review" label once a PR is no longer awaiting review.
+#
+# `pull_request_target` (rather than `pull_request`) is required so that the job
+# has a read/write token for pull requests opened from forks. This is safe here
```
jsturtevant, line 8: Did we try this with pull_request first?  its only a write for issues and I don't think `pull_request_target` is needed for that: https://github.com/hyperlight-dev/hyperlight/blob/69c362b6ea2f6e273d9559d4baf18091f07134b3/.github/workflows/RustNightly.yml#L133-L134

pull_request also has types of closed/converted_to_draft https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request

ludfjig: agree pull_request_target is a code smell. Although it might be better soon https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/

.github/workflows/ready-for-review-label.yml
ludfjig: TIL too :) https://docs.github.com/en/actions/how-tos/reuse-automations/share-with-your-organization

nit: the comment above sounds very AI-generated

.github/workflows/ready-for-review-label.yml
simongdavies: OK I will change it, it will probably mean that this wont work for `close` or `converted_to_draft` on a PR from a fork though so we might have to change this back since quite a few people open PRs from their forks , we will soon see.

.github/workflows/ready-for-review-label.yml
simongdavies: fixed the comments

.github/workflows/ready-for-review-label.yml
```diff
@@ -0,0 +1,33 @@
+name: Ready-for-review label
+
+# When a pull-request changes state (closed, converted to draft) or a review is submitted/dismissed,
+# run the  `manage-ready-for-review.yml` workflow, to remove the 
+# "ready-for-review" label once a PR if its no longer needed.
```
danbugs:
```suggestion
# run the `manage-ready-for-review.yml` workflow, to remove the
# "ready-for-review" label from a PR if its no longer needed.
```

danbugs (dismissed):
LGTM–just one nit on documentation

ludfjig (dismissed):

ludfjig (approved):

PR #1578, opened 2026-06-24
feat(whp): support no-surrogate mode via HYPERLIGHT_MAX_SURROGATES=0

danbugs:
- When `HYPERLIGHT_MAX_SURROGATES=0`, skip surrogate process creation entirely and use `VirtualAlloc` + `WHvMapGpaRange` instead of `CreateFileMappingA` + surrogate + `WHvMapGpaRange2`
- This is a single-VM-per-process mode (`WHvMapGpaRange` returns `ERROR_VID_PARTITION_ALREADY_EXISTS` when called from multiple partitions in the same process)
- `compute_surrogate_counts()` now accepts 0 as a valid minimum, and `surrogates_disabled()` checks the env var at runtime
- `WhpVm::surrogate_process` is now `Option<SurrogateProcess>`, with `map_memory`/`unmap_memory` branching at runtime
- `ExclusiveSharedMemory::new()` uses `VirtualAlloc` (via new `DirectAllocation` RAII type) when surrogates are disabled, `CreateFileMappingA` otherwise

### DIFF
--- a/.github/workflows/dep_build_test.yml
+++ b/.github/workflows/dep_build_test.yml
@@ -118,6 +118,38 @@ jobs:
           # with only one driver enabled (kvm/mshv3 features are unix-only, no-op on Windows)
           just test ${{ inputs.config }} ${{ inputs.hypervisor == 'mshv3' && 'mshv3' || 'kvm' }}
 
+      - name: Run no-surrogate VM tests (WHP only)
+        if: runner.os == 'Windows'
+        env:
+          HYPERLIGHT_MAX_SURROGATES: "0"
+          HYPERLIGHT_INITIAL_SURROGATES: "0"
+        run: |
+          PROFILE=${{ inputs.config == 'debug' && 'dev' || inputs.config }}
+          # Verify each expected test actually ran (guards against silent
+          # renames where cargo test exits 0 with 0 matches).
+          # NOTE: keep in sync with `just test-no-surrogate`.
+          assert_ran() {
+            local output="$1"; shift
+            for name in "$@"; do
+              if ! echo "$output" | grep -q "$name \.\.\. ok"; then
+                echo "::error::Expected test '$name' did not run — may have been renamed"
+                return 1
+              fi
+            done
+          }
+          OUT=$(cargo test -p hyperlight-host --profile=$PROFILE --lib -- no_surrogate_tests --test-threads=1 2>&1) || { echo "$OUT"; exit 1; }
+          echo "$OUT"
+          assert_ran "$OUT" single_vm_lifecycle
+          OUT=$(cargo test -p hyperlight-host --profile=$PROFILE --test integration_test -- guest_malloc guest_panic corrupt_output_size_prefix_rejected --test-threads=1 2>&1) || { echo "$OUT"; exit 1; }
+          echo "$OUT"
+          assert_ran "$OUT" guest_malloc guest_panic corrupt_output_size_prefix_rejected
+          OUT=$(cargo test -p hyperlight-host --profile=$PROFILE --test sandbox_host_tests -- --exact callback_test float_roundtrip --test-threads=1 2>&1) || { echo "$OUT"; exit 1; }
+          echo "$OUT"
+          assert_ran "$OUT" callback_test float_roundtrip
+          OUT=$(cargo test -p hyperlight-host --profile=$PROFILE --lib -- snapshot_evolve_restore_handles_state_correctly restore_from_loaded_snapshot --test-threads=1 2>&1) || { echo "$OUT"; exit 1; }
+          echo "$OUT"
+          assert_ran "$OUT" snapshot_evolve_restore_handles_state_correctly restore_from_loaded_snapshot
+
       - name: Run Rust tests with hw-interrupts
         run: |
           # with hw-interrupts feature enabled (+ explicit driver on Linux)
--- a/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
+++ b/src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
@@ -19,6 +19,7 @@ use std::fs::File;
 use std::io::Write;
 use std::mem::size_of;
 use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
 use std::sync::atomic::{AtomicUsize, Ordering};
 
 use crossbeam_channel::{Receiver, Sender, TryRecvError, unbounded};
@@ -86,17 +87,20 @@ fn surrogate_binary_name() -> Result<String> {
 /// (or `None` when the variable is unset or unparsable).
 ///
 /// Resolution order:
-/// 1. `max` is clamped to `1..=HARD_MAX_SURROGATE_PROCESSES`, defaulting
+/// 1. `max` is clamped to `0..=HARD_MAX_SURROGATE_PROCESSES`, defaulting
 ///    to `HARD_MAX_SURROGATE_PROCESSES` when `None`.
-/// 2. `initial` is clamped to `1..=max`, defaulting to `max` when `None`.
+/// 2. `initial` is clamped to `0..=max`, defaulting to `max` when `None`.
 ///    This guarantees `initial <= max` without an extra conditional.
+///
+/// When `max == 0`, surrogates are disabled entirely and the system
+/// falls back to `WHvMapGpaRange` (single-VM-per-process mode).
 fn compute_surrogate_counts(raw_initial: Option<usize>, raw_max: Option<usize>) -> (usize, usize) {
     let max = raw_max
-        .map(|n| n.clamp(1, HARD_MAX_SURROGATE_PROCESSES))
+        .map(|n| n.clamp(0, HARD_MAX_SURROGATE_PROCESSES))
         .unwrap_or(HARD_MAX_SURROGATE_PROCESSES);
 
-    // Clamp initial to 1..=max so it can never exceed the authoritative limit.
-    let initial = raw_initial.map(|n| n.clamp(1, max)).unwrap_or(max);
+    // Clamp initial to 0..=max so it can never exceed the authoritative limit.
+    let initial = raw_initial.map(|n| n.clamp(0, max)).unwrap_or(max);
 
     (initial, max)
 }
@@ -104,8 +108,8 @@ fn compute_surrogate_counts(raw_initial: Option<usize>, raw_max: Option<usize>) 
 /// Returns the (initial, max) surrogate process counts from environment
 /// variables, applying validation and clamping.
 ///
-/// - `HYPERLIGHT_INITIAL_SURROGATES`: clamped to `1..=max`, default `max`.
-/// - `HYPERLIGHT_MAX_SURROGATES`: clamped to `1..=512`, default 512.
+/// - `HYPERLIGHT_INITIAL_SURROGATES`: clamped to `0..=max`, default `max`.
+/// - `HYPERLIGHT_MAX_SURROGATES`: clamped to `0..=512`, default 512.
 fn surrogate_process_counts() -> (usize, usize) {
     let raw_initial = std::env::var(INITIAL_SURROGATES_ENV_VAR)
         .ok()
@@ -351,6 +355,21 @@ pub(crate) fn get_surrogate_process_manager() -> Result<&'static SurrogateProces
             Err(new_error!("Failed to get SurrogateProcessManager {}", e))
         }
     }
+}
+
+/// Returns `true` when `HYPERLIGHT_MAX_SURROGATES=0`, meaning surrogate
+/// processes are disabled and the system should use `WHvMapGpaRange`
+/// (single-VM-per-process mode) instead of `WHvMapGpaRange2`.
+///
+/// The result is cached on first call — the env var is read only once.
+pub(crate) fn surrogates_disabled() -> bool {
+    static DISABLED: OnceLock<bool> = OnceLock::new();
+    *DISABLED.get_or_init(|| {
+        std::env::var(MAX_SURROGATES_ENV_VAR)
+            .ok()
+            .and_then(|v| v.parse::<usize>().ok())
+            .is_some_and(|n| n == 0)
+    })
 }
 
 // Creates a job object that will terminate all the surrogate processes when the struct instance is dropped.
@@ -885,9 +904,9 @@ mod tests {
             "initial should be clamped down to max when it exceeds it"
         );
 
-        // --- initial below minimum → clamped to 1 ---
+        // --- initial at zero → allowed (surrogates disabled when max is also 0) ---
         let (initial, max) = compute_surrogate_counts(Some(0), None);
-        assert_eq!(initial, 1, "initial should be clamped to minimum of 1");
+        assert_eq!(initial, 0, "initial of 0 should be allowed");
         assert_eq!(
             max, HARD_MAX_SURROGATE_PROCESSES,
             "max should default when unset"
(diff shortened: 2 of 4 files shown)


src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -634,55 +672,74 @@ impl ExclusiveSharedMemory {
...
+        let mut unused = PAGE_PROTECTION_FLAGS(0);
+
+        // If these calls are changed, update VirtualProtectEx in
+        // surrogate_process_manager.rs too.
+        if let Err(e) = unsafe {
+            VirtualProtect(base, PAGE_SIZE_USIZE, PAGE_NOACCESS, &mut unused)
+        } {
+            log_then_return!(WindowsAPIError(e.clone()));
+        }
+        let last = unsafe { base.add(total_size - PAGE_SIZE_USIZE) };
```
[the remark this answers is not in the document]
danbugs: Fixed by casting to `*mut u8` before the arithmetic: `(base as *mut u8).add(total_size - PAGE_SIZE_USIZE) as *mut c_void`.

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -133,6 +135,11 @@ enum WindowsMapping {
         trailing: Placeholder,
         file_mapping: FileMapping,
     },
+    /// `VirtualAlloc`-backed allocation used with `whp-no-surrogate`.
+    /// No file mapping or surrogate process needed — memory is mapped
+    /// into the partition with `WHvMapGpaRange` from the host address.
+    #[cfg(feature = "whp-no-surrogate")]
+    DirectAlloc(DirectAllocation),
```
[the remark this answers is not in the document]
danbugs: Added a `DirectAlloc(_)` arm to `ReadonlySharedMemory::host_region_base()`, following the same layout as the `Anonymous` variant (full region with `PAGE_SIZE_USIZE` offset).

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -101,13 +107,9 @@ pub(crate) struct WhpVm {
...
-// Safety: `WhpVm` is !Send because it holds `SurrogateProcess` which contains a raw pointer
-// `allocated_address` (*mut c_void). This pointer represents a memory mapped view address
-// in the surrogate process. It is never dereferenced, only used for address arithmetic and
-// resource management (unmapping). This is a system resource that is not bound to the creating
-// thread and can be safely transferred between threads.
-// `file_mappings` contains raw pointers that are also kernel resource handles,
-// safe to use from any thread.
+// Safety: `WhpVm` is !Send because it holds raw pointers (in `file_mappings` and,
+// when present, `SurrogateProcess`). These are kernel resource handles / addresses
+// that are not bound to the creating thread and can be safely transferred.
```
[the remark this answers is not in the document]
danbugs: Reworded to clarify that `WhpVm` is auto-`!Send` (due to raw pointers) and the manual `Send` `impl` is sound because the pointers are kernel resource handles that are never dereferenced and not bound to the creating thread.

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -709,48 +764,43 @@ impl ExclusiveSharedMemory {
...
+            }),
+        })
+    }
 
-        // If the following calls to VirtualProtect are changed make sure to update the calls to VirtualProtectEx in surrogate_process_manager.rs
+    /// Create a new region of shared memory backed by `VirtualAlloc`.
+    /// Used with `whp-no-surrogate` — no file mapping or surrogate
+    /// process needed.
+    #[cfg(all(target_os = "windows", feature = "whp-no-surrogate"))]
+    #[instrument(skip_all, parent = Span::current(), level= "Trace")]
```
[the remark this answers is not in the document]
danbugs: Added test coverage in CI.

ludfjig (commented):
Have you considered making this a runtime option instead, for example on SandboxConfiguration?  If possible I think I would prefer it

danbugs:
> Have you considered making this a runtime option instead, for example on SandboxConfiguration? If possible I think I would prefer it

Had a chat w/ @simongdavies and modified this PR to integrate w/ `HYPERLIGHT_MAX_SURROGATES`. Now, if you set that to zero, it just doesn't spawn a surrogate process–so, essentially, a runtime config 👍

simongdavies (dismissed):
LGTM, prefer this to having a feature.

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
```diff
@@ -353,6 +356,16 @@ pub(crate) fn get_surrogate_process_manager() -> Result<&'static SurrogateProces
     }
 }
 
+/// Returns `true` when `HYPERLIGHT_MAX_SURROGATES=0`, meaning surrogate
+/// processes are disabled and the system should use `WHvMapGpaRange`
+/// (single-VM-per-process mode) instead of `WHvMapGpaRange2`.
+pub(crate) fn surrogates_disabled() -> bool {
```
ludfjig, line 365: should we cache this?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -103,6 +105,24 @@ pub struct HostMapping {
...
+struct DirectAllocation {
+    addr: *mut c_void,
+    size: usize,
+}
+
+#[cfg(target_os = "windows")]
+impl Drop for DirectAllocation {
+    fn drop(&mut self) {
+        unsafe {
+            let _ = VirtualFree(self.addr, 0, MEM_RELEASE);
```
ludfjig: can we trace errors here, mirroring existing patterns

ludfjig (commented):
LGTM, but I think we definitely need some testing that exercises the no-surrogate-process code? Can we rerun existing tests setting it to 0? And should we error gracefully if multiple sandboxes are created in this mode?

src/hyperlight_host/src/hypervisor/surrogate_process_manager.rs
danbugs: Done.

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Done. Matches the `FileMapping`/`Placeholder` pattern with `tracing::error!`.

danbugs:
> Can we rerun existing tests setting it to 0? And should we error gracefully if multiple sandboxes are created in this mode?

Added testing. Opted to not re-run the full CI to save time and also because some tests just don't make sense (e.g., full tests validate surrogate process machinery).

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -665,57 +697,95 @@ impl ExclusiveSharedMemory {
...
+                size: total_size,
+            };
 
-        if std::mem::size_of::<usize>() == 8 {
-            dwmaximumsizehigh = (total_size >> 32) as u32;
-            dwmaximumsizelow = (total_size & 0xFFFFFFFF) as u32;
-        }
+            Self::setup_guard_pages(alloc.addr, total_size)?;
 
-        // Allocate the memory use CreateFileMapping instead of VirtualAlloc
```
syntactically, line 676: Is there any downside to always using CreateFileMapping, even though it isn't strictly necessary in this use case?

syntactically (commented):
Looks fine to me

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: `CreateFileMappingA` just felt like extra kernel overhead that isn't really used for the single-VM-per-process scenario this is targeting, so I opted for `VirtualAlloc`, which is a bit lighter-weight. Not much more to it, tbh.

syntactically: If we did use CreateFileMappingA, would we be able to basically avoid the changes to shared_memory.rs entirely? I doubt there is that much difference in the kernel structures for the two, since anonymous memory probably still ends up in a page cache.

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -143,11 +155,23 @@ impl WhpVm {
...
+        let surrogate_process = if surrogates_disabled() {
+            if NO_SURROGATE_VM_ACTIVE.swap(true, Ordering::SeqCst) {
+                return Err(CreateVmError::SurrogateProcess(
+                    "HYPERLIGHT_MAX_SURROGATES=0 limits the process to a single VM; \
+                     a VM is already active"
+                        .into(),
+                ));
+            }
+            None
+        } else {
```
ludfjig: Shouldn't this be checked at the top of the func? Otherwise the stuff above will leak?

ludfjig (dismissed):
I think there's a leak possible. And I don't think any tests exercises the new memory apis still. Can we at least run a subset of existing tests to make sure guest calls/hostcall and snapshotting/persist snapshotting still works as expected?

We should also add this to changelog, up to you if you want to do it in this PR or future PR though

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: You are right--done!!

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
danbugs: Moved the `NO_SURROGATE_VM_ACTIVE` check to the top of `WhpVm::new()`, before any partition/vCPU allocation. Also wrapped the rest of the function in a closure so the guard resets on any creation failure.

danbugs:
> And I don't think any tests exercises the new memory apis still. Can we at least run a subset of existing tests to make sure guest calls/hostcall and snapshotting/persist snapshotting still works as expected?

 Expanded the CI no-surrogate step to run `guest_malloc`, `guest_panic`, `corrupt_output_size_prefix_rejected` (integration tests covering guest calls + snapshot/restore) and `float_roundtrip`, `callback_test`. 

> We should also add this to changelog, up to you if you want to do it in this PR or future PR though

Will update the changelog in a follow-up before the release.

cc: @ludfjig

ludfjig:
> > And I don't think any tests exercises the new memory apis still. Can we at least run a subset of existing tests to make sure guest calls/hostcall and snapshotting/persist snapshotting still works as expected?
> 
> Expanded the CI no-surrogate step to run `guest_malloc`, `guest_panic`, `corrupt_output_size_prefix_rejected` (integration tests covering guest calls + snapshot/restore) and `float_roundtrip`, `callback_test`.
> 
> > We should also add this to changelog, up to you if you want to do it in this PR or future PR though
> 
> Will update the changelog in a follow-up before the release.
> 
> cc: @ludfjig

Looks good but I don't think snapshotting nor saving/loading snapshot from/to disk is covered by these added tests. And minor nit is that renaming these tests will not fail ci, it will just report running 0 tests, maybe we can grep for them in output to make sure they ran and passed. I fear that we might rename the tests later and forget updating the ci

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
```diff
@@ -114,48 +126,66 @@ impl WhpVm {
...
+            Ok(WhpVm {
+                partition,
+                surrogate_process,
+                file_mappings: Vec::new(),
+                #[cfg(feature = "hw-interrupts")]
+                timer: None,
+            })
+        })();
+
+        if result.is_err() && no_surrogate {
```
syntactically: Nit: instead of doing this you could use a RAII-style guard that sets it back until the change is committed, which would also work if unwinding a panic in the binding code, and makes `?` work nicely.

.github/workflows/dep_build_test.yml
```diff
@@ -118,6 +118,18 @@ jobs:
           # with only one driver enabled (kvm/mshv3 features are unix-only, no-op on Windows)
           just test ${{ inputs.config }} ${{ inputs.hypervisor == 'mshv3' && 'mshv3' || 'kvm' }}
 
+      - name: Run no-surrogate VM tests (WHP only)
```
syntactically, line 121: If changing the GH workflow, can we update `just test-like-ci` as well?

syntactically (dismissed):

danbugs:
> Looks good but I don't think snapshotting nor saving/loading snapshot from/to disk is covered by these added tests.

Current coverage:
- Guest calls: guest_malloc (calls TestMalloc in guest)
- Host calls: callback_test (guest calls back into a registered host function)
- Snapshot/restore: guest_panic triggers a panic, then the sandbox auto-restores from snapshot to handle the next call. corrupt_output_size_prefix_rejected similarly exercises restore after a poisoned sandbox.
- Snapshot save/load from disk: restore_from_loaded_snapshot (saves snapshot to disk, loads it, restores from it)
- In-memory snapshot: snapshot_evolve_restore_handles_state_correctly (explicit snapshot/restore cycle)

> And minor nit is that renaming these tests will not fail ci, it will just report running 0 tests, maybe we can grep for them in output to make sure they ran and passed. I fear that we might rename the tests later and forget updating the ci

Added a rename guard for this.

.github/workflows/dep_build_test.yml
danbugs: Good catch–updated!!

src/hyperlight_host/src/hypervisor/virtual_machine/whp.rs
danbugs: Done.

ludfjig:
> > Looks good but I don't think snapshotting nor saving/loading snapshot from/to disk is covered by these added tests.
> 
> Current coverage:
> 
> * Guest calls: guest_malloc (calls TestMalloc in guest)
> * Host calls: callback_test (guest calls back into a registered host function)
> * Snapshot/restore: guest_panic triggers a panic, then the sandbox auto-restores from snapshot to handle the next call. corrupt_output_size_prefix_rejected similarly exercises restore after a poisoned sandbox.
> * Snapshot save/load from disk: restore_from_loaded_snapshot (saves snapshot to disk, loads it, restores from it)
> * In-memory snapshot: snapshot_evolve_restore_handles_state_correctly (explicit snapshot/restore cycle)
> 
> > And minor nit is that renaming these tests will not fail ci, it will just report running 0 tests, maybe we can grep for them in output to make sure they ran and passed. I fear that we might rename the tests later and forget updating the ci
> 
> Added a rename guard for this.

Tests look perfect thanks!

For the guard, why not have the guard itself set the static boolean when created (or fail if already created), and have the drop unset it, and store on the vm as a field. I think this would be simpler (and no need to std::mem::forget)

syntactically:
> For the guard, why not have the guard itself set the static boolean when created (or fail if already created), and have the drop unset it, and store on the vm as a field. I think this would be simpler (and no need to std::mem::forget)

Much better idea, thanks.

Actually, with this change, the surrogate guard and the surrogate process have the exact same lifetime and exactly one of them is used. So maybe you could even make the SurrogateProcess an enum or a dyn Trait and push the difference in behaviour all the way down to that level, so whp.rs does not need to know or care whether there is a surrogate process.

To unify the `WHvMapGpaRange` vs `WHvMapGpaRange2` distinction, you could either always use `WHvMapGpaRange2` and just use the current process handle when you are not using a surrogate, or you could move the mapping call into the `surrogate_process_manager.rs` or similar.

ludfjig (approved):
LGTM (nit last bullet on pr description is stale)

PR #1593, opened 2026-06-26
fix: update tar-headers sysroot path for cargo-hyperlight 0.1.12

danbugs:
Fixes the publish job failure on main: https://github.com/hyperlight-dev/hyperlight/actions/runs/28255787499/job/83728732967

cargo-hyperlight 0.1.12 moved the sysroot include directory from `target/sysroot/lib/rustlib/x86_64-hyperlight-none/include` to `target/sysroot/include`.


ludfjig (approved):
lgtm, (I didn't double check cargo-hyperlight change)

PR #1600, opened 2026-06-29
Adds optional repo variables for job timeout values

simongdavies:
The timeout values on GH jobs are hardcoded, this means that if we need to change any of these values even temporarily then we need to submit a PR, this PR allows the timeout value for each job to be optionally read from a repo variable. The hardcoded (fallback) values are not changed.


jprendes (dismissed):
LGTM

jprendes (approved):

PR #1603, opened 2026-06-30
update PR validation and daily runs to reduce PR dependance on ARM64 runners

simongdavies:
Helps to speed up PR validation whilst we have limited ARM runners, some validation/tests are moved to a daily job that runs at 5am UTC daily, failures will result in a release blocking issue

### DIFF
--- a/.github/workflows/DailyArm64.yml
+++ b/.github/workflows/DailyArm64.yml
@@ -1,0 +1,106 @@
+# yaml-language-server: $schema=https://json.schemastore.org/github-workflow.json
+
+name: Daily aarch64
+
+# Full aarch64 coverage runs here on a schedule rather than on every PR, because
+# there are only limited aarch64 runners. PRs run a reduced aarch64
+# build-test (build + default Rust tests only) and skip aarch64 examples; this
+# workflow restores Miri, the single-driver tests, and the examples once a day.
+
+on:
+  schedule:
+    - cron: '0 5 * * *' # Runs at 05:00 UTC every day
+  workflow_dispatch: # Allow manual triggering
+
+env:
+  CARGO_TERM_COLOR: always
+  RUST_BACKTRACE: full
+
+permissions:
+  contents: read
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  # Build aarch64 guest binaries once and upload them as artifacts for the
+  # build-test and run-examples jobs to download.
+  build-guests:
+    strategy:
+      fail-fast: false
+      matrix:
+        config: [debug, release]
+    uses: ./.github/workflows/dep_build_guests.yml
+    secrets: inherit
+    with:
+      arch: arm64
+      config: ${{ matrix.config }}
+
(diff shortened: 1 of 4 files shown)


.github/workflows/DailyArm64.yml
```diff
@@ -0,0 +1,109 @@
...
+  # File (or update) a release-blocking GitHub issue if any job fails. The first
+  # label (area/ci-periodics-aarch64) is the de-duplication key used by
+  # notify-ci-failure.sh, so this workflow maintains its own issue lineage and is
+  # guaranteed to carry release-blocker, instead of commenting on another periodic
+  # job's issue. NOTE: that label must exist in the repo (Issues -> Labels) or the
+  # `gh issue create` call will fail.
+  notify-failure:
+    runs-on: ubuntu-latest
+    needs: [build-guests, build-test, run-examples, fuzzing]
+    if: always() && (needs.build-guests.result == 'failure' || needs.build-test.result == 'failure' || needs.run-examples.result == 'failure' || needs.fuzzing.result == 'failure')
```
danbugs, line 96: Does this also catch timeouts?

danbugs (commented):
Mostly LGTM. Just one question.

.github/workflows/ValidatePullRequest.yml
```diff
@@ -138,10 +138,9 @@ jobs:
...
           hypervisor: mshv3
-        - cpu_vendor: amd
-          arch: arm64
-        - cpu_vendor: intel
-          arch: arm64
+        # aarch64 examples are exercised by the daily schedule (DailyArm64.yml)
+        # only, to keep the limited arm64 runners free on PRs.
+        - arch: arm64
         - cpu_vendor: apple
           arch: X64
```
ludfjig: is it possible to define `exclude` like this to simplify if we never run arm64/apple on PRs?
```
exclude:
- arch: arm64
- cpu_vendor: apple
```

.github/workflows/DailyArm64.yml
ludfjig: yeah I'm wondering why this looks different to `report-ci-status` in ValidatePullRquest.yml (the `if` statement that is)

ludfjig (dismissed):

.github/workflows/DailyArm64.yml
simongdavies: No, I modelled this on the other jobs we have that do this, the problem with cancellation is that any sort of cancellation would trigger this (i.e. a manual cancellation or a concurrency cancellation) , I think its OK to skip this?

simongdavies:
> yeah I'm wondering why this looks different to report-ci-status in ValidatePullRquest.yml (the if statement that is)

All the daily jobs failure notify steps are defined like this e.g. https://github.com/hyperlight-dev/hyperlight/blob/main/.github/workflows/DailyBenchmarks.yml

ludfjig (approved):
Ah I see didn't see that other workflow already use similar pattern

PR #1605, opened 2026-07-01
perf(mem): avoid eager scratch re-zero on restore

simongdavies:
On snapshot restore the guest scratch region is re-zeroed to reset state. On Linux/KVM SharedMemory::zero() is an O(1) madvise(MADV_DONTNEED), but on mshv (which maps guest memory up-front without host-MM notification) and Windows/WHP it degrades to O(size) memset. 

This change swaps in a fresh demand-zero scratch region and remaps it into the guest (the same path already taken when the scratch size changes) for Windows and mshv. The OS zero-fills lazily on fault and the old region is released once it has been unmapped from the VM. Restore still resets while the eager memset becomes an O(1) allocation . 

Temporary scaffolding for CI validation (to be reduced to the one-line change before review): a HYPERLIGHT_SCRATCH_ZERO_STRATEGY env var to force either mechanism, unit tests for the strategy resolver, and tests/scratch_restore_perf.rs which measures per-restore timing and asserts no mapping/handle leak across 50 restores of large scratch regions on both strategies (captured to the CI step summary).

### DIFF
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -113,10 +113,17 @@ hyperlight-component-macro = { workspace = true }
 [target.'cfg(windows)'.dev-dependencies]
 windows = { version = "0.62", features = [
     "Win32_System_Diagnostics_ToolHelp",
+    # Process working-set / private-usage and handle-count probes used by the
+    # scratch-restore perf/leak integration test.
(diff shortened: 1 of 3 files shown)


simongdavies:
Closing this, apporach does not work, is more expensive than just resetting the entire scratch to zero

PR #1606, opened 2026-07-01
make sure nightly job failure adds release blocker label

### LINKED ISSUE #1526: #1522 did not get release-blocker label
#1522 should have gotten it to prevent release (I added the label manually)

simongdavies:
fixes #1526

### DIFF
--- a/.github/workflows/RustNightly.yml
+++ b/.github/workflows/RustNightly.yml
@@ -140,6 +140,6 @@ jobs:
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
 
       - name: Notify Nightly Failure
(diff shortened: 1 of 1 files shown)


ludfjig (commented):
How does this fix it?

simongdavies:
> How does this fix it?

 The script that runs creates an issue if it cannot find an open issue with the first label passed as an argument, both this job and the coverage job share a first label, the coverage job does not apply the release-blocker label, so if the coverage job fails it creates an issue WITHOUT the release-blocker label, then this job fails finds the issue and adds to it, but doesn't add the label.
 
 I guess we could change the script but this PR gives a unique label to this job so the issue will always have the release-blocker label.

ludfjig (approved):

PR #1634, opened 2026-07-09
feat(virtq): add virtio-villain inspired packed virtq coverage

andreiltd:
Map applicable packed ring cases into deterministic in process virtq tests, and add a packed-ring fuzz target with binary seed inputs.

See: https://github.com/weltling/virtio-villain/tree/main/tests/packed

### DIFF
--- a/fuzz/fuzz_targets/virtq_packed_ring.rs
+++ b/fuzz/fuzz_targets/virtq_packed_ring.rs
@@ -1,0 +1,271 @@
+/*
+Copyright 2026  The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+#![no_main]
+
+use std::cell::UnsafeCell;
+use std::num::NonZeroU16;
+use std::ops::Range;
+use std::rc::Rc;
+
+use hyperlight_common::virtq::{Descriptor, Layout, MemOps, RingConsumer};
+use libfuzzer_sys::{Corpus, fuzz_target};
+
+const DEFAULT_QUEUE_SIZE: usize = 16;
+const MAX_QUEUE_SIZE: usize = 64;
+const MAX_DESCS: usize = 64;
+const PAYLOAD_SIZE: usize = 4096;
+const BASE_ADDR: u64 = 0x1000;
+const HEADER_SIZE: usize = 16;
+const DESC_SIZE: usize = 12;
+
+#[derive(Clone, Debug)]
+struct FuzzDesc {
+    addr_offset: u32,
+    len: u32,
+    id: u16,
+    flags: u16,
+}
+
+#[derive(Clone, Debug)]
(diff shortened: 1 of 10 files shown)


fuzz/fuzz_targets/virtq_packed_ring.rs
```diff
@@ -0,0 +1,276 @@
...
+struct FuzzMemInner {
+    storage: UnsafeCell<Vec<u8>>,
+    base_addr: u64,
+}
+
+// SAFETY: The fuzz target is single-threaded. `FuzzMem` uses interior
+// mutability only to satisfy `MemOps`' shared-reference API.
+unsafe impl Send for FuzzMemInner {}
+unsafe impl Sync for FuzzMemInner {}
+
```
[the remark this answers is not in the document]
andreiltd: Fixed

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -3421,6 +3433,795 @@ pub(crate) mod tests {
...
+        let ring = make_ring(8);
+        let mut consumer = make_consumer(&ring);
+
+        write_driver_event(&ring, 0, 3);
+        write_avail(&ring, 0, 0, DescFlags::empty());
+
+        let (id, _) = consumer.poll_available().unwrap();
+
+        assert!(!consumer.submit_used_with_notify(id, 0).unwrap());
+    }
```
ludfjig, line 3850: nit: seems identical to p0014

src/hyperlight_common/src/virtq/ring.rs
```diff
@@ -3421,6 +3433,795 @@ pub(crate) mod tests {
...
+        let ring = make_ring(8);
+        let mut consumer = make_consumer(&ring);
+
+        write_driver_event(&ring, 0x8000, EventFlags::DESC.bits());
+        write_avail(&ring, 0, 0, DescFlags::empty());
+
+        let (id, _) = consumer.poll_available().unwrap();
+
+        assert!(consumer.submit_used_with_notify(id, 0).unwrap());
+    }
```
ludfjig, line 3837: nit: looks identical to p0023

ludfjig (dismissed):
LGTM!

src/hyperlight_common/src/virtq/ring.rs
andreiltd: Copy/Paste mistake. Thanks!

ludfjig (dismissed):

fuzz/corpus/fuzz_virtq_packed_ring/event-off-qsize.bin
jsturtevant, line 1: do these bin's need to be checked in?  looks like gitignore as added after?

andreiltd: Yes, the intention was to have a valid queue as starting point for fuzzer and mutate memory from there. The seed is just binary form of the queue. I think it's more realistic than just starting from completely random mapping.

jsturtevant: Is there another way to produce and maybe cache them without checking in binaries? Having binaries makes it hard to identify changes to the files.

andreiltd: I removed binary seeds for now, I'm not sure what's the best way to pregenerate them but probably have a binary crate that reads the format from json/toml and call that from justfile before fuzz targets. For now, let's hope fuzzer learns virtio format quickly without providing meaningful state. 🤞

jsturtevant (approved):

ludfjig (approved):

PR #1636, opened 2026-07-09
Rewrite copilot instructions

ludfjig:
Rewriting it slightly for hopefully better results

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,82 +1,132 @@
-This repository contains hyperlight. Hyperlight is a lightweight Virtual Machine Manager (VMM) designed to be embedded within applications.
-It enables safe execution of untrusted code within micro virtual machines with very low latency and minimal overhead.
+# Hyperlight: Copilot instructions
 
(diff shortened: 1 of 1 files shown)


.github/copilot-instructions.md
```diff
@@ -1,82 +1,132 @@
...
+
+## Code conventions
+
+* Follow idiomatic Rust and match the surrounding code style.
+* Do not add new crates or dependencies without discussing first.
+* Maintain the existing structure and organization. Do not refactor beyond
+  what the task requires.
+* Do not commit large binary files.
+
+## Writing
```
jsturtevant, line 46: To keep the copilot instructions smaller in the Context, we could split these out to separate files like .github/copilot-instructions.writing.md and then link to it here so it can selectively load this when needed.  

We could do this for "Reviewing code" section below.

jsturtevant (approved):
LGTM, minor suggestion but overall should help

andreiltd (approved):

.github/copilot-instructions.md
ludfjig: I considered this but I think it's still pretty small. Can change in future if we need to

Issue #1581, opened 2026-06-26
Unify surrogate/no-surrogate mapping behind a single abstraction


danbugs:
## Context

PR #1578 adds a no-surrogate mode (`HYPERLIGHT_MAX_SURROGATES=0`) that uses `WHvMapGpaRange` instead of `WHvMapGpaRange2`. Currently `whp.rs` branches on `Option<SurrogateProcess>` to choose the mapping path.

## Proposal

Unify the surrogate and no-surrogate paths so `whp.rs` doesn't need to know or care which mode is active. Two possible approaches:

1. **Enum / trait abstraction**: Make `SurrogateProcess` an enum (or `dyn Trait`) with two variants — one backed by a real surrogate process (`WHvMapGpaRange2`) and one that maps from the host process directly (`WHvMapGpaRange`). The `NoSurrogateGuard` could be folded into the no-surrogate variant.

2. **Always use `WHvMapGpaRange2`**: Use the current process handle as the source when surrogates are disabled, eliminating the branch entirely. This needs validation — it's unclear whether `WHvMapGpaRange2` with `GetCurrentProcess()` is a supported pattern.

Either way, the mapping call would move into the surrogate module so `whp.rs` just calls a single `map(...)` method.

## Origin

Suggestion from @syntactically  in [PR #1578 review](https://github.com/hyperlight-dev/hyperlight/pull/1578).

PR #1638, opened 2026-07-12
feat(aarch64): add WHP backend for ARM64 Windows

### LINKED ISSUE #1544: Support AArch64 on Windows

cshung:
## Summary

Implements the WHP (Windows Hypervisor Platform) hypervisor backend for aarch64, enabling hyperlight to run micro-VMs on Windows ARM64 systems.

Resolves #1544

### Structural
- Restructure `whp.rs` into `whp/` directory (`mod.rs` + `x86_64.rs`) to support per-architecture implementations, matching the existing `kvm/` and `mshv/` pattern
- Fix `super::x86_64::hw_interrupts` module path after directory restructure

### New: `whp/aarch64.rs`
- Manual FFI bindings for ARM64 WHP register names, exit reasons, and exit context layout (extracted from Windows SDK `WinHvPlatformDefs.h`)
- Full `VirtualMachine` trait implementation with:
  - MMIO-based I/O handling (ARM64 has no IO ports)
  - Register get/set via `WHvGet/SetVirtualProcessorRegisters`
  - Surrogate process support (same pattern as x86_64)

### Integration
- Wire `WhpVm` into `hyperlight_vm/aarch64.rs` for Windows platform
- Add `WindowsInterruptHandle` for aarch64 Windows
- Add `CpuVendor::current()` for aarch64 Windows target

### Cross-compilation fix
- Move `vmm-sys-util` to unix-only dependencies (it doesn't compile on Windows)

## Verification

Verified compilation on three targets:
- **x86_64-pc-windows-msvc** (native) — `just clippy debug/release` pass
- **aarch64-pc-windows-msvc** (cross-compile) — `cargo check` passes
- **aarch64-unknown-linux-gnu** (cross-compile) — `cargo check --features kvm` passes

> **Note:** No runtime testing was possible — requires actual ARM64 Windows hardware with Hyper-V enabled. CI has no ARM64 Windows runners.

## Why manual FFI bindings?

The `windows` crate (v0.62) does not expose ARM64 WHP types (register names, exit reasons, exit context structs). All definitions were extracted from the Windows SDK header `WinHvPlatformDefs.h` (SDK 10.0.26100.0) which has full ARM64 support behind `#ifdef _ARM64_`.

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/aarch64.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/aarch64.rs
@@ -12,8 +12,10 @@ use super::{
 #[cfg(hvf)]
 use crate::hypervisor::HvfInterruptHandle;
 use crate::hypervisor::InterruptHandleImpl;
-#[cfg(any(kvm, mshv3))]
+#[cfg(target_os = "linux")]
 use crate::hypervisor::LinuxInterruptHandle;
+#[cfg(target_os = "windows")]
+use crate::hypervisor::WindowsInterruptHandle;
 #[cfg(gdb)]
 use crate::hypervisor::gdb::{DebugCommChannel, DebugMsg, DebugResponse};
 use crate::hypervisor::hyperlight_vm::get_guest_log_filter;
@@ -22,6 +24,8 @@ use crate::hypervisor::regs::{CommonFpu, CommonRegisters, CommonSpecialRegisters
 use crate::hypervisor::virtual_machine::hvf::HvfVm;
 #[cfg(kvm)]
 use crate::hypervisor::virtual_machine::kvm::KvmVm;
+#[cfg(target_os = "windows")]
+use crate::hypervisor::virtual_machine::whp::WhpVm;
 use crate::hypervisor::virtual_machine::{
     HypervisorType, RegisterError, ResetVcpuError, VirtualMachine, VmError,
     get_available_hypervisor,
@@ -46,7 +50,7 @@ impl HyperlightVm {
         next_action: NextAction,
         rsp_gva: u64,
         page_size: usize,
-        config: &SandboxConfiguration,
+        #[cfg_attr(target_os = "windows", allow(unused_variables))] config: &SandboxConfiguration,
         #[cfg(gdb)] _gdb_conn: Option<DebugCommChannel<DebugResponse, DebugMsg>>,
         #[cfg(crashdump)] _rt_cfg: SandboxRuntimeConfig,
         #[cfg(feature = "mem_profile")] _trace_info: MemTraceInfo,
@@ -59,13 +63,14 @@ impl HyperlightVm {
         let mut vm: VmType = match get_available_hypervisor() {
             #[cfg(kvm)]
             Some(HypervisorType::Kvm) => Box::new(KvmVm::new().map_err(VmError::CreateVm)?),
-            // TODO: mshv support
             #[cfg(mshv3)]
             Some(HypervisorType::Mshv) => return Err(CreateHyperlightVmError::NoHypervisorFound),
             #[cfg(hvf)]
             Some(HypervisorType::Hvf) => {
                 Box::new(HvfVm::new(interrupt_handle.clone()).map_err(VmError::CreateVm)?)
             }
+            #[cfg(target_os = "windows")]
+            Some(HypervisorType::Whp) => Box::new(WhpVm::new().map_err(VmError::CreateVm)?),
             None => return Err(CreateHyperlightVmError::NoHypervisorFound),
         };
         vm.set_sregs(&CommonSpecialRegisters::defaults(root_pt_addr))
@@ -73,6 +78,10 @@ impl HyperlightVm {
         #[cfg(any(kvm, mshv3))]
         let interrupt_handle: Arc<dyn InterruptHandleImpl> =
             Arc::new(LinuxInterruptHandle::new(config));
+
+        #[cfg(target_os = "windows")]
+        let interrupt_handle: Arc<dyn InterruptHandleImpl> =
+            Arc::new(WindowsInterruptHandle::new(vm.partition_handle()));
 
         let snapshot_slot = 0u32;
         let scratch_slot = 1u32;
@@ -193,7 +202,7 @@ impl HyperlightVm {
             self.vm_can_reset_vcpu,
             "No fallback path for vcpu reset on aarch64"
         );
-        self.vm.reset_vcpu()?;
+        self.interrupt_handle.reset_vcpu(self.vm.as_mut())?;
         self.apply_sregs(cr3, sregs)?;
         Ok(())
     }
--- a/src/hyperlight_host/src/hypervisor/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/mod.rs
@@ -32,6 +32,9 @@ use std::sync::atomic::{AtomicU8, Ordering};
 #[cfg(any(kvm, mshv3, hvf))]
 use std::time::Duration;
 
+#[cfg(target_arch = "aarch64")]
+use self::virtual_machine::{ResetVcpuError, VirtualMachine};
+
 #[derive(Debug)]
 pub(crate) struct InterruptHandleStateMachine(AtomicU8);
 impl InterruptHandleStateMachine {
@@ -129,6 +132,12 @@ pub(crate) trait InterruptHandleImpl: InterruptHandle {
 
     /// Mark the handle as dropped
     fn set_dropped(&self);
+
+    /// Reset the vCPU while honoring platform lifecycle synchronization.
+    #[cfg(target_arch = "aarch64")]
+    fn reset_vcpu(&self, vm: &mut dyn VirtualMachine) -> Result<(), ResetVcpuError> {
+        vm.reset_vcpu()
+    }
 }
 
 pub(crate) trait InterruptHandleInternal {
@@ -361,6 +370,20 @@ impl<T: SynchronousInterruptState> InterruptHandleImpl for SynchronousInterruptH
                 tracing::error!("Failed to acquire partition_state write lock: {}", e);
             }
         }
+    }
+
+    #[cfg(target_arch = "aarch64")]
+    fn reset_vcpu(&self, vm: &mut dyn VirtualMachine) -> Result<(), ResetVcpuError> {
+        let guard = self
+            .dropped_state
+            .write()
+            .map_err(|e| ResetVcpuError::Unknown(e.to_string()))?;
+        if guard.0 {
+            return Err(ResetVcpuError::Unknown(
+                "cannot reset a dropped partition".to_string(),
+            ));
+        }
+        vm.reset_vcpu()
     }
 }
 
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
@@ -1,18 +1,5 @@
-/*
-Copyright 2025 The Hyperlight Authors.
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2025 The Hyperlight Authors.
 
 //! Windows Hypervisor Platform (WHP) backend for AArch64.
 //!
--- a/src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
+++ b/src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
@@ -1,0 +1,950 @@
+/*
+Copyright 2025 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Windows Hypervisor Platform (WHP) backend for AArch64.
+//!
+//! This module provides the [`VirtualMachine`] trait implementation using the
+//! WHP APIs on Windows ARM64 systems. Because the `windows` crate does not yet
+//! expose ARM64 WHP structures, we define our own FFI bindings derived from
+//! the Windows SDK header `WinHvPlatformDefs.h` (10.0.26100.0).
+
+use std::os::raw::c_void;
+use std::sync::atomic::Ordering;
+
+use hyperlight_common::outb::VmAction;
+#[cfg(feature = "trace_guest")]
+use tracing::Span;
+#[cfg(feature = "trace_guest")]
+use tracing_opentelemetry::OpenTelemetrySpanExt;
+use windows::Win32::System::Hypervisor::*;
+use windows_result::HRESULT;
+
+use crate::hypervisor::regs::{
+    CommonDebugRegs, CommonFpu, CommonRegisters, CommonSpecialRegisters,
+};
+use crate::hypervisor::surrogate_process::SurrogateProcess;
+use crate::hypervisor::surrogate_process_manager::{
+    get_surrogate_process_manager, surrogates_disabled,
+};
+use crate::hypervisor::virtual_machine::{
+    CreateVmError, MapMemoryError, RegisterError, ResetVcpuError, RunVcpuError, UnmapMemoryError,
+    VirtualMachine, VmExit,
+};
+use crate::hypervisor::wrappers::HandleWrapper;
+use crate::mem::memory_region::{MemoryRegion, MemoryRegionFlags, MemoryRegionType};
+#[cfg(feature = "trace_guest")]
+use crate::sandbox::trace::TraceContext as SandboxTraceContext;
+
+// ============================================================================
+// ARM64 WHP FFI bindings
+//
+// These are manually translated from WinHvPlatformDefs.h (Windows SDK 10.0.26100.0)
+// because the `windows` crate does not expose ARM64 WHP types.
+// ============================================================================
+
+const WHV_PARTITION_PROPERTY_CODE_ARM64_IC_PARAMETERS: WHV_PARTITION_PROPERTY_CODE =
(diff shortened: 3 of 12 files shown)


src/hyperlight_host/src/sandbox/snapshot/file/config.rs
```diff
@@ -132,6 +132,12 @@ impl CpuVendor {
             // `0x` prefix padded to width 4, e.g. Apple `0x61`, Arm `0x41`.
             Self(format!("{implementer:#04x}"))
         }
+        #[cfg(all(target_arch = "aarch64", target_os = "windows"))]
+        {
+            // Windows does not expose MIDR_EL1 directly; use a fixed vendor
+            // string. All current Windows ARM64 devices use Qualcomm/ARM cores.
```
syntactically: This is not true---the surface laptop ultra uses an NVidia SoC.

However, the reason why we differentiate Intel and AMD for x86-64 is that they have totally different virtualisation extensions (VT-x vs AMD-v), which have significantly different capabilities/interfaces. This is not the case on aarch64, so I don't necessarily think that we need to record the cpu implementor at all in that case. Instead, if we rely on specific FEAT_x or other capabilities, we can introduce specific feature bits for those capabilities in the future.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,850 @@
...
+                                ),
+                            })?;
+
+                        if port == VmAction::Halt as usize {
+                            return Ok(VmExit::Halt());
+                        } else {
+                            return Ok(VmExit::IoOut(
+                                port as u16,
+                                (off as u64).to_le_bytes().to_vec(),
+                            ));
```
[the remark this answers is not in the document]
cshung: Fixed in 1d3a2996. The MMIO path decodes ISV, SAS, SRT, and WnR from the syndrome, reads the source register, and returns only the written bytes.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,850 @@
...
+        let fpsr = self.get_reg64(WHV_ARM64_REGISTER_FPSR).map_err(|_| {
+            RegisterError::GetFpu(super::super::HypervisorError::WindowsError(
+                windows_result::Error::from_hresult(HRESULT(0)),
+            ))
+        })? as u32;
+        let fpcr = self.get_reg64(WHV_ARM64_REGISTER_FPCR).map_err(|_| {
+            RegisterError::GetFpu(super::super::HypervisorError::WindowsError(
+                windows_result::Error::from_hresult(HRESULT(0)),
+            ))
+        })? as u32;
```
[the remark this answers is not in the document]
cshung: Fixed in 05d53509. GetRegs errors are converted to GetFpu while preserving the original HypervisorError.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,850 @@
...
+                RegisterError::SetFpu(super::super::HypervisorError::WindowsError(
+                    windows_result::Error::from_hresult(HRESULT(0)),
+                ))
+            })?;
+        self.set_reg64(WHV_ARM64_REGISTER_FPCR, fpu.fpcr as u64)
+            .map_err(|_| {
+                RegisterError::SetFpu(super::super::HypervisorError::WindowsError(
+                    windows_result::Error::from_hresult(HRESULT(0)),
+                ))
+            })?;
```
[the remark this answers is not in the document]
cshung: Fixed in 05d53509. SetRegs errors are converted to SetFpu while preserving the original HypervisorError.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,850 @@
...
+    let module = unsafe { LoadLibraryA(s!("winhvplatform.dll"))? };
+    let proc = unsafe { GetProcAddress(module, s!("WHvMapGpaRange2")) };
+    match proc {
+        Some(f) => Ok(unsafe { std::mem::transmute(f) }),
+        None => {
+            unsafe {
+                windows::Win32::Foundation::FreeLibrary(module).ok();
+            }
+            Err(windows_result::Error::from_hresult(HRESULT(-1)))
+        }
```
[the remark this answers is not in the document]
cshung: Fixed in 05d53509. ARM64 uses LoadLibraryExA with LOAD_LIBRARY_SEARCH_DEFAULT_DIRS and reports ERROR_PROC_NOT_FOUND, matching the x64 backend.

src/hyperlight_host/src/sandbox/snapshot/file/config.rs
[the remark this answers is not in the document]
cshung: Fixed in 05d53509. The comment describes the architectural virtualization interface and makes no hardware-vendor claim.

cshung: Agreed. Fixed in 05d53509. Windows ARM64 snapshots use one platform token because the virtualization interface is architectural. Feature-specific compatibility can be represented separately when needed.

cshung:
End-to-end `cargo-hyperlight` validation completed on the ARM64 WHP hardware.

Using this PR's 0.16 crates, a freshly scaffolded sample successfully built `aarch64-hyperlight-none` guests and ran through WHP in debug and release. Host callbacks, typed calls, persistent state, snapshot, and restore passed. I also ran the release workflow 10 times with surrogate mappings and 10 times with surrogates disabled. All runs produced the expected output.

The backend works. The published scaffold is the remaining user-experience gap: `cargo-hyperlight` pins Hyperlight 0.15 and hardcodes the generated host's x64 guest path. Filed hyperlight-dev/cargo-hyperlight#70 with the reproduction and results.

src/hyperlight_host/src/hypervisor/hyperlight_vm/aarch64.rs
```diff
@@ -64,13 +71,19 @@ impl HyperlightVm {
         let vm: VmType = match get_available_hypervisor() {
             #[cfg(kvm)]
             Some(HypervisorType::Kvm) => Box::new(KvmVm::new().map_err(VmError::CreateVm)?),
-            // TODO: mshv support
             #[cfg(mshv3)]
-            Some(HypervisorType::Mshv) => return Err(CreateHyperlightVmError::NoHypervisorFound),
+            Some(HypervisorType::Mshv) => {
```
syntactically: This line looks like it's the same before/after?

src/hyperlight_host/src/hypervisor/hyperlight_vm/aarch64.rs
```diff
@@ -79,6 +92,15 @@ impl HyperlightVm {
             dropped: AtomicBool::new(false),
         });
 
+        #[cfg(target_os = "windows")]
+        let interrupt_handle: Arc<dyn InterruptHandleImpl> = Arc::new(WindowsInterruptHandle {
```
syntactically, line 93: Just a heads up that there is a big refactor/simplification to the interrupt handle machinery in #1674 that you may want to take a look at.

src/hyperlight_host/src/hypervisor/hyperlight_vm/aarch64.rs
```diff
@@ -214,7 +236,7 @@ impl HyperlightVm {
             self.vm_can_reset_vcpu,
             "No fallback path for vcpu reset on aarch64"
         );
-        self.vm.reset_vcpu()?;
+        self.interrupt_handle.reset_vcpu(self.vm.as_mut())?;
```
syntactically: What's this change about? It seems like the wrong place for this.

If it's related to the locking that happens in the interrupt handle: that's required because the interrupt handle can outlive the `WhpVm` object that it is created from (and the partition is destroyed when the latter is dropped). Here, you have a reference to the `WhpVm` on hand, so you can be sure that it hasn't been dropped (and then destroyed).

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+//! Windows Hypervisor Platform (WHP) backend for AArch64.
+//!
+//! This module provides the [`VirtualMachine`] trait implementation using the
+//! WHP APIs on Windows ARM64 systems. Because the `windows` crate does not yet
+//! expose ARM64 WHP structures, we define our own FFI bindings derived from
+//! the Windows SDK header `WinHvPlatformDefs.h` (10.0.26100.0).
```
syntactically, line 22: Is there an issue on the upstream `windows` crate about getting these exposed?

Also, I remember that Windows APIs in general often have a bunch of random __aligned__(16) or similar things---are those accounted for here? I remember that the bindgen that the windows crate uses used to get that wrong.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+//
+// These are manually translated from WinHvPlatformDefs.h (Windows SDK 10.0.26100.0)
+// because the `windows` crate does not expose ARM64 WHP types.
+// ============================================================================
+
+const WHV_PARTITION_PROPERTY_CODE_ARM64_IC_PARAMETERS: WHV_PARTITION_PROPERTY_CODE =
+    WHV_PARTITION_PROPERTY_CODE(0x00001012);
+const WHV_ARM64_REGISTER_GICR_BASE_GPA: WHV_REGISTER_NAME = WHV_REGISTER_NAME(0x00063000);
+
+#[repr(C)]
```
syntactically, line 60: Ouch, is it really a hard requirement from WHP to have the interrupt controller configured? I think I have done some experiments with mshv on Linux in the past and had success without an ic.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+        gic_ppi_performance_monitors_interrupt: 0x17,
+        reserved1: [0; 6],
+    },
+};
+
+const GICR_BASE_GPA: u64 = 0xeffee000;
+
+/// ARM64 WHP register name constants.
+/// Mapped from `WHV_REGISTER_NAME` enum values in the SDK header under `_ARM64_`.
+mod arm64_regs {
```
syntactically: For consistency with the KVM code, you could move this into `src/hypervisor/regs/aarch64/whp_reg.rs`?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+
+// ============================================================================
+// Surrogate process guard (same as x86_64 version)
+// ============================================================================
+
+use std::sync::atomic::AtomicBool as StdAtomicBool;
+
+/// RAII guard: when surrogates are disabled, only one no-surrogate WHP VM
+/// can exist at a time. This flag prevents a second one from being created.
+static NO_SURROGATE_VM_ACTIVE: StdAtomicBool = StdAtomicBool::new(false);
```
syntactically, line 220: While we are in here, and since it is a bit unpleasant to have to duplicate all this logic, how would you feel about #1581?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+        Ok(_) => unsafe { capability.HypervisorPresent.as_bool() },
+        Err(_) => {
+            tracing::info!("Windows Hypervisor Platform is not available on this system");
+            false
+        }
+    }
+}
+
+/// Helper: release a host-side file mapping view and its handle.
+fn release_file_mapping(view_base: *mut c_void, mapping_handle: HandleWrapper) {
```
syntactically: This also could be moved to the architecture-independent mod.rs so it isn't duplicated?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+                release_file_mapping(view, handle);
+            }
+        }
+
+        Ok(())
+    }
+
+    fn run_vcpu(
+        &mut self,
+        #[cfg(feature = "trace_guest")] tc: &mut SandboxTraceContext,
```
syntactically: Did you validate any of the `trace_guest` stuff? I would be surprised if that worked on aarch64 right now (#1615). If not, do you think it's worth keeping the code here because the eventual implementation will be very similar, or should it be removed for clarity?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+                            }
+                        })?;
+
+                    if port == VmAction::Halt as usize {
+                        Ok(VmExit::Halt())
+                    } else {
+                        let syndrome = exit_context.memory_access_syndrome();
+                        const ISV: u64 = 1 << 24;
+                        const WNR: u64 = 1 << 6;
+                        if syndrome & (ISV | WNR) != ISV | WNR {
```
syntactically, line 586: The hvf implementation in this case falls back to `VmExit::Mmio{Read,Write}` in this case instead of erroring. It should never be encountered in the current setup, but idk if it seems fatal enough to raise an error here. Up to you.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+                mgr.get_surrogate_process()
+                    .map_err(|e| CreateVmError::SurrogateProcess(e.to_string()))?,
+            );
+        }
+
+        Ok(vm)
+    }
+
+    /// Get a single 64-bit register value.
+    fn get_reg64(&self, name: WHV_REGISTER_NAME) -> Result<u64, RegisterError> {
```
syntactically: The code on other hvs has similar functions, but has them return the `HypervisorError` that is wrapped into a `RegisterError` at the next layer up callers, so that we don't have to do the "wrap as `GetRegs`, unwrap and rewrap as `GetFpu`" type of dance.

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+            })? as u32;
+
+        Ok(CommonFpu { v, fpsr, fpcr })
+    }
+
+    fn set_fpu(&self, fpu: &CommonFpu) -> Result<(), RegisterError> {
+        use arm64_regs::*;
+
+        for i in 0..32u32 {
+            self.set_reg128(qreg(i), fpu.v[i as usize])?;
```
syntactically: Is there a reason that this one isn't batched, when the others are?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
syntactically: Also, it seems like there is a mix of batched and un-batched users here but the underlying whp call is the same. Maybe it would make sense to restructure these to have one helper function that just gives a more Rust-y interface to that underlying batched API call, and then either have this a tiny special case wrapper (or not at all if it isn't needed---it seems like all but a single use sets/gets at least 2 regs and would be better-off with the batched call?)

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+        Ok(())
+    }
+
+    fn can_reset_vcpu(&self) -> bool {
+        true
+    }
+
+    fn reset_vcpu(&mut self) -> Result<(), ResetVcpuError> {
+        unsafe {
+            WHvDeleteVirtualProcessor(self.partition, 0)
```
syntactically: I think `WHvResetPartition` is also perhaps an option here?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/aarch64.rs
```diff
@@ -0,0 +1,950 @@
...
+                tracing::error!("Failed to delete virtual processor: {e:?}");
+            }
+            if let Err(e) = WHvDeletePartition(self.partition) {
+                tracing::error!("Failed to delete partition: {e:?}");
+            }
+        }
+    }
+}
+
+// ============================================================================
```
syntactically: This can also perhaps be shared with the x86_64 variant?

src/hyperlight_host/src/hypervisor/virtual_machine/whp/x86_64.rs
```diff
@@ -586,7 +593,11 @@ impl VirtualMachine for WhpVm {
                             if self.handle_hw_io_out(port, &data) {
                                 continue;
                             }
-                        } else if let Some(val) = super::x86_64::hw_interrupts::handle_io_in(port) {
+                        } else if let Some(val) =
+                            crate::hypervisor::virtual_machine::x86_64::hw_interrupts::handle_io_in(
```
syntactically, line 577: Perhaps `use crate::hypervisor::virtual_machine::x86_64` or something to reduce the formatting churn?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -67,6 +70,12 @@ pub(crate) trait InterruptHandleImpl: InterruptHandle {
     /// Clear the cancellation request flag
     fn clear_cancel(&self);
 
+    /// Reset the vCPU while honoring platform lifecycle synchronization.
+    #[cfg(target_arch = "aarch64")]
```
syntactically: As mentioned above, this seems like the wrong place for this.

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -1471,7 +1474,11 @@ fn interrupt_infinite_moving_loop_stress_test() {
     use std::sync::Arc;
     use std::thread;
 
-    // We have a high thread count to stress test and to have interesting interleavings
+    // ARM64 WHP partitions include GICv3 state, limiting practical partition
```
syntactically: Can you elaborate on this a little bit?

src/hyperlight_host/tests/integration_test.rs
```diff
@@ -935,6 +935,9 @@ fn interrupt_random_kill_stress_test() {
     use hyperlight_host::sandbox::snapshot::Snapshot;
     use tracing::{error, trace};
 
+    #[cfg(target_arch = "aarch64")]
+    const POOL_SIZE: usize = 64;
```
syntactically, line 942: Are you planning to run the tests in single-threaded mode in CI? I initially had a similar change in #1674, but I found that CI ran multiple tests at once and went past thread limits, so I [changed](https://github.com/hyperlight-dev/hyperlight/commit/082f17c4f093c8ca16e9aa4c1883b6f60d3b309b) that code to work correctly (if slowly) with any amount of concurrency.

syntactically (changes requested):
This looks great! I have a bunch of little nits, and also I (more majorly) think that the addition of `reset_vcpu` to `InteruptHandle` seems wrong.

Issue #1560, opened 2026-06-18
Redesign public hyperlight API


ludfjig:
This issue is meant to cover redesigning the public hyperlight API for a 1.0 release. It's separate from https://github.com/hyperlight-dev/hyperlight/issues/765, because this one is referring to redesigning the existing API, while #765 is concerned about removing public items that should not be public.

Some things I can think of right now:
- Remove Uninitiated sandbox and `evolve`
- Combine and/or `SandboxConfiguration` and `SandboxRuntimeConfig`
- Add builder-pattern where makes sense
- Rename `MultiUseSandbox` (name is outdated)
- Possibly add some light-weight type-safety to MultiUseSandbox to prevent host functions from being added (see https://github.com/hyperlight-dev/hyperlight/pull/1392)
- Narrow method-specific errors and error variants, rather than a catch-all HyperlightError

Please add more!

jprendes:
AFAICT, `SandboxRuntimeConfig` is not a public struct so it's not part of the public API.

syntactically:
> AFAICT, SandboxRuntimeConfig is not a public struct so it's not part of the public API.

I don't think that was intentional, if that is the case. @ludfjig ?

PR #1652, opened 2026-07-17
fix: use PathBuf instead of String for guest binary and crash dump paths

### LINKED ISSUE #1518: `GuestBinary::FilePath` takes path by String
Should be path or pathbuf

midsterx:
Currently, we do not allow non-UTF-8 paths for loading guest binaries and crash dumps. This PR replaces `String`-backed filesystem paths with `PathBuf`, thus allowing non-UTF-8 paths.

Resolves https://github.com/hyperlight-dev/hyperlight/issues/1518

### DIFF
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
@@ -16,8 +16,6 @@ limitations under the License.
 
 #[cfg(gdb)]
 use std::collections::HashMap;
-#[cfg(crashdump)]
-use std::path::Path;
 #[cfg(any(kvm, mshv3))]
 use std::sync::atomic::AtomicBool;
 use std::sync::atomic::AtomicU8;
@@ -600,13 +598,6 @@ impl HyperlightVm {
             regs[25] = sregs.fs.selector as u64; // fs
             regs[26] = sregs.gs.selector as u64; // gs
 
-            // Get the filename from the binary path
-            let filename = self.rt_cfg.binary_path.clone().and_then(|path| {
-                Path::new(&path)
-                    .file_name()
-                    .and_then(|name| name.to_os_string().into_string().ok())
-            });
-
             // Use the stored entry point address from the runtime config.
             // This is the original entry point (load_addr + ELF entry offset)
             // which GDB needs for AT_ENTRY to compute the PIE load offset.
@@ -631,7 +622,6 @@ impl HyperlightVm {
                 xsave.to_vec(),
                 initialise,
                 self.rt_cfg.binary_path.clone(),
-                filename,
             )))
         } else {
             Ok(None)
--- a/src/hyperlight_host/src/sandbox/uninitialized.rs
+++ b/src/hyperlight_host/src/sandbox/uninitialized.rs
@@ -16,7 +16,7 @@ limitations under the License.
 
 use std::fmt::Debug;
 use std::option::Option;
-use std::path::Path;
+use std::path::PathBuf;
 use std::sync::{Arc, Mutex};
 
 use tracing::{Span, instrument};
@@ -39,7 +39,7 @@ use crate::{MultiUseSandbox, Result, new_error};
 #[derive(Clone, Debug, Default)]
 pub(crate) struct SandboxRuntimeConfig {
     #[cfg(crashdump)]
-    pub(crate) binary_path: Option<String>,
+    pub(crate) binary_path: Option<PathBuf>,
     #[cfg(gdb)]
     pub(crate) debug_info: Option<super::config::DebugInfo>,
     #[cfg(crashdump)]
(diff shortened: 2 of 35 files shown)


src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -557,13 +555,6 @@ impl HyperlightVm {
             regs[25] = sregs.fs.selector as u64; // fs
             regs[26] = sregs.gs.selector as u64; // gs
 
-            // Get the filename from the binary path
-            let filename = self.rt_cfg.binary_path.clone().and_then(|path| {
```
jsturtevant, line 604: Why was this removed?

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -414,15 +410,32 @@ mod tests {
...
+    #[cfg(unix)]
+    #[test]
+    fn guest_binary_loads_from_non_utf8_path() {
+        use std::ffi::OsString;
+        use std::os::unix::ffi::OsStringExt;
+
+        let temp_dir = tempfile::tempdir().unwrap();
+        let guest_path = temp_dir
+            .path()
+            .join(OsString::from_vec(b"guest-\xff".to_vec()));
```
jsturtevant, line 428: Does this file exist?

midsterx: The file does not exist yet. I intended to create it for testing purposes, which would automatically get deleted (by TempDir's Drop) at the end of the test function’s scope.

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
midsterx: It was removed because the filename duplicated information already contained in the binary PathBuf.

Previously, we stored both binary and filename. Now that we are using PathBuf, I removed filename from being passed.

The filename is derived later in crashdump.rs, where the ELF metadata actually requires text:
```
  let filename = ctx
      .binary
      .as_deref()
      .and_then(|path| path.file_name())
      .map_or("<unknown>".to_string(), |name| {
          name.to_string_lossy().into_owned()
      });
```

jsturtevant: ah yes, makes sense. Thanks!

src/hyperlight_host/src/sandbox/uninitialized.rs
jsturtevant: Oh, I see you copy the simple guest to that file path.  I think that works 👍

jsturtevant:
LGTM @ludfjig for a quick look

ludfjig (dismissed):
LGTM. @jprendes for quick look :) The only concern I might have that we perhaps want to consider pairing this break with others at the same time to lessen pain for users (but not a huge deal, the update-path is straight forward)

jprendes (commented):
I think we could take `impl AsRef<Path>` wherever we take a a `String` or `&str` (or `Option<impl AsRef<Path>>`, so that we stay backwards compatible.

midsterx:
> I think we could take `impl AsRef<Path>` wherever we take a a `String` or `&str` (or `Option<impl AsRef<Path>>`, so that we stay backwards compatible.

AsRef<Path> works for function parameters, but not enum variant fields. generate_crashdump_to_dir already accepts strings through Into<PathBuf>. The remaining breaking change is GuestBinary::FilePath, where callers can migrate with GuestBinary::FilePath(path.into()).

From what I understand, we can get the backwards compatibility even as is. What am I missing?

jprendes (dismissed):
This looks good!
Thank you for working on this! :-)

Maybe we could have `GuestBinary` implement
```rust
impl<T: AsRef<std::path::Path>> From<T> for GuestBinary<'_> {
    fn from(path: T) -> Self {
        GuestBinary::FilePath(path.as_ref().to_path_buf())
    }
}
```

But we should leave that for a different PR.
We should consider this as part of #1560

jprendes:
Needs a rebase

midsterx:
> Needs a rebase

I did rebase it, and fixed a bunch of newly added tests (from the time I raised the PR). 
I think it needs an approval now.

jprendes:
@midsterx, there's been some further changes since your last push.
I can't merge until it's rebased again.

jprendes:
While you rebase you could also change the commit message with a one line explanation of the change?
You could use the title of this PR as the commit message.

jprendes (approved):

jprendes:
Sorry, you are right, it was just missing the approval

midsterx:
Thank you Jorge, I'll keep this in mind for the future PRs. Meanwhile, I have raised an issue which would potentially avoid folks from facing this problem - https://github.com/hyperlight-dev/hyperlight/issues/1708
Given that this is a settings change, you are probably the best person to make the change.

PR #1655, opened 2026-07-18
feat: enable ASLR for PIE guest binaries

cshung:
## Summary

Randomize the virtual base address for PIE guest code regions instead of using identity mapping. This provides address space layout randomization (ASLR) for PIE guests, making the code region virtual address unpredictable across sandbox instantiations.

## Changes

- **layout.rs**: \code_virt_base()\ now randomizes VA for PIE guests using \
and::rng()\, choosing a page-aligned address within 47-bit canonical user space. Conflict validation now applies to both PIE and non-PIE.
- **mgr.rs**: Thread \code_virt_base\ through \SandboxMemoryManager\ so it is preserved across snapshot/restore.
- **snapshot/mod.rs**: Store \code_virt_base\ in \Snapshot\, pass it to \xe_info.load()\ for correct relocation processing, and expose via \code_virt_base()\ accessor.
- **config.rs**: Relax entrypoint validation to accept non-identity-mapped virtual addresses (required for ASLR and non-PIE).
- **initialized_multi_use.rs**: \	race_guest\ tests now use \code_virt_base\ instead of assuming GVA == GPA.

## Design

- The random base is chosen from \[0x1000000, 0x7FFFFFFFFF * PAGE_SIZE - code_size)\, ensuring it is above all identity-mapped layout regions and within 47-bit canonical user space.
- Non-PIE binaries continue to use their declared ELF base VA (no change).
- Conflict validation ensures the randomized code region does not overlap other memory regions.

## Testing

- All unit tests pass (300 passed, 6 ignored)
- All integration tests pass (39 passed, 1 ignored)
- Clippy clean (debug + release)

## Dependencies

This PR is based on #1530 (non-PIE ELF loading support).


ludfjig:
How does this interact with snapshots and restoring snapshots into existing sandboxes? Is it transparent?

syntactically:
> How does this interact with snapshots and restoring snapshots into existing sandboxes? Is it transparent?

I believe from discussions we had in the past that this only does randomisation at the time of the elf -> initial snapshot creation, as it's quite difficult to do after that without some way to walk the heap, which is one reason I'm not sure whether it makes sense to move forward with this.

cshung:
The current PR is for testing purposes: to verify the code correctly distinguishes GVA from GPA throughout, rather than to ship ASLR to production.

PR #1667, opened 2026-07-20
Reseed libc rng on snapshot-restore

### LINKED ISSUE #1745: Reseed randomness on Sandbox::from_snapshot(&snapshot) and Sandbox::restore(&snapshot)

ludfjig:
Restoring snapshot now unconditionally reseeds libc rand() and random() to prevent multiple snapshots from sharing prng state. 

For simplicity sake, decided against adding new config or policy regarding this, since this PR will fix the immediate issue. If we in the future want host to modify it, we can add it later. Users can also add guest functions to call srand manually if they wish.

Perf: adds unconditional ~7ns to restore path on host side on my local machine. We could store a cached ThreadRng handle on sandbox, which brings this down to ~2ns (avoid TLS lookup), but I didn't think this was worth it as it's already so fast, and most users don't care about randomness. On guest side, `srand()` is just 1 direct memory store (negligible, a couple ns at most)

Compatibility: old snapshot load fine but the guests will ignore the host's request to reseed. This is fine as #1674 will be a breaking change anyway. 

closes #1745

### DIFF
--- a/src/hyperlight_guest_bin/src/lib.rs
+++ b/src/hyperlight_guest_bin/src/lib.rs
@@ -210,6 +210,23 @@ unsafe extern "C" {
     fn srand(seed: u32);
 }
 
+#[cfg(feature = "libc")]
+pub(crate) fn refresh_libc_rng() {
+    let seed_ptr = hyperlight_guest::layout::libc_rng_seed_gva();
+    // SAFETY: The host maps this aligned u64 scratch slot for the guest's
+    // lifetime and writes it only while the guest is stopped.
+    let request = unsafe { seed_ptr.read_volatile() };
+    if request >> 32 != 0 {
+        // SAFETY: The scratch slot has the validity and exclusivity described
+        // above. The libc feature provides srand with a u32 seed.
+        unsafe {
+            srand(request as u32);
+            // clear request u32 and zero u32 seed
+            seed_ptr.write_volatile(0u64);
+        }
+    }
+}
+
 #[tracing::instrument(skip_all, parent = tracing::Span::current(), level= "Trace")]
 extern "C" fn hyperlight_main_default() {
     // no-op
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 * Fix symbol resolution in guest core dumps for sandboxes created from snapshots by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/1618
 * Reject malformed OCI snapshot metadata and non-regular artifact files during load.
 * Reset XCR0 during x86 snapshot restore.
+* Reseed guest libc `rand()` and `random()` after restoring a snapshot to avoid multiple sandboxes sharing PRNG state.
 
 ## [v0.16.0] - 2026-06-26
 
--- a/src/hyperlight_common/src/layout.rs
+++ b/src/hyperlight_common/src/layout.rs
@@ -27,6 +27,7 @@ pub const SCRATCH_TOP_SIZE_OFFSET: u64 = 0x08;
 pub const SCRATCH_TOP_ALLOCATOR_OFFSET: u64 = 0x10;
 pub const SCRATCH_TOP_SNAPSHOT_PT_GPA_BASE_OFFSET: u64 = 0x18;
 pub const SCRATCH_TOP_SNAPSHOT_GENERATION_OFFSET: u64 = 0x20;
+pub const SCRATCH_TOP_LIBC_RNG_SEED_OFFSET: u64 = 0x28;
 pub const SCRATCH_TOP_EXN_STACK_OFFSET: u64 = 0x30;
 
 pub fn scratch_base_gpa(size: usize) -> u64 {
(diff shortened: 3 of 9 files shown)


syntactically:
> However if somebody passes Refresh with an old snapshot it will be silently ignored

I haven't looked at the implementation, but from the description this concerns me a bit. It also feels to me like "whether I need access to actual randomness" is more a semantic property of the guest being run than it is a property of the sandbox, so I wonder if it would make more sense to make it a (persisted) property of the snapshot that the sandbox honours. Snapshots which don't set the rng-needed flag could leave it out of their manifest for backwards compatibility.

I also wonder whether or not there is really any benefit to the current rng initialisation that is shared between everyone. I could imagine that rather than looking at it as "do I need randomness re-initialised on snapshot restore or not", which might be somewhat opaque to users, the semantically simpler thing to consider, which is basically identical, is "do I need randomness at all or not". If we recast the API from that perspective, we could change to having a snapshot-level flag that switches between "not initialising randomness at all (/ideally also erroring if you try to use it)" and "initialising randomness on every restore".

ludfjig:
>> However if somebody passes Refresh with an old snapshot it will be silently ignored

>I haven't looked at the implementation, but from the description this concerns me a bit. It also feels to me like "whether I need access to actual randomness" is more a semantic property of the guest being run than it is a property of the sandbox, so I wonder if it would make more sense to make it a (persisted) property of the snapshot that the sandbox honours. Snapshots which don't set the rng-needed flag could leave it out of their manifest for backwards compatibility.

Agreed, I made it property of the snapshot now. And made it a break change which should be fine given we get this in before release

> I also wonder whether or not there is really any benefit to the current rng initialisation that is shared between everyone. I could imagine that rather than looking at it as "do I need randomness re-initialised on snapshot restore or not", which might be somewhat opaque to users, the semantically simpler thing to consider, which is basically identical, is "do I need randomness at all or not". If we recast the API from that perspective, we could change to having a snapshot-level flag that switches between "not initialising randomness at all (/ideally also erroring if you try to use it)" and "initialising randomness on every restore".

This would be nice, but how does this cover same/different rng for multiple sandboxes from the same snapshot? From each guest's perspective, they both could get randomness, but it might be the same as the other sandbox's

syntactically:
> This would be nice, but how does this cover same/different rng for multiple sandboxes from the same snapshot? From each guest's perspective, they both could get randomness, but it might be the same as the other sandbox's

My instinct would be to say that if you need randomness at all you should get reseeded randomness, and if you don't need it all you don't get anything ofc. So there is no case where 2 restores of the same snapshot can call rand() and think it works and get the same value. I don't think there is probably almost any case where you want that. Maybe if you do have randomness enabled, you could have host API to mock it with a fixed seed, but idk if that's useful or not. Either way, that would be a decision of the host.

 It seems like "do I need randomness" is a property of the code-being-run-in-the-sandbox, and changes its ABI to require a randomness reseed on snapshot restore, whereas "do I want to mock the randomness for a test" is a property of the host code, and does not change the ABI (it just means a fixed value is used on that reseed).

ludfjig:
> > This would be nice, but how does this cover same/different rng for multiple sandboxes from the same snapshot? From each guest's perspective, they both could get randomness, but it might be the same as the other sandbox's
> 
> My instinct would be to say that if you need randomness at all you should get reseeded randomness, and if you don't need it all you don't get anything ofc. So there is no case where 2 restores of the same snapshot can call rand() and think it works and get the same value. I don't think there is probably almost any case where you want that. Maybe if you do have randomness enabled, you could have host API to mock it with a fixed seed, but idk if that's useful or not. Either way, that would be a decision of the host.
> 
> It seems like "do I need randomness" is a property of the code-being-run-in-the-sandbox, and changes its ABI to require a randomness reseed on snapshot restore, whereas "do I want to mock the randomness for a test" is a property of the host code, and does not change the ABI (it just means a fixed value is used on that reseed).

Simplified this PR to its core essence, please see new updated PR description.

src/hyperlight_guest_bin/src/lib.rs
```diff
@@ -210,6 +210,23 @@ unsafe extern "C" {
     fn srand(seed: u32);
 }
 
+#[cfg(feature = "libc")]
+pub(crate) fn refresh_libc_rng() {
+    let seed_ptr = hyperlight_guest::layout::libc_rng_seed_gva();
+    // SAFETY: The host maps this aligned u64 scratch slot for the guest's
+    // lifetime and writes it only while the guest is stopped.
+    let request = unsafe { seed_ptr.read_volatile() };
```
jprendes, line 218: can we use `Option<u32>` instead of `u64`?
something like:
```
   let seed_ptr = X.as_ptr() as *mut Option<u32>;
   // SAFETY: The host maps this aligned Option<u32> scratch slot for the guest's
   // lifetime and writes it only while the guest is stopped.
   let seed = unsafe { seed_ptr.read_volatile() };
   if let Some(seed) = seed {
        unsafe { srand(seed) };
        unsafe { seed_ptr.write_volatile(None) };
   }
```

jprendes: hm, I see that this would ties us to a stable binary representation of an Option<u32> between the guest and the host, which might not be true.

jprendes (approved):
LGTM

PR #1674, opened 2026-07-21
Add initial single-sandbox-executing-at-a-time support for Hypervisor.framework

syntactically:


src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -28,6 +28,9 @@ use crate::mem::memory_region::MemoryRegion;
 #[cfg(feature = "trace_guest")]
 use crate::sandbox::trace::TraceContext as SandboxTraceContext;
 
+/// Hypervisor.framework functionality (MacOS)
+#[cfg(hvf)]
+pub(crate) mod hvf;
```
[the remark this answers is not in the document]
syntactically: There is no intention to support non-aarch64 hvf builds

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -77,6 +80,12 @@ pub fn get_available_hypervisor() -> &'static Option<HypervisorType> {
                 } else {
                     None
                 }
+            } else if #[cfg(hvf)] {
+                if hvf::is_hypervisor_present() {
+                    Some(HypervisorType::Hvf)
+                } else {
+                    None
+                }
```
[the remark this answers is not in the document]
syntactically: There is no intention to support non-aarch64 hvf builds

src/hyperlight_host/src/hypervisor/virtual_machine/mod.rs
```diff
@@ -102,6 +111,9 @@ pub(crate) enum HypervisorType {
 
     #[cfg(target_os = "windows")]
     Whp,
+
+    #[cfg(hvf)]
+    Hvf,
```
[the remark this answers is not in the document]
syntactically: There is no intention to support non-aarch64 hvf builds

siennathesane:
I submitted #1681, which has full surrogate support for parallel VMs. Would you be able to take a look at that before merging this?

(Also, your copyright is out of date in several files)

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
```diff
@@ -789,7 +759,7 @@ pub(super) mod debug {
 
                 let read_len = std::cmp::min(
                     data.len(),
-                    (PAGE_SIZE - (gpa & (PAGE_SIZE - 1))).try_into().unwrap(),
+                    page_size::get() - (gpa as usize & (page_size::get() - 1)),
```
ludfjig, line 807: I assume `page_size` is cached?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -654,7 +666,7 @@ impl SandboxMemoryLayout {
 impl SandboxMemoryLayout {
     /// Offset of the PEB struct within the snapshot region.
     pub(crate) fn peb_offset(&self) -> usize {
-        self.code_size.next_multiple_of(PAGE_SIZE_USIZE)
+        self.code_size.next_multiple_of(PAGE_SIZE)
```
ludfjig, line 669: My first instinct was why not use `page_size::get()` here then I realize it's the guest's page size. Do you think we should rename variable to indicate this?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -2767,7 +2763,12 @@ mod tests {
             #[cfg(unix)]
             {
                 use std::os::unix::process::ExitStatusExt;
-                status.signal() == Some(libc::SIGSEGV)
+                let expected_signal = if cfg!(target_os = "macos") {
```
ludfjig, line 2853: maybe stale comment on top should be updated

src/hyperlight_host/build.rs
```diff
@@ -101,13 +102,64 @@ fn main() -> Result<()> {
...
+            .header(format!(
+                "{manifest_dir}/src/hypervisor/virtual_machine/hvf/bindings.h"
+            ))
+            .generate()
+            .context("Unable to generate hvf bindings")?
+            .write_to_file(out_path.join("hvf_bindings.rs"))
+            .context("Couldn't write hvf bindings")?;
+
+        cc::Build::new()
+            .opt_level(3)
```
ludfjig, line 155: should we unconditionally do opt level 3?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -727,14 +732,16 @@ mod tests {
         CommonSpecialRegisters::default()
     }
 
-    const SIMPLE_PT_BASE: usize = PAGE_SIZE + SandboxMemoryLayout::BASE_ADDRESS;
+    fn simple_pt_base() -> usize {
+        page_size::get() + SandboxMemoryLayout::BASE_ADDRESS
+    }
```
ludfjig, line 754: What is the initial chunk of page size for? Isn't this already included in base_address?

src/hyperlight_host/src/sandbox/snapshot/mod.rs
```diff
@@ -550,6 +550,11 @@ impl Snapshot {
                 }
                 pt_buf.set_root(pt_buf.initial_root());
 
+                snapshot_memory.resize(
+                    snapshot_memory.len().next_multiple_of(page_size::get()),
+                    0u8,
+                );
+
```
ludfjig, line 566: Why this change?

ludfjig (commented):
Looks really good. I haven't reviewed the changes to InterruptHandle yet. Btw do you know if `just snapshot-goldens-pull` will work on mac once golden snapshto images areadded? I haven't tested it but I think there mgiht some hardcoded stuff there

src/hyperlight_host/src/hypervisor/hyperlight_vm/x86_64.rs
syntactically: Yes, it's a `OnceLock` internally

src/hyperlight_host/src/mem/layout.rs
syntactically: The import path is in `hyperlight_common::vmem` which is where all the other stuff that is related to the guest memory translation stuff lives, which I think is disambiguating enough? We have a tonne of other constants in there that we would probably want to rename too if we changed this one, and I think it would get a bit clunky.

src/hyperlight_host/build.rs
syntactically: It's a tiny file with ABI wrapper functions, so I think unconditionally optimising is fine.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
syntactically: This is matching what the test was doing before, I think; IIRC the tests here lay out the guest memory with 1 page of real data/code and then n pages of page tables.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,938 @@
...
+}
+
+impl HvfCpu {
+    /// Must be called only once per thread.
+    fn new() -> Result<Self, HypervisorError> {
+        use core::mem::MaybeUninit;
+        let mut vcpu: MaybeUninit<bindings::hv_vcpu_t> = MaybeUninit::zeroed();
+        let mut exit: *mut bindings::hv_vcpu_exit_t = core::ptr::null_mut();
+        unsafe {
+            let config = bindings::hv_vcpu_config_create();
```
jsturtevant, line 220: do we need to release this when done since these are FFI?

https://developer.apple.com/documentation/hypervisor/hv_vcpu_config_create()#Discussion

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,938 @@
...
+                            if ipa >= io_page_gpa
+                                && let off = (ipa - io_page_gpa) as usize
+                                && off < hyperlight_common::vmem::PAGE_SIZE
+                            {
+                                let port = off / core::mem::size_of::<u64>();
+                                if port == VmAction::Halt as usize {
+                                    VmExit::Halt()
+                                } else {
+                                    let srt = bits::<20, 16>(esr);
+                                    let data = self.get_vcpu_regs()?.x[srt as usize];
```
jsturtevant: this looks like `srt` could be `11111`(31) and this indexes into https://github.com/hyperlight-dev/hyperlight/blob/f45ee4bc0d41ea7006fcf266449896abe7863fe6/src/hyperlight_host/src/hypervisor/regs/aarch64/common_regs.rs#L19 which could cause a panic?

jsturtevant (commented):
Amazing work! We should be able to run this on CI now?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,938 @@
...
+                            }
+                        } else {
+                            VmExit::MmioRead(ipa)
+                        }
+                    } else {
+                        unknown_exit
+                    }
+                } else {
+                    unknown_exit
+                }
```
jsturtevant, line 689: would be it be possible to flip these if statements to exit early instead of nesting and make it easier to read?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,938 @@
...
+                    HvfCpu::new()
+                        .map_err(HvfSyncError::CreateVcpu)
+                        .map_err(RunVcpuError::HvfSync)?,
+                ),
+                Some(v) => v,
+            };
+            self.interrupt_handle.set_vcpu(vcpu.id);
+            vcpu.sync_state_from_vm(self)
+                .map_err(RunVcpuError::HvfSync)?;
+            // TODO: replace unwrap()
```
jsturtevant: looks like this was taken care of and comment needs to be removed?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,938 @@
...
+    }
+
+    #[cfg(target_arch = "aarch64")]
+    fn can_reset_vcpu(&self) -> bool {
+        true
+    }
+
+    #[cfg(target_arch = "aarch64")]
+    fn reset_vcpu(&mut self) -> std::result::Result<(), ResetVcpuError> {
+        self.id.epoch += 1;
```
jsturtevant, line 1245: out of curiosity how does this reset the vcpu?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -96,25 +178,71 @@ pub trait InterruptHandle: Send + Sync + Debug {
...
+
 #[cfg(any(kvm, mshv3))]
 #[derive(Debug)]
-pub(super) struct LinuxInterruptHandle {
-    /// Atomic value packing vcpu execution state.
-    ///
-    /// Bit layout:
-    /// - Bit 2: DEBUG_INTERRUPT_BIT - set when debugger interrupt is requested
-    /// - Bit 1: RUNNING_BIT - set when vcpu is actively running
-    /// - Bit 0: CANCEL_BIT - set when cancellation has been requested
```
jsturtevant, line 113: did we loose this doc?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: Thanks, fixed in d12c2b1c1949d6066a7afc2efc5d240c830b4d7a.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: Good catch! 31 would usually mean the zero register in this context (though for some instructions it can be the stack pointer). We don't do anything that could make that happen in hyperlight-guest, so I've changed it to fall back to reporting it as a non-hyperlight fault in fe60ed4372efd32fba3899e347adefdb3ba65d84.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: Even though there is only one syndrome that we bother to decode more fully right now, I don't think I like the idea of giving structural primacy to that code path in that way, because it's not really privileged as the "right path"---there's no reason why we mightn't switch to using a mix of different exits if we thought it was more performant or easier for some reason. If we were to add another case for parsing ec, right now that would look like just going from
```rust
  if ec == ... {
    ...
  } else {
    unknown_exit
  }
```
to
```rust
  if ec == ... {
    ...
  } else if ec == ... {
    ...
  } else {
    unknown_exit
  }
```
which is the correct idea (we're just adding a 2nd case to the 1 we already supported), whereas going to that from
```rust
  if ec != ... { return unknown_exit; }
  ...
```
feels like you are totally changing the character of the logic.

However, I did realise that we already have some code parsing essentially the same thing in hyperlight-guest-bin in the exception handler, which parses into some nice enum structures which are easy and clear to destructure on. So, I've moved that into hyperlight-common (b38d5cdaf1cfcaadb797ae8f0fb077406b07d373), and simplified this decode logic a little bit using it (fe60ed4372efd32fba3899e347adefdb3ba65d84).

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: Thanks, fixed in 0d018f486c506150a46ae3906699e2b98e6044b1.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: It means that when we `sync_state_from` the sandbox to the vcpu that will actually be used for executing the sandbox, even if that vcpu previously executed the same sandbox, it will see the CPU-state-epoch mismatch and call `reset_vcpu()`.

src/hyperlight_host/src/hypervisor/mod.rs
syntactically: It seemed to me that this was basically a duplicate of the constant names above, probably because the same pattern was being used in so many places. Since this refactor pushes the management of these actual bits all into one smaller place where the constants are right there, I thought it was likely no longer necessary. I can add it back if you prefer, though.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: Actually, with some more changes, we now do run into this in release builds which are optimised to use xzr, so fixed in fa1367fadaee27c524e4e6f4f71341fe0ea4988c.

src/hyperlight_host/src/sandbox/snapshot/mod.rs
syntactically: Needed to make sure that the visible-to-guest-when-pt-tail-is-unmapped portion is page aligned.

syntactically:
After running into some problems with tests failing in CI due to too many test threads ending up breaking the 64 vcpu limit, I made a change in 082f17c4f093c8ca16e9aa4c1883b6f60d3b309b to try to avoid breaking that limit automatically.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
jsturtevant: this is more a code style thing so I am fine with it if we have solid reasoning. I just have a knee jerk reaction to many nested if statements :-)

jsturtevant (dismissed):

src/hyperlight_common/src/arch/aarch64/exn.rs
```diff
@@ -0,0 +1,105 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+ */
+
+/// Utility function to extract an (inclusive on both ends) bit range
+/// from a quadword.
+#[inline(always)]
+fn bits<const HIGH_BIT: u8, const LOW_BIT: u8>(x: u64) -> u64 {
+    (x & ((1 << (HIGH_BIT + 1)) - 1)) >> LOW_BIT
+}
```
ludfjig: this duplicate already exists in hyperlight_common/src/vmem.rs, should we reuse it?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -96,25 +178,71 @@ pub trait InterruptHandle: Send + Sync + Debug {
...
-    ///
-    /// CANCEL_BIT persists across vcpu exits/re-entries within a single `VirtualCPU::run()` call
-    /// (e.g., during host function calls), but is cleared at the start of each new `VirtualCPU::run()` call.
-    state: AtomicU8,
```
ludfjig, line 117: did you intend to remove these comments?

src/hyperlight_host/src/hypervisor/mod.rs
```diff
@@ -125,67 +253,42 @@ pub(super) struct LinuxInterruptHandle {
...
+                tid: AtomicU64::new(unsafe { libc::pthread_self() as u64 }),
+                #[cfg(not(all(
+                    target_arch = "x86_64",
+                    target_vendor = "unknown",
+                    target_os = "linux",
+                    target_env = "musl"
+                )))]
+                tid: AtomicU64::new(unsafe { libc::pthread_self() as u64 }),
+                sig_rt_min_offset: config.get_interrupt_vcpu_sigrtmin_offset(),
+                dropped: AtomicBool::new(false),
```
ludfjig, line 271: these cfg'd blocks looks identical

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,1038 @@
...
+    pub(crate) fn new(
+        interrupt_handle: Arc<dyn InterruptHandleImpl>,
+    ) -> Result<Self, CreateVmError> {
+        static NEXT_AVAILABLE_ID: AtomicU64 = AtomicU64::new(0);
+        // If the vm for this process has not yet been created, create it
+        let mut created = HV_VM_CREATED.lock();
+        if !*created {
+            unsafe {
+                let cfg = bindings::hv_vm_config_create();
+                let ret = bindings::hv_vm_create(cfg);
```
ludfjig, line 738: do you know why this states that cfg must be nil? https://developer.apple.com/documentation/hypervisor/hv_vm_create(_:)?language=objc#parameters

CHANGELOG.md
```diff
@@ -13,6 +13,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   resets to a clean default. On KVM the guest may only read or write declared
   MSRs, on MSHV and WHP this is not enforced. by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/991
 
+Certain fixed guest addresses were changed on AArch64 to more easily
+accommodate 16k pages without wasting memory. Snapshots taken from
+sandboxes using the old addresses will not be loadable by new
+hyperlight versions.
+
```
ludfjig, line 21: To me it sounds like only old aarch64 snapshot are no longer loadable which might be a bit confusing. Also should we make this a bullet?

Justfile
```diff
@@ -622,6 +622,7 @@ snapshot-goldens-pull target=default-target which="all" image=default-snapshot-g
     if [[ -e /dev/mshv ]]; then hv=mshv
     elif [[ -e /dev/kvm ]]; then hv=kvm
     elif [[ "${OS:-}" == "Windows_NT" ]]; then hv=whp
+    elif [[ "$(sysctl -n kern.hv.supported)" == "1" ]]; then hv=hvf
```
ludfjig: should this be kern.hv_support?

ludfjig (changes requested):
LGTM with some nits (maybe)

src/hyperlight_host/src/hypervisor/mod.rs
syntactically: I did, because I think these were basically just the same as the constants and really there because the uses were far away from the constants and spread around; I think now it is much more self-evident what is going on, but I am open to adding them back if you feel strongly.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: I'm not sure why the documentation says that; I don't think it's correct (anymore?) since one definitely needs to pass an object here for some features (e.g. to use nested virtualisation you need to call `hv_vm_config_set_el2_enabled` on the `config`).

CHANGELOG.md
syntactically: The aarch64 ones are the only ones that actually changed in layout, so I think this is more accurate.  (I bumped the global version because it seems difficult with the current infrastructure to make the change arch-specific, and as far as I know nobody is depending on snapshot compatibility yet (since we don't really have any guest ABI compatibility guarantees pre-1.0))

Justfile
syntactically: I have always used them interchangeably, but good point about consistency.

src/hyperlight_host/src/hypervisor/mod.rs
syntactically: Hm, I carried over the distinction from the old code, where non-musl branch did not have the `as u64`, I assume due to old clippy warnings? I didn't get the warning on this, though, so I guess I can just get rid of the distinction totally. Thanks!

src/hyperlight_common/src/arch/aarch64/exn.rs
syntactically: Fixed, thanks. It would be nice to move this somewhere more neutral since it's not vmem related at all, but I think that is for later.

syntactically:
@ludfjig @jsturtevant  There are a few substantive changes in the last few commits (starting from 83b696a872ac8b37f861ff343778b58c945e69da Convert mem::shared_mem module to use thiserror structured errors), so please take a quick skim if you have a chance.

src/hyperlight_host/src/hypervisor/mod.rs
ludfjig: sounds good I dont' feel strongly

CHANGELOG.md
ludfjig: ah ok

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,1248 @@
...
+    }
+
+    fn map_memory(&mut self, region: (u32, &MemoryRegion)) -> Result<(), HypervisorError> {
+        let slot = region.0 as usize;
+        if self.mappings.len() <= slot {
+            self.mappings.resize(slot + 1, None);
+        }
+        self.mappings[slot] = Some(region.1.clone());
+        self.opportunistically_sync_slot(slot)
+    }
```
ludfjig, line 1085: Should we do `self.opportunistically_sync_slot(slot)` before updating `self.mappings` in case it fails? Same for `fn unmap_memory`. Or do some roll back if it fails

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,1248 @@
...
+//! process by using nested virtualisation on platforms where it is
+//! available, making each sandbox a nested guest VM. This does,
+//! unfortunately, of course have some performance implications.
+//!
+//! ## Supporting [`core::marker::Send`] on Sandboxes
+//!
+//! The Hyperlight public API constraints sandboxes (and, by
+//! extension, hypervisor API implementations) to implement
+//! [`core::marker::Send`]. Other hypervisors have one vCPU but allow
+//! the vCPU handle to be mgirated across threads, although this comes
```
ludfjig, line 36: nit: typo `mgirated`

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,1248 @@
...
+    /// loaded memory space.
+    ///
+    /// TODO[before push]: we should probably do the optimisation of
+    /// updating this to be a Vec and setting it when
+    /// creating/destroying vcpus.
+    valid_vcpu: Option<bindings::hv_vcpu_t>,
+    /// For the memory region with a TLBI in it used in tlbi_vmalle1
+    /// below, we really want the semantics of a global
+    /// [`OnceLock`]. Unfortunately, [`OnceLock::get_or_try_init`] is
+    /// not stable, so we can'} use it, so we would have to use an
```
ludfjig, line 827: nit: typo `can'}`

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,1248 @@
...
+impl MemorySpace {
+    fn new() -> Self {
+        static NEXT_AVAILABLE_ID: AtomicU64 = AtomicU64::new(0);
+        Self {
+            id: NEXT_AVAILABLE_ID.fetch_add(1, Ordering::Relaxed),
+            mappings: Vec::new(),
+        }
+    }
+
+    /// This function is used both as a performance optimisation e(when
```
ludfjig, line 1030: nit: typo `e(when`

ludfjig (dismissed):
I like the typed error changes. Was there something particular that made you change it?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -641,29 +767,7 @@ impl ExclusiveSharedMemory {
...
-
-        // usize and isize are guaranteed to be the same size, and
-        // isize::MAX should be positive, so this cast should be safe.
-        if total_size > isize::MAX as usize {
-            return Err(HyperlightError::MemoryRequestTooBig(
-                total_size,
-                isize::MAX as usize,
-            ));
-        }
+        let total_size = Self::total_size(min_size_bytes);
```
jsturtevant, line 770: this looks like its missing the `?` for `total_size`

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1655,9 +1728,14 @@ impl ReadonlySharedMemory {
     fn map_file(file: &std::fs::File, len: usize) -> Result<Arc<HostMapping>> {
         use std::os::windows::io::AsRawHandle;
 
-        let total_size = len.checked_add(2 * PAGE_SIZE_USIZE).ok_or_else(|| {
-            new_error!("Memory required for file-backed mapping exceeded usize::MAX")
-        })?;
+        let total_size =
+            len.checked_add(2 * page_size::get())
+                .ok_or(SharedMemoryError::MemoryRequest(
+                    total_size,
```
jsturtevant, line 1734: does this need to be intialized?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -811,36 +879,147 @@ impl LoadedMemorySpace {
...
+                };
+
+                // The access/unwrap can't panic, since we just set it above.
+                Self::do_unmap(self.mappings[0].as_ref().unwrap()).itme()?;
+                self.mappings = Vec::new();
+                Ok(())
+            };
+
+            // Even if the actual tlbi failed or was cancelled,
+            // unconditionally try to restore the state that we saved
```
jsturtevant, line 958: above we call `return` in the result block direction above this, which would by pass this unconditional?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
```diff
@@ -0,0 +1,1248 @@
...
+    }
+
+    fn reset_vcpu(&mut self) -> Result<(), HypervisorError> {
+        // TODO: figure out if there is a more efficient way to clear
+        // all the state
+        if !self.destroyed_in_reset_vcpu {
+            unsafe { bindings::hv_vcpu_destroy(self.id) }.is_success()?;
+            self.destroyed_in_reset_vcpu = true;
+        }
+        *self = HvfCpu::new_unconditional()?;
```
jsturtevant, line 271: Is it ok to leave the cpu around on this call failing? I guess that would signal that the reset failed?

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: I think it is OK to be in a not super consistent state here if you this call fails; I don't think we make any guarantee about what the state is in case of failure on other backends

src/hyperlight_host/src/mem/shared_mem.rs
syntactically: Yeah, sorry, there are a variety of stupid bugs in the windows and/or x86 code because my local machine for testing them was down. Will be fixed soon

src/hyperlight_host/src/mem/shared_mem.rs
syntactically: Same as above

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: Very good catch, thanks, I missed that when doing some cleanup/inlining.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
syntactically: I think it is probably fine. We have already set self.destroyed_in_reset_vcpu so we don't need to worry about the Drop impl trying to re-destroy. The only issue is if we somehow get someone trying to use the same cpu for something without calling reset again later, but we have not yet remedied whatever caused the reset to happen the last time, so it should just happen again I think? I also do not know that there are many errors that are likely to be temporary / worth retrying that could materialise here at all.

syntactically:
> I like the typed error changes. Was there something particular that made you change it?

The VM errors are already typed errors, and now the `HvfSyncError` needs to be able to include an error from shared-memory. So, unless we want to break the original plan for transition and let the already-converted subsystem reference an anyhow error, I had to convert the shared memory subsystem that is now a dependency of the VM errors.

syntactically:
@jsturtevant  @ludfjig addressed your comments. I pushed granular commits at 6c6a304fea60cec1eb1cbda24315395054ad142e, but also pushed again the result of --autosquash so that we won't have to wait for CI + re-approval if automatically dismissed if this is approved with no changes.

src/hyperlight_host/src/hypervisor/virtual_machine/hvf/mod.rs
jsturtevant: > I also do not know that there are many errors that are likely to be temporary / worth retrying that could materialise here at all.

This was my thought too.

ludfjig (dismissed):

jsturtevant (dismissed):

syntactically:
Since I had to rebase this anyway, I did make a tiny change to include ludvig's comment. The fixup commit as at 692514e6, but, like before, I also pushed the result of autosquashing so that we can merge this after CI runs if everyone is happy.

ludfjig (approved):
🚀

PR #1681, opened 2026-07-22
feat: macOS support via Hypervisor.framework (aarch64)

siennathesane:
## macOS support via Hypervisor.framework (aarch64)

Adds a new `hvf` backend for Apple silicon, with multi-sandbox support via per-sandbox surrogate processes.

### Backend

`HvfVm` implements the `VirtualMachine` trait over a new shared `hyperlight-hvf` crate (vCPU lifecycle, register access, ESR-decoded exits: IO-page MMIO writes → `IoOut`/`Halt`, aborts → `MmioRead`/`MmioWrite`, WFI skipped, vtimer ignored). Cancellation uses `hv_vcpus_exit`, the only HVF call allowed from a non-owning thread. Requires macOS 26 (4 KiB IPA granule; a clear `IpaGranuleUnsupported` error is raised on older systems) and the `com.apple.security.hypervisor` entitlement (see `dev/hvf-entitlements.plist`).

### Surrogate architecture

HVF allows only one VM per process and has no cross-process mapping API, so each sandbox's VM lives in a pooled `hvf_surrogate` helper process. Unlike the WHP surrogate (an empty suspended address-space container), this is a real server: vCPU thread + reader/writer threads speaking a small JSON-over-unix-socket protocol (`hyperlight_hvf::proto`). Guest memory is shared by named POSIX shm objects, never copied. The surrogate binary is nested-built by `hyperlight-host`'s build script, embedded via rust-embed, and ad-hoc codesigned with the entitlement at extraction time.

Pool sizing honors `HYPERLIGHT_INITIAL_SURROGATES` / `HYPERLIGHT_MAX_SURROGATES` (defaults 0/128; HVF allows 127 concurrent VMs system-wide). `HYPERLIGHT_MAX_SURROGATES=0` selects the direct in-process backend, limited to one live sandbox per process.

### Also in this PR

- Guard pages sized to the host page size (16 KiB on Apple silicon) — also fixes the same latent bug on 16K/64K-page aarch64 Linux kernels
- aarch64 host builds now compile with no hypervisor backend enabled
- `map_region`/mmap tests made host-page-size aware (pre-existing aarch64 test debt)
- macOS guest toolchain support: pinned `cargo-hyperlight` 0.1.13 (CI installs a pinned git rev because the crates.io release doesn't compile on macOS); `llvm-ar` for C guest libc archiving
- `build-test-macos-hvf` CI job on `macos-26` runners; `just test-hvf` recipe that builds, ad-hoc codesigns, and runs the suite
- Docs: macOS platform section in `getting-started.md`, HVF section in the surrogate development notes, README hypervisor list

### Testing

On macOS 26 (Apple silicon), all via the surrogate path by default:

- lib unit suite: **279 passed / 0 failed**
- integration: **36 passed / 0 failed**
- sandbox_host: **14 passed / 0 failed**
- wit: **43 passed / 0 failed**
- snapshot goldens (`v1.0-aarch64-hvf-apple`): 5/5 verified locally — publishing to the golden registry is a follow-up
- `clippy-exhaustive` (debug + release), nightly `fmt-check`, MSRV 1.89: green

### Follow-ups (out of scope)

- Publish hvf snapshot goldens to the registry so golden verification runs in CI
- cargo-hyperlight release > 0.1.13 (unblocks crates.io installs on macOS)
- Surrogate death detection beyond kill-on-drop; IPC codec benchmarking (JSON → flatbuffers if profiling justifies it)
- gdb/crashdump/trace_guest/mem_profile on aarch64 (x86_64-gated upstream)

I am happy to be the maintainer for macOS code, if needed :)

### DIFF
--- a/.github/workflows/ValidatePullRequest.yml
+++ b/.github/workflows/ValidatePullRequest.yml
@@ -179,6 +179,61 @@ jobs:
       config: ${{ matrix.config }}
       should_regen_goldens: ${{ needs.check-golden-label.outputs.should_regen_goldens }}
 
+  # Build and test the macOS HVF (Hypervisor.framework) backend on Apple silicon.
+  # Self-contained: guests are built and tested on the same GitHub-hosted arm64
+  # runner (no artifact sharing with the other build-test jobs). Requires
+  # macOS 26 for the 4 KiB IPA granule API.
+  build-test-macos-hvf:
+    needs: [docs-pr, update-guest-locks]
+    # Required because update-guest-locks is skipped on non-dependabot PRs,
+    # and a skipped dependency transitively skips all downstream jobs.
+    # See: https://github.com/actions/runner/issues/2205
+    if: ${{ !cancelled() && !failure() }}
+    timeout-minutes: 45
+    runs-on: macos-26
+    env:
+      # cargo-hyperlight builds a custom target/sysroot; run everything through
+      # the rustup proxy with the pinned toolchain (newer rustc rejects custom
+      # target specs, Homebrew rustc shadows the proxy otherwise).
+      RUSTUP_TOOLCHAIN: "1.94"
+      RUSTC: ~/.cargo/bin/rustc
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0  # v7.0.0
+        with:
+          submodules: true
+
+      - uses: hyperlight-dev/ci-setup-workflow@f6bd9cc86d0737976d2128c8b8ced8edc017cbb4  # v1.9.0
+        with:
+          rust-toolchain: "1.94"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      # crates.io cargo-hyperlight 0.1.13 does not compile on macOS (the ar
+      # detection fix is only on main). Pin the git rev; switch back to the
+      # crates.io pin (just ensure-cargo-hyperlight) once a release > 0.1.13
+      # ships. The Justfile's ensure-cargo-hyperlight accepts this binary
+      # because it reports version 0.1.13.
+      - name: Install cargo-hyperlight from git
+        run: cargo install --locked --git https://github.com/hyperlight-dev/cargo-hyperlight --rev 33384c0c4ed9dea4f0525943809fc444c41a27df cargo-hyperlight
+
+      # llvm-ar/llvm-ranlib are needed to archive the picolibc objects for C
+      # guests; Xcode's ar silently drops ELF members.
+      - name: Install LLVM
+        run: brew install llvm
+
+      - name: Build Rust and C guests
+        run: |
+          just build-and-move-rust-guests
+          just build-c-guests debug
+          just move-c-guests debug
+
+      # Builds, ad-hoc codesigns (com.apple.security.hypervisor) and runs the
+      # hyperlight-host lib + integration test binaries with the hvf feature.
+      # Snapshot-golden verification is a no-op here until the hvf goldens are
+      # published to the golden registry (see RegenSnapshotGoldens.yml).
+      - name: Test with the HVF backend
+        run: just test-hvf debug
+
   # Run examples - needs guest artifacts, runs in parallel with build-test
   run-examples:
     needs:
@@ -255,6 +310,7 @@ jobs:
       - code-checks
       - check-golden-label
       - build-test
+      - build-test-macos-hvf
       - run-examples
       - fuzzing
       - spelling
--- a/.github/workflows/dep_code_checks.yml
+++ b/.github/workflows/dep_code_checks.yml
@@ -83,7 +83,7 @@ jobs:
         run: just clippy-exhaustive release
 
       - name: Verify MSRV
-        run: ./dev/verify-msrv.sh hyperlight-host hyperlight-guest hyperlight-guest-bin hyperlight-common
+        run: ./dev/verify-msrv.sh hyperlight-host hyperlight-guest hyperlight-guest-bin hyperlight-common hyperlight-hvf
 
       - name: Check cargo features compile
         run: just check
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,6 +114,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "330a5ed07fa54e4702c9d6c4174f74427fc0ef6e214bbd677ae50a5099946470"
 
 [[package]]
+name = "applevisor-sys"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "14ae37676989768c5b21414857b782a5e513d93129cda34854f5f78c423a63be"
+dependencies = [
+ "libc",
+]
+
+[[package]]
 name = "arbitrary"
 version = "1.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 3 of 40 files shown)


syntactically:
Thanks for the interest in Hyperlight! We've actually been taking a look at Hypervisor.framework on-and-off for some time, although primarily for developer experience (non-production-service) use cases---if you have any other use cases, please feel free to drop into one of our [community meetings](https://aka.ms/hyperlight-agenda) to share.

We specifically want to avoid doing out-of-process RPCs at the hypervisor-trait level, because we expect this will have a significant performance impact, since it requires a kernel scheduler entry for each minor operation of which there are several per call. We've recently publicly put up the implementation of the first piece of our approach in #1674. This supports making multiple sandboxes in a single process, but only running a single one at a time, and has some overhead due to the way that the shared memory space is managed. We will shortly be extending this approach to add support for running multiple simultaneously, with less overhead.

We have talked about providing a more-or-less transparent way to offer remote sandboxes in another process. Our vague plan for that (not prioritised as nobody has asked for it) is to design it as a wrapper around the Hyperlight crate, sending high-level operations like restores and function calls to the remote process as a single action.  This should require no changes in hyperlight-core itself, which is what we consider our TCB, would work on every supported platform, and should be a bit more performant as well, since it does not require sending multiple RPCs for every high-level action.

siennathesane:
Sure sure. I don't forsee, at least for myself, any use case that's non-development focused. I won't be able to use #1674 until there's parallel support, unfortunately.

> We specifically want to avoid doing out-of-process RPCs at the hypervisor-trait level, because we expect this will have a significant performance impact, since it requires a kernel scheduler entry for each minor operation of which there are several per call. We've recently publicly put up the implementation of the first piece of our approach in https://github.com/hyperlight-dev/hyperlight/pull/1674. This supports making multiple sandboxes in a single process, but only running a single one at a time, and has some overhead due to the way that the shared memory space is managed. We will shortly be extending this approach to add support for running multiple simultaneously, with less overhead.

The macOS HVF framework is extremely frustrating, so I understand the challenges. From my perspective, due to macOS's insane limitations, I'm more than willing to take performance hits, as well as memory hits, if it means I don't need to run a VM just to do Hyperlight development. Having to manage the lifecycle of cloud-init VMs just for development isn't something I'm excited about.

I will join the community meetings moving forward.

.github/workflows/ValidatePullRequest.yml
```diff
@@ -179,6 +179,61 @@ jobs:
...
+  # runner (no artifact sharing with the other build-test jobs). Requires
+  # macOS 26 for the 4 KiB IPA granule API.
+  build-test-macos-hvf:
+    needs: [docs-pr, update-guest-locks]
+    # Required because update-guest-locks is skipped on non-dependabot PRs,
+    # and a skipped dependency transitively skips all downstream jobs.
+    # See: https://github.com/actions/runner/issues/2205
+    if: ${{ !cancelled() && !failure() }}
+    timeout-minutes: 45
+    runs-on: macos-26
```
ludfjig, line 193: unfortunately I don't think github mac runners support nested so this won't work :/

siennathesane: Sad :(

jsturtevant:
> I will join the community meetings moving forward.

heads up this meeting is tomorrow, info at https://github.com/hyperlight-dev/hyperlight#community

syntactically:
> Our vague plan for that (not prioritised as nobody has asked for it) is to design it as a wrapper around the Hyperlight crate, sending high-level operations like restores and function calls to the remote process as a single action. This should require no changes in hyperlight-core itself, which is what we consider our TCB, would work on every supported platform, and should be a bit more performant as well, since it does not require sending multiple RPCs for every high-level action

I've made an issue #1706 for this approach. Would you be at all interested in pursuing it? It should be very similar to the remoting design you have here, just up a couple layers in the API, making it more composable & bringing it outside the TCB.

> I won't be able to use https://github.com/hyperlight-dev/hyperlight/pull/1674 until there's parallel support, unfortunately.

I am a little bit curious about the dev & test use case that requires true parallelism (i.e. actual compute happening in two sandboxes simultaneously), if you're able to share?  The support in #1674 should be API compatible and is enough to run the whole test suite; you can use whatever access pattern you want, but the hvf backend code holds a single global lock when guest code is executing. Since the sandboxes can't interact with each other except via host function calls, and since the global lock is unlocked when a host function is executing, I wouldn't have expected that the difference between true-parallel and the concurrency supported on that PR should generally be semantically observable, except as performance degradation.

siennathesane:
> > Our vague plan for that (not prioritised as nobody has asked for it) is to design it as a wrapper around the Hyperlight crate, sending high-level operations like restores and function calls to the remote process as a single action. This should require no changes in hyperlight-core itself, which is what we consider our TCB, would work on every supported platform, and should be a bit more performant as well, since it does not require sending multiple RPCs for every high-level action
> 
> I've made an issue #1706 for this approach. Would you be at all interested in pursuing it? It should be very similar to the remoting design you have here, just up a couple layers in the API, making it more composable & bringing it outside the TCB.

Omg absolutely!! Thank you for thinking of me 🫶🏼 I would love to get any other design notes from you beforehand, if you can think of any. Sorry for missing this last community meeting, it got buried under my workload, I meant to be there. For some reason the LF invite system won't send meeting invites to either my LF email or my company email.

> 
> > I won't be able to use #1674 until there's parallel support, unfortunately.
> 
> I am a little bit curious about the dev & test use case that requires true parallelism (i.e. actual compute happening in two sandboxes simultaneously), if you're able to share? The support in #1674 should be API compatible and is enough to run the whole test suite; you can use whatever access pattern you want, but the hvf backend code holds a single global lock when guest code is executing. Since the sandboxes can't interact with each other except via host function calls, and since the global lock is unlocked when a host function is executing, I wouldn't have expected that the difference between true-parallel and the concurrency supported on that PR should generally be semantically observable, except as performance degradation.

Of course!! I am the studio head & lead engineer of a small Portuguese startup called [Sunbeam Studios](https://github.com/sunbeamdotpt), and me and one of my other two cofounders are former operations leads for AWS Lambda & AWS API Gateway from wayback when. Our core product is a next-generation competitor to Lambda/Azure Funcs/etc., but it's way different than those products in it's architecture. We picked Hyperlight for our sandbox engine because we're doing GPU virtualization (not time-slicing or MIPS) and we needed a VMM that gives us that blank sandbox. Our execution framework is designed for massive parallelization (for obv reasons) so testing it on an M1 Macbook Pro is a bit convoluted. I would really like to be able to test our scheduler and parallel execution framework on my Macbook without having to always be copying files remotely or doing remote development just to get parallelization on Linux, and running sandboxes one-at-a-time won't really let me test things in parallel. I do understand that macOS imposes arbitrary usage requirements (wish I knew the PM at Apple in charge of those decisions...) so my expectations are low, but it's fine for "are the unit tests passing".

So the parallelization needs we have - specifically for macOS - are "we need to be able to run unit tests and some small scheduler benchmarks to ensure our framework is actually doing what we think it's doing". And without sandboxes of different execution lengths running in parallel, we can't mimic or model contention or scheduling unit tests if we're limited to one-at-a-time. Does that help or is there something different you were looking for?

jsturtevant:
@siennathesane I did dig up Simons old branch but I don't know that it would be all that useful here as it was long time ago and a lot has change in the code. Next steps would be to sketch out a plan and design to discuss then we could get some consensus.  We've used hackmd's in the past to do the initial design proposals and collect feedback. Would that be something you are interested in?  This is something I think we generally would like but don't have it on the immediate roadmap

Issue #1698, opened 2026-07-30
max_guest_log_level is not plumbed through snapshot load


jsturtevant:
### What happened?

After https://github.com/hyperlight-dev/hyperlight/pull/1465 is merged,

max_guest_log_level is not plumbed through snapshot load.

### Steps to Reproduce

1. Run '...'
2. And then do '...'
3. Check logs for '...'

### Expected Results

log level to be set when working with snapshot loads.

### Actual Results

TODO: What actually happened?

### Versions and Environment

Hyperlight version or commit: TODO

#### OS Version

Run the following to find your OS version:

Linux:
```console
cat /etc/os-release && uname -a
```

Windows (PowerShell):
```powershell
cmd /c ver
```

#### Hypervisor

Run the following to check hypervisor access:

Linux:
```console
ls -la /dev/kvm /dev/mshv 2>&1; getfacl /dev/kvm /dev/mshv 2>&1; id
[ -r /dev/kvm ] && [ -w /dev/kvm ] && echo "KVM: OK" || echo "KVM: FAIL"
[ -r /dev/mshv ] && [ -w /dev/mshv ] && echo "MSHV: OK" || echo "MSHV: FAIL"
```

Windows (Admin PowerShell):
```powershell
Get-WindowsOptionalFeature -Online | Where-Object {$_.FeatureName -match 'Hyper-V|HypervisorPlatform|VirtualMachinePlatform'} | Format-Table
```

### Extra Info

Anything else you'd like to add?

sethryanrollins:
Thanks for splitting this out, @jsturtevant!

PR #1699, opened 2026-07-31
plumb `max_guest_log_level` through `MultiUseSandbox::from_snapshot`

sethryanrollins:
draft for #1698

### DIFF
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -166,6 +166,14 @@ impl MultiUseSandbox {
     /// [`SandboxConfiguration::guest_msrs`](crate::sandbox::SandboxConfiguration::guest_msrs)),
     /// or the load fails with an MSR mismatch.
     ///
+    /// [`SandboxConfiguration::set_max_guest_log_level`](crate::sandbox::SandboxConfiguration::set_max_guest_log_level)
+    /// sets the maximum log level passed to the guest. This only takes effect
+    /// for snapshots that still need their guest entrypoint run
+    /// (`NextAction::Initialise`). For a snapshot taken from an
+    /// already-initialized guest, the level was baked into the captured memory
+    /// when the guest first ran, so a configured value has no effect and a
+    /// warning is logged.
+    ///
     /// # Examples
     ///
     /// From a snapshot taken on another sandbox:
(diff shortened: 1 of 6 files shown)


src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -207,6 +219,7 @@ impl MultiUseSandbox {
         snapshot: Arc<Snapshot>,
         host_funcs: crate::HostFunctions,
         config: Option<crate::sandbox::SandboxConfiguration>,
+        max_guest_log_level: Option<LevelFilter>,
```
jsturtevant: I think it would make more sense to move this into the sandbox configuration instead of adding a new field

sethryanrollins: You're right! It would make much more sense

jsturtevant:
this looks in the right direction, @sethryanrollins could you sign the DCO or fix the commits and rebase? then we can run CI

ludfjig:
thanks for your contribution and you interest in hyperlight @sethryanrollins! What I was hoping to do is to address the issue for regular snapshots (taken after init has ran), rather than for these pre-init snapshots (which are not public anyway). It looks like the current state of the PR only addresses pre-init snapshot. Would you be willing to rewrite the PR to deal with post-init snapshots instead? Please let me know if you have any questions as it probably won't be trivial!

yoshuawuyts:
https://github.com/hyperlight-dev/hyperlight/pull/1725 was merged yesterday, renaming `max_guest_log_level` to `guest_log_level`. Given we're introducing a new public API in this PR here, we may want to update the API name here to match. (cc/ @jprendes)

dblnz (dismissed):
LGTM!

jprendes:
> we may want to update the API name here to match

I think this is fine, as we are going to mark the SandboxConfig as deprecated. It will stay as an internal implementation detail.
Currently the SandboxBuilder will build if you set the log level on a snapshot, this will let us remove that failure mode, nice!

jprendes:
Rebased the changes onto latest main.

jprendes (approved):

ludfjig (approved):
lgtm! as my other comment pointed out this doens't resolve the issue for initialized snapshots so let's keep the issue open

PR #1709, opened 2026-08-07
feat: host_bindgen! returns Result from guest-implemented functions

### LINKED ISSUE #1316: `host_bindgen!` generated code panics on guest errors instead of returning `Result`

jsturtevant:
fixes: #1316

`host_bindgen!` export wrappers no longer panic when `Callable::call()` returns `Err` (guest abort/trap/timeout). Host callers now receive the `HyperlightError`.

## Breaking change

Calls to guest exports return `hyperlight_host::Result`. Implementations of host imports are unchanged.

Example WIT:

```wit
world test {
  import roundtrip;
  export roundtrip;
}

interface roundtrip {
  echo: func(x: string) -> string;
}
```

```rust
// Implementing an import: unchanged.
impl test::wit::Roundtrip for Host {
    fn echo(&mut self, x: String) -> String { x }
}

// Calling an export: now fallible.
let value: Result<String, HyperlightError> = sandbox.roundtrip().echo("hi".into());
```

## Interface direction

The trait and its two markers live in `hyperlight_host::component`, so nothing is emitted into the wit namespaces where a package name could collide:

```rust
pub trait InterfaceDirection: Sealed {
    type CallResult<T>;      // Imported: T      Exported: hyperlight_host::Result<T>
    type Borrow<'a, T: 'a>;  // Imported: guard  Exported: &'a T
}
pub struct Imported;  // host implements it, calls run guest to host
pub struct Exported;  // guest implements it, calls run host to guest
```

Generated traits are written against that parameter:

```rust
pub trait Roundtrip<D: InterfaceDirection = Imported> {
    fn echo(&mut self, x: String) -> <D as InterfaceDirection>::CallResult<String>;
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
 ### Added
 
 ### Changed
+* **Breaking:** `host_bindgen!` returns `hyperlight_host::Result` from every
+  guest-implemented function instead of panicking when the guest traps or the
+  call fails.
 * **Breaking:** Guest MSR state is now saved and restored across snapshots.
   `SandboxConfiguration::guest_msrs` declares the MSRs a guest depends on:
   declared MSRs are captured in a snapshot and restored, while every other MSR
--- a/src/hyperlight_component_util/src/emit.rs
+++ b/src/hyperlight_component_util/src/emit.rs
@@ -132,6 +132,28 @@ fn component_first_camel(s: &str) -> String {
     result
 }
 
+/// Which `hyperlight_host::component::InterfaceDirection` marker to name at
+/// the point we are emitting.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum InterfaceDirection {
+    /// Written `Imported`.
+    Imported,
+    /// Written `Exported`.
+    Exported,
+    /// Inside a trait generic over its direction, so name that trait's own
+    /// parameter, `D`.
+    Generic,
+}
+impl InterfaceDirection {
+    pub fn ident(&self) -> Ident {
+        match self {
+            InterfaceDirection::Imported => format_ident!("Imported"),
+            InterfaceDirection::Exported => format_ident!("Exported"),
+            InterfaceDirection::Generic => format_ident!("D"),
+        }
+    }
+}
+
 /// A representation of a trait definition that we will eventually
 /// emit. This is used to allow easily adding onto the trait each time
 /// we see an extern decl.
@@ -149,6 +171,9 @@ pub struct Trait {
     pub tvs: BTreeMap<Ident, (Option<u32>, TokenStream)>,
     /// Raw tokens of the contents of the trait
     pub items: TokenStream,
+    /// The trailing direction parameter declaration, if this trait has one.
+    /// Last so that it can carry a default.
+    pub direction: Option<TokenStream>,
 }
 impl Trait {
     pub fn new() -> Self {
@@ -156,6 +181,7 @@ impl Trait {
             supertraits: BTreeMap::new(),
             tvs: BTreeMap::new(),
             items: TokenStream::new(),
+            direction: None,
         }
     }
     /// Collect the component tyvar indices that correspond to the
@@ -194,11 +220,17 @@ impl Trait {
     /// Build a token stream for the type variable part of the trait
     /// declaration
     pub fn tv_toks(&mut self) -> TokenStream {
+        let mut toks = Vec::new();
         if !self.tvs.is_empty() {
-            let toks = self.tv_toks_inner();
-            quote! { <#toks> }
-        } else {
+            toks.push(self.tv_toks_inner());
+        }
+        if let Some(d) = &self.direction {
+            toks.push(d.clone());
+        }
+        if toks.is_empty() {
             quote! {}
+        } else {
+            quote! { <#(#toks),*> }
         }
     }
     /// Build a token stream for this entire trait definition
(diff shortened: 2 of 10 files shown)


jsturtevant:
@syntactically, please feel free to push any changes you would perfer directly here.  I did my best with how I understand components but it isn't at the same level as you 😃

syntactically:
I will take a look at this and see if there are semantic changes that look important. 

> (positivity and Negative were hard for me to track as I was reading and Export/Import essentially track the existing ideas in the code. I also left out the `NegativeOfThis` as we don't have a use case for it right now (that i am aware of).

@jsturtevant I don't like the import/export naming, because what we care about is the input-or-output-ness from the perspective of the top-level component type, i.e. if we are looking at 
```
(component
  (type $I (component
    (import "positive-import" (instance))
    (export "negative-export" (instance)))
  (import  "foo" (type $I)))
```
then the name is very confusing when it comes to the import and export of the nested type.  I am open to alternative suggestions because I do not want to go with the positivity/negativity if you find it confusing, but I would like it to be something that makes sense with higher-order components.

(In the same vein, since the type code more-or-less handles higher order components correctly, I want to keep that, so I think there is no alternative to having `NegativeOfThis` or something equivalent and having the trait for a component itself parameterized by whether it is itself in a positive or negative of position.)

jsturtevant:
> I will take a look at this and see if there are semantic changes that look important.
> 
> > (positivity and Negative were hard for me to track as I was reading and Export/Import essentially track the existing ideas in the code. I also left out the `NegativeOfThis` as we don't have a use case for it right now (that i am aware of).
> 
> @jsturtevant I don't like the import/export naming, because what we care about is the input-or-output-ness from the perspective of the top-level component type, i.e. if we are looking at
> 
> ```
> (component
>   (type $I (component
>     (import "positive-import" (instance))
>     (export "negative-export" (instance)))
>   (import  "foo" (type $I)))
> ```
> 
> then the name is very confusing when it comes to the import and export of the nested type. I am open to alternative suggestions because I do not want to go with the positivity/negativity if you find it confusing, but I would like it to be something that makes sense with higher-order components.
> 
> (In the same vein, since the type code more-or-less handles higher order components correctly, I want to keep that, so I think there is no alternative to having `NegativeOfThis` or something equivalent and having the trait for a component itself parameterized by whether it is itself in a positive or negative of position.)

Thanks, this is a consideration that I missed. I am fine going with the positive/negative naming with a small comment that helps other future maintainers understand the concept if they are not familiar with it.

syntactically:
Most of this is now #1724.

PR #1720, opened 2026-08-12
Minor quality-of-life improvements to component test/debug infrastructure

syntactically:


jsturtevant (dismissed):
Having WAT and wit makes testing easier!

ludfjig (approved):
lgtm. Should we exercise the wat somewhere in a test? And do we want to update the public doc on the macro to include the new syntax?

syntactically:
> lgtm. Should we exercise the wat somewhere in a test? And do we want to update the public doc on the macro to include the new syntax?

The inline version of the wat is used elsewhere which I think is good enough---I don't think we have any testing of the file path variants other than the wit one right now, so maybe it's worth adding that in the future, but I don't think worth doing right now.

I don't believe there is presently any documentation in this repo that covers the other syntax, either, so again it may be worth adding later, but IMO not right now.

PR #1725, opened 2026-08-12
Add SandboxBuilder

jprendes:
## Description
`SandboxBuilder` is the entry point for creating a sandbox. It gathers machine configuration, host functions, init data and memory mappings, then builds a `MultiUseSandbox` from a guest binary on disk, a guest binary in memory, or a snapshot.

It merges the roles of `SandboxConfiguration`, `GuestEnvironment` and `UninitializedSandbox` into a single type, so all three can become implementation details.

Every setting has an in place accessor taking `&mut self` and a `with_*` counterpart consuming `self` for chaining. Machine configuration values are readable through `get_*`.

`map_memory_region` is `unsafe`, matching `MultiUseSandbox::map_region`. The mapped region must stay valid for the lifetime of the built sandbox.

`build_from_snapshot` errors when `init_data` or `max_guest_log_level` are set, as a snapshot already carries both.

## Scope
This PR only adds the new API, it does not deprecate any old API, and does not replace the use of the old API throughout the repo.

### DIFF
--- a/src/hyperlight_host/src/sandbox/builder.rs
+++ b/src/hyperlight_host/src/sandbox/builder.rs
@@ -15,7 +15,7 @@ limitations under the License.
 */
 
 use std::path::Path;
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 #[cfg(target_os = "linux")]
 use std::time::Duration;
 
@@ -41,6 +41,44 @@ use crate::{
 /// the `build_from_*` methods to create the sandbox from a guest binary on
 /// disk, a guest binary in memory, or a [`Snapshot`]. Every setting has a
 /// default, so a builder with no adjustments is valid.
+///
+/// # Examples
+///
+/// From a guest binary on disk:
+///
+/// ```no_run
+/// # use hyperlight_host::{Result, SandboxBuilder};
+/// # fn example() -> Result<()> {
+/// let mut sandbox = SandboxBuilder::new()
+///     .with_heap_size(1024 * 1024)
+///     .with_host_function("Add", |a: i32, b: i32| a + b)
+///     .build_from_file("guest.bin")?;
+///
+/// let result: String = sandbox.call("Echo", "hello".to_string())?;
+/// # Ok(())
+/// # }
+/// ```
+///
+/// From a snapshot. The snapshot carries the guest binary and the state it was
+/// taken in, so no guest binary is given here. The builder must still register
+/// every host function the snapshot was taken with:
+///
+/// ```no_run
+/// # use hyperlight_host::{Result, SandboxBuilder};
+/// # fn example() -> Result<()> {
+/// let mut sandbox = SandboxBuilder::new()
+///     .with_host_function("Add", |a: i32, b: i32| a + b)
+///     .build_from_file("guest.bin")?;
+/// let snapshot = sandbox.snapshot()?;
+///
+/// let mut restored = SandboxBuilder::new()
+///     .with_host_function("Add", |a: i32, b: i32| a + b)
+///     .build_from_snapshot(snapshot)?;
+///
+/// let result: String = restored.call("Echo", "hello".to_string())?;
+/// # Ok(())
+/// # }
+/// ```
 #[derive(Default)]
 pub struct SandboxBuilder {
     cfg: SandboxConfiguration,
@@ -52,7 +90,12 @@ pub struct SandboxBuilder {
 }
 
 impl SandboxBuilder {
-    /// Create a builder with the default configuration and no host functions.
+    /// Create a builder with the default configuration and the default host
+    /// functions.
+    ///
+    /// By default only the `HostPrint` host function is registered, which
+    /// writes guest output to the host's stdout. Replace it with
+    /// [`Self::host_print`].
     pub fn new() -> Self {
         Self::default()
     }
@@ -82,17 +125,7 @@ impl SandboxBuilder {
 
         let mut uninitialized_sandbox = UninitializedSandbox::new(env, Some(self.cfg))?;
 
-        let mut func_registry = uninitialized_sandbox
-            .host_funcs
-            .try_lock()
-            .map_err(|_| new_error!("Error locking"))?;
-
-        let host_funcs = self.host_funcs.into_inner().functions_map;
-        for (func_name, func_entry) in host_funcs {
-            func_registry.register_host_function(func_name, func_entry);
-        }
-
-        drop(func_registry);
+        uninitialized_sandbox.host_funcs = Arc::new(Mutex::new(self.host_funcs.into_inner()));
 
         for (path, guest_base) in self.mapped_file_cow {
             uninitialized_sandbox.map_file_cow(&path, guest_base)?;
@@ -105,7 +138,7 @@ impl SandboxBuilder {
         let mut sandbox = uninitialized_sandbox.evolve()?;
 
         for region in self.mapped_memory_regions {
-            // SAFETY: the caller of `map_memory_region` guaranteed each region
+            // SAFETY: the caller of `mapped_memory_region` guaranteed each region
             // stays valid and unmodified for the lifetime of this sandbox.
             unsafe { sandbox.map_region(&region)? };
         }
@@ -139,7 +172,7 @@ impl SandboxBuilder {
         }
 
         for region in self.mapped_memory_regions {
-            // SAFETY: the caller of `map_memory_region` guaranteed each region
+            // SAFETY: the caller of `mapped_memory_region` guaranteed each region
             // stays valid and unmodified for the lifetime of this sandbox.
             unsafe { sandbox.map_region(&region)? };
         }
@@ -171,15 +204,15 @@ impl SandboxBuilder {
     /// `guest_base` must be page-aligned and lie outside the sandbox's primary
     /// shared memory region. Violations surface as an error from the
     /// `build_from_*` call, not here. Call this once per file to map several.
-    pub fn map_file_cow(&mut self, path: impl AsRef<Path>, guest_base: u64) -> &mut Self {
+    pub fn mapped_file_cow(&mut self, path: impl AsRef<Path>, guest_base: u64) -> &mut Self {
         self.mapped_file_cow
             .push((path.as_ref().to_path_buf(), guest_base));
         self
     }
 
-    /// Like [`Self::map_file_cow`], but consumes and returns `self` for chaining.
+    /// Like [`Self::mapped_file_cow`], but consumes and returns `self` for chaining.
     pub fn with_mapped_file_cow(mut self, path: impl AsRef<Path>, guest_base: u64) -> Self {
-        self.map_file_cow(path, guest_base);
+        self.mapped_file_cow(path, guest_base);
         self
     }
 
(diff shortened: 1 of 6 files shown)


src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+    cfg: SandboxConfiguration,
+    host_funcs: HostFunctions,
+    init_data: Option<(Vec<u8>, MemoryRegionFlags)>,
+    mapped_file_cow: Vec<(std::path::PathBuf, u64)>,
+    mapped_memory_regions: Vec<MemoryRegion>,
+    max_guest_log_level: Option<LevelFilter>,
+}
+
+impl SandboxBuilder {
+    /// Create a builder with the default configuration and no host functions.
```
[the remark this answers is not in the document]
jprendes: HostPrint is always available, this is the same as the existing behaviour.

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+            guest_binary,
+        };
+
+        let mut uninitialized_sandbox = UninitializedSandbox::new(env, Some(self.cfg))?;
+
+        let mut func_registry = uninitialized_sandbox
+            .host_funcs
+            .try_lock()
+            .map_err(|_| new_error!("Error locking"))?;
+
```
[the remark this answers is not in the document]
jprendes: This map_err is here for completeness, but it should actually be unreachable as the uninitialized_sandbox is created in the line above, so it can't be poisoned or contended.

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -32,7 +32,7 @@ use crate::func::host_functions::TypeErasedHostFunction;
 #[derive(Default)]
 /// A Wrapper around details of functions exposed by the Host
 pub struct FunctionRegistry {
-    functions_map: HashMap<String, FunctionEntry>,
+    pub(super) functions_map: HashMap<String, FunctionEntry>,
```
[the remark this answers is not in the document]
jprendes: I believe there's low risk of doing this here

jsturtevant:
> This PR only adds the new API, it does not deprecate any old API, and does not replace the use of the old API throughout the repo.

I like that this just simplifies the API, and would be an almost drop in for https://github.com/hyperlight-dev/hyperlight-sandbox which I would be able to reduce the code there.  

What do propose for plans to depreciate the older apis?  I think this would provide a bit more flexibility to do things internally

andreiltd (dismissed):
A big fan of this PR 🚀

jprendes:
I should have dropped this link in the PR description.
Here's the document describing the API https://hackmd.io/SdWE6fKSShifX-2_DlY4Fw

At the end of the document I added how we can deprecate the old API

We have 3 deprecation levels, in increasing severity:

* `#[doc(hidden)]`: the types disappear from rustdoc and from code completion,
  so new code stops reaching for them. Nothing downstream breaks.
* `#[deprecated]`: existing users get a compiler warning pointing at the
  builder, or an error where warnings are denied.
* `pub(crate)`: the types leave the public API. A breaking change, and the only
  one of the three that guarantees no downstream use remains.

Applying `#[doc(hidden)]` and `#[deprecated]` together in one release, then
`pub(crate)` in the next, is the recommended path. The first release hides the
types from new code and warns existing users, the second removes them from the
public API.

I'd be happy to hear different opinions.

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+        self.mapped_file_cow
+            .push((path.as_ref().to_path_buf(), guest_base));
+        self
+    }
+
+    /// Like [`Self::map_file_cow`], but consumes and returns `self` for chaining.
+    pub fn with_mapped_file_cow(mut self, path: impl AsRef<Path>, guest_base: u64) -> Self {
+        self.map_file_cow(path, guest_base);
+        self
+    }
```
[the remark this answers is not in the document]
jprendes: I think we need a better renaming, maybe:
```
cow_mapped_file(path, base) / with_cow_mapped_file(path, base)
mapped_memory_region(region)  / with_mapped_memory_region(region)
```
but that could be a follow up discussion.

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+        self.mapped_memory_regions.push(region);
+        self
+    }
+
+    /// Like [`Self::map_memory_region`], but consumes and returns `self` for chaining.
+    ///
+    /// # Safety
+    ///
+    /// Same as [`Self::map_memory_region`].
+    pub unsafe fn with_mapped_memory_region(mut self, region: MemoryRegion) -> Self {
```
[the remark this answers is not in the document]
jprendes: See answer to previous comment

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+
+/// Builds a [`Sandbox`].
+///
+/// Start from [`SandboxBuilder::new`], adjust settings through the `with_*`
+/// (consuming, chainable) or bare-named (in place) accessors, then call one of
+/// the `build_from_*` methods to create the sandbox from a guest binary on
+/// disk, a guest binary in memory, or a [`Snapshot`]. Every setting has a
+/// default, so a builder with no adjustments is valid.
+#[derive(Default)]
+pub struct SandboxBuilder {
```
ludfjig, line 84: Maybe a rustdoc example showing off relevant apis could be a good addition

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+    ///
+    /// Note: [`Self::build_from_snapshot`] errors if this setting is set, as the log level is
+    /// already captured in the snapshot.
+    pub fn max_guest_log_level(&mut self, level: LevelFilter) -> &mut Self {
+        self.max_guest_log_level = Some(level);
+        self
+    }
+
+    /// Like [`Self::max_guest_log_level`], but consumes and returns `self` for chaining.
+    pub fn with_max_guest_log_level(mut self, level: LevelFilter) -> Self {
```
ludfjig: thinking about if we can name this something better...

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -0,0 +1,525 @@
...
+
+        let mut func_registry = uninitialized_sandbox
+            .host_funcs
+            .try_lock()
+            .map_err(|_| new_error!("Error locking"))?;
+
+        let host_funcs = self.host_funcs.into_inner().functions_map;
+        for (func_name, func_entry) in host_funcs {
+            func_registry.register_host_function(func_name, func_entry);
+        }
```
ludfjig: this will re-register default hostprint host function (which is probably fine...?)

ludfjig (dismissed):
I like this! Only one concern:

- Looks like pr is missing newly added `guest_msrs` config.

Some minor things: 

- Would it be possible to somehow allow multiple sandboxes to be created from a builder? Or have builder be clonable? Just thinking about ergonomics if multiple identical sandboxes are to be created (I understand this might not be easy with host functions...)
- Also bikeshedding if any apis should be renamed (my candiadates are `max_guest_log_level`, `map_file_cow`, `map_memory_region`) but not blocking for this pr...
- Could you add this builder addition to CHANGELOG.md?
- Do we want to consider adding the possiblity of having sandboxes without any host functions (including no hostprint?)

src/hyperlight_host/src/sandbox/builder.rs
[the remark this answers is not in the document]
ludfjig: true, but maybe we can word it something like ".. and no **additional** host functions" to avoid confusion?

src/hyperlight_host/src/sandbox/builder.rs
jprendes: It will, assuming that host_funcs contains a HostPrint.
That's also the behaviour if you call `host_function("HostPrint", ...)`.

jprendes: I'm adding a comment to make this clear.

src/hyperlight_host/src/sandbox/builder.rs
jprendes: done

src/hyperlight_host/src/sandbox/builder.rs
jprendes: any suggestions?

yoshuawuyts:
I love the work you're doing on making sandboxes easier to construct! I think that will go a long way in how people experience using Hyperlight. However in Rust, the `builder` suffix typically does refer to a chaining constructor. These typically take `self` by-value or by-(mutable)-reference. I would prefer it if we just kept the `with_` variants of the APIs here, but without the `with_` prefix (so e.g. `with_mapped_file_cow` -> `mapped_file_cow`).

As provided I think the mutable accessors on a type named `*Builder` will probably confuse first-time users of the crate.

jprendes:
I was mostly following the `std::process::Command`, which is the builder I most frequently use. But I agree it's not called `*Builder`.
As for things `*Builder` in the rust std, there's:
* [`DirBuilder`](https://doc.rust-lang.org/std/fs/struct.DirBuilder.html) -> takes `&mut self`
* [`ContextBuilder`](https://doc.rust-lang.org/std/task/struct.ContextBuilder.html) -> takes `self`
* [`thread::Builder`](https://doc.rust-lang.org/std/thread/struct.Builder.html) -> takes `self`
* [`ProcThreadAttributeListBuilder`](https://doc.rust-lang.org/std/os/windows/process/struct.ProcThreadAttributeListBuilder.html) -> takes `self`

outside rust's std:
* [`tokio::runtime::Builder`](https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html) -> takes `&mut self`
* [`LogHistogramBuilder`](https://docs.rs/tokio/latest/tokio/runtime/struct.LogHistogramBuilder.html) -> takes `self`
* [`rayon::ThreadPoolBuilder`](https://docs.rs/rayon/latest/rayon/struct.ThreadPoolBuilder.html) -> takes `self`
* [`rustls`](https://docs.rs/rustls/latest/rustls) -> many builder consistently taking `self`

So yeah, I think the most common pattern is taking `self`.

How would you feel about having method `xxx(self, ...)` and in a trait `SandboxBuilderMut` (name TBD) that exposes the `with_xxx(&mut self, ...)` (`with_` prefix TBD) methods?

yoshuawuyts:
> How would you feel about having method `xxx(self, ...)` and in a trait `SandboxBuilderMut` (name TBD) that exposes the `with_xxx(&mut self, ...)` (`with_` prefix TBD) methods?

I'm a fan of working on designs incrementally: perhaps let's start with the `xxx(self, ...)` variant in this PR and get some experience with that. If we then find that this design is too restrictive and we need additional kinds of builders, let's add those separately. Adding a second kind of builder is a strictly _additive_ change, so if we feel there is a need for it, it's something we can even do post 1.0.

yoshuawuyts (dismissed):

src/hyperlight_host/src/sandbox/builder.rs
jprendes: I've dropped the with_ prefixed methods, and we can drop the max_ so that it becomes `guest_log_level`, I think that's a good name.

yoshuawuyts (approved):

jprendes:
> Looks like pr is missing newly added guest_msrs config.

Added

> Would it be possible to somehow allow multiple sandboxes to be created from a builder? Or have builder be clonable?

No, and you are right on spot, the main reason is the host functions, they are not clonable (we don't require the closures to be clonable, and requiring it would be a big ask). Also the host functions are behind a mutex, but we also always assume that the mutex is uncontested, and error otherwise (IIRC).

> Also bikeshedding if any apis should be renamed (my candiadates are max_guest_log_level, map_file_cow, map_memory_region) but not blocking for this pr...

Changed `max_guest_log_level` -> `guest_log_level`, `map_*` -> `mapped_*`.

> Could you add this builder addition to CHANGELOG.md?

Done

> Do we want to consider adding the possiblity of having sandboxes without any host functions (including no hostprint?)

My intuition is no, but we can discuss it :-)

PR #1727, opened 2026-08-12
Introduce Unrecoverable sandbox state

ludfjig:
Some failures can leave sandboxes in a state that cannot be restored. The ones this PR is concerned with is if mapping/unmapping memory hv-calls. In certain cases, if these fails for example during call to `restore()` the state of the sandbox can be unknown, and we should prevent further use. 

This PR introduces such state as Unrecoverable. Unrecoverable sandboxes can no longer be used at all, and users need to create new one. I don't expect that this will be common at all, but it was an existing gap in our codebase that should be addressed.

This pr also tries to be a bit more transactional in certain operations, meaning it either fully completes successfully or errors cleanly without leaving sandbox in invalid state.

### DIFF
--- a/src/hyperlight_host/src/error.rs
+++ b/src/hyperlight_host/src/error.rs
@@ -201,6 +201,10 @@ pub enum HyperlightError {
     #[error("The sandbox was poisoned")]
     PoisonedSandbox,
 
+    /// The sandbox cannot safely perform further operations and must be discarded.
+    #[error("The sandbox is unrecoverable and must be discarded")]
+    UnrecoverableSandbox,
+
     /// Raw pointer is less than base address
     #[error("Raw pointer ({0:?}) was less than the base address ({1})")]
     RawPointerLessThanBaseAddress(RawPtr, u64),
@@ -383,6 +387,7 @@ impl HyperlightError {
             | HyperlightError::UnexpectedNoOfArguments(_, _)
             | HyperlightError::UnexpectedParameterValueType(_, _)
             | HyperlightError::UnexpectedReturnValueType(_, _)
+            | HyperlightError::UnrecoverableSandbox
             | HyperlightError::UTF8StringConversionFailure(_)
             | HyperlightError::VectorCapacityIncorrect(_, _, _)
             | HyperlightError::SharedMemory(_) => false,
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
@@ -19,6 +19,10 @@ mod x86_64;
 
 #[cfg(target_arch = "aarch64")]
 mod aarch64;
+
+#[cfg(all(test, not(gdb)))]
+pub(crate) mod test_support;
+
 #[cfg(gdb)]
 use std::collections::HashMap;
 use std::str::FromStr;
@@ -532,11 +536,14 @@ impl HyperlightVm {
         let guest_base = crate::mem::layout::SandboxMemoryLayout::BASE_ADDRESS as u64;
         let rgn = snapshot.mapping_at(guest_base, MemoryRegionType::Snapshot);
 
-        if let Some(old_snapshot) = self.snapshot_memory.replace(snapshot) {
+        // Keep the backing memory alive until its VM mapping is removed.
+        if let Some(old_snapshot) = self.snapshot_memory.as_ref() {
             let old_rgn = old_snapshot.mapping_at(guest_base, MemoryRegionType::Snapshot);
             self.vm.unmap_memory((self.snapshot_slot, &old_rgn))?;
         }
+        self.snapshot_memory = None;
         unsafe { self.vm.map_memory((self.snapshot_slot, &rgn))? };
+        self.snapshot_memory = Some(snapshot);
 
         Ok(())
     }
@@ -549,12 +556,15 @@ impl HyperlightVm {
         let guest_base = hyperlight_common::layout::scratch_base_gpa(scratch.mem_size());
         let rgn = scratch.mapping_at(guest_base, MemoryRegionType::Scratch);
 
-        if let Some(old_scratch) = self.scratch_memory.replace(scratch) {
+        // Keep the backing memory alive until its VM mapping is removed.
+        if let Some(old_scratch) = self.scratch_memory.as_ref() {
             let old_base = hyperlight_common::layout::scratch_base_gpa(old_scratch.mem_size());
             let old_rgn = old_scratch.mapping_at(old_base, MemoryRegionType::Scratch);
             self.vm.unmap_memory((self.scratch_slot, &old_rgn))?;
         }
+        self.scratch_memory = None;
         unsafe { self.vm.map_memory((self.scratch_slot, &rgn))? };
+        self.scratch_memory = Some(scratch);
 
         Ok(())
     }
--- a/src/hyperlight_host/src/hypervisor/hyperlight_vm/test_support.rs
+++ b/src/hyperlight_host/src/hypervisor/hyperlight_vm/test_support.rs
@@ -1,0 +1,291 @@
+/*
+Copyright 2025 The Hyperlight Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+use std::collections::VecDeque;
+
+use super::*;
+#[cfg(target_arch = "x86_64")]
+use crate::hypervisor::regs::MsrEntry;
+use crate::hypervisor::regs::{
+    CommonDebugRegs, CommonFpu, CommonRegisters, CommonSpecialRegisters,
+};
+#[cfg(target_arch = "x86_64")]
+use crate::hypervisor::virtual_machine::CreateVmError;
+use crate::hypervisor::virtual_machine::{HypervisorError, VirtualMachine};
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub(crate) enum VmOperation {
+    Map(MemoryRegionType),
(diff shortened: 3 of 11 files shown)


src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -532,11 +534,12 @@ impl HyperlightVm {
         let guest_base = crate::mem::layout::SandboxMemoryLayout::BASE_ADDRESS as u64;
         let rgn = snapshot.mapping_at(guest_base, MemoryRegionType::Snapshot);
 
-        if let Some(old_snapshot) = self.snapshot_memory.replace(snapshot) {
+        if let Some(old_snapshot) = self.snapshot_memory.as_ref() {
             let old_rgn = old_snapshot.mapping_at(guest_base, MemoryRegionType::Snapshot);
             self.vm.unmap_memory((self.snapshot_slot, &old_rgn))?;
         }
         unsafe { self.vm.map_memory((self.snapshot_slot, &rgn))? };
+        self.snapshot_memory = Some(snapshot);
```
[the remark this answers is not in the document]
ludfjig: Not an issue, the sandbox will is unrecoverable in this case

src/hyperlight_host/src/hypervisor/hyperlight_vm/test_support.rs
jsturtevant, line 1: should this be only avaliable when ` #[test]`?

jsturtevant:
LGTM

src/hyperlight_host/src/hypervisor/hyperlight_vm/test_support.rs
ludfjig: That shoud already be the case if you look at `mod.rs`

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
```diff
@@ -532,11 +536,12 @@ impl HyperlightVm {
         let guest_base = crate::mem::layout::SandboxMemoryLayout::BASE_ADDRESS as u64;
         let rgn = snapshot.mapping_at(guest_base, MemoryRegionType::Snapshot);
 
-        if let Some(old_snapshot) = self.snapshot_memory.replace(snapshot) {
+        if let Some(old_snapshot) = self.snapshot_memory.as_ref() {
```
jprendes, line 540: **nit:** could you add a comment here so that future us doesn't try to be smart and refactor it back to the old implementation?
same below for scratch.

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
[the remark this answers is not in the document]
jprendes: I think the comment from Copilot is valid, this is true now that we introduce unrecoverable errors, but if that ever changes for this operation, it would be a bug.
And especially since it's easy to fix, just before `map_memory` you can do
```rust
self.snapshot_memory = None;
```

jprendes (commented):
I like this PR! Good job!
LGTM.

Just some tiny nits

jprendes (dismissed):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -108,6 +119,20 @@ pub struct MultiUseSandbox {
 pub type PtRootFinder = Box<dyn Fn(&[u8], &[u8], u64) -> Vec<u64> + Send>;
 
 impl MultiUseSandbox {
+    fn ensure_usable(&self) -> Result<()> {
```
syntactically: `ensure_usable` is a little bit ambiguous as a name, since it could also mean a function that makes sure the present sandbox is in fact usable.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -531,36 +581,29 @@ impl MultiUseSandbox {
...
-                .update_scratch_mapping(gscratch)
-                .map_err(|e| HyperlightError::HyperlightVmError(e.into()))?;
-        }
-
         let sregs = snapshot.sregs().ok_or_else(|| {
             HyperlightError::Error("snapshot from running sandbox should have sregs".to_string())
         })?;
-        // TODO (ludfjig): Go through the rest of possible errors in this `MultiUseSandbox::restore` function
-        // and determine if they should also poison the sandbox.
+
```
syntactically, line 596: Is this moved after the sregs reset on purpose?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -531,36 +581,29 @@ impl MultiUseSandbox {
...
+        if let Err(error) = self.restore_memory_and_mappings(&snapshot) {
+            self.status = SandboxStatus::Unrecoverable;
```
syntactically, line 598: Would it be preferable to extract this out into a `should_make_unrecoverable`, like we have `should_poison`?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -1741,6 +2076,7 @@ mod tests {
             let path = simple_guest_as_pathbuf();
             let mut cfg = SandboxConfiguration::default();
             cfg.set_heap_size(0x20_000);
+            cfg.set_scratch_size(0x60_000);
```
syntactically: Do you have any idea why we needed to change this?

syntactically (commented):
This looks good. My one high-level question is whether it we could/it would make sense to handle the unrecoverable state transparently (making the API the same as poisoned) by changing the restore operation to do something like  `if self.status == SandboxStatus::Unrecoverable { *self = MultiUseSandbox::new(/* copy over host functions and runtime config */); }`?

(Or, if that's too much to do automatically, is it worth providing an explicit helper for either the "replace this sandbox with a new one with the same configuration" and/or "reset to snapshot, first replacing this sandbox if needed" operations?)

src/hyperlight_host/src/error.rs
```diff
@@ -201,6 +201,10 @@ pub enum HyperlightError {
     #[error("The sandbox was poisoned")]
     PoisonedSandbox,
```
andreiltd: Port

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: whoops good catch we don't need this at all. This test is getting removed in next pr anyway

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: good point renamed to `check_ready`

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: yes it is, I thought sregs check should fail first

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: I considered it but given that it's not really dependent on the variant of the error I didn't do it, and since we only make the sandbox unrecoverable in one single place I didn't. If in the future we add more, I think it would be a good idea

ludfjig:
> This looks good. My one high-level question is whether it we could/it would make sense to handle the unrecoverable state transparently (making the API the same as poisoned) by changing the restore operation to do something like `if self.status == SandboxStatus::Unrecoverable { *self = MultiUseSandbox::new(/* copy over host functions and runtime config */); }`?
> 
> (Or, if that's too much to do automatically, is it worth providing an explicit helper for either the "replace this sandbox with a new one with the same configuration" and/or "reset to snapshot, first replacing this sandbox if needed" operations?)

I do like the first idea, but I am a little bit worried about it doing too much automatically. It could be a major perf hit if say somebody runs it to it on each restore, and they wouldn't even know that it's happening.

I do like the second idea better, my only hesitation is I haven't looked through it enough to know if it's even possible right now? But it would be something we should consider adding, since it probably could be useful for users.

syntactically:
> I do like the first idea, but I am a little bit worried about it doing too much automatically. It could be a major perf hit if say somebody runs it to it on each restore, and they wouldn't even know that it's happening.

The "invisible perf hit" nature could be mitigated a little bit by emitting metrics tracking it? But I agree that it's not ideal.  We could also update the existing restore and add a "restore_unless_unrecoverable" method or something?

> I do like the second idea better, my only hesitation is I haven't looked through it enough to know if it's even possible right now? But it would be something we should consider adding, since it probably could be useful for users.

I think it might be worth checking if a really simple `*self = Self::new(self.host_functions, self.rt_fg)` equivalent works and adding it to this PR if it does, but up to you.

jprendes:
> restore_unless_unrecoverable

my 2 cents, I vote for keeping plain `restore`, every other method is basically `_unless_poisoned_or_unrecoverable`, but we don't make that explicit either.

jprendes (approved):

ludfjig:
> I think it might be worth checking if a really simple `*self = Self::new(self.host_functions, self.rt_fg)` equivalent works and adding it to this PR if it does, but up to you.

Since it would be a pure addition, I'd rather merge this first and do that separate

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
ludfjig: done

src/hyperlight_host/src/hypervisor/hyperlight_vm/mod.rs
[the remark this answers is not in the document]
ludfjig: addressed

syntactically:
> my 2 cents, I vote for keeping plain `restore`, every other method is basically `_unless_poisoned_or_unrecoverable`, but we don't make that explicit either.

The idea would be to keep `restore` (and have it emit a metric + automatically replace the sandbox if unrecoverable) as the default, simpler path for most users, and then find some sensible name for an alternative for users who care about handling it specially.

> Since it would be a pure addition, I'd rather merge this first and do that separate

I suggested waiting mostly because we are planning a release soon and I would like to make sure we have thought through the quality of life (especially "should restore do anything by default", since that probably is a breaking change) before making a release with this, since most downstream consumers of hyperlight will probably make the change to replace unrecoverable sandboxes when we do this release and then be unlikely to move to take advantage of anything we do to abstract away the reinitialisation.

PR #1728, opened 2026-08-12
Relax layout requirements for MultiUseSandbox::restore

### LINKED ISSUE #1592: Reconsider SandboxMemoryLayout::is_compatible_with on snapshot restore
`SandboxMemoryLayout::is_compatible_with` gates snapshot restore on a strict
match of all configured layout fields. This looks more restrictive than needed,
and possibly unnecessary, because restore already replaces the sandbox's live
layout with the snapshot's.

`MultiUseSandbox::restore` calls `validate_compatibility`, which rejects a
snapshot whose layout differs in any of `input_data_size`, `output_data_size`,
`heap_size`, `code_size`, `init_data_size`, `init_data_permissions`, or
`scratch_size`.

But `SandboxMemoryManager::restore_snapshot` then does
`self.layout = *snapshot.layout()` and rebuilds `shared_mem` and `scratch_mem`
from the snapshot. The VM remaps both regions, and `reset_vcpu`,
`set_stack_top`, and `set_entrypoint` all take their values from the snapshot.
So the live layout is fully adopted from the snapshot on every restore. Every
consumer reads sizes and offsets through this layout object.

Given that, the layout compatibility check may be redundant: any snapshot could
load into any sandbox, with the sandbox taking on the snapshot's layout. Host
function validation (`validate_host_functions`) is independent and stays.

## Proposed work

* Confirm nothing outside the layout/mem state assumes the pre-restore layout
  after `restore_snapshot` runs.
* If safe, drop the layout check (or reduce it to the few fields that actually
  must match, if any).

ludfjig:
This makes it possible to restore snapshots from different guest binaries and memory layouts into a sandbox, provided the registered host functions are compatible.

closes #1592

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   MSRs, on MSHV and WHP this is not enforced. by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/991
 * **Breaking:** Filesystem paths are now represented using `PathBuf`. `GuestBinary::FilePath` now stores a `PathBuf` instead of a `String`, and `MultiUseSandbox::generate_crashdump_to_dir` accepts `Into<PathBuf>` instead of `Into<String>`. Callers passing a `String` to `GuestBinary::FilePath` must convert it using `.into()`.
 * Deprecate `MultiUseSandbox::poisoned` in favor of `MultiUseSandbox::status().is_poisoned()`.
+* `MultiUseSandbox::restore` has been made more flexible and now accepts snapshots from any guest binary or memory layout when host functions are compatible.
 
 Certain fixed guest addresses were changed on AArch64 to more easily
 accommodate 16k pages without wasting memory. Snapshots taken from
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -331,10 +331,9 @@ impl MultiUseSandbox {
     /// Creates a snapshot of the sandbox's current memory state.
     ///
     /// The returned snapshot can be applied to any
-    /// [`MultiUseSandbox`] whose memory layout is structurally
-    /// compatible with this sandbox's layout and whose registered
-    /// host functions are a superset of those registered here at the
-    /// time of capture. See [`MultiUseSandbox::restore`] and
+    /// [`MultiUseSandbox`] whose registered host functions are a
+    /// superset of those registered here at the time of capture. See
+    /// [`MultiUseSandbox::restore`] and
     /// [`MultiUseSandbox::from_snapshot`] for the exact compatibility
     /// rules and the error variants returned on mismatch.
     ///
@@ -436,10 +435,6 @@ impl MultiUseSandbox {
     }
 
     /// Restores the sandbox's memory to a previously captured snapshot state.
-    ///
-    /// The snapshot's memory layout must be structurally compatible
-    /// with this sandbox's layout, otherwise this returns
-    /// [`SnapshotLayoutMismatch`](crate::HyperlightError::SnapshotLayoutMismatch).
     ///
     /// The sandbox's registered host functions must be a superset of
     /// those required by the snapshot (matched by name and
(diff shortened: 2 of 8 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
   MSRs, on MSHV and WHP this is not enforced. by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/991
 * **Breaking:** Filesystem paths are now represented using `PathBuf`. `GuestBinary::FilePath` now stores a `PathBuf` instead of a `String`, and `MultiUseSandbox::generate_crashdump_to_dir` accepts `Into<PathBuf>` instead of `Into<String>`. Callers passing a `String` to `GuestBinary::FilePath` must convert it using `.into()`.
 * Deprecate `MultiUseSandbox::poisoned` in favor of `MultiUseSandbox::status().is_poisoned()`.
+* `MultiUseSandbox::restore` has been made more flexible and now accepts snapshots from any guest binary or memory layout when host functions are compatible.
```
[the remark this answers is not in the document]
ludfjig: don't agree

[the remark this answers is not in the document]
jprendes: I think this is technically correct, but I also think we don't care too much about it since we bump or version in each release (not just the patch version), and we haven't hit 1.0 yet. Once we hit 1.0 we will need to be more careful.

jprendes (dismissed):
LGTM

jprendes (dismissed):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -587,23 +582,29 @@ impl MultiUseSandbox {
...
 
+        // Errors below leave the sandbox poisoned unless base mapping updates make it unrecoverable.
+        self.status = SandboxStatus::Poisoned;
+        self.snapshot = None;
+
+        let current_regions: Vec<MemoryRegion> = self.vm.get_mapped_regions().cloned().collect();
+        for region in &current_regions {
+            self.vm
+                .unmap_region(region)
+                .map_err(HyperlightVmError::UnmapRegion)?;
```
syntactically, line 576: Does this also need to make the sandbox unrecoverable if it fails? Or is it intended that this just re-poisoning the sandbox and letting the user retry is fine? I would worry a little that an error here is likely to be about the health of the system as a whole, and I feel like this encourages a retry loop `while sandbox.status().is_poisoned() { sandbox.restore(...); }` which might end up spinning.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -619,18 +620,15 @@ impl MultiUseSandbox {
...
             self.vm
-                .unmap_region(region)
-                .map_err(HyperlightVmError::UnmapRegion)?;
+                .set_crashdump_entry_point(snapshot.original_entrypoint());
+            self.vm.clear_crashdump_binary_path();
         }
 
         self.mem_mgr
             .request_libc_rng_reseed(rand::random::<u32>())?;
+        self.pt_root_finder = None;
```
syntactically: Until we properly add the root finder to the snapshot, I think it is defensible to leave it in here and say "the user should write a root finder which works with any guest that they load", since (a) most users probably only use 1 or 2 different "base" guests and (b) the root finder has the opportunity to inspect the guest state and dispatch on what guest it thinks it is.

ludfjig: agree

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: My idea was that it shouldn't make it `Unrecoverable` because it can be retried without risk and doesn't leave sandbox in a bad state. 

And yeah I worry that most unexpected hypervisor error are likely about the system as a whole... Unsure if there's something better we can do here though

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: fixed in most recent commit a537f7968c9d8feb03399993ab3816941b909385

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: and note this was not introduced in this pr, just moved up earlier in code

syntactically: > because it can be retried without risk and doesn't leave sandbox in a bad state.

If we did put the attempt-to-recreate-sandbox into restore, the same would apply to the other errors that cause poisoning, I think?  I agree it is debatable and am happy to go either way.

Not too important since as you say it was introduced earlier---this is just the first time that it was introduced since we introduced unrecoverability, so I thought it might be worth thinking about.

syntactically (approved):
I did not look at the new tests in any detail, but apart from that this looks good!

PR #1730, opened 2026-08-12
Retain page table root finder for in-memory snapshots

ludfjig:
restoring a snapshot into another sandbox could retain the target sandbox’s finder or lose the source guest’s finder. Later snapshots might then discover the wrong page-table roots. This pr fixes that (for in-memory snapshots only as the pt-finder cannot be serialized to disk)

### DIFF
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -348,10 +348,9 @@ impl MultiUseSandbox {
     /// Creates a snapshot of the sandbox's current memory state.
     ///
     /// The returned snapshot can be applied to any
-    /// [`MultiUseSandbox`] whose memory layout is structurally
-    /// compatible with this sandbox's layout and whose registered
-    /// host functions are a superset of those registered here at the
-    /// time of capture. See [`MultiUseSandbox::restore`] and
+    /// [`MultiUseSandbox`] whose registered host functions are a
+    /// superset of those registered here at the time of capture. See
+    /// [`MultiUseSandbox::restore`] and
     /// [`MultiUseSandbox::from_snapshot`] for the exact compatibility
     /// rules and the error variants returned on mismatch.
     ///
@@ -456,10 +455,6 @@ impl MultiUseSandbox {
     }
 
     /// Restores the sandbox's memory to a previously captured snapshot state.
-    ///
-    /// The snapshot's memory layout must be structurally compatible
-    /// with this sandbox's layout, otherwise this returns
-    /// [`SnapshotLayoutMismatch`](crate::HyperlightError::SnapshotLayoutMismatch).
     ///
     /// The sandbox's registered host functions must be a superset of
     /// those required by the snapshot (matched by name and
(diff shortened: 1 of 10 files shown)


src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -1222,6 +1230,23 @@ mod tests {
...
+    impl<T: ?Sized> AmbiguousIfSync<()> for T {}
+    impl<T: ?Sized + Sync> AmbiguousIfSync<u8> for T {}
+
+    #[test]
+    fn snapshot_and_sandbox_thread_safety() {
+        fn assert_send<T: Send>() {}
+        fn assert_send_sync<T: Send + Sync>() {}
+
+        assert_send::<MultiUseSandbox>();
+        let _ = <MultiUseSandbox as AmbiguousIfSync<_>>::assert_not_sync;
```
jprendes, line 1255: nice trick!

jprendes (dismissed):
LGTM!

What is the consequence of not setting the PtRootFinder?
If I have a sandbox, snapshot it, save to disk, then restore from disk, and don't set the PtRootFinder, what could go wrong?

syntactically:
I'm a little bit skeptical of diverging the in-memory snapshots from the disk snapshots (at least making a release with a semantic gap there), although I agree that logically the root finder feels like a property of the guest image more than the sandbox host. Can we serialise a tag/hash/something/somehow and do validation (or lookup from a user-provided registry) on load? (Or switch to the root finder running in the guest.... but probably there are sometimes cases where the host can do better w/o needing to enter the guest? hard to say for sure.)

ludfjig:
> I'm a little bit skeptical of diverging the in-memory snapshots from the disk snapshots (at least making a release with a semantic gap there), although I agree that logically the root finder feels like a property of the guest image more than the sandbox host. Can we serialise a tag/hash/something/somehow and do validation (or lookup from a user-provided registry) on load? (Or switch to the root finder running in the guest.... but probably there are sometimes cases where the host can do better w/o needing to enter the guest? hard to say for sure.)

How do you think about serializing a flag in snapshot if pt finder is set, and use the new SandboxBuilder to error out deserializing a snapshot if pt root finder is not set on finishing the sandbox builder?

syntactically:
> How do you think about serializing a flag in snapshot if pt finder is set, and use the new SandboxBuilder to error out deserializing a snapshot if pt root finder is not set on finishing the sandbox builder?

That would be an improvement, but I still don't love the idea of having different representations of something in the memory vs file snapshots. I would rather use a tag and compare/look it up in a registry on load, uniformly for both the memory and file paths.

syntactically:
> That would be an improvement, but I still don't love the idea of having different representations of something in the memory vs file snapshots. I would rather use a tag and compare/look it up in a registry on load, uniformly for both the memory and file paths.

Using a tag would also potentially let us get rid of the Arc requirement that is on there now (although adding a finder registry/auto resolving would still need it).

> Or switch to the root finder running in the guest.... but probably there are sometimes cases where the host can do better w/o needing to enter the guest? hard to say for sure

Thinking on this a little more, I think the main reason we put this in the host at all was to save the cost of a vm enter/exit transition if it wasn't needed (i.e. if there were no alternate roots, which is the case for most (all?) of the current shipping guests, or if there was a simple structure to the roots that was easy to read in the host).  On the other hand, with #1717, we in practice will pretty much always enter the guest on snapshot anyway. So I could also believe that the right answer here was to change to calling into the guest to get the roots, if there is a flag on the snapshot saying "I have a custom root finder".

ludfjig:
Closing this for now since we're not clearing it on `restore()` as was the original plan, and isntead https://github.com/hyperlight-dev/hyperlight/pull/1728 added a clarfication to `set_pt_root_finder` that 
>    The callback must support every guest restored into this sandbox.

PR #1736, opened 2026-08-13
Simplify GDB memory access by borrowing sandbox memory manager

ludfjig:
MemoryManager already contains everything needed for gdb so no need to clone it as we can borrow it instead. Also fixes potential stale memory access since snapshot restore can replace snapshot and scratch memory, and the cloned manager could potentially be out of sync. Also removed some now unnecessary synchronization and plumbing

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -289,7 +289,7 @@ test-no-surrogate target=default-target:
 # runs tests that exercise gdb debugging
 test-rust-gdb-debugging target=default-target features="":
     {{ cargo-cmd }} test --profile={{ if target == "debug" { "dev" } else { target } }}  {{ target-triple-flag }} --example guest-debugging {{ if features =="" {'--features gdb'} else { "--features gdb," + features } }}
-    {{ cargo-cmd }} test --profile={{ if target == "debug" { "dev" } else { target } }}  {{ target-triple-flag }} {{ if features =="" {'--features gdb'} else { "--features gdb," + features } }} -- test_gdb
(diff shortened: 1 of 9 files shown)


syntactically (approved):
This looks reasonable to me. Getting rid of the dbg_mem_access_fn clone has been on my todo list for a while because I had a sneaking suspicion that something was off with it even though I never had the time to actually check. It's a shame that the guest_mmap_regions vector is still unsafe to access.

The only real question I have with this approach is whether we want to support the idea of running the gdb thread in parallel with other sandbox threads at any point. This would I think make that difficult? I think that that is likely only useful if we are running multiple sandbox threads in parallel (or the sandbox and the host in parallel), in which case we would have to have some other solution to make the SandboxMemoryManager more shareable, so maybe this will not stay a single &mut reference forever. But, at least, getting rid of the separate dbg_mem_access_fn that was not staying in sync is a major improvement!

ludfjig:
> The only real question I have with this approach is whether we want to support the idea of running the gdb thread in parallel with other sandbox threads at any point. This would I think make that difficult? I think that that is likely only useful if we are running multiple sandbox threads in parallel (or the sandbox and the host in parallel), in which case we would have to have some other solution to make the SandboxMemoryManager more shareable, so maybe this will not stay a single &mut reference forever. But, at least, getting rid of the separate dbg_mem_access_fn that was not staying in sync is a major improvement!

Agreed, for the future this might make it a bit trickier if we want to run it in parallel, but I think that's fine.

PR #1746, opened 2026-08-19
Use SandboxBuilder across examples, tests and docs

jprendes:
`SandboxBuilder` is the entry point for creating a sandbox, so examples, benchmarks, fuzz targets, integration tests and documentation all go through it.

Reshape the shared test helpers in `tests/common` around the builder: `build_rust_sandbox`, `with_rust_sandbox_from`, `with_c_sandbox_from` and `with_all_guests` replace the helpers that handed out an `UninitializedSandbox`.

Call sites that test internals below the public API keep using `UninitializedSandbox` and `SandboxConfiguration` directly. So does `wit_test`, because the generated `Test::instantiate` takes an `UninitializedSandbox`.

### DIFF
--- a/src/hyperlight_host/benches/benchmarks.rs
+++ b/src/hyperlight_host/benches/benchmarks.rs
@@ -10,9 +10,9 @@ use flatbuffers::FlatBufferBuilder;
 use hyperlight_common::flatbuffer_wrappers::function_call::{FunctionCall, FunctionCallType};
 use hyperlight_common::flatbuffer_wrappers::function_types::{ParameterValue, ReturnType};
 use hyperlight_common::flatbuffer_wrappers::util::estimate_flatbuffer_capacity;
-use hyperlight_host::GuestBinary;
+use hyperlight_host::SandboxBuilder;
 use hyperlight_host::mem::shared_mem::ExclusiveSharedMemory;
-use hyperlight_host::sandbox::{MultiUseSandbox, SandboxConfiguration, UninitializedSandbox};
+use hyperlight_host::sandbox::MultiUseSandbox;
 use hyperlight_testing::sandbox_sizes::{LARGE_HEAP_SIZE, MEDIUM_HEAP_SIZE, SMALL_HEAP_SIZE};
 use hyperlight_testing::{c_simple_guest_as_pathbuf, simple_guest_as_pathbuf};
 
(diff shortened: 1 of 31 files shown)


src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -100,7 +99,7 @@ impl Default for HostFunctions {
     /// stdout in green).
     ///
     /// This matches the default registry installed by
-    /// `UninitializedSandbox::new()`, so a snapshot taken from a
+    /// `SandboxBuilder::new()`, so a snapshot taken from a
     /// regular sandbox can be loaded with
     /// `MultiUseSandbox::from_snapshot(snap, HostFunctions::default(), None)`
     /// without registering anything else.
```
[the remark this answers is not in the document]
jprendes: fixed

src/hyperlight_host/tests/common/mod.rs
```diff
@@ -31,127 +29,84 @@ fn c_guest_path() -> PathBuf {
...
+/// Runs a test once per guest binary, passing the path to it.
+///
+/// Use this when the test needs to configure the sandbox itself, for instance
+/// to register a host function that owns per-guest state.
+pub fn with_all_guests<F>(f: F)
 where
-    F: Fn(MultiUseSandbox),
+    F: Fn(PathBuf),
 {
     for path in [rust_guest_path(), c_guest_path()] {
```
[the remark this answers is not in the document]
jprendes: not needed

src/hyperlight_host/benches/benchmarks.rs
```diff
@@ -127,19 +97,6 @@ fn bench_create_initialized_and_drop(b: &mut criterion::Bencher, size: SandboxSi
...
-        });
-    }
-
-    for size in SandboxSize::all() {
-        group.bench_function(
-            format!("create_uninitialized_and_drop/{}", size.name()),
-            |b| bench_create_uninitialized_and_drop(b, size),
-        );
-    }
-
```
ludfjig, line 129: should we really remove these?

ludfjig (dismissed):
LGTM with 1 question about keeping benchmark

src/hyperlight_host/benches/benchmarks.rs
jprendes: Good question, the main reasoning was that this is not possible anymore, but I guess it still gives valuable benchmarking information. How different is creating an `UninitializedSandbox` from creating a `MultiUseSandbox` from a `Snapshot`?

jprendes: Ok, I think the way to go would be to add a new builder terminator `snapshot_from_file` and `snapshot_from_bytes`, and use those to benchmark all the setup that we were measuring for the uninitialized sandbox here (maybe except for the memory allocation and memcpy, but that's not a big loss).

I'll create an issue to add this and reintroduce the new benchmarks to compensate what we loose here.

jprendes: Created #1755 to track this

yoshuawuyts (approved):
Very nice! This looks good to me.

In terms of ordering, I suggest we merge this PR first, and then follow up with the changes in https://github.com/hyperlight-dev/hyperlight/issues/1748 next. That way we don't end up with needing to wait and rebase one on top of the other.

PR #1750, opened 2026-08-20
Use short-form license pragmas

yoshuawuyts:
As I've been reading the Hyperlight source, I noticed we're using long-form license pragmas in each file. I assume we have a reason for that, so I don't want to suggest we remove these outright. But what I figured we could do is move to the short-form pragma format that is a bit more compact.

The format I'm introducing here is for example also used by the Linux kernel ([example](https://github.com/torvalds/linux/blob/a4ff2be345d0abc943da8dd8da98151843b750dc/mm/Kconfig)). As I understand it, tools like FOSSA also know how to ingest this, and can ensure appropriate attribution. As far as I can tell there are no downsides to adopting this shorter format, other than the cost of making the change itself (e.g. this PR). But admittedly this is the first time I'm involved in a migration of this kind, so it might be that I've missed something.

Anyway, I figured this might be a relatively easy way to delete some code with no real downsides. As well as to familiarize myself with the Hyperlight contribution process. Keen to hear what folks think; thanks!

### DIFF
--- a/dev/check-license-headers.sh
+++ b/dev/check-license-headers.sh
@@ -5,27 +5,15 @@
 REPO_ROOT="$(git rev-parse --show-toplevel)"
 cd "$REPO_ROOT" || exit 1
 
-# Define the license header pattern to look for
-LICENSE_PATTERN="Copyright .* The Hyperlight Authors..*Licensed under the Apache License, Version 2.0"
+# Define the license patterns to look for
+SPDX_PATTERN="^// SPDX-License-Identifier: Apache-2\.0$"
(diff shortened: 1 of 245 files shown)


jprendes (approved):
LGTM.

As we are a CNCF project, I checked the CNCF guidelines
https://github.com/cncf/foundation/blob/main/license-notices.md#original-content-for-cncf-projects

> These notices can be included using either the standard language from the license steward, and/or using an [SPDX short-form identifier](https://spdx.dev/ids).

So I think we are good with this change

jprendes:
btw, I am **not** a lawyer :-D

PR #1751, opened 2026-08-21
Add method for recreating unrecoverable sandbox

### LINKED ISSUE #1747: Add API to "replace this sandbox with a new one with the same configuration"
is it worth providing an explicit helper for either the "replace this sandbox with a new one with the same configuration" and/or "reset to snapshot, first replacing this sandbox if needed" operations?

_Originally posted by @syntactically in https://github.com/hyperlight-dev/hyperlight/pull/1727#pullrequestreview-4971549530_

ludfjig:
Closes https://github.com/hyperlight-dev/hyperlight/issues/1747

### DIFF
--- a/src/hyperlight_host/src/sandbox/host_funcs.rs
+++ b/src/hyperlight_host/src/sandbox/host_funcs.rs
@@ -69,6 +69,11 @@ impl HostFunctions {
         self.0
     }
 
+    #[cfg(not(gdb))]
+    pub(crate) fn from_inner(inner: FunctionRegistry) -> Self {
+        Self(inner)
+    }
+
     /// Borrow the inner registry mutably.
     pub(crate) fn inner_mut(&mut self) -> &mut FunctionRegistry {
         &mut self.0
--- a/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
+++ b/src/hyperlight_host/src/sandbox/initialized_multi_use.rs
@@ -37,7 +37,8 @@ pub enum SandboxStatus {
     Ready,
     /// The sandbox requires a successful restore before further use.
     Poisoned,
-    /// The sandbox cannot be used and must be discarded.
+    /// The sandbox cannot be used and cannot be restored.
+    /// [`MultiUseSandbox::recreate()`] can be used to replace an unrecoverable sandbox with a new one.
     Unrecoverable,
 }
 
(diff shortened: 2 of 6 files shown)


ludfjig:
@syntactically first draft at https://github.com/hyperlight-dev/hyperlight/issues/1747

src/hyperlight_host/src/sandbox/host_funcs.rs
```diff
@@ -69,6 +69,11 @@ impl HostFunctions {
         self.0
     }
 
+    #[cfg(not(gdb))]
```
syntactically, line 72: Why is so much stuff not(gdb) guarded?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -618,11 +646,104 @@ impl MultiUseSandbox {
...
+    /// callback, interrupt retry delay, interrupt signal offset, x86_64 guest
+    /// MSR declarations, and crashdump enablement from the unrecoverable sandbox
+    /// to the rebuilt sandbox.
+    ///
+    /// Memory regions added through [`map_region()`](Self::map_region) or
+    /// [`map_file_cow()`](Self::map_file_cow) are not transferred. Map them
+    /// again on the rebuilt sandbox.
+    ///
+    /// Existing interrupt handles do not apply to the rebuilt sandbox. Call
+    /// [`interrupt_handle()`](Self::interrupt_handle) on the rebuilt sandbox to
```
syntactically, line 675: This seems like a nasty footgun. Is there really no way to get around this? I guess we should just need to update the partition handle here on the WHP sandboxes? I would expect the linux/hvf ones might just work, since the tid update happens just before run? If the latter is true, maybe we can just check/update the partition id on the whp sandboxes before run so that the lifecycle is the same?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -569,13 +595,15 @@ impl MultiUseSandbox {
         })?;
 
         if let Err(error) = self.restore_memory_and_mappings(&snapshot) {
-            self.status = SandboxStatus::Unrecoverable;
+            self.lifecycle = SandboxLifecycle::Unrecoverable {
+                recreation_snapshot: snapshot,
```
syntactically, line 599: Why do we want to save the snapshot here rather than in self.snapshot?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -52,12 +53,33 @@ impl SandboxStatus {
...
         matches!(self, Self::Unrecoverable)
     }
 }
 
+enum SandboxLifecycle {
+    Ready,
+    Poisoned,
+    Unrecoverable {
+        #[cfg_attr(gdb, allow(dead_code))]
+        recreation_snapshot: Arc<Snapshot>,
```
syntactically, line 69: I think with the current lifecycle of `MultiUseSandbox::snapshot`, this means we can only ever transition into the unrecoverable state from `restore()`, since we are not guaranteed to have a valid snapshot around otherwise? That doesn't seem right to me---surely we should be able to end up in unrecoverable from e.g. `map_region` as well, even if we don't have a current valid snapshot? Like with poisoned, I would expect that the new snapshot should be supplied _on the recreate operation_?

syntactically (commented):
I left a few comments inline.

I'm still trying to work out what the best API here is. It feels to me like it is a smaller/less breaking change to move this into `restore()` and add a new way to opt-out of the implicit recreate, but the possibility of recreation itself failing (which is probably pretty likely if we've recently failed to restore, since the relevant mapping failures seem likely to be resource exhaustion) is a little difficult to handle ergonomically. It seems inescapable to me that client code, if it is going to handle this locally, will need to do something like
```rust
while should_keep_trying() {
  match sb.<restore/recreate/whatever>(snapshot) {
    Ok(()) => break,
    Err(HyperlightError::UnrecoverableSandbox) => {
      // maybe try to free up resources somehow, try a different snapshot, etc
      continue;
    },
    Err(e) => return e,
}
```
which is just kind of nasty, whether you have one function for restore/recreate or multiple.

(The other option for clients I guess would be to just back out all the way to some top level retry when they get an unrecoverable error. That probably doesn't benefit at all from having recreate functionality at all, though...)

PR #1754, opened 2026-08-21
Update cargo-hyperlight version to 0.1.14

### LINKED ISSUE #1749: macOS tests depend on unreleased `cargo-hyperlight` version
https://github.com/hyperlight-dev/cargo-hyperlight/pull/69 was merged last month but not yet published. Now that https://github.com/hyperlight-dev/hyperlight/pull/1674  has been merged we now have macOS support. However running `just guests` against the main branch on macOS fails to build, which is needed for `just test` to pass and validate that the main branch is actually working as intended.

The currently checked in dep on `cargo-hyperlight` is 0.12.0. The latest release is 0.13.0. We should publish a new version 0.14.0 which includes the fix and pin our dep to that. Thanks!

yoshuawuyts:
Closes https://github.com/hyperlight-dev/hyperlight/issues/1749, thanks!

**Edit:** this PR also adds an automatic update script to automate the `cargo-hyperlight` update bump. The most complicated part of that was updating the `flake.nix` files (see https://github.com/hyperlight-dev/hyperlight/pull/1754#pullrequestreview-4991829464 from review), hence the number of changes.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -8,7 +8,7 @@ bin-suffix := if os() == "windows" { ".bat" } else { ".sh" }
 nightly-toolchain := "nightly-2026-02-27"
 # Pinned cargo-hyperlight version used to build the guest sysroot. Keep this in
 # lockstep with the version pinned in flake.nix.
-cargo-hyperlight-version := "0.1.12"
+cargo-hyperlight-version := "0.1.14"
 
 ################
 ### cross-rs ###
--- a/dev/update-cargo-hyperlight-version.sh
+++ b/dev/update-cargo-hyperlight-version.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+set -Eeuo pipefail
+
+## DESCRIPTION:
+##
+## Pins the latest published cargo-hyperlight in the Justfile and in
+## flake.nix. The Justfile only needs the version, flake.nix also needs
+## the SRI hash of the .crate tarball. Both come from the crates.io
+## sparse index, so no download and no Nix are needed.
+##
+## PRE-REQS:
+##
+## curl, jq, and sed.
+
+CRATE=cargo-hyperlight
+ROOT="$(git rev-parse --show-toplevel)"
+cd "$ROOT"
+
+for tool in curl jq sed; do
+    command -v "$tool" >/dev/null || { echo "error: $tool is required" >&2; exit 1; }
+done
+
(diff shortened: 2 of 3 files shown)


yoshuawuyts:
Oh heh, Copilot review flagged a good issue. I feel like I should probably write a script for that so that updating cargo-hyperlight can be automated. I'll try and get to that later today.

yoshuawuyts:
@syntactically As the owner of our Nix flake integration, I'm assigning this to you for review.

I'm on macOS and don't have Nix installed, so I relied on AI to write the Nix update script. To get that to work I had it change the `flake.nix` file to no longer require having a local `nix` installation and pulling hash information from Crates.io instead. From what I can tell it looks fine, but I have no way of actually testing it right now - so if you could review this that would be great. Thanks!

ludfjig (approved):
lgtm (no idea about the nix stuff though))

jprendes (approved):
Same as @ludfjig , no idea about nix, but otherwise it looks good.
Maybe we split it in 2 PRs, one for the non-nix changes and another one for nix?

yoshuawuyts:
> Maybe we split it in 2 PRs, one for the non-nix changes and another one for nix?

If the Nix changes end up being incorrect or broken, I'll do that - good suggestion. I'd like to hear from @syntactically first though to see if we can land this outright.

syntactically (approved):
@yoshuawuyts Thank you for thinking of the nix support as a first class citizen!  I took a quick look and I think the changes look good, and I did go ahead and test them locally and it seems to work fine.

The change to importCargoLock does mean using import-from-derivation, which can be a problem at large scale but is probably fine here.

> to no longer require having a local `nix` installation 

Not sure what you meant by that?

yoshuawuyts:
> Not sure what you meant by that?

Updating the `flake.nix` file prior to this file seemed to require access to a local `nix(1)` install, which I don't have. Apparently the SHA's are computed based on some kind of Nix-specific directory layout.

syntactically:
> Updating the `flake.nix` file prior to this file seemed to require access to a local `nix(1)` install, which I don't have. Apparently the SHA's are computed based on some kind of Nix-specific directory layout.

Ah, I understand what you meant.  Yes, the hash that was in there previously was the hash (via a nix-specific algorithm) of a vendor directory containing downloaded copies of all the dependencies, which is probably a pain to compute the correct thing for without nix.

PR #1756, opened 2026-08-21
WHP: releasing file mappings after deleting partition in WhpVm::Drop

ludfjig:
Keep mapped-file views alive while WHP may still reference them. With surrogate processes disabled (`HYPERLIGHT_MAX_SURROGATES=0`), `WHvMapGpaRange` uses the host view directly. Releasing that view during VM teardown leaves an active GPA mapping backed by an invalid virtual address.

This is a defensive change, I don't know of any current impact


ludfjig:
Closing, this was wrong! Whoops

PR #1760, opened 2026-08-24
Construct a SandboxBuilder from its guest source

jprendes:
## Naming the guest source

The builder used to start sourceless, with the terminator naming the guest:

```rust
let mut sandbox = SandboxBuilder::new()
    .heap_size(1024 * 1024)
    .host_function("Add", |a: i32, b: i32| a + b)
    .build_from_file("guest.bin")?;
```

It now takes the source up front, and `build` is the only terminator:

```rust
let mut sandbox = SandboxBuilder::from_file("guest.bin")
    .heap_size(1024 * 1024)
    .host_function("Add", |a: i32, b: i32| a + b)
    .build()?;
```

`from_bytes` and `from_snapshot` are the other two constructors. A builder names
its source once, at construction, and nothing changes it afterwards. `from_file`
takes the path of a guest ELF file, `from_bytes` the contents of one.

## Owning the guest bytes

`GuestBinary` used to borrow:

```rust
pub enum GuestBinary<'a> {
    Buffer(&'a [u8]),
    FilePath(PathBuf),
}
```

It now owns, and carries no lifetime:

```rust
pub enum GuestBinary {
    Buffer(Vec<u8>),
    FilePath(PathBuf),
}
```

A caller hands its bytes over rather than keeping them alive for as long as the
sandbox, and the builder stores them without a lifetime of its own.
`GuestEnvironment` keeps its `'b`, the lifetime of the borrowed init data.
`ElfInfo` takes the payload by value, which drops a copy of the guest binary
that the borrowed form forced.

This is the only breaking change, and it carries a changelog entry.

## Removed

| Was | Is |
| --- | --- |
| `SandboxBuilder::new().build_from_file(p)` | `SandboxBuilder::from_file(p).build()` |
| `SandboxBuilder::new().build_from_bytes(b)` | `SandboxBuilder::from_bytes(b).build()` |
| `SandboxBuilder::new().build_from_snapshot(s)` | `SandboxBuilder::from_snapshot(s).build()` |
| `MultiUseSandbox::builder()` | `SandboxBuilder::from_file(p)` and friends |
| `SandboxBuilder::new()` and `SandboxBuilder::default()` | name a source up front |

`SandboxBuilder::new` and its `Default` impl go too, as neither could name a
source. All of these arrived unreleased in #1725 and #1746, so none needs a
changelog entry.

## Commits

Four commits, each building and testing on its own.

### DIFF
--- a/src/hyperlight_host/src/mem/elf.rs
+++ b/src/hyperlight_host/src/mem/elf.rs
@@ -97,8 +97,9 @@ impl framehop::ModuleSectionInfo<Vec<u8>> for &UnwindInfo {
 }
 
 impl ElfInfo {
-    pub(crate) fn new(bytes: &[u8]) -> Result<Self> {
-        let elf = Elf::parse(bytes)?;
+    pub(crate) fn new(bytes: impl Into<Vec<u8>>) -> Result<Self> {
+        let bytes = bytes.into();
+        let mut elf = Elf::parse(&bytes)?;
         let relocs = elf.dynrels.iter().chain(elf.dynrelas.iter()).collect();
         if !elf
             .program_headers
@@ -110,25 +111,32 @@ impl ElfInfo {
 
         // Look for the hyperlight version note embedded by
         // hyperlight-guest-bin.
-        let guest_bin_version = Self::read_version_note(&elf, bytes);
+        let guest_bin_version = Self::read_version_note(&elf, &bytes);
+
+        let phdrs = std::mem::take(&mut elf.program_headers);
+        let entry = elf.entry;
+        #[cfg(feature = "mem_profile")]
+        let shdrs = elf
+            .section_headers
+            .iter()
+            .filter_map(|sh| {
+                Some(ResolvedSectionHeader {
+                    name: elf.shdr_strtab.get_at(sh.sh_name)?.to_string(),
+                    addr: sh.sh_addr,
+                    offset: sh.sh_offset,
+                    size: sh.sh_size,
+                })
+            })
+            .collect();
+
+        drop(elf);
 
         Ok(ElfInfo {
-            payload: bytes.to_vec(),
-            phdrs: elf.program_headers,
+            payload: bytes,
+            phdrs,
             #[cfg(feature = "mem_profile")]
-            shdrs: elf
-                .section_headers
-                .iter()
-                .filter_map(|sh| {
-                    Some(ResolvedSectionHeader {
-                        name: elf.shdr_strtab.get_at(sh.sh_name)?.to_string(),
-                        addr: sh.sh_addr,
-                        offset: sh.sh_offset,
-                        size: sh.sh_size,
-                    })
-                })
-                .collect(),
-            entry: elf.entry,
+            shdrs,
+            entry,
             relocs,
             guest_bin_version,
         })
--- a/src/hyperlight_host/src/mem/exe.rs
+++ b/src/hyperlight_host/src/mem/exe.rs
@@ -66,9 +66,9 @@ impl ExeInfo {
         let mut file = File::open(path)?;
         let mut contents = Vec::new();
         file.read_to_end(&mut contents)?;
-        Self::from_buf(&contents)
+        Self::from_buf(contents)
     }
-    pub fn from_buf(buf: &[u8]) -> Result<Self> {
+    pub fn from_buf(buf: impl Into<Vec<u8>>) -> Result<Self> {
         ElfInfo::new(buf).map(ExeInfo::Elf)
     }
     pub fn entrypoint(&self) -> Offset {
@@ -187,7 +187,7 @@ mod tests {
     fn patched_version_reports_mismatch() {
         let bytes = simpleguest_with_patched_version();
 
-        let info = ExeInfo::from_buf(&bytes).expect("failed to load patched ELF");
+        let info = ExeInfo::from_buf(bytes).expect("failed to load patched ELF");
         assert_eq!(info.guest_bin_version(), Some("0.0.0"));
         assert_ne!(
             info.guest_bin_version().unwrap(),
@@ -217,7 +217,7 @@ mod tests {
         let bytes = simpleguest_with_patched_version();
 
         let result = crate::sandbox::snapshot::Snapshot::from_env(
-            crate::GuestBinary::Buffer(&bytes),
+            crate::GuestBinary::Buffer(bytes),
             crate::sandbox::SandboxConfiguration::default(),
         );
 
--- a/src/hyperlight_host/src/sandbox/builder.rs
+++ b/src/hyperlight_host/src/sandbox/builder.rs
@@ -95,9 +95,8 @@ impl SandboxBuilder {
     }
 
     /// Build a sandbox running the guest binary held in `buffer`.
-    pub fn build_from_bytes(self, buffer: impl AsRef<[u8]>) -> Result<Sandbox> {
-        let buffer = buffer.as_ref();
-        self.build_from_guest_binary(GuestBinary::Buffer(buffer))
+    pub fn build_from_bytes(self, buffer: impl Into<Vec<u8>>) -> Result<Sandbox> {
+        self.build_from_guest_binary(GuestBinary::Buffer(buffer.into()))
     }
 
     fn build_from_guest_binary(self, guest_binary: GuestBinary) -> Result<Sandbox> {
--- a/src/hyperlight_host/src/sandbox/builder.rs
+++ b/src/hyperlight_host/src/sandbox/builder.rs
@@ -46,9 +46,6 @@ impl Source {
 /// chain the settings you need, then call [`SandboxBuilder::build`]. Every
 /// setting has a default, so a builder with no adjustments is valid.
 ///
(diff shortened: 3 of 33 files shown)


jprendes:
@ludfjig, @yoshuawuyts, PTAL :-)

src/hyperlight_host/src/sandbox/uninitialized.rs
```diff
@@ -128,25 +128,25 @@ impl<'a> From<&'a [u8]> for GuestBlob<'a> {
 /// This struct combines a guest binary (either from a file or memory buffer) with
 /// optional data that will be available to the guest during execution.
 #[derive(Debug)]
-pub struct GuestEnvironment<'a, 'b> {
+pub struct GuestEnvironment<'b> {
     /// The guest binary, which can be a file path or a buffer.
-    pub guest_binary: GuestBinary<'a>,
+    pub guest_binary: GuestBinary,
     /// An optional guest blob, which can be used to provide additional data to the guest.
     pub init_data: Option<GuestBlob<'b>>,
```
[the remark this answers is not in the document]
jprendes: That's not correct, the GuestBinary is the one that's owned, GuestEnvironment still holds the init_data un-owned. Added a comment explaining this.

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -78,88 +97,99 @@ pub struct SandboxBuilder {
...
+            mapped_memory_regions: Vec::new(),
+            guest_log_level: None,
+        }
     }
 
     /// Build a sandbox running the guest binary at `path`.
-    pub fn build_from_file(self, path: impl AsRef<Path>) -> Result<Sandbox> {
-        let path = path.as_ref().to_path_buf();
-        self.build_from_guest_binary(GuestBinary::FilePath(path))
+    pub fn from_guest_file(path: impl AsRef<Path>) -> Self {
```
yoshuawuyts: I would have expected this to be called:


```suggestion
    pub fn from_file(path: impl AsRef<Path>) -> Self {
```

yoshuawuyts (commented):
I'm surprised by the names chosen here? Are we expecting e.g. `from_host_bytes`/`from_host_file` or anything similar in the future? If not, the additional `*_guest*` addition to the paths feel superfluous.

If users are unsure what the encoding of the arguments ought to be, I feel like comments would be the right place to clarify that?

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -78,88 +97,99 @@ pub struct SandboxBuilder {
...
+        Self::with_source(Source::guest_file(path))
     }
 
     /// Build a sandbox running the guest binary held in `buffer`.
-    pub fn build_from_bytes(self, buffer: impl AsRef<[u8]>) -> Result<Sandbox> {
-        let buffer = buffer.as_ref();
-        self.build_from_guest_binary(GuestBinary::Buffer(buffer))
+    pub fn from_guest_bytes(buffer: impl Into<Vec<u8>>) -> Self {
```
yoshuawuyts: Same here:

```suggestion
    pub fn from_bytes(buffer: impl Into<Vec<u8>>) -> Self {
```

jprendes:
> Are we expecting e.g. from_host_bytes/from_host_file or anything similar in the future? If not, the additional *_guest* addition to the paths feel superfluous.

No, the intention was to make it clearer what "file" means. But I see your point.

> If users are unsure what the encoding of the arguments ought to be, I feel like comments would be the right place to clarify that?

Will do that

src/hyperlight_host/src/sandbox/builder.rs
jprendes: done

src/hyperlight_host/src/sandbox/builder.rs
jprendes: done

src/hyperlight_host/src/mem/elf.rs
```diff
@@ -97,8 +97,9 @@ impl framehop::ModuleSectionInfo<Vec<u8>> for &UnwindInfo {
 }
 
 impl ElfInfo {
-    pub(crate) fn new(bytes: &[u8]) -> Result<Self> {
-        let elf = Elf::parse(bytes)?;
+    pub(crate) fn new(bytes: impl Into<Vec<u8>>) -> Result<Self> {
```
yoshuawuyts, line 100: did you mean to change this? I don't believe `Self::read_version_note` changed it signature?

ludfjig (commented):
Looks reasonable! What's the motivation for having GuestBinary own the buffer? Avoiding a lifetime in the builder? It will make creating multiple sandboxes from the same buffer slightly slower. Also rip https://github.com/hyperlight-dev/hyperlight/pull/559 :D

src/hyperlight_host/src/mem/exe.rs
```diff
@@ -66,9 +66,9 @@ impl ExeInfo {
         let mut file = File::open(path)?;
         let mut contents = Vec::new();
         file.read_to_end(&mut contents)?;
-        Self::from_buf(&contents)
+        Self::from_buf(contents)
     }
-    pub fn from_buf(buf: &[u8]) -> Result<Self> {
+    pub fn from_buf(buf: impl Into<Vec<u8>>) -> Result<Self> {
```
yoshuawuyts, line 71: Same here - is this change needed? I'm asking because I don't see any other uses for it that would suggest we need to make this change?

yoshuawuyts (commented):
Some API changes stood out to me that I have questions about

src/hyperlight_host/src/mem/elf.rs
ludfjig: I guess it's to get rid of the `bytes.to_vec()` mem copy

jprendes:
> What's the motivation for having GuestBinary own the buffer? Avoiding a lifetime in the builder?

Yes

> It will make creating multiple sandboxes from the same buffer slightly slower

No, `ElfInfo` was already making an owning copy of the data (it used to call `bytes.to_vec()`).
Now if the data is a `Vec`, we just transfer ownership, and avoid an allocation.
If the data is a `&[]`, we are still incurring in the same allocation, just a bit earlier in the function.

> Also rip https://github.com/hyperlight-dev/hyperlight/pull/559 :D

Arguably we could make `ElfInfo` take a lifetime, so that we can avoid ownership there too, but that's not what #559 was doing, that one was still paying the allocation cost per sandbox. All of that falls down however if we enable debugging features, as that code path is still doing a copy of the buffer. Maybe it should be `Arc<[u8]>` instead of `Vec`, or we could use the `Bytes` type from the `bytes` crate, which IIUC, does ref-counting internally.

@yoshuawuyts, what do you think of the builder having a lifetime?
This is something else where the `.build_from_bytes(...)` was nicer, as we don't have to impose the lifetime on the builder, only on the method.

src/hyperlight_host/src/mem/elf.rs
jprendes: `ElfInfo` was doing `bytes.to_vec()`, paying the cost of an allocation. This change is so that we don't have to pay that cost, as we are now passing a `Vec`.
Arguably it should be `bytes: Vec<u8>` as this is an internal API and with this change we only ever pass a `Vec`... but this works as well

src/hyperlight_host/src/mem/exe.rs
jprendes: This is where `ElfInfo::new` is called, and is changed so that we can avoid the extra llocation in there.

jprendes:
if we do introduce lifetimes into the builder, should we also introduce a lifetime for `init_data` ? We currently hold it as a `Vec`, and that one really doesn't have a strong reason to be owned by the builder. Make me thing again that we should be using `Bytes`!

src/hyperlight_host/src/mem/elf.rs
yoshuawuyts: From https://github.com/hyperlight-dev/hyperlight/pull/1760#issuecomment-5408875754:

> No, ElfInfo was already making an owning copy of the data (it used to call bytes.to_vec()).
Now if the data is a Vec, we just transfer ownership, and avoid an allocation.
If the data is a &[], we are still incurring in the same allocation, just a bit earlier in the function.

yoshuawuyts:
> @yoshuawuyts, what do you think of the builder having a lifetime?

I don't mind it honestly. As we discussed sync yesterday: in our current design our builders cannot be reused. That means that builders are not long-lived objects, but intended to be used once and then discarded. An extra lifetime should go unnoticed by most users.

> if we do introduce lifetimes into the builder, should we also introduce a lifetime for `init_data` ?

Maybe that's worth picking up as a follow-up? I believe the answer is... maybe? I'm for sure curious how that would play out.

jprendes:
> Maybe it should be Arc<[u8]> instead of Vec, or we could use the Bytes type from the bytes crate, which IIUC, does ref-counting internally.

Ok, I'm going into premature optimization here. I'm fine with having extra copies until we profile and these copies come back as the bottleneck. When/if that happens we can consider having all these as borrowed or ref-counted slices.

yoshuawuyts (approved):

PR #1765, opened 2026-08-25
fix(restore): replace scratch zero with fresh allocation on Windows

danbugs:
When restoring a snapshot, `restore_snapshot()` can fall into two paths:

1. **Scratch size unchanged** → call `zero()` on the existing region
2. **Scratch size changed** → allocate a fresh new scratch region

Unless you exclude default features and use KVM only, `zero()` falls to a codepath that `memset`s the entire scratch region with zeroes. With big scratch regions like is often the case with Hyperlight+Unikraft guests (e.g., 448 MiB for Node.js), this causes a massive RSS spike and significantly increases restore time.

| | Peak RSS | Restore time |
|---|---|---|
| **Before (Windows)** | ~453 MiB | ~138 ms |
| **After (Windows)** | ~12 MiB | ~11 ms |

With this change, the default path on Windows uses the same fresh-allocation strategy as when the scratch sizes differ — the OS provides demand-zero pages (page-file-backed), so physical memory is consumed only as the guest touches pages.

This is gated behind Windows only. The MSHV path on Linux is left unchanged for now to avoid accidentally regressing small MSHV workloads, and may be updated in the future after more thorough benchmarking.

### DIFF
--- a/src/hyperlight_host/src/mem/mgr.rs
+++ b/src/hyperlight_host/src/mem/mgr.rs
@@ -490,8 +490,9 @@ impl SandboxMemoryManager<HostSharedMemory> {
         };
         let new_scratch_size = snapshot.layout().get_scratch_size();
         let gscratch = if new_scratch_size == self.scratch_mem.mem_size() {
-            self.scratch_mem.zero()?;
-            None
+            // zero_or_replace picks the fastest zeroing strategy for
+            // the current platform (see SharedMemory::zero_or_replace).
+            self.scratch_mem.zero_or_replace()?
         } else {
             let new_scratch_mem = ExclusiveSharedMemory::new(new_scratch_size)?;
             let (hscratch, gscratch) = new_scratch_mem.build();
@@ -501,7 +502,6 @@ impl SandboxMemoryManager<HostSharedMemory> {
             // mapping, so it won't actually be deallocated until it
             // has been unmapped from the VM.
             self.scratch_mem = hscratch;
-
             Some(gscratch)
         };
         self.layout = *snapshot.layout();
--- a/src/hyperlight_host/src/mem/shared_mem.rs
+++ b/src/hyperlight_host/src/mem/shared_mem.rs
@@ -590,31 +590,6 @@ pub trait SharedMemory {
     fn with_contents<T, F: FnOnce(&[u8]) -> T>(&mut self, f: F) -> Result<T> {
         self.with_exclusivity(|m| f(m.as_slice()))
     }
-
-    /// Zero a shared memory region
-    fn zero(&mut self) -> Result<()> {
-        self.with_exclusivity(|e| {
-            #[allow(unused_mut)] // unused on some platforms, although not others
-            let mut do_copy = true;
-            // TODO: Compare & add heuristic thresholds: mmap, MADV_DONTNEED, MADV_REMOVE, MADV_FREE (?)
-            // TODO: Find a similar lazy zeroing approach that works on MSHV.
-            //       (See Note [Keeping mappings in sync between userspace and the guest])
-            #[cfg(all(target_os = "linux", feature = "kvm", not(any(feature = "mshv3"))))]
-            unsafe {
-                let ret = libc::madvise(
-                    e.region.ptr() as *mut libc::c_void,
-                    e.region.size(),
-                    libc::MADV_DONTNEED,
-                );
-                if ret == 0 {
-                    do_copy = false;
-                }
-            }
-            if do_copy {
-                e.as_mut_slice().fill(0);
-            }
-        })
-    }
 }
 
 fn mapping_at(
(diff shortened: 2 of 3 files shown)


jsturtevant (dismissed):

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -489,21 +489,29 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
+        // Linux (KVM and MSHV) falls to the zero() path.  zero()
+        // either uses MADV_DONTNEED (KVM-only builds) or fill(0)
+        // (mshv3 builds).  On Windows, where guests like
+        // Hyperlight-Unikraft use large scratch regions (e.g.,
+        // 448 MiB for Node.js), fill(0) is very suboptimal: it
+        // memsets the entire region, spiking RSS and restore time.
+        // Instead, we freshly allocate a new scratch region, same as
+        // we do when the sizes differ.
+        let gscratch =
+            if new_scratch_size == self.scratch_mem.mem_size() && !cfg!(target_os = "windows") {
```
syntactically: It's a bit annoying that this otherwise-fairly-generic code ends up needing to make decisions about what method of zeroing is fastest on each platform. I wonder whether it would be slightly cleaner to change the return value of the zero function to indicate that it would be better to use delete+recreate as a strategy, or have the new scratch mem creation be done in zero entirely, assigning hscratch to Self and returning gscratch. (Although in the latter case, at least, it would probably make sense to also add a size parameter and basically move this whole let block into a renamed `SharedMemory::resize_and_zero(&mut self, size)`...)

src/hyperlight_host/src/mem/mgr.rs
```diff
@@ -489,21 +489,29 @@ impl SandboxMemoryManager<HostSharedMemory> {
...
-            // Even though this destroys the reference to the host
-            // side of the old scratch mapping, the VM should still
-            // own the reference to the guest side of the old scratch
-            // mapping, so it won't actually be deallocated until it
-            // has been unmapped from the VM.
-            self.scratch_mem = hscratch;
-
-            Some(gscratch)
-        };
+        // Linux (KVM and MSHV) falls to the zero() path.  zero()
```
syntactically: I think this comment could be a little clearer about the central confusing thing of this code: "here's why we're using this weird proxy check instead of doing something clever that is actually linked to whether it will be fast or not". How about something more like:

```rust
// Minor hack: assume that unmapping and remapping the scratch
// region is more efficient than zeroing it if and only if we
// are building for WHP.  There is a break-even point between
// delete+recreate and zero that depends on both the
// hypervisor and the size of the memory region, but this is a
// very simple heuristic that ensures certain common WHP
// workloads with a large scratch region avoid a great deal of
// unnecessary expense.
```

Or, if you go with the comment below and move this code into the existing `SharedMemory::zero()`, then a very short TODO-ish comment is probably enough, since that method is already where the platform-specific performance tradeoffs are being made and already has TODO comments about e.g. working out the heuristic threshold where one is faster than the other.

syntactically (dismissed):
This looks like the functionally correct change to unblock the Windows unikraft performance issues. I do have a couple of minor nits below, especially around communicating why we are using (and that we should replace) this fairly odd heuristic.

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1503,6 +1503,37 @@ impl HostSharedMemory {
...
+    /// On Windows, zeroing via `fill(0)` is prohibitively expensive
+    /// for large regions (e.g. 448 MiB scratch).  Instead, the
+    /// mapping is replaced with a fresh demand-zero allocation.
+    /// Returns `Some(GuestSharedMemory)` when the mapping was
+    /// replaced (the caller must update the VM mapping), or `None`
+    /// when zeroed in place.
+    ///
+    // TODO: Find the break-even point between zero-in-place and
+    // replace for each hypervisor and use a size-based heuristic
+    // instead of a platform check.
```
ludfjig: Is there a tracking issue for this?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1503,6 +1503,37 @@ impl HostSharedMemory {
...
+    pub(crate) fn zero_or_replace(&mut self) -> Result<Option<GuestSharedMemory>> {
+        if cfg!(target_os = "windows") {
```
ludfjig: Is there a reason you're not doing compile-time cfg-check?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1503,6 +1503,40 @@ impl HostSharedMemory {
...
+    ///
+    // TODO: Find the break-even point between zero-in-place and
+    // replace for each hypervisor and use a size-based heuristic
+    // instead of a compile-time platform check.
+    pub(crate) fn zero_or_replace(&mut self) -> Result<Option<GuestSharedMemory>> {
+        #[cfg(target_os = "windows")]
+        {
+            let new_mem = ExclusiveSharedMemory::new(self.mem_size())?;
+            let (hscratch, gscratch) = new_mem.build();
+            self.region = hscratch.region;
```
syntactically: Instead of setting region and lock individually can we do `*self = hscratch`?

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1503,6 +1503,40 @@ impl HostSharedMemory {
...
+    /// for large regions (e.g. 448 MiB scratch).  Instead, the
+    /// mapping is replaced with a fresh demand-zero allocation.
+    /// Returns `Some(GuestSharedMemory)` when the mapping was
+    /// replaced (the caller must update the VM mapping), or `None`
+    /// when zeroed in place.
+    ///
+    // TODO: Find the break-even point between zero-in-place and
+    // replace for each hypervisor and use a size-based heuristic
+    // instead of a compile-time platform check.
+    pub(crate) fn zero_or_replace(&mut self) -> Result<Option<GuestSharedMemory>> {
```
syntactically, line 1498: The old zero function in the SharedMemory trait is I think no longer used and should go away?

syntactically (dismissed):

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Nop, changed to compile-time.

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Done #1766

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: It's still called by `zero_or_replace`, but it is dead code on Windows, so I agree that it is cleaner to inline.

src/hyperlight_host/src/mem/shared_mem.rs
danbugs: Done

ludfjig (approved):

Issue #1690, opened 2026-07-28
Add architecture-independent exception handling APIs sufficient for hyperlight-wasm


syntactically:
Currently, hyperlight-guest exposes an architecture-specific way of overriding certain raw exception handler on x86-64 that is used by hyperlight-wasm. Really, we should instead have architecture-independent hooks for
- Intercepting certain kinds of misbehaviour-indicating exceptions (e.g. FP divide by zero, access to guard regions)
- Support for the virtual memory primitives that hyperlight-wasm needs in hyperlight-guest (e.g. zero-on-first-read regions, copy-on-write images, guard regions)

syntactically:
These should likely be split up a little bit:
- For virtual memory:
  + For CoW images, hl-wasm already uses vmem::MappingKind::Cow and does not need any extra help
  + We should add vmem::MappingKind::LazyZeroInitialised (or something of the sort), which would also be useful for the heap, potentially replacing the existing stack growth, etc, which populates invalid PTEs with flags on them that let the normal fault handler know to replace them with zero initialised pages
  + We also need some way to trap on violations of the permissions on pages, which are currently not being used. That should likely actually be handled like the next category, instead of like other virtual memory regions
- For other exceptions:
  + Currently (already) the hl-wasm support is a little lacking: it only handles #UD on x86-64 which is used because that is the explicit trap instruction that wasmtime uses. We should instead have e.g. `hyperlight_guest::exception::handle_undefined_instr`, `hyperlight_guest::exception::handle_permission_fault`, etc., and the rewriting of registers to actually move into that handler (which should for the moment abort or longjmp) should move into the exception handler in `hyperlight_guest`.

PR #1771, opened 2026-08-26
Add AArch64 guest exception callback

jsturtevant:
Adds an AArch64-specific guest exception callback before Hyperlight's default abort. This enables guest runtimes such as Wasmtime to resolve lazy translation faults or redirect hardware traps while preserving Hyperlight's stack and CoW fault handling.

The callback receives the decoded exception, raw ESR and FAR, mutable ELR, and saved x0 through x30 registers.


syntactically:
Actually, this was purposefully left out in favour of #1690

Issue #1697, opened 2026-07-30
Remove legacy heap region from snapshot


syntactically:
For historical reasons, we have a "heap" region of memory in the snapshot, which is just a bunch of zero-initialised pages that bound the size of the heap. These are, in theory, never read from, only written to (upon which they will get CoW'd anyway), and have no real purpose existing, but bloat the size of the snapshot image. They are also all distinct physical pages when they are implicitly created in from_env (the space for them is included in `SandboxMemoryLayout::get_memory_size()`, so they are created when we make a memory buffer of that size), so the logic that deduplicates pages by GPA does not even work on them.

We should remove this region, and replace it with a VA region outside the snapshot image region that is unmapped by default, and maps to zero when a fault hits it. We can essentially reuse the logic from the stack region for this.

syntactically:
There are a number of options for how we could do this:
- Keep the current situation at the API, where there is a config-controlled heap size knob that sets the size of a VA reservation, but just find a way to avoid actually materialising it in the snapshot when it isn't needed
  + This could itself be done either by moving the VA region out of the currently-originally-identity-mapped-copy-of-snapshot space; or
  + We could leave it in place in VA space but change the PA space to not have that space, and making the original snapshot copy not-identity-mapped. There might be some slight perf benefit to this due to keeping the VA space more compact
- We could make the heap region "unbounded" (well bounded only by the scratch size available in a given run) by moving it to a new very large VA space and populating it on demand. Because the current allocator is not very good, I am not sure whether this might have some negatives for its performance/minimizing its memory demands.
- We could also make the heap region "unbounded" by moving it but only initialise the `LockedHeap` with a small subset of it, and then wrap the `LockedHeap`'s allocator to hurriedly call `add_to_heap` whenever an allocation failed. We would have to double-check that the allocator's internal invariants allow ths.
- We could switch our allocator entirely to a different one (see #343) that supports the more common large-system allocator pattern of, rather than having a fixed backing memory region, asking the "kernel" to provide some more memory when it runs out.

PR #1788, opened 2026-09-02
perf(snapshot): write the guest memory blob sparsely

jprendes:
`Snapshot::save` wrote the guest memory image with `write_all`, which allocates blocks for zeros like any other data: the kernel cannot know the caller would accept a hole. A guest memory image is mostly pages the guest never touched, so saving a snapshot wrote hundreds of MB to reach a final artifact of a few tens of MB.

Skip all-zero blocks instead, seeking over them and extending the file with `set_len`, so the untouched pages become holes. Runs of adjacent non-zero blocks are coalesced into one write, so a blob with no zeros costs a single `write_all` as before.

This is invisible to readers. The digest is computed from the in-RAM buffer before the write, and reads of a hole return zeros, so a sparse blob is byte-for-byte identical to a dense one and hashes the same. The format is unchanged and existing snapshots still load.

Unix filesystems make a file sparse implicitly. NTFS does not, and zero-fills any range skipped by a seek unless the file is explicitly marked sparse, so request that first with FSCTL_SET_SPARSE. The request is best-effort: a filesystem without sparse support (FAT32, exFAT, HFS+) rejects it, and the write is then simply dense, which is still correct.

Only the large snapshot blob takes this path; the manifest and config blobs are small enough that scanning them for zeros would not pay.

Measured on btrfs with a 559 MB image whose pages are ~96% zero: 3.55s and 586 MB on disk, down to 1.70s and 23 MB.

### DIFF
--- a/src/hyperlight_host/Cargo.toml
+++ b/src/hyperlight_host/Cargo.toml
@@ -73,7 +73,13 @@ windows = { version = "0.62", features = [
     "Win32_System_JobObjects",
     "Win32_System_SystemServices",
 ] }
-windows-sys = { version = "0.61", features = ["Win32"] }
+windows-sys = { version = "0.61", features = [
+    "Win32",
+    # FSCTL_SET_SPARSE and DeviceIoControl, used to mark the snapshot
+    # blob sparse. Not implied by "Win32".
+    "Win32_System_IO",
+    "Win32_System_Ioctl",
+] }
 windows-result = "0.4"
 rust-embed = { version = "8.11.0", features = ["debug-embed", "include-exclude", "interpolate-folder-path"] }
 windows-version = "0.1"
--- a/src/hyperlight_host/src/sandbox/snapshot/file/fsutil.rs
+++ b/src/hyperlight_host/src/sandbox/snapshot/file/fsutil.rs
@@ -1,27 +1,141 @@
 // SPDX-License-Identifier: Apache-2.0
 // Copyright 2025 The Hyperlight Authors.
 
-use std::io::{Read, Write};
+use std::io::{Read, Seek, SeekFrom, Write};
 use std::path::Path;
 
 use tempfile::NamedTempFile;
 
 use super::digest::{Digest256, verify_blob_file};
 
+/// Granularity at which [`write_sparse`] looks for all-zero regions.
+/// A hole can only exist on a filesystem-block boundary, and 4 KiB is
+/// the block size of every filesystem we target, so scanning finer
+/// would cost comparisons that can never turn into a hole.
+const SPARSE_BLOCK: usize = 4096;
+
+/// Ask the filesystem to treat `file` as sparse, so that ranges left
+/// unwritten by [`write_sparse`] become holes.
+///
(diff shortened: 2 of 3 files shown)


src/hyperlight_host/Cargo.toml
```diff
@@ -73,7 +73,13 @@ windows = { version = "0.62", features = [
...
     "Win32_System_SystemServices",
 ] }
-windows-sys = { version = "0.61", features = ["Win32"] }
+windows-sys = { version = "0.61", features = [
+    "Win32",
+    # FSCTL_SET_SPARSE and DeviceIoControl, used to mark the snapshot
+    # blob sparse. Not implied by "Win32".
+    "Win32_System_IO",
+    "Win32_System_Ioctl",
+] }
```
ludfjig: looks good but why is windows-rs and windows-sys not the same version 🤔

src/hyperlight_host/src/sandbox/snapshot/file/fsutil.rs
```diff
@@ -1,27 +1,141 @@
...
+    // which are skipped above and so never extend it.
+    file.set_len(len as u64)?;
+    Ok(())
+}
+
 /// Replace `target` atomically: a reader either sees the old
 /// contents or the full new contents, never a partial write. A
 /// failure before commit leaves `target` untouched and removes the
 /// staging file.
 pub(super) fn replace_file_atomic(target: &Path, bytes: &[u8]) -> crate::Result<()> {
```
jsturtevant, line 115: is this still used anywhere beyond tests?

jsturtevant (approved):

syntactically:
This is a nice improvement!

However, I think that the fact that we are seeing snapshots that are e.g. "~96% zero" at all is reflective of deeper issues which we consider addressing additionally or instead.

For one thing, I think it might be more impactful to change the snapshot code to simple not include all-zero pages in the snapshot blob in the first place. They could either all be CoW aliases of a single zero page (we could make this a more general page deduplication a la KSM), or be a new "zero-allocate on first read/write" page region type (which we need to introduce for #1690 anyway).

And, I'm also still curious where these all-zero pages are coming from in the first place. One source is probably .bss in the executable, which we could change to be a zero-allocate-on-first-read/write mapping as discussed above. The other source is probably the legacy heap region, which as pointed out in #1697, is basically useless and should go away.

src/hyperlight_host/src/sandbox/snapshot/file/fsutil.rs
jprendes: It is used to write small blobs like the manifest.

jprendes:
> The other source is probably the legacy heap region

This is the main reason. IIUC, we still need to use heap as our allocator is tied to this heap.

jprendes:
> change the snapshot code to simple not include all-zero pages in the snapshot blob in the first place

That's probably true, but IIUC it would also have a perf cost on restore, as we can't just mmap a single blob.
I would like differential snapshots :-)

PR #1791, opened 2026-09-03
docs: add external byte stream proposal

andreiltd:
[Rendered](https://github.com/andreiltd/hyperlight/tree/tandr/hip-ext-bufs/proposals/0002-envelope-external-byte-stream)

PR #1792, opened 2026-09-03
feat: add external byte codecs

andreiltd:
See HIP document for motivation for those changes: https://github.com/hyperlight-dev/hyperlight/pull/1791

This PR provides the codec and API groundwork for the optimization. Legacy stack based host-guest communication continues to embed byte payloads in flatbuffers. The patch 
- adds external value sources and sinks to function call and result codecs,
- keeps embedded codecs available for the existing stack transport,
- exposes `ByteChunks` parameters and typed guest returns through the C API.


src/hyperlight_common/src/func/ret_type.rs
```diff
@@ -8,6 +8,10 @@ use super::error::Error;
 use crate::flatbuffer_wrappers::function_types::{ReturnType, ReturnValue};
 
 /// This is a marker trait that is used to indicate that a type is a valid Hyperlight return type.
+///
+/// `Vec<u8>` and `Vec<Bytes>` are distinct logical return types. `Vec<u8>`
+/// carries contiguous bytes, while `Vec<Bytes>` preserves local chunk
+/// ownership. Chunk boundaries are not application framing.
```
jsturtevant: I've seen this comment a couple times and I don't really know what it means to be "application framing"?

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
```diff
@@ -268,6 +396,100 @@ pub enum ReturnType {
...
+    S: ExternalValueSource + ?Sized,
+{
+    if parameter.value_type() != FbParameterValue::hlexternalbytes {
+        return parameter.try_into();
+    }
+
+    let marker = parameter
+        .value_as_hlexternalbytes()
+        .ok_or_else(|| anyhow!("Failed to get external byte parameter marker"))?;
+    let length = usize::try_from(marker.length()).map_err(|_| {
```
jsturtevant, line 417: Does this length need to be bounded? (because it might allocate on the host later?)

src/hyperlight_guest_capi/src/types/vec.rs
```diff
@@ -16,6 +16,13 @@ pub struct FfiVec {
 }
 
 impl FfiVec {
+    pub(crate) fn from_mut_slice(value: &mut [u8]) -> Self {
```
ludfjig, line 23: maybe a comment on this could be useful

src/hyperlight_guest_capi/src/types/parameter.rs
```diff
@@ -25,46 +26,113 @@ pub union FfiParameterValue {
...
+                let ptr = value.as_ptr().cast_mut();
+                (
+                    ParameterType::String,
+                    FfiParameterValue { String: ptr },
+                    FfiParameterOwner::String { _value: value },
+                )
             }
             ParameterValue::VecBytes(v) => {
-                let leaked = unsafe { FfiVec::from_vec(v) };
+                let mut value = v;
```
ludfjig: tiny optional nit: slightly cleaner
```suggestion
    ParameterValue::VecBytes(mut v) => {
        let view = FfiVec::from_mut_slice(&mut v);
        (
            ParameterType::VecBytes,
            FfiParameterValue { VecBytes: view },
            FfiParameterOwner::VecBytes { _value: v },
        )
    }
```

ludfjig (dismissed):
this looks great

src/hyperlight_common/src/func/ret_type.rs
yoshuawuyts: My understanding of what @andreiltd is trying to say by this is that chunk boundaries are not semantically meaningful. That means you can't assume that one chunk == one message on the wire. Instead messages boundaries are independent of the way reads are chunked over the network.

src/hyperlight_common/src/flatbuffer_wrappers/codec.rs
```diff
@@ -0,0 +1,32 @@
...
+pub trait ExternalValueSource {
+    /// Take the next external value as contiguous bytes.
+    fn take_bytes(&mut self, length: usize) -> Result<Vec<u8>>;
+
+    /// Take the next external value as owned byte chunks.
+    fn take_chunks(&mut self, length: usize) -> Result<Vec<Bytes>>;
+
+    /// Finish decoding and reject any unused external values.
+    fn finish(&mut self) -> Result<()>;
+}
```
yoshuawuyts, line 35: Could you elaborate on why Rust's standard `Read` and `Write` traits are insufficient for this purpose? I'm personally not a fan of the `bytes` crate, but I recognize the crate can occasionally beneficial. I would like to understand the motivation for us using it in this case, as the [accompanying HIP](https://github.com/andreiltd/hyperlight/tree/tandr/hip-ext-bufs/proposals/0002-envelope-external-byte-stream) does not appear to cover that part of the design.

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
```diff
@@ -259,6 +289,61 @@ pub fn estimate_flatbuffer_capacity(function_name: &str, args: &[ParameterValue]
     estimated_capacity.next_power_of_two()
 }
 
+/// Wrap contiguous bytes as one chunk without copying.
+pub fn byte_chunks_from_bytes(value: Bytes) -> Vec<Bytes> {
```
yoshuawuyts, line 293: I'm a bit confused by this API? `Bytes` is a bit like an `Arc<Vec<u8>>`. What this seems to be doing is converting that to a `Vec<Arc<Vec<u8>>>`. I'm not sure why?

Also afaik it isn't right to call `Bytes` contiguous as they're internally represented by a rope structure. Not that that matters too much, but it doesn't make me less confused 😅

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
```diff
@@ -259,6 +289,61 @@ pub fn estimate_flatbuffer_capacity(function_name: &str, args: &[ParameterValue]
...
+pub(crate) fn try_byte_chunks_len(value: &[Bytes]) -> Option<usize> {
+    value
+        .iter()
+        .try_fold(0usize, |length, chunk| length.checked_add(chunk.len()))
+}
+
+/// Materialize byte chunks as contiguous [`Bytes`].
+///
+/// This is O(1) for zero or one chunk and copies once for multiple chunks.
+pub fn byte_chunks_to_bytes(value: &[Bytes]) -> Bytes {
```
yoshuawuyts, line 321: Actually: I can't help but feel that this is an anti-pattern, and we're doing something wrong here. I can see why we'd have a byte stream (e.g. `impl Read`), but a logical `Vec<Vec<u8>>` - that to me feels like something isn't quite right in our factoring?

yoshuawuyts: The reason why I'm saying this also is because earlier on in the PR it's stated that chunk boundaries are unrelated to message framing.

yoshuawuyts (commented):
left some comments; I've reached out to @andreiltd offline to talk these through because that might be quickest

src/hyperlight_common/src/flatbuffer_wrappers/codec.rs
andreiltd: First problem is that `common` crate is `no_std` -- it is shared between host and guest and `Read`/`Write` traits are `std::io` only. 

Sink here really serves as a collector of borrowed slices; it retains slices until the envelope serialization is done and the message is sent. Source on the other hand, is framed reader. It preserves transport boundaries, which normal `Read` stream of bytes would erase.

tl;dr the strongest reason for custom traits is that they are tailored to model side channel scatter gather lists -- sink is the scatter gather collector, the source is the inverse.

yoshuawuyts:
@andreiltd and I had an offline chat; I now much better understand the decisions made in this PR. Resolved all my comments and happy for us to proceed with this!

src/hyperlight_common/src/func/ret_type.rs
andreiltd: Thanks, yes that is exactly right. I will reword this but the bottom line is that input chunking may not survive the transport and have different boundaries on the receiver side.

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
andreiltd: Right, the code in this PR does not allocate. Later, virtual queue implementation verify the length before it reaches allocation:
- length has to be less than total readable length indicated by buffer chain,
- length has to be less then remaining bytes,
- length has to be less then configured total g2h size,

First two are just protocol sanity checks, because both compared values comes from guest anyway. The way to allocation looks something like this

`consumer.poll(max_recv_len)` rejects chains larger than the configured g2h pool
=> `read_message_header()`  requires the declared payload length to equal the chain remaining length
=> `take_bytes()`  and  `take_chunks()`  check  `length <= request.remaining()`
=> `zeroed_vec()`  uses  `try_reserve_exact()` , so allocation failure returns an error rather than panicking

Comparing to legacy stack io, previous implementation checks if the flatbufffer payload is less than configured region. That is the equivalent of checking g2h pool. Slight improvement in safety in virtual queue implementation is to use `try_reserve_exact` so allocation never panics.

Let me know if clarifying the contract in the docs would be enough, something along those lines I would imagine:

"implementations must validate length against available input and the resource budget before allocating".

There is also an interesting question whether guest could mutate data after `poll` succeeded and before reaching validation code: the answer to that question is currently no, because poll snapshots (copies) the data into owned buffers and vmexit is a strong guarantee that the source does not change. Running host/guest concurrently will probably require synchronized "handoff" period where metadata like length etc, can be copied atomically before guest resumes.

andreiltd:
I think copilot comments about leaks are technically correct, although they are not introduced by this PR. This code has been moved so they appear as added. I would prefer fixing it in a separate PR: expose paired `dealloc` function so that c users don't use `free()` for rust allocated content.

ludfjig (approved):

src/hyperlight_common/src/flatbuffer_wrappers/function_types.rs
jsturtevant: sounds good, Sounds like we have checks for this coming later.  Thanks for verifying you've thought through this!

jsturtevant:
> @andreiltd and I had an offline chat; I now much better understand the decisions made in this PR. Resolved all my comments and happy for us to proceed with this!

would you put the understanding in the comments?  I am interested in hear the response

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
andreiltd: `Vec<Bytes>` is one logical byte value stored as scatter gather segments. This maps nicely to to virt queue buffer chains.

This function is utility adapter. Although it is public, it is only used in internal code. It is only for saying: give me segments from the value that has been already flattened without copying, that is, represent contiguous data as single segment.

```
Bytes::from_static([b"foo"]) => vec![Bytes::from_static(b"foo")]
```

I admit that having a new type such as `ByteChunks(Vec<Bytes>)` , together with a conversion implementation, might be cleaner.

`Bytes` is a buffer that guarantees contiguous memory, that is explicitly stated in crate docs: https://docs.rs/bytes/1.12.1/bytes/index.html#buf-bufmut. `Buf` and BufMut` traits can represent reading/writing over scatter gather lists which is very useful for buffer chains.

src/hyperlight_common/src/flatbuffer_wrappers/util.rs
andreiltd: Here,  `&[Bytes]` represents scatter/gather storage and the function flattens that sequence for the API that requires continuous memory. A useful (illustrative only) example here would be a large string that arrives split across several buffers in a virt queue chain but receiver API requires owned string, so contigous memory has to be reconstructed.

andreiltd:
Hey @jsturtevant, I added comments to Yosh's original questions. Let me know if that clarifies things a bit.

jsturtevant (approved):

jprendes (approved):

PR #1793, opened 2026-09-03
feat: add virtqueue transport foundations

andreiltd:
The patch implements shared layout and ownership model for virtq based host and guest communication. This is a preparation work for replacing stack based IO with virtq, in the meantime function calls and logging use the legacy transport. Actual takeover, buffer mappings, guest allocated pools and retained buffer snapshots are separate stages.

- queue configurations and a shared scratch arena for rings and buffer pools.
- `SlotPool` allocation, 
- stateful chain I/O,
- shared message framing,
- guest queue initialization,
- canonical ring state validation


src/hyperlight_common/src/layout.rs
```diff
@@ -25,4 +110,257 @@ pub fn scratch_base_gva(size: usize) -> u64 {
...
+        let addr = |offset: usize| base_addr.checked_add(u64::try_from(offset).ok()?);
+        let _end_addr = addr(len)?;
+
+        Some(Self {
+            g2h_ring_addr: base_addr,
+            h2g_ring_addr: addr(h2g_ring_offset)?,
+            g2h_pool_addr: addr(g2h_pool_offset)?,
+            h2g_pool_addr: addr(h2g_pool_offset)?,
+            ring_span_len: g2h_pool_offset,
+            len,
```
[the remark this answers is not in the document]
andreiltd: It looks to me like offsets are correct, also changeset suggestion is nonsensical.

src/hyperlight_guest/src/transport/mem.rs
```diff
@@ -0,0 +1,135 @@
...
+        Ok(unsafe { &*ptr.cast::<AtomicU16>() })
+    }
+}
+
+// SAFETY: Every address is restricted to the scratch mapping. Payload
+// references rely on descriptor ownership, and ring flags use aligned atomics.
+unsafe impl MemOps for GuestMemOps {
+    type Error = GuestMemError;
+
+    fn read(&self, addr: u64, dst: &mut [u8]) -> Result<(), Self::Error> {
```
ludfjig, line 73: The following satisfies unsafe preconditions but is UB due to src and dst overlapping in a call to `copy_to_nonoverlapping`.

```rust
let mut backing = [1u8, 2, 3, 4];
let base = backing.as_mut_ptr() as u64;
// SAFETY: Backing stays mapped and alive. No peer accesses it.
let mem = unsafe { GuestMemOps::from_raw_parts(base, 4) };
mem.read(base, &mut backing).unwrap();
```

src/hyperlight_common/src/virtq/consumer.rs
```diff
@@ -462,63 +597,111 @@ impl<M: MemOps + Clone, N: Notifier> VirtqConsumer<M, N> {
...
+    /// Reset ring and inflight state to initial values.
+    ///
+    /// Fails while a polled chain has not yet been completed, preventing a
+    /// live [`RecvChain`] from reading descriptors after reset and reuse.
+    ///
+    /// # Errors
+    ///
+    /// - [`VirtqError::InvalidState`] - one or more chains are still in flight
+    /// - [`VirtqError::RingError`] - device-event normalization failed
+    pub fn reset(&mut self) -> Result<(), VirtqError> {
```
ludfjig, line 608: `reset()` can release a slot while the consumer still holds a writable handle to it. With initialized producer/consumer endpoints and a one-slot pool, these safe calls let the stale handle overwrite a new message:

```rust
let sent = producer.chain().writable(4).build().unwrap();
producer.submit(sent).unwrap();
let (_recv, reply) = consumer.poll(0).unwrap().unwrap();
let ReplyChain::Writable(mut stale) = reply else { unreachable!() };

producer.reset().unwrap();
let mut replacement = producer.chain().readable(4).build().unwrap();
replacement.write_all(b"GOOD").unwrap();
stale.write_all(b"BAD!").unwrap();
```

src/hyperlight_host/src/sandbox/config.rs
```diff
@@ -316,6 +357,98 @@ impl SandboxConfiguration {
...
+        self.g2h_queue_size
+    }
+
+    /// Set the G2H virtqueue descriptor count.
+    ///
+    /// Values are rounded up to a power of two in `2..=32768`.
+    #[instrument(skip_all, parent = Span::current(), level= "Trace")]
+    pub fn set_g2h_queue_size(&mut self, size: usize) {
+        self.g2h_queue_size = Self::normalize_queue_size(size);
+    }
```
ludfjig, line 372: Could we add these configs to src/hyperlight_host/src/sandbox/builder.rs as well?

ludfjig (dismissed):
I think this looks great together with #1793. I had a look and also had AI take a look too. I'm not too familiar with virtio but I'm learning a lot 😅. Btw, I am curious if have you compared benchmarks on this branch compared to `main`, for example for creating a sandbox and taking a snapshot

src/hyperlight_host/src/sandbox/config.rs
andreiltd: They are added as part of https://github.com/hyperlight-dev/hyperlight/pull/1794

src/hyperlight_common/src/virtq/consumer.rs
andreiltd: Yes. I think reset method should be unsafe and it should document this invariant. Because end of the queue are living in guest and host there is no really programmatic way to make it safe. I also don't think this can be exploited to crash the host but I will double check.

src/hyperlight_guest/src/transport/mem.rs
andreiltd: I added checks for this.

ludfjig (approved):

docs/snapshot-oci-format.md
```diff
@@ -31,11 +31,11 @@ Three blob kinds per tag:
 * **manifest** (`application/vnd.oci.image.manifest.v1+json`). Tiny JSON
   pointer record selected via `index.json`. References one config and
   one layer by digest.
-* **config** (`application/vnd.hyperlight.snapshot.config.v1+json`). The
+* **config** (`application/vnd.hyperlight.snapshot.config.v2+json`). The
```
jsturtevant, line 34: This is breaking right?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -1381,23 +1430,23 @@ mod tests {
...
+    // Checks that 1,000 calls work with constrained guest memory.
+    // This catches guest stack reset and heap leaks.
     #[test]
     fn test_with_small_stack_and_heap() {
-        const HEAP_SIZE: u64 = 32 * 1024;
-        // min_scratch_size already includes 1 page (4k on most
-        // platforms) of guest stack, so add 20k more to get 24k
-        // total, and then add some more for the eagerly-copied page
-        // tables on amd64
+        const HEAP_SIZE: u64 = 128 * 1024;
```
jsturtevant, line 1437: is there a reason we increased the guest size here?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -1216,6 +1260,11 @@ mod tests {
         assert!(SandboxStatus::Unrecoverable.is_unrecoverable());
     }
 
+    fn assert_virtq_attached(sbox: &MultiUseSandbox) {
+        assert!(sbox.mem_mgr.g2h_consumer.is_some());
+        assert!(sbox.mem_mgr.h2g_consumer.is_some());
+    }
+
     #[test]
```
jsturtevant, line 1268: I noticed alot of the tests in here had heap and scratch sized changed.  What was the reasoning here?

docs/snapshot-oci-format.md
simongdavies: If it is we should call out the breaking change in the changelog.md

src/hyperlight_common/src/virtq/pool/slot.rs
```diff
@@ -0,0 +1,446 @@
...
+
+/// Exact memory layout for one [`SlotPool`] tier.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SlotLayout {
+    /// Start of the first slot.
+    pub base_addr: u64,
+    /// Capacity of each slot. Must fit in [`Allocation::len`].
+    pub slot_size: usize,
+    /// Number of slots.
+    pub slot_count: usize,
```
simongdavies: Do these need to be pub?

src/hyperlight_common/src/virtq/pool/slot.rs
```diff
@@ -0,0 +1,446 @@
...
+        if layout.slot_size == 0 {
+            return Err(AllocError::InvalidArg);
+        }
+        let slot_size = u32::try_from(layout.slot_size).map_err(|_| AllocError::InvalidArg)?;
+
+        if layout.slot_count == 0 {
+            return Err(AllocError::EmptyRegion);
+        }
+
+        layout.end_addr()?;
```
simongdavies: This looks odd, is it validating the values in SlotLayout? should that be part of SlotLayout::new instead? Also not sure why clippy is not catching the non use of return value?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
andreiltd: This is very conservative headroom but also a default heap limit. In general the default heap size hasn't change although the requirement for guest heap slightly increased. The guest side of the queue needs some bookkeeping that use heap allocations: buffer pool state and inflight chains tracking. Those tests heap configuration was close enough for the bookkeeping to cause oom. I increased the size, but I can find the real minimum requirement for this test if we care about using something different than default here.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
andreiltd: I think it's best to understand new scratch requirements using comaprison with stack based input/ouput region. Main reserves two 16K scratch regions for stack io. The input stack carries guest-call requests and host-call replies. The output stack carries guest-call results and host-call requests. Stack bookkeeping are backpointers stored inside these regions.

Virtq uses two descriptor rings and two buffer pools. H2G contains receive buffers posted by the guest for host requests. G2H carries guest requests and results. A guest-to-host request also reserves writable G2H buffers for the host's reply, so request and reply storage must coexist.

Guest code can retain incoming external bytes as pool-backed `ByteChunks` after a call returns. Requests that are carrying external bytes may leave one H2G buffer available for a small control call. Without that reserve, retained values could deplate all receive buffers and deadlock the host as it cannot ask the guest to release them.

With 4K buffers, a sizing baseline for roughly 16K of encoded request and reply capacity is:
- H2G: 5 pages. 4 receive buffers plus one control reserve.
- G2H: 9 pages. 4 request buffers, 4 reply buffers, and one page of 256-byte slots for small control messages and logs.

The defaults capacity is rounded: 8 H2G pages and 12 G2H pages, totalling 80K. The 64 descriptor G2H ring, 32 descriptor H2G ring, and alignment padding occupy another 4K. That gives us transport footprint is 84K, compared with main's 32K.

That adds 52K to the old 288K scratch budget, giving 340K. Rounding to a 16K gives the 352K default that is compatible with macos host pages. We could bring thise memory closer to main with 5 H2G and 9 G2H pool pages reducing the transport arena to 60K (-20K) and smaller descriptor rings could trim bookkeeping overhead, although it is now 4K only. We could make reservation optional, that will give us another 8K back for the workflows for which bytes retention is not important.

See also: https://github.com/andreiltd/hyperlight/blob/virtq-takeover/docs/virtio-host-guest-communication.md#transport-arena

For heap increase see my other response: https://github.com/hyperlight-dev/hyperlight/pull/1793#discussion_r4061746009

src/hyperlight_common/src/virtq/pool/slot.rs
```diff
@@ -0,0 +1,446 @@
...
+    pub const fn new(base_addr: u64, slot_size: usize, slot_count: usize) -> Self {
+        Self {
+            base_addr,
+            slot_size,
+            slot_count,
+        }
+    }
+
+    /// Total bytes occupied by the slots.
+    pub fn byte_len(self) -> Result<usize, AllocError> {
```
simongdavies: Is it OK for slot_size and slot_count to be unbounded?

src/hyperlight_host/src/mem/layout.rs
```diff
@@ -383,6 +437,44 @@ impl SandboxMemoryLayout {
         self.scratch_size
     }
 
+    #[allow(dead_code)]
```
simongdavies: #[allow(dead_code)] needed?

simongdavies (commented):
A few minor comments

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jsturtevant: Thanks for the detailed response, the biggest thing I wanted to cover was that this was thought about and not just a config change to make tests pass.  Obviously you've thought and covered this very deeply.

jsturtevant (dismissed):

src/hyperlight_common/src/virtq/pool/slot.rs
andreiltd: Removed pub.

src/hyperlight_common/src/virtq/pool/slot.rs
andreiltd: I moved all the validation to constructor.

src/hyperlight_host/src/mem/layout.rs
andreiltd: Removed.

src/hyperlight_common/src/virtq/pool/slot.rs
andreiltd: The constructor verifies that slot_size and slot_count are non zero and they fit into u32. If we talk about resource limits, those params are derived from host buffer configuration for the virtq. The pool is managed entirely by the guest. The guest could of course ignore the host settings but it would oom pretty fast on the guest side.

simongdavies (approved):

docs/snapshot-oci-format.md
andreiltd: Yes, I added this to CHANGELOG.md: 

```
**Breaking:** Virtqueue rings and pools occupy host-owned scratch before page
  tables. Snapshots use ABI 4 and config schema v2. Existing snapshots must be
  regenerated.
```

PR #1794, opened 2026-09-03

andreiltd:
See [rendered](https://github.com/andreiltd/hyperlight/blob/virtq-takeover/docs/virtio-host-guest-communication.md) design doc. For broader context see HIP PR: https://github.com/hyperlight-dev/hyperlight/pull/1112

This patch replaces stack based communication with function calls transported over virtual queues. The documentation covers two current limitations: snapshots do not support retained buffers, and the host owns the transport arena. Addressing both is deferred to future work.


ludfjig (commented):
I am curious if you have run any of the benchamrks compared to main and could share?

Question about ergonomics orthogonal to this pr: Could the `call` on the host accept slice like `&[u8]`  (or `&[&[u8]]` and `&str` instead of the owned counterparts? At least 1 copy is necessary on the host I believe, so looks doable on the surface maybe? I think it could get rid of some clones if host only has access to a slice and then needs to clone it unnecessarily.

Guest functions which take a `String` have the bytes copied from the flatbuffer into a new String if I understand correctly. Could we consider supporting guest functions which take something similar to a `&str` pointing directly into the flatbuffer to get rid of copy? Flatbuffer already supports it I think. An alternative could perhaps be to document that if perf is important, use `ByteChunks` instead of String. 

Will get back reviewing this in more detail shortly!

jsturtevant:
> I am curious if you have run any of the benchamrks compared to main and could share?

I am also very interested in this.  We might want to take this branch and run it on https://github.com/hyperlight-dev/hyperlight-bench as well?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -137,6 +137,7 @@ impl MultiUseSandbox {
             mem_mgr: mgr,
             vm,
             snapshot: None,
+            transport_dirty: false,
```
jprendes, line 140: This initializes `transport_dirty` to `false`. However, `hyperlight_main` can already log or call host functions. Those operations leave noncanonical G2H descriptors. An immediate `snapshot()` skips the checkpoint and fails canonical validation. Mark transport dirty after guest initialization, or canonicalize it before declaring it clean. Add a startup-host-call-then-snapshot regression test.

src/hyperlight_host/src/sandbox/snapshot/file/media_types.rs
jprendes, line 22: This retains ABI 3, config v2, and golden version v3.0, despite removing layout fields and requiring another OCI layer. Old snapshots fail structural validation under the same identifiers. Coordinate ABI, schema, and golden version bumps across this stacked PR. The current check-golden-label failure confirms the missing golden bump.

I think this is a consequence of the recent tracing changes, which also bumped the snapshot ABI.

jprendes (commented):
Comments from my agent.

andreiltd:
@ludfjig @jsturtevant The benchmarks are very nuanced, it's hard to compare apples to apples because `virtq` has a lot more configuration knobs that you can tune to optimize transport based on your payload sizes etc. Without any adjustement the results looks like this:

| Benchmark | stack I/O | Virtq I/O |
|---|---:|---:|
| Small guest call | 13.10 µs | 14.49 µs (+10.6%) |
| Guest call with host callback | 23.75 µs | 26.24 µs (+10.5%) |
| Guest call with restore | 41.90 µs | 48.61 µs (+16.0%) |
| Snapshot create, dirty sandbox | 320.55 µs | 353.67 µs (+10.3%) |
| Snapshot restore | 13.12 µs | 14.97 µs (+14.1%) |
| Sandbox from resident snapshot | 749.57 µs | 768.01 µs (+2.5%) |
| Create initialized sandbox | 2.433 ms | 2.710 ms (+11.4%) |
| Create and drop sandbox | 25.449 ms | 24.677 ms (-3.0%) |
| 24 KiB input / 8 KiB output, C | 23.27 µs | 21.10 µs (-9.3%) |
| 24 KiB input / 8 KiB output, Rust | 23.21 µs | 21.81 µs (-6.0%) |
| 50 MiB Vec + 50 MiB String, cold call | 557.883 ms | 751.244 ms (+34.7%) |

This is quite consistent with what I would expect without measuring anything: for tiny payload transport complexity with virtq dominates and delay is slightly worse. Snapshot creation and restore are slightly worse due to increased scratch requirement.

---
Here are the results of the benchmarks that are taking advantage of `virtq` optimizations. Each call sends the full 1M input to the guest. **Length** returns its byte count. **Echo** returns the entire payload.

| Guest operation | stack I/O | Virtq: 4K slots | Virtq: 64K slots | Virtq: 2M slots |
|---|---:|---:|---:|---:|
| Length, `Vec` | 146.9 | 258.5 (+76.0%) | 99.2 (-32.5%) | 89.8 (-38.9%) |
| Length, `ByteChunks` | 143.8 | 226.4 (+57.4%) | 70.8 (-50.8%) | 62.6 (-56.5%) |
| Echo, `Vec` | 1109.0 | 360.1 (-67.5%) | 183.1 (-83.5%) | 172.1 (-84.5%) |
| Echo, `ByteChunks` | 513.1 | 347.8 (-32.2%) | 156.2 (-69.6%) | 144.0 (-71.9%) |

You can clearly see where chunking dominates and where full zero copy path without chunking kicks in. That is almost 80% gain for large payloads.

Note, that I want to give the virtq implementation another pass that is entirely focused on performance, that is after all the correctness is figured out.

jsturtevant:
That all looks very reasonable to me.  We also discussed this at length today in the community meeting: https://hackmd.io/blCrncfOSEuqSbRVT9KYkg#2026-09-21 with the main conclusion being that those numbers come out to about 10% but it is 1microsecond which is well with our variance. At some point in the future we might be able to gain that back.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -925,13 +954,8 @@ impl MultiUseSandbox {
         args: Vec<ParameterValue>,
     ) -> Result<ReturnValue> {
         self.check_ready()?;
-        // ===== KILL() TIMING POINT 1 =====
```
jsturtevant, line 935: Are these timing points still valid?  fyi @ludfjig

jsturtevant (dismissed):
I am really impressed with the quality of the tests and thought that went into the validation.   This is a large code change but given the performance and the number of tests _added_ (with very few changed/removed) I have high confidence.  

I've spent some time segmenting off parts of the code to use agents to find issues and didn't find anything substantial. Also went over the user facing changes they look good.

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -60,12 +62,8 @@ pub(crate) fn call_guest_function(function_call: FunctionCall) -> Result<Vec<u8>
     } else {
         // The given function is not registered. The guest should implement a function called
         // guest_dispatch_function to handle this.
-
-        // TODO: ideally we would define a default implementation of this with weak linkage so the guest is not required
```
ludfjig, line 64: Did you intend to remove this commetn?

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
ludfjig: yes but looks like the comment was lost. It would be good to keep the comment, so the points can easily be found when searching the repo since there are 5 more with the exact phrasing `===== KILL() TIMING POINT <nr here> `. Mind putting this comment back to where clear_cancel() was moved to?

ludfjig (commented):
This looks great! Just one thing:

Previously, taking a snapshot involved only the host and didn't rely on guest. Now we enter the guest and rely (?) on the guest to do some specific work. A malicious guest could either not do the required work, do it incorrectly, but also do something else completely (spin infinitely, abort, etc). This seems like a big change that we should make sure we understand.

src/hyperlight_guest_bin/src/guest_function/call.rs
andreiltd: Yes, it looks like this TODO is obsolette and we are using weak linkage already. For example https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_guest_bin/src/guest_function/call.rs#L18

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
andreiltd: I've changes this to transport is always dirty when in initialization phase.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
andreiltd: I restored the comment.

andreiltd:
Hey @ludfjig,

> Could the call on the host accept slice like &[u8] (or &[&[u8]] and &str instead of the owned counterparts? 

It looks feasible, yes. I think currently `call` require `'static` (probably because of closures?) so there will be some borrow checker wrestling but alternative borrowed path should be possible. Guest side `str` would require type system that ensure the reference does not outlive the call. That is a bit more involved. This implementation is mapping `Bytes` into shared memory and is using type erased owner that returns backing memory to the pool when bytes are dropped: see https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner

> Previously, taking a snapshot involved only the host and didn't rely on guest. Now we enter the guest and rely (?) on the guest to do some specific work. A malicious guest could either not do the required work, do it incorrectly, but also do something else completely (spin infinitely, abort, etc). This seems like a big change that we should make sure we understand.

I think this concern is valid. The preparation work in this implementation is a tradeoff: the problem is that we need to rebuild the host side state of the queue from guest controlled shared memory. If we support any arbitrary state, the complexity of validation that state blows up. Canonical state gives the host very well defined and small invariant.

We could move that preparation to the restore step, but that adds a host to guest roundtrip for first call and adds delay to the  restore. 

I think it would be great if we could at least limit the ability to call host functions during snapshot checkpoint. That would eliminate side effects on the host. I could also add tests/examples for handling spinning snapshot checkpoint, poisoning sandbox when guest fails to canonicalize queues? Let me know what you think. I also vaguely remember we did similar checkpoint for nanvix but I'm very fuzzy on the details.

ludfjig:
> Hey [@ludfjig](https://github.com/ludfjig),
> 
> > Could the call on the host accept slice like &[u8] (or &[&[u8]] and &str instead of the owned counterparts?
> 
> It looks feasible, yes. I think currently `call` require `'static` (probably because of closures?) so there will be some borrow checker wrestling but alternative borrowed path should be possible. Guest side `str` would require type system that ensure the reference does not outlive the call. That is a bit more involved. This implementation is mapping `Bytes` into shared memory and is using type erased owner that returns backing memory to the pool when bytes are dropped: see https://docs.rs/bytes/latest/bytes/struct.Bytes.html#method.from_owner

Sounds good maybe we can try to do this in a follow up. And yes I was mostly concerned with the host side api ergonomics, although if a copy could be removed on the guest-side that would be bonus too (seems hard though)
 
> > Previously, taking a snapshot involved only the host and didn't rely on guest. Now we enter the guest and rely (?) on the guest to do some specific work. A malicious guest could either not do the required work, do it incorrectly, but also do something else completely (spin infinitely, abort, etc). This seems like a big change that we should make sure we understand.
> 
> I think this concern is valid. The preparation work in this implementation is a tradeoff: the problem is that we need to rebuild the host side state of the queue from guest controlled shared memory. If we support any arbitrary state, the complexity of validation that state blows up. Canonical state gives the host very well defined and small invariant.

Makes sense. I'm not sure there are any other alternatives to this then

> 
> We could move that preparation to the restore step, but that adds a host to guest roundtrip for first call and adds delay to the restore.

Yes this is probably not idea

> I think it would be great if we could at least limit the ability to call host functions during snapshot checkpoint. That would eliminate side effects on the host. 

I think this would be useful, although it would need to be done by the host (don't accept host calls during this period) since guest is untrusted

>I could also add tests/examples for handling spinning snapshot checkpoint, poisoning sandbox when guest fails to canonicalize queues? Let me know what you think. I also vaguely remember we did similar checkpoint for nanvix but I'm very fuzzy on the details.

Sounds good. I don't remember either for nanvix unfortunately 🤔

jsturtevant:
> > I could also add tests/examples for handling spinning snapshot checkpoint, poisoning sandbox when guest fails to canonicalize queues? Let me know what you think. I also vaguely remember we did similar checkpoint for nanvix but I'm very fuzzy on the details.
> 
> Sounds good. I don't remember either for nanvix unfortunately 🤔

paging @danbugs who might have some insight

jprendes:
@ludfjig would it be fine to merge this PR as is and address the remaining concerns as a follow up?

We can create issues for them with release blocker label.

What do you think?

ludfjig:
> [@ludfjig](https://github.com/ludfjig) would it be fine to merge this PR as is and address the remaining concerns as a follow up?

fine by me :)

simongdavies (commented):
> Previously, taking a snapshot involved only the host and didn't rely on guest. Now we enter the guest and rely (?) on the guest to do some specific work. A malicious guest could either not do the required work, do it incorrectly, but also do something else completely (spin infinitely, abort, etc). This seems like a big change that we should make sure we understand.

Would doing something like this alleviate some of the concerns here:

1. Set a host-owned  Checkpointing  execution mode before entering the guest.
2. During that mode,  outb_virtq_call  must reject ordinary G2H function requests before  call_host_function .
3. Permit only operations required by checkpointing, such as logs, completion notification, abort, and cancellation.
4. Clear the mode on every exit path.
5. Validate the queues and mailbox after the guest returns. Do not trust the dedicated guest routine to have behaved correctly.

I also added a couple of agent generated issues, so I dont have particularly high confidence in them

src/hyperlight_guest/src/transport/codec.rs
```diff
@@ -0,0 +1,239 @@
...
+
+    let payload_len =
+        size_prefix_payload_len(&prefix).context("virtqueue message has an invalid prefix")?;
+
+    let control_len =
+        size_prefixed_len(payload_len).context("virtqueue message control length overflow")?;
+    let control = segments
+        .split_to(control_len)
+        .context("virtqueue message control data is truncated")?
+        .into_bytes();
```
simongdavies, line 78: Agent Generated:

Fragmented message decoding is quadratic. `Segments::split_to` repeatedly removes index zero from a `SmallVec`, shifting all remaining segments each time. Large valid messages split across thousands of descriptors can require hundreds of millions of element moves. Consume via an index/drain or rebuild the remainder once.

Same comment applies to  src/hyperlight_common/src/virtq/buffer.rs:65-81 which is not new  code in this PR

src/hyperlight_host/src/mem/virtq/mod.rs
```diff
@@ -0,0 +1,227 @@
...
+            || self.h2g_ring.len() != layout.get_h2g_queue_dims().ring_len()
+        {
+            return Err(new_error!(
+                "virtqueue snapshot ring lengths do not match layout"
+            ));
+        }
+
+        let (g2h, h2g) = ring_layouts(layout)?;
+        let g2h_mem = ImageMem::new(g2h.desc_table_addr(), &self.g2h_ring);
+
```
simongdavies, line 167: Agent generated:

Snapshot validation does not verify H2G descriptor addresses. Writable descriptors of the expected length pass even if duplicated or placed outside the H2G pool, including over  transport metadata. Restore can then corrupt queue state when the host writes a request. Require every descriptor to be a unique, aligned slot in the configured H2G pool.

simongdavies:
> @ludfjig would it be fine to merge this PR as is and address the remaining concerns as a follow up?
> 
> We can create issues for them with release blocker label.
> 
> What do you think?

I just added a couple of comments, not sure if a couple of these are even valid since the are agent generated, happy to address these in a follow up as they don't alter the functionality, the comments on guest entry at snapshot liekwise could be addressed as a follow up

andreiltd:
I thought I would add some context to the discussion and justify the strategy for snapshoting in this PR.

Each queue has two endpoints: a producer in the guest and a consumer on the host. After restore, their state must agree before communication can resume.

The rings live in shared scratch memory, but they are only part of the queue state; both endpoints keep private state, including cursors and inflight chains. The guest also keeps buffer pool allocation and ownership state on the heap.

H2G has an additional requirement that is before the host can send a request, the guest must post writable buffers. These fixed-size buffers tell the host which memory slots it can use. They must be available immediately after restore if we want to avoid a preparatory VM entry.

Transport pools are configurable and can be several megabytes. We do not want to capture their entire capacity with every snapshot. Only payloads still held by guest code should survive snapshotting. Those payloads must remain consistent with their captured ownership state and guest pointers.

This creates a tension: we want compact snapshots and fast restore, but omitting transport storage means some preparation is necessary. Where should that work happen?

1. **Make every top-level call a snapshot-ready boundary**

The first idea is to leave the queues ready for snapshotting whenever a top-level guest call finishes. The problem is the final G2H response; tthe guest submits it before halting, but the host processes and completes it after the VM exit. The guest cannot reclaim that entry before the host completes it.

With the current protocol, fully resetting the queues would require another round trip. The host would process the response, then re-enter the guest to reclaim completed entries and reset its state. This would tax every function call for snapshot preparation, even when no snapshot is requested. That cost is not acceptable.

2. **Capture the entire transport arena**

The guest's producer and pool bookkeeping already survive through the ordinary memory snapshot as part of the heap. We could also capture the rings, both buffer pools, and the host consumer cursors.

That would preserve matching state without resetting the queues but snapshot size and restore copying is scaling with configured pool capacity, including unused slots and stale payloads. This really conflicts with the goal of preserving only live transport data.

3. **Reset the queues from the host**

The host can rewrite shared ring memory, but that does not reset the guest's private state. Its cursors, inflight chains, and allocation ownership would still match the old rings.

The host could try to modify those guest objects too. That would couple snapshot correctness to private guest data structures, rust layouts, and ownership rules. It feels super fragile. 💀 

4. **Rebuild the transport during restore**

Another option is to restore guest memory and ask the guest to construct fresh queues before normal execution resumes.

This keeps the snapshot compact, but adds guest execution to restore. Deferring reconstruction until the first call only moves the latency. The host still needs writable H2G buffers before it can publish that call. This does not meet the goal of keeping restore immediately usable.

5. **Capture only buffers referenced by the rings**

Ring descriptors do not identify every live payload. Guest code can retain a buffer after its descriptor has been consumed and reclaimed. Its lifetime is then represented by guest ownership state, not by the ring. Selective capture therefore needs an ownership-aware mechanism. Scanning the rings alone cannot safely tell us which payloads to preserve or discard.

6. **The strategy in this PR is an explicit snapshot checkpoint**

Preparation should happen when a snapshot is requested and the transport is dirty. Ordinary calls should not pay for it, and restore should not need to execute guest preparation code.

The checkpoint establishes a canonical transport state:

1. The host sends a checkpoint request.
2. The guest reclaims completed work and resets its producers.
3. The guest checks retained allocations and prefills H2G with writable buffers.
4. After preparation finishes, the guest publishes its status and halts.
5. The host resets its consumers and validates the captured ring images before admitting the snapshot.

Successful capture preserves matching guest bookkeeping and canonical rings. Pool payloads remain outside the snapshot. Restore installs the rings and attaches consumers at the agreed initial positions, without a preparatory VM entry.

I would also include in this, proposed by @simongdavies, checkpoint execution restricitons.

src/hyperlight_guest/src/transport/codec.rs
andreiltd: Fixed with proposed index/drain.

src/hyperlight_host/src/mem/virtq/mod.rs
andreiltd: I will add those checks but I would rather add it in `validate_canon_image` callsites for H2G. The function takes the closure that allows caller to validate the descriptors content as well. Currently it only checks if buffers are writable, but I can easily extend this to cover ranges, duplicates etc. 

See: https://github.com/hyperlight-dev/hyperlight/blob/main/src/hyperlight_common/src/virtq/ring/canonical.rs#L163

andreiltd: Fixed with github.com/hyperlight-dev/hyperlight/pull/1794/commits/a70d6a0c6400ae5c6a88014228d1e930e33a6f1d

simongdavies (approved):

ludfjig (approved):

PR #1812, opened 2026-09-09
Fix unused sandbox test import on ARM64

simongdavies:
## Summary

Gate the `MultiUseSandbox` test import with `cfg(any(target_arch = "x86_64", feature = "trace_guest"))`. Its uses are limited to those configurations. This prevents an unused-import warning in default ARM64 builds.


ludfjig (approved):
lgtm but why didn't ci fail?

simongdavies:
> lgtm but why didn't ci fail?

Because the Code checks job does not run on ARM64 #1819

PR #1820, opened 2026-09-10
fix: use checked arithmetic in push_buffer (defense in depth)

JM00NJ:
`push_buffer` computed `data.len() + 8` and `stack_pointer_rel + data.len() + 8` with unchecked arithmetic. 
In a release build, if `data.len()` is near `usize::MAX`, these additions silently wrap around, causing `size_required` to appear smaller than `size_available` and bypassing the buffer-full check.

The subsequent `copy_from_slice` call catches this via `bounds_check!` today, so there is no currently exploitable path. However, as `try_pop_buffer_into` already uses `checked_add` for its size prefix, this is an inconsistency worth closing before future refactors introduce a caller where the downstream safety net is absent (defense in depth).

Changes:
- Add `StackError::PushSizeOverflow` variant with a clear error message
- Replace unchecked `+` with `checked_add` in both arithmetic sites
- Add a test that constructs a maximally-sized slice and asserts `PushSizeOverflow` is returned before any memory access occurs

### DIFF
--- a/fuzz/Cargo.toml
+++ b/fuzz/Cargo.toml
@@ -55,6 +55,13 @@ test = false
 doc = false
 bench = false
 
+[[bin]]
+name = "fuzz_push_pop_buffer"
+path = "fuzz_targets/fuzz_push_pop_buffer.rs"
+test = false
+doc = false
+bench = false
+
 [features]
 default = []
 trace = ["hyperlight-host/trace_guest", "hyperlight-common/trace_guest"]
--- a/fuzz/fuzz_targets/fuzz_push_pop_buffer.rs
+++ b/fuzz/fuzz_targets/fuzz_push_pop_buffer.rs
@@ -1,0 +1,55 @@
+#![no_main]
+
+use hyperlight_host::mem::shared_mem::ExclusiveSharedMemory;
+use libfuzzer_sys::fuzz_target;
+
+// try_pop_buffer_into requires a generic T: TryFrom<&[u8]>
+struct FuzzBytes(#[allow(dead_code)] Vec<u8>);
+
+impl TryFrom<&[u8]> for FuzzBytes {
+    type Error = std::convert::Infallible;
+    fn try_from(v: &[u8]) -> Result<Self, Self::Error> {
+        Ok(FuzzBytes(v.to_vec()))
+    }
+}
+
+fuzz_target!(|data: &[u8]| {
+    if data.is_empty() {
+        return;
+    }
+
+    const MEM_SIZE: usize = 65536;  // 64KB
+
+    // === TARGET 1: try_pop_buffer_into ===
+    // Write fuzzer-generated data directly into shared memory, simulating a guest.
+    // This is the primary target since it exercises the guest-controlled code path.
+    {
+        let Ok(eshm) = ExclusiveSharedMemory::new(MEM_SIZE) else {
+            return;
+        };
+        let (mut hshm, _) = eshm.build();
+
+        let write_len = data.len().min(MEM_SIZE);
+        let _ = hshm.copy_from_slice(&data[..write_len], 0);
+
+        // Any panic or UB will be caught by the fuzzer
+        let _: Result<FuzzBytes, _> = hshm.try_pop_buffer_into(0, MEM_SIZE);
+    }
+
+    // === TARGET 2: push_buffer → try_pop_buffer_into roundtrip ===
+    // Verifies the correctness of the checked_add fix introduced in this PR.
+    {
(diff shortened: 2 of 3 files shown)


JM00NJ:
## Why the fuzz target is included

The fuzz target (`fuzz/fuzz_targets/fuzz_push_pop_buffer.rs`) 
was written alongside this fix to validate the checked_add 
guards and to provide ongoing regression coverage.

**Key differences from existing fuzz targets:**
- Exercises the memory stack API directly, no sandbox or 
  hypervisor required
- ~35K exec/s vs existing targets which require KVM spin-up
- Targets the guest-controlled code path in try_pop_buffer_into,
  which is the actual trust boundary

**Fuzz results across three configurations:**

| Input size | Iterations | Crashes |
|------------|------------|---------|
| 4KB        | 184M+      | 0       |
| 64KB       | 10M+       | 0       |
| 256KB      | 57M+       | 0       |

At 256KB, the fuzzer autonomously identified `0xFFFFFFFB` 
(u32::MAX - 4) as a critical boundary value and exercised 
it 3.28M times — the exact overflow point that 
`checked_add(4)` in `try_pop_buffer_into` guards against.

Coverage plateau reached at 399 branches / 872 features, 
confirming all reachable code paths were exercised.

As the guest protocol evolves and new message types are 
added to try_pop_buffer_into, the fuzzer will automatically 
exercise new code paths without any modifications.

fuzz/fuzz_targets/fuzz_push_pop_buffer.rs
```diff
@@ -0,0 +1,55 @@
...
+
+        let write_len = data.len().min(MEM_SIZE);
+        let _ = hshm.copy_from_slice(&data[..write_len], 0);
+
+        // Any panic or UB will be caught by the fuzzer
+        let _: Result<FuzzBytes, _> = hshm.try_pop_buffer_into(0, MEM_SIZE);
+    }
+
+    // === TARGET 2: push_buffer → try_pop_buffer_into roundtrip ===
+    // Verifies the correctness of the checked_add fix introduced in this PR.
```
simongdavies: Nit: "introduced in this PR" doesn't really make sense as a comment as the context is missing.

simongdavies (dismissed):
Thanks , LGTM, one minor nit on the comments.

simongdavies (approved):
@JM00NJ LGTM thanks!!

fuzz/Cargo.toml
```diff
@@ -55,6 +55,13 @@ test = false
 doc = false
 bench = false
 
+[[bin]]
+name = "fuzz_push_pop_buffer"
```
jsturtevant, line 59: Do we need to update CI to run this somewhere?

jsturtevant (approved):

fuzz/Cargo.toml
JM00NJ: If you ask me, it could have benefits for future code changes on push_buffer/try_pop_buffer_into since it doesn't require a hypervisor — making it cheaper to run in CI than the existing targets.
Also from what I can see in the Justfile, fuzz-like-ci only runs for 1 second in release mode — mainly to verify the harness compiles. So adding it there should be very low overhead.

simongdavies: I'll merge this and we can add that in a new PR.

src/hyperlight_host/src/mem/shared_mem.rs
```diff
@@ -1392,10 +1399,11 @@ impl HostSharedMemory {
...
 
         // update stack pointer to point to the next free address
-        self.write::<u64>(
-            buffer_start_offset,
-            (stack_pointer_rel + data.len() + 8) as u64,
-        )?;
+        let new_sp = stack_pointer_rel
+            .checked_add(data.len())
+            .and_then(|v| v.checked_add(8))
+            .ok_or(StackError::PushSizeOverflow(data.len()))? as u64;
```
jprendes, line 1405: shouldn't this be using `.try_into::<u64>()` instead of `as u64`?
usize is the same or smaller than u64 in the current platform, but as this change is about correctness and future proofing the code, I think it would make sense.

JM00NJ: You're right, .try_into::<u64>() would be more correct and future-proof. Since the project currently targets x86_64 only usize and u64 are equivalent, but I'm happy to open a follow-up PR with that fix if you'd like.

jsturtevant: We also support ARM64 so we should follow up

JM00NJ: #1825

PR #1822, opened 2026-09-10
ci: add fuzz_push_pop_buffer to the CI fuzzing workflow

JM00NJ:
As discussed in #1820, adds fuzz_push_pop_buffer to the 
like-ci fuzzing recipe.

The hypervisor parameter is not used in fuzz-like-ci, so 
this target runs cleanly regardless of hypervisor config.
In release mode it runs for 1 second; compile verification only.


simongdavies (approved):

simongdavies:
Thanks @JM00NJ

PR #1833, opened 2026-09-15
feat: support guest log level overrides for initialized snapshots

### LINKED ISSUE #1698: max_guest_log_level is not plumbed through snapshot load

jprendes:
Closes #1698

Adds guest log-level overrides for initialized snapshots.

### Changes

- Adds a host-to-guest runtime log-level update protocol.
- Updates guest `log` and tracing filters before guest calls.
- Supports log-level overrides when restoring snapshots.
- Adds `MultiUseSandbox::log_level`.
- Reapplies the configured level after `restore()`.
- Adds coverage for initialized snapshots and restore persistence.
- Update documentation comments.

### Validation

- Focused snapshot and log-level tests


src/hyperlight_guest_tracing/src/subscriber.rs
```diff
@@ -19,17 +21,22 @@ pub(crate) struct GuestSubscriber {
...
     /// Creates a new `GuestSubscriber` with the given guest start TSC and maximum log level
     pub(crate) fn new(guest_start_tsc: u64, filter: LevelFilter) -> Self {
         Self {
             state: Arc::new(Mutex::new(GuestState::new(guest_start_tsc))),
-            max_log_level: filter,
+            max_log_level: AtomicU64::new(u64::from(GuestLogFilter::from(filter))),
         }
     }
+
+    pub(crate) fn set_max_log_level(&self, filter: LevelFilter) {
```
ludfjig, line 36: Updating this value does not rebuild tracing’s cached callsite interest. A callsite registered under ERROR can remain disabled after switching to INFO. The reverse also applies.

Please rebuild the interest cache tracing_core::callsite::rebuild_interest_cache() after updating the filter, or return Interest::sometimes() from register_callsite.

Please also test one tracing callsite before and after raising and lowering the filter. The integration tests only exercise log::set_max_level. The subscriber.rs:188 calls enabled() directly and bypasses callsite caching.

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -4752,6 +4757,138 @@ mod tests {
...
+                LOGGER.clear_log_calls();
+                count
+            };
+
+            let before_restore = count_guest_logs(&mut sandbox);
+            sandbox.restore(snapshot).unwrap();
+            let after_restore = count_guest_logs(&mut sandbox);
+
+            assert!(before_restore >= 1);
+            assert_eq!(before_restore, after_restore);
```
ludfjig, line 4889: This test can pass when log_level(ERROR) has no effect because both counts can retain the original TRACE value. Please assert the expected ERROR-filtered count before and after restore.

The test is also ignored but absent from test-isolated, so CI does not run it.

src/hyperlight_host/src/sandbox/builder.rs
```diff
@@ -174,10 +173,8 @@ impl SandboxBuilder {
                     ));
                 }
 
-                if guest_log_level.is_some() {
-                    return Err(new_error!(
-                        "guest_log_level has no effect when building from a snapshot, as the snapshot already contains it"
-                    ));
+                if let Some(log_level) = guest_log_level {
```
ludfjig, line 176: The old error is still mentioned in rustdoc on `SandboxBuilder::guest_log_level`

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -138,9 +141,20 @@ impl MultiUseSandbox {
...
             snapshot: None,
             pt_root_finder: None,
+            max_guest_log_level: None,
         }
     }
 
+    /// Sets the maximum log level used by future guest calls.
+    ///
+    /// The setting is reapplied after restoring a snapshot.
+    pub fn log_level(&mut self, log_level: LevelFilter) -> Result<()> {
```
ludfjig, line 151: should this be `set_log_level`?

ludfjig (commented):
Looks great. I think we can update docs/hyperlight-metrics-logs-and-traces.md as well

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: I believe in general we don't use the set/get prefixes much in Rust. I'll leave it as it is now, but we can have a follow up discussion and change it with a follow up PR.

@yoshuawuyts , I'm almost sure you have some opinion on this.

jprendes: Created #1834 so that we can discuss it :-)

ludfjig (commented):
looks good the only thing I'm uncertain about is whether bumping the snapshot version is really necessary. What could go wrong if we don't bump any versions, and load an old snapshot with these changes?

jprendes:
I think on that case we could risk overriding the exception stack with log level information, right?
But then, the exception stack shouldn't have any valuable information IIUC. Not sure.

jprendes:
Here's the comment that resulted in the version bump: https://github.com/hyperlight-dev/hyperlight/pull/1833#discussion_r4017923451

ludfjig:
> Here's the comment that resulted in the version bump: [#1833 (comment)](https://github.com/hyperlight-dev/hyperlight/pull/1833#discussion_r4017923451)

Ok I think let's bump it, there will be breakage next release anyway since #1793 will bumps it too

jprendes:
@ludfjig would you ~~mum be~~ *mind* approving the PR if you are happy with it?

**Edit:** Autocorrect typo

src/hyperlight_guest_bin/src/guest_function/call.rs
```diff
@@ -73,6 +73,8 @@ pub(crate) fn call_guest_function(function_call: FunctionCall) -> Result<Vec<u8>
 }
 
 pub(crate) fn internal_dispatch_function() {
+    crate::refresh_guest_log_level();
```
ludfjig, line 76: nit: not necessarily related to this PR, but it would be nice if setting level to OFF removed all the tracing overhead. Then we could also remove the tracing cargo feature in the guest and just always enable it. But happy to merge this pr as is

ludfjig (approved):
Send it!!!!

src/hyperlight_guest_bin/src/guest_function/call.rs
jprendes: that would be nice, although I guess the feature not only helps with perf, but also with binary size

jprendes: Created #1837

PR #1840, opened 2026-09-17
perf: skip guest tracing work while the log level is off

### LINKED ISSUE #1837: setting the tracing level to OFF should remove all the tracing overhead
nit: not necessarily related to this PR, but it would be nice if setting level to OFF removed all the tracing overhead. Then we could also remove the tracing cargo feature in the guest and just always enable it. But happy to merge this pr as is

_Originally posted by @ludfjig in https://github.com/hyperlight-dev/hyperlight/pull/1833#discussion_r4033215533_

jprendes:
Closes #1837.

A guest log level of `OFF` skips the per-call tracing work: TSC reads, span
setup, trace state locking, and flush plumbing.

## Behavior

* `is_trace_enabled` and `accepts_trace_events` derive from a single relaxed
  atomic holding the encoded filter, so neither upgrades a `Weak`.
* `register_callsite` caches interest per level. A level change rebuilds the
  interest cache, so runtime level updates take effect.
* `generic_init` and the dispatch path open their spans only while tracing is
  active.
* `end_trace`, `flush`, `new_call`, `reset`, and `serialized_data` return early
  while tracing is off, which keeps `out32` clear of trace state locking.

`hyperlight_guest_tracing::is_trace_enabled` reports whether the configured
level is above `OFF`. It is public, so `CHANGELOG.md` carries an entry.

## Why `trace_guest` stays

The feature and the log level control different things, and both earn their
place.

`trace_guest` is a build-time choice about whether a guest carries the tracing
machinery at all. It gates the subscriber, the trace state, and the batch
encoder. A guest built without it pays no binary size for code it never runs,
which matters for small guests.

The log level is a runtime choice. `MultiUseSandbox::log_level` raises and
lowers it while the guest runs, so the call sites keep a check that survives
compilation. That check is one relaxed atomic load at the boundaries above,
plus the callsite interest `tracing` already caches. Dropping the feature would
compile the machinery into every guest and still leave that check in place, so
it trades binary size for nothing.

Fixing the level at build time with `tracing/max_level_off` is the only way to
erase the check, and it costs the runtime control that
`MultiUseSandbox::log_level` provides.

## Testing

* `max_guest_log_level_can_be_disabled_and_reenabled` covers TRACE to OFF to
  TRACE on a live sandbox.
* `max_guest_log_level_can_be_enabled_after_off_init` covers a sandbox built at
  OFF and raised at runtime, and asserts it matches a sandbox built at TRACE.
* `callsite_interest_tracks_filter_updates` covers the interest contract.
* The existing `log_message` integration test asserts exact span enter and exit
  counts, which covers the span gating.


ludfjig (dismissed):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
```diff
@@ -4889,6 +4889,88 @@ mod tests {
             assert_eq!(before_restore, after_restore);
         }
 
+        /// Runtime log-level updates can disable and re-enable guest tracing.
+        ///
+        /// Ignored because it installs a process-global `log` logger. Run in
+        /// isolation via the `test-isolated` Justfile recipe.
+        #[test]
+        #[ignore]
+        fn max_guest_log_level_can_be_disabled_and_reenabled() {
```
ludfjig, line 4898: these tests are good but only run the simpleguest with trace_guest feature disabled, so it's only testing logging not tracing. I noticed we don't have any tests guests complied with trace_guest except the tracing fuzzing

ludfjig (approved):

src/hyperlight_host/src/sandbox/initialized_multi_use.rs
jprendes: created #1842

